PR #15, opened 2015-02-16
takeaway_challenge 

Pau1fitz:


lib/dishes.rb
```diff
@@ -0,0 +1,10 @@
+class Dish
+
+  attr_accessor :name, :price
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+  end
+
```
danldb: This works. If I am being very picky, I could suggest that you use an attr_reader and then manually create methods to allowing amending the name and price of a dish - this way you can control the input and ensure that e.g. the name is always a string, the price is always an integer.

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
+class Menu
+
+  attr_reader :food_holder
+
+  def initialize
+    @food_holder = []
+  end
+
```
danldb: The convention would be to call an array of dishes @dishes instead of @food_holder

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+  def add_dish(dish)
+    if @food_holder.include?(dish)
+    raise "Dish already added"
```
danldb: line 11 should be indented

lib/order.rb
```diff
@@ -0,0 +1,21 @@
+class Order
+
+  attr :basket
+
```
danldb: attr is now depracated - you should use attr_accessor instead. In this instance I would actually use a reader instead of an accessor - again, you want to ensure that any additions or subtractions to the order are controlled.

lib/order.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def total_cost
+    total = 0
+    @basket.each do |dish, quantity|
+      total += dish.price * quantity
+    end
+    total
+  end
+
```
danldb: Consider using inject to add to a total

lib/restaurant.rb
```diff
@@ -0,0 +1,13 @@
+class Restaurant
+
+  def take_order(order, payment, person)
+
+    if payment < order.total
+      puts "not enough grade"
+    else
```
danldb: You don't need to write an interface - not part of the brief. 
Do not handle errors with a string - raise an error instead. 
Do not use puts except when debugging.

spec/dishes_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'dishes'
+
+describe Dish do
+
+  it 'has a name and a price' do
+    dish = Dish.new("pizza", 5)
+    expect(dish.name && dish.price).to eq "Pizza" && 5
+  end
```
danldb: Preferable to make 2 tests, one for price and one for name.

spec/menu_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'menu'
+
+describe Menu do
+
+  let(:menu) { Menu.new }
+  let(:dish) {double :dish}
+
```
danldb: If you give the dish a name and a price you can use the same dish for all your tests.

spec/menu_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  it 'can print the contents of the menu' do
+    dish = double(dish, name: "Pizza", price: 5)
+    menu.add_dish(dish)
+    expect(menu.print_menu).to eq "Pizza -- $5"
+  end
+
+  it 'prevents the same dishes being added to menu twice' do
+    dish = double(dish, name: "Pizza", price:5)
+    menu.add_dish(dish)
```
danldb: Watch out for repetition - you could use a before block to reduce this.

spec/order_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'order'
+
+describe Order do
+
+  let(:order) { Order.new }
+  let(:dish) { double :dish}
+
+  it 'has a basket which dishes can be added to' do
+    dish = double(dish)
+    order.add_dish(dish, 4)
```
danldb: You already have a double of a dish in your let statement

spec/order_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    expect(order.basket.count).to eq 1
+  end
+
+  it 'can calculate the total cost of order' do
+    dish = double(dish, price: 4)
+    order.add_dish(dish, 4)
```
danldb: If your original let(:dish) had a name and a price, you could use the same dish double for all your tests, reducing repetition.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:order) { double :order, total_cost: 10}
+
+  it 'raises an error if the incorrect amount is paid' do
+    expect{takeaway.process_order(order, 15)}.to raise_error "Wrong amount paid"
+  end
+
+  it 'confirms a text will be sent if amount paid is correct' do
+    expect(takeaway.process_order(order, 10)).to eq "Your order is processed, text confirming this will be sent shortly"
+  end
+
```
danldb: Returning a string is unnecessary - maybe just check that no error is raised.

danldb:
Very strong submission, thanks Paul!

I notice that you didn't test your sms message - this needs to be stubbed out so that you are not sending text messages when running your tests. You should also be testing to ensure that the correct message is sent to the Twilio client object. 

Let us know if you have any questions.

PR #16, opened 2015-02-16
Pull request on Monday morning

wardymate:
Dish, menu, message_sender, customer and order class created and tests passed and created. project almost complete. Need to finish customer class and finish writing tests with the twilio message sender.


lib/customer.rb
```diff
@@ -0,0 +1,14 @@
...
+		if input[0] == "0"
+			input[0] =''
+			input.insert(0,"+44")
+		elsif input[0] == "+" && input.length == 13
+			input
+		else
+			input.insert(0,"+")
+		end
+	end	
+end
```
tansaku: you have tab indenting here, please prefer 2 space indenting - also you have added lots of functionality not requested in the brief.  In general only add complexity like this when you find you really really really need it.

lib/menu.rb
```diff
@@ -0,0 +1,21 @@
...
+
+	def add_dish(dish)
+		@dishes << dish
+	end
+
+	def select_dish(dish)
+		@dishes.select{|listed_dish| listed_dish == dish}.first
+	end
+
+end
```
tansaku: this select_dish method seems redundant.  If you already have a reference to the object you already have a reference to the object.  You might be checking that a dish is on the menu, but that is a different thing.  The way you have this here it will throw an error when the dish is not on the menu - not quite what you want - or at least the type of error will be difficult to understand

lib/dish.rb
```diff
@@ -0,0 +1,10 @@
+class Dish
+
+	attr_reader :name, :price
+
+	def initialize name, price=300
+		@name = name
+		@price = price
+	end
+
+end
```
tansaku: good - although could be a struct if no other functionality needed

lib/message_sender.rb
```diff
@@ -0,0 +1,39 @@
...
+
+	def set_order_time(time)
+		@order_time = time.to_s.slice(11,5)
+	end
+
+	def order_time
+		@order_time
+	end
+
+	def delivery_time
```
tansaku: not sure what this is doing or why you need it - again don't do anything not requested in brief

lib/message_sender.rb
```diff
@@ -0,0 +1,39 @@
...
+
+	def sms_message
+		@sms_message = "Thank you! Your order was placed and will be delivered before #{delivery_time}"
+	end
+		
+	def send_message
+		require 'rubygems'
+		require 'twilio-ruby'
+		# Get your Account Sid and Auth Token from twilio.com/user/account
+		account_sid = 'AC224c736fce467ca8cf9b8b3bfa73a034'
```
tansaku: all ids etc. should be in env variables and accessed by ENV['TWILLIO_ACCOUNT_SID'] statements

lib/message_sender.rb
```diff
@@ -0,0 +1,39 @@
...
+		array.insert(1,":")
+		array.join
+	end
+
+	def sms_message
+		@sms_message = "Thank you! Your order was placed and will be delivered before #{delivery_time}"
+	end
+		
+	def send_message
+		require 'rubygems'
```
tansaku: require statements at the top of the file please :-)

lib/message_sender.rb
```diff
@@ -0,0 +1,39 @@
+class MessageSender
```
tansaku: well done moving this functionality into it's own class

lib/order.rb
```diff
@@ -0,0 +1,36 @@
+class Order
+
+	attr_accessor :order_items
```
tansaku: do you really need an attr_accessor here? prefer attr_reader if at all possible

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+	def initialize
+		@order_items = []
+	end
+
+	def total_item_count
+		@order_items.count
+	end
+
+	def add_item(menu,item)
+		order_items << menu.dishes.select{|dish| dish == item}.first
```
tansaku: seems more complex than necessary

spec/customer_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+	it 'should be able to provide a telephone number in a different format' do
+		customer = Customer.new
+		expect(customer.telephone_number("+447738567183")).to eq "+447738567183"
+	end
+	
+	it 'should be able to provide a telephone number in a different format' do
+		customer = Customer.new
+		expect(customer.telephone_number("447738567183")).to eq "+447738567183"
+	end
+end
```
tansaku: yes, but entirely unnecessary to the project ... things will get complicated enough on their own :-)

spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+		curry = Dish.new "curry"
+		expect(curry.name).to eq "curry"
+	end
+
+	it 'should know it\'s price' do
+		curry = Dish.new("curry",400)
+		expect(curry.price).to eq 400
+	end
+
+
```
tansaku: avoid this dead white space

spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'dish'
+
+describe Dish do 
+
+	it 'should know its name' do
+		curry = Dish.new "curry"
```
tansaku: could be in a let

spec/menu_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'menu'
+
+describe Menu do
+
+	let(:curry) {double :dish, name: "curry", price: 500}
+	let(:kebab){double :dish, name: "kebab", price: 450}
+
+
+	it 'should be able to accept dishes' do
+		menu = Menu.new
```
tansaku: could be in a let

spec/message_sender_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+
+describe MessageSender do
+
+	let(:time) {double :time, to_s: "2015-02-15 20:48:22 +0000"}
+	let(:time2) {double :time, to_s: "2015-02-15 23:48:22 +0000"}
+	let(:message_sender) {MessageSender.new}
+	let(:twilio) {double :twilio, account_sid: 'AC224c736fce467ca8cf9b8b3bfa73a034', auth_token: 'b4c6b04027185a9860c8e9b429038012', sid: 'SM9acd879d0458458d8bb82fa9cfd22f15'}
+
+	it 'should know what time it is' do
```
tansaku: okay, understand what you are doing, but all surplus to requirement ...

spec/order_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+	it 'should be able to select items from the menu and add them to the list' do
+		order.add_item(menu,curry)
+		order.add_item(menu,kebab)
+		expect(order.total_item_count).to eq (2)
+		expect(order.order_items).to eq [curry,kebab]
+	end
+
+	it 'should be able to total the cost of one item in the order list' do
+		order.add_item(menu,curry)
+		expect(curry).to receive(:price)
```
tansaku: I think it would make more sense to not specify the price in the double above and specifiy the return value here like expect(curry).to receive(:price).and_return(500)

spec/order_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+		order.add_item(menu,curry)
+		expect(order.item_count(curry)).to eq 2
+	end
+	
+	it 'should raise an error if the total cost of the items in the order is incorrect' do
+		order.add_item(menu,curry)
+		order.add_item(menu,curry)
+		expect(curry).to receive(:price)
+		expect{order.order_total_check(950)}.to raise_error(RuntimeError, 'Order total is wrong')
+  end
```
tansaku: do end matching gone funny here - must fix this or employers will look at it and not even consider you for an interview

spec/message_sender_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+		expect(message_sender.sms_message).to eq "Thank you! Your order was placed and will be delivered before 21:48"
+	end
+
+	it 'should send the outbound message' do
+		expect(time).to receive(:to_s)
+		message_sender.set_order_time(time)
+		expect(message_sender.send_message).to eq 34
+	end
+
+end
```
tansaku: main problem here is that your test is hitting the real Twillio API - you need to stub out twilio and test that the right things are happening under the hood

tansaku:
Some good work here, but you did too much work on things not required by the brief, and then missed out other things such as properly stubbing out Twillio.  I'm also a little concerned that things don't all pull together at the high level.  For example the order total check is only tested - it's not called anywhere else, i.e. here is no check that the order total is correct before sending the SMS message

PR #17, opened 2015-02-16
takeaway-challenge(Louise)

louisebeh:


lib/customer.rb
```diff
@@ -0,0 +1,28 @@
+require 'twilio-ruby'
+
+module Customer
+
+  attr_reader :send_text
+
```
danldb: send_text is not an instance variable, so an attr_reader is not appropriate.

lib/customer.rb
```diff
@@ -0,0 +1,28 @@
...
+ def send_text
+    # put your own credentials here
+    account_sid = 'ACdff620df7376a7affc0d0ae499f860d1'
+    auth_token = 'e2db70e02bb0b68a9994bd0893d5a875'
+
```
danldb: Account credentials are now publicly available to anyone who wants to impersonate you on twilio! These should be extracted to environment variables.

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
...
+  attr_reader :items, :food, :price
+
+  def initialize
+    @menu_items ={"Duck Rice" => 8, "Tonkotsu Ramen" => 8, "Gyoza" => 5, "Jalapeno Hamachi" => 8, "Summer Rolls" => 5, "Mango" => 3}
+  end
+
+  def items
+    @menu_items
+  end
+
```
danldb: This is an exact replica of the attr_reader

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+module Menu
+
+  attr_reader :items, :food, :price
+
```
danldb: You are not using the attr_reader :food or :price. Should delete them.

lib/nil.rb
```diff
@@ -0,0 +1,6 @@
+class NilClass
+ def to_hash
+   to_h
+ end
+end
+data = {test: :t}.merge!(nil)
```
danldb: I respect your ingenuity! However, I think you are overcomplicating things - if you are getting a nil value error, figure out a way to rescue the error or to prevent it from occuring, rather than monkeypatching extra methods onto NilClass.

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+module Menu
+
```
danldb: should be a class

lib/customer.rb
```diff
@@ -0,0 +1,28 @@
+require 'twilio-ruby'
+
+module Customer
+
```
danldb: should be a class

lib/order.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'menu'
+require_relative 'customer'
+include Menu
+include Customer
+
```
danldb: By including menu and customer modules in your Order class, you have put everything into one class, thus removing almost all OOP from your program. Menu and Customer should be separate classes.

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+class Order
+
+attr_reader :basket, :select_items, :change_quantity, :basket_count, :order_value, :total_values
+
```
danldb: attr_readers refer to instance variables

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+  def initialize
+    @order = {}
+    @min_order_value = 10
+  end
+
+  def select_items(item)
+    Menu.items {|k,v| k.match item}
+      @order.merge!(item)
+  end
+
```
danldb: Line 17 is referring to the entire Menu class, rather than an instance of the class. You do not actually have a menu class, as Menu is currently a module.

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+  def change_quantity(item)
+  end
+
+  def basket
+    @order
+  end
+
```
danldb: If your instance variable is called order, and you have an attr_reader :order, then you do not need another method to read the order. If you wanted to manually write the attr_reader, the convention would be for the method to have the same name as the variable.

``` attr_reader :order ```

is the same as:

```
def order
  @order 
end
```

File: spec/customer_spec.rb [hunk 12]
```diff
@@ -0,0 +1,15 @@
+require 'customer'
+require 'menu'
+
+class Customer_Class; include Customer; end
+
```
danldb: Tests should not be inside the Customer class, and should not include Customer

spec/customer_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe Customer do
+  let(:customer){Customer_Class.new}
+
```
danldb: if Customer was a class instead of a module, you could use
``` let(:customer){Customer.new} ```

File: spec/customer_spec.rb [hunk 14]
```diff
@@ -0,0 +1,15 @@
...
+it 'should receive a Twilio text with delivery time' do
+  customer.send_text
+  #check phone to confirm that a customer can receive text
+end
+
```
danldb: You should be using a stub, so that you are not spamming the Twilio API every time you run your tests.

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'menu'
+
+class Menu_Class; include Menu; end
+
```
danldb: Menu should be its own class, so you don't have to create these pseudo-classes.

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
...
+  attr_reader :items, :food, :price
+
+  def initialize
+    @menu_items ={"Duck Rice" => 8, "Tonkotsu Ramen" => 8, "Gyoza" => 5, "Jalapeno Hamachi" => 8, "Summer Rolls" => 5, "Mango" => 3}
+  end
```
danldb: You have an attr_reader for items, but you have called the instance variable @menu_items. Call the variable @items so that the attr_reader works. You have put an initialize method inside a module - this should be within a class, as you cannot have an instance of a module.

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  let(:menu){Menu_Class.new}
+
+it 'should not be empty on initialization' do
+  expect(menu.items).not_to be_empty
+end
+
+it 'should contain food' do
+  expect(menu.food).to eq menu.items.keys
+end
+
```
danldb: better to put the exact object you are expecting after the eq. 

for instance ``` exect(menu.food).to eq ["Duck Rice", "Tonkotsu Ramen"] ```

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+it 'should contain prices' do
+  expect(menu.price).to eq menu.items.values
+end
+
+
+it 'should contain food with corresponding prices' do
+  expect(menu.items.keys).to include{values :true}
+end
+
```
danldb: Again, not really testing anything here.

spec/order_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+require 'customer'
+require 'nil'
+
+
+describe Order do
+  let(:order){Order.new}
+  let(:customer){double :Customer}
+  let(:customer_order){ order.select_items({"Duck Rice" => 8, "Gyoza" => 5, "Mango" => 3}) }
+  let(:small_order){ order.select_items({"Duck Rice" => 8}) }
+
```
danldb: lines 9 and 10 do not really belong in let statements - writing a method would be better.

spec/order_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  it 'should allow customer to select items' do
+    customer_order
+    expect(order.select_items(items)).to eq order.basket
+  end
+
```
danldb: use the actual expected values after the eq

spec/order_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  it 'should reject an order that does not meet a minimum order value' do
+    small_order
+    expect{order.submit_order}.to raise_error(RuntimeError, "You haven't reached the minimum order value of £10, please add to your order.")
+  end
+
+  xit 'should send the customer a text using Twillio' do
+    customer_order
+    expect(order.submit_order).to eq order.send_text #text successfully sends to phone
+  end
```
danldb: The sms should not be sent whenever you run your test suite - imagine if you were being charged for every test, you wouldn't want to run your tests very regularly!

danldb:
Hi Louise - thanks for your submission! 

You have obviously put a lot of work into this, so please do not be disheartened by all the comments! You have come up with some ingenious solutions, and I feel that you will be a very strong coder. Unfortunately in this case, you have been building creative solutions to cover some gaps in your understanding. 

I would recommend you to revisit Boris Bikes - when looking at a brief, go through and underline all the nouns, then look at each of them and decide whether they could be a potential class. Then write CRC cards for each class, and be clear in head what their responsibilities are. Remind yourself what a Class actually is, what is an instance of a class, and what is a module. You got into a muddle because you should not have used modules for Menu and Customer - they should be classes.

I hope that helps - let us know if you have any questions.

PR #18, opened 2015-02-16
Takeaway challenge - working?

katebeavis:
Can add to menu, check total but when checking if money matches total, cannot access the correct total so have to 'stub' it in.


lib/menu.rb
```diff
@@ -0,0 +1,11 @@
+class Dish
+
+
+attr_accessor :name, :price
+
+  def initialize name, price=1
+    @name = name
+    @price = price
+  end
+
```
danldb: Use attr_reader here instead of attr_accessor. Better to write individual setter methods so that you can control the input to ensure that the name is always a string, the price is always an integer, and any malicious or faulty code in other classes cannot delete your data.

lib/order.rb
```diff
@@ -0,0 +1,46 @@
+require './lib/menu'
+
+class Order
+
+  DEFAULT_MAXIMUM = 10
+
+  attr_accessor :orders, :count, :total
+
```
danldb: Use a reader instead of an accessor.

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+  def order_total
+    orders.each {|dish, amount| @total += (dish.price * amount)}
+    @total #this seems to have a bug where it keeps adding everytime
+    #you call it
+  end
+  
+end
+
+
```
danldb: Every time you call order_total, you are adding to the @total, but you are never resetting it - that is why it adds again and again. Maybe a better idea not to use an instance variable for the total, but just to have the order_total method return the current total at the time it is called.

lib/take_order.rb
```diff
@@ -0,0 +1,41 @@
...
+require './lib/order.rb'
+
+class Checkout
+
+  def initialize(money, total)
+    @money = money
+    @total = total#doesn't access order total, have to insert total
+    #upon initialising checkout
+  end
+
```
danldb: You could consider passing an order to the checkout, so that the checkout can access order total.

lib/take_order.rb
```diff
@@ -0,0 +1,41 @@
...
+def pay_money
+  if @money != @total
+    raise "You have entered an incorrect amount, please reenter"
+  else
+    puts "Thanks for your order! A text has been sent with your estimated delivery time."
+  end
+  send_message
```
danldb: Avoid using puts. The brief does not ask for an interface. Does not really need an else.

lib/take_order.rb
```diff
@@ -0,0 +1,41 @@
...
+
+ def delivery_time
+  @hourahead = Time.new + 3600
+ end
+ 
+  def send_message
+    delivery_time
+    account_sid = 'ACb869fa6a0e0d8bcbca82368079985d07' 
+    auth_token = '8ca316bf9c0c14d31739896ba2825085' 
+
```
danldb: Tokens are now published on github, so anyone can impersonate you on twilio. Avoid this by using environment variables. Google Dotenv gem for an easy way to do this.

lib/take_order.rb
```diff
@@ -0,0 +1,41 @@
...
+    # set up a client to talk to the Twilio REST API 
+    @client = Twilio::REST::Client.new account_sid, auth_token 
+     
+    @client.account.messages.create({
+      :from => '+441412804334', 
+      :to => '+447935294331', 
+      :body => "Thanks for your order. Your cereal will be with you before #{delivery_time.strftime("%I:%M%p")}"  
+    })
+  end
+
```
danldb: Hard coded customer telephone number - could have had a Customer class.

spec/order_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    order.add_cereal(coco_pops, 3)
+    expect(order.item_count).to eq (5)
+  end
+
+  it 'can calculate the total of an order' do
+    order.add_cereal(frosties, 2)
+    order.add_cereal(coco_pops, 3)
+    expect(order.order_total).to eq (17.60)
+  end
+
```
danldb: Could have added tests here for your order_total bug.

spec/take_order_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:checkout2){ double :checkout, :pay_money => @money != @total }
+
+  it 'should tell the customter that their food will be delivered with an hour' do
+    expect(checkout.delivery_time).to eq (Time.new + 3600)
+  end #when this is tested it fails as the time is less than a millisecond apart
+
+  it 'should raise an error if the money doesn\'t match the total' do
+    expect(checkout2.pay_money).to raise_error(RuntimeError, "You have entered an incorrect amount, please reenter")
+  end
+
```
danldb: No tests for the text message - you should be stubbing out the Twilio client to ensure that the correct messages are being sent, but the actual API will not be contacted every time you run your tests.

danldb:
Hey Kate - thanks for the submission!

You have done a great job here. You have a good understanding of RSpec, and have clearly understood the problems in the brief. 

You ran into a couple of problems with your instance variables, that you could have targeted better with your test suite - this would have helped you to debug the problem. 

Your OOD could be improved by diligent use of CRC cards at the planning stage. You had a very powerful Checkout class, that should have handed some of its responsibilities on to a Customer and/or Messenger class.

Also very important that you should never be hitting a real api call from within your tests - using a stub for the Twilio client would mean you can test your code without actually sending a text message and potentially wasting time and money sending unnecessary texts.

Hope that helps - let us know if you have any questions!

PR #19, opened 2015-02-16
takeaway challenge - Loris

loris-fo:


lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+class Order
+
+  def dishes_choice
+    $dishes = {"carbonara" => 11,
+    	"bolognesa" => 9,
+    	"pesto" => 10,
+    	"sea food" => 13
+    }
+  end
+
```
danldb: Consider a menu class

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+  def menu_list
+   puts "here is the list of the menu:"
+   dishes_choice.each {|k, v| puts "#{k}: #{v}"}
+   puts "Please choose the items you'd like to purchase by entering the name of the item followed by the quantity."
+  end
+
```
danldb: Avoid puts - the brief did not ask for an interface.

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+  def order_count
+    if @items == nil
+      0
+    else
+      @items.order_count
+    end
+  end
+
```
danldb: This looks like an infinite loop?

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+  def add_dish(item, amount)
+    selected_dishes << item
+    selected_quantities << amount
+  end
+
```
danldb: Consider a Dish class

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+    def cost(item, amount)
+    amount * $dishes[item].to_i
+  end
+
```
danldb: Bad indentation, line 40.

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+ 
+		@client = Twilio::REST::Client.new account_sid, auth_token 
+ 
+		@client.account.messages.create({
+		:from => '+441254790223', 
+		:to => '+447474718811', 
+		:body => '"Thank you! Your order was placed and will be delivered before #{time}"',  
+		})
+  end
+
```
danldb: Bad indentation. The Order should not be responsible for knowing the customer phone number (A Customer class would be ideal). The Order class should not be responsible for sending a text message - a Messaging class would be better. Your Twilio tokens are now public on Github, so anyone can impersonate you on Twilio. Your tokens should be extracted to environment variables instead - Dotenv is a good gem that makes this relatively easy.

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+let(:order)  {Order.new}
+
+	it 'should have a list of dishes' do
+		expect(order.dishes_choice.length).to eq (4)
+	end
+
+	it 'should set the order items to 0' do
+		expect(order.order_count).to eq (0)
+	end
+
```
danldb: order_count should have been tested for other numbers - you have missed edge cases. Always try to test the "good path" and the "bad path".

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+	it 'should be able to add quantities' do
+		order.add_dish("pesto", 5)
+		expect(order.selected_quantities).to eq([5])
+	end
+
+	it 'should be able to calculate the cost of the order' do
+		expect(order.cost("carbonara",3)).to eq (33)
+	end
+
```
danldb: You have not tested the messenger - you should be testing to ensure the correct messages are sent from the code you have written yourself, whilst stubbing out the Twilio client object to ensure that it the API call is never made from your tests.

danldb:
Hi Loris - thanks for the submission!

You seem to have a good grasp of the problem and have successfully figured out a working solution. 

However, you could make improvements in the planning stages - I would recommend reading through the brief and underlining all the nouns, then identifying all the potential classes. Diligent use of CRC cards will help you to spot oddities in the responsibilities (i.e. the order being responsible for sending a text message).

Are you using TDD to write your tests first? If you were correctly focussing on red-green-refactor, and always using only the simplest possible way to make a test pass, you would have spotted that your test coverage was not sufficient.

I hope that helps - let us know if you have any questions.

PR #22, opened 2015-02-16
Takeaway Challenge

tommasobratto:
Unfortunately, I couldn't verify my number with Twilio and the challenge is still rough in design and code... maybe I also needed more tests

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,13 @@
+class Takeaway
+
+  attr_reader :menu
+
+  def initialize
+    @menu = Hash.new
+    menu[:dish]
+  end
+
+  def menu_update(dish)
+    menu[:dish] = dish.price
+  end 
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -2,7 +2,7 @@ require 'twilio-ruby'
 
 class Takeaway
 
-  attr_accessor :menu, :dishes_selected, :total_price
+  attr_accessor :menu, :dishes_selected, :total_price, :dishes
 
   def initialize
     @menu = Hash.new
@@ -38,5 +38,9 @@ class Takeaway
     @total_price = price.to_i * quantity.to_i
     raise 'Total price not correct' if @total_price != price.to_i * quantity.to_i
     puts total_price.to_s + '£'
+  end
+
+  def interactive_menu
+
   end
 end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -2,7 +2,7 @@ require 'twilio-ruby'
 
 class Takeaway
 
-  attr_accessor :menu, :dishes_selected, :total_price, :dishes_available
+  attr_accessor :menu, :dishes_selected, :total_price, :total_price_array, :dishes_available, :dish_price
 
   def initialize
     @menu = Hash.new
@@ -16,6 +16,7 @@ class Takeaway
   def order
     puts self.menu
     puts "To begin the selection, press enter. To cancel the action, type 'exit'."
+    @total_price_array = []
 
     while gets.chomp != 'exit' 
       puts "Please choose a dish."
@@ -26,20 +27,28 @@ class Takeaway
       quantity = gets.chomp
 
       puts "Please press enter to select another dish or type 'exit'."
+
+      dish_price_calculator(dish, quantity)
     end
 
     puts "Here is your order."
     puts dishes_selected
-    total_price_calculator(dish, quantity)
-    send_reply_message
+    total_price_calculator
   end
 
-  def total_price_calculator(dish, quantity)
+  def dish_price_calculator(dish, quantity)
     price = menu["#{dish}"]
-    @total_price = price.to_i * quantity.to_i
-    raise 'Total price not correct' if @total_price != price.to_i * quantity.to_i
-    puts total_price.to_s + '£'
+    dish_price = price.to_i * quantity.to_i 
+    total_price_array << dish_price
   end
+
+  def total_price_calculator
+    @total_price = total_price_array.inject {|result, i| result + i}
+    p total_price.to_s  + '£'
+    raise 'Total price is wrong' if total_price != total_price_array.inject {|result, i| result + i}
+  end
+
+  # 'send_reply_message' method not implemented, it should have been part of the 'order' method
 
   def send_reply_message
   account_sid = ''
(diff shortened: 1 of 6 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,65 @@
+require 'twilio-ruby'
+
+class Takeaway
+
+  attr_accessor :menu, :dishes_selected, :total_price, :total_price_array, :dishes_available, :dish_price
+
```
danldb: Many instance variables, most are necessary. For example, total_price should just be a method rather than an instance variable.

lib/takeaway.rb
```diff
@@ -0,0 +1,65 @@
...
+      puts "Please press enter to select another dish or type 'exit'."
+
+      dish_price_calculator(dish, quantity)
+    end
+
+    puts "Here is your order."
+    puts dishes_selected
+    total_price_calculator
+  end
+
```
danldb: Avoid puts. The brief deliberately does not include an interface, so all the puts statements are just creating spam in the terminal. Using gets.chomp means that it is now very hard to adapt your code to a different medium such as the web.

lib/takeaway.rb
```diff
@@ -0,0 +1,65 @@
...
+class Takeaway
+
+  attr_accessor :menu, :dishes_selected, :total_price, :total_price_array, :dishes_available, :dish_price
+
+  def initialize
+    @menu = Hash.new
+    @dishes_selected = []
+  end
+
```
danldb: Prefer {} to Hash.new

lib/takeaway.rb
```diff
@@ -0,0 +1,65 @@
...
+  @client = Twilio::REST::Client.new account_sid, auth_token
+
+  @client.messages.create(
+  from: '',
+  to: '',
+  body: 'Your order has been accepted and is already on the way. Thanks for choosing Takeaway.rb.'
+)
+
+  end
+end
```
danldb: Remember Single Responsibility Principle. The Takeaway has too many responsibilities. It is not logical that the takeaway is responsible for the customers telephone number. Consider adding classes for Customer, Messenger, Menu, Dish.

lib/takeaway.rb
```diff
@@ -0,0 +1,65 @@
...
+    p total_price.to_s  + '£'
+    raise 'Total price is wrong' if total_price != total_price_array.inject {|result, i| result + i}
+  end
+
+  # 'send_reply_message' method not implemented, it should have been part of the 'order' method
+
+  def send_reply_message
+  account_sid = ''
+  auth_token = ''
+  @client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: Tokens should be stored in the environment variables. Have a look at the dotenv gem for an easy way to do this.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    expect(takeaway.menu).to eq({'pizza' => '4£'})
+  end 
+
+  it "should be able to show the dishes selected and the total price" do
+    takeaway.menu_update(pizza)
+    takeaway.menu_update(pasta)
+    takeaway.order  
+    expect(takeaway.dishes_selected).to eq(['pizza', 'pasta'])
+    expect(takeaway.total_price).to eq(8)
+  end
```
danldb: 2 expectations - this should be 2 separate tests.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  it "should be able to send a text message when the order is placed" do
+
+  end
```
danldb: No test for the messenger? You should stub out the Twilio client to ensure that you can fully test all your own code, but that an SMS is never sent when you are running your tests,

takeaway_script.rb
```diff
@@ -0,0 +1,23 @@
+require './lib/takeaway.rb'
```
danldb: unnecessary script

lib/takeaway.rb
tommasobratto: Ok, but wouldn't using more classes potentially create more dependencies? For example if I needed to make various objects interact, such as the customer selecting the dish and reporting its quantity to the takeaway, wouldn't  I need to use more than one object as arguments of an hypothetical customer.order method? (like customer.order(dish, takeaway))

lib/takeaway.rb
tommasobratto: I'm not sure I understand what tokens are. I did a quick google search on them regarding Ruby but I didn't come up with anything.

lib/takeaway.rb
tommasobratto: OK, so I should be relying on actual methods between objects for example to make the order, then assigning the "user interaction" responsibilities to the web framework part?

danldb:
Hey Tommaso - thanks for your submission!

You have clearly got a good understanding of the brief and have solved a lot of the problems that were presented. You have a good grasp of the Ruby language, and demonstrated that you can write clean, concise code.

I would focus in Object Oriented Design - a good way to approach this is to have a look at the brief and underline all the nouns, then assess each of those nouns as potential classes. Once you have your list of potentials, write CRC cards for each of them, do further clarify each classes responsibilities. You will find that your CRC cards become a basis for writing your tests.

Be careful to follow the brief - you spent a lot of time building the wrong thing - there was no call for an interface, and by riting your code around gets.chomp and puts, you have limited its use to a single environment - a unix terminal.

You have a good grasp of testing, and I was happy to see you using doubles in a meaningful and careful manner.

Let us know if you have any questions - I hope this helps!

spec/takeaway_spec.rb
tommasobratto: I couldn't test the messenger because I couldn't make Twilio accept my number during registration, so I couldn't make the send_message method work. I didn't write out the test because I was trying to figure out how to make the method work within the "order" method.

lib/takeaway.rb
danldb: Hey Tommaso,

I wouldn’t say that adding classes adds dependencies. 

Classes should be encapsulated so that they are not dependent on another class to function. That is one of the hardest skills to develop in OOP, but I am sure you will pick it up. 

I see you have sent a bunch of emails - maybe the best would be for us to have a 121 so that we can talk over your code?

Let me know when works for you.

Cheers,

___________________________

Dan Le Dosquet-Bergquist, Meta Maker

+44 753 900 8999 | www.makersacademy.com <http://www.makersacademy.com/>

GitHub <https://github.com/danldb>    |    Twitter <http://www.twitter.com/danldb>

tommasobratto: Ok, thank you for your clarification. I think it would be best to have a 121, maybe tomorrow?

PR #26, opened 2015-03-29
Richard Lynch Take Away Challenge

rjlynch:
I can't get travis to view the feature test. It passes in rspec fine but I have to pass the path to rspec first.


stephenlloyd:
:checkered_flag:

PR #27, opened 2015-03-29
takeaway_challenge

armi1189:


tansaku:
Some great work here, although my main cause for concern is that you implemented features that were not in the spec, and failed to implement one that was required.  Some more detailed notes:

* Dish class looking good
* Good stubbing of print and puts in menu class, but notice that menu is now locked to a particular UI, similarly for Order
* In Order notice the line "@restaurant.menu.list.keys.include?(name)" - this is reaching a bit too deeply into the Restaurant class - the Restaurant should define some method to allow this to happen - avoid reaching across like this
* Great to see you playing around with all the adding and removing functionality - but that wasn't in the specification ... and conversely you haven't implemented to total checking functionality that WAS in the specification
* Also I think the test in menu_feature_spec is somewhat vacuous in that it will pass with functionality removed. 
* restaurant --> avoid putting your secrets in your git file - use ENV
* watch out for naming methods 'send' --> this is special in Ruby and you shouldn't override it.

Implement that last bit of functionality (the error raising) and clean up and you'll be eligible for a badge :-)

:construction:

tansaku:
:checkered_flag: great work!

PR #28, opened 2015-03-29
take-away

SammyAbukmeil:
Wasn't able to implement Twilio, ran out of time trying to set it up. Will continue during the week.


lib/customer.rb
```diff
@@ -0,0 +1,23 @@
...
+  def add!(dishes, total_price)
+    @order = @order.merge(dishes)
+    @total_price += total_price
+
+    total_value = 0
+    dishes.each do |dish, quantity|
+      total_value += quantity * dishes_list[dish]
+    end
+  end
+end
```
MihaiLiviuCojocar: Single Responsibility Principle!
How many things does the customer do/know? He knows about the business logic of the order, calculates something, has a menu that it's not being used ... He should only be able to place an order, without knowing how this is implemented.

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  def dishes_names
+    dishes_list.keys
+  end
+
+  def dishes_prices
+    dishes_list.values
+  end
+end
```
MihaiLiviuCojocar: Maybe the menu should be a class and it shouldn't be hardcoded. Everything has to be reusable :)

spec/customer_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'customer'
+
+describe Customer do
+
+  let(:menu) { DishesList.new }
+  let(:customer) { Customer.new(menu) }
+
```
MihaiLiviuCojocar: The described class should be the only real object and everything else should be mocked.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'restaurant'
+
+describe Restaurant do
+
+  let(:menu) { DishesList.new }
+  let(:restaurant) { Restaurant.new(menu) }
+  let(:customer) { Customer.new(menu) }
+
```
MihaiLiviuCojocar: Doubles for menu and customer! :)

MihaiLiviuCojocar:
Try to follow the hints that I gave you! :)

MihaiLiviuCojocar:
:construction:

PR #29, opened 2015-03-29
Pull request

sanjsanj:


lib/send_text.rb
```diff
@@ -0,0 +1,27 @@
+require 'twilio-ruby'
+
+class MessageKlass
```
MihaiLiviuCojocar: "class Message" is enough :)

lib/send_text.rb
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    @client.account.messages.create(conf_msg)
+  end
+
+  def d_time
+    delivery_time = Time.now + 1 * 60 * 60
+    delivery_time.strftime('%H:%M')
+  end
+end
```
MihaiLiviuCojocar: I like that you wrapped the sms into it's own class. :+1:

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+    out << "Total = £#{total_price}"
+    out
+  end
+
+  def delivery_conf
+    fail 'Error' if @customer_order.empty? || !customer_number
+    MessageKlass.new.send_text @customer_number, @customer_order
+  end
+end
```
MihaiLiviuCojocar: This class deals with too many things. SRP! :)

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+    @customer_order.inject(out) do |accumulator, (item, qty)|
+      accumulator << "#{qty} x #{item} = £#{@menu[item] * qty}\n"
+    end
+    out << "Total = £#{total_price}"
+    out
+  end
+
+  def delivery_conf
+    fail 'Error' if @customer_order.empty? || !customer_number
+    MessageKlass.new.send_text @customer_number, @customer_order
```
MihaiLiviuCojocar: We have some coupling here, which can be replaced by injecting the dependency.

MihaiLiviuCojocar:
SRP! :)
:construction:

lib/menu.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def dish_prices
+    @dish_prices ||= []
+  end
+
+  def add dish
+    dishes << dish.name
+    dish_prices << dish.price
+  end
+
```
MihaiLiviuCojocar: If add() takes a dish object as an argument, that object disappears because we are storing just the names and the prices into some lists.

lib/order.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def add_line line_item
+    sub_total = line_item.sub_total
+    dish_name = line_item.dish
+    dish_qty = line_item.quantity
+    @grand_total += line_item.sub_total
+    @lines << "#{dish_qty} x #{dish_name} = £#{sub_total}\n"
+  end
+
```
MihaiLiviuCojocar: Same thing like in Menu class ... we should be dealing with objects and not just some strings.

lib/takeaway.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def initialize menu
+    @menu = menu
+  end
+
+  def customer name, number
+    @customer_name = name
+    @customer_number = number
+  end
+
```
MihaiLiviuCojocar: Same thing ... If we want the takeaway to have a customer, why don't use the customer object instead of strings. The customer class becomes redundant... We should be dealing only with objects and passing them around ...

MihaiLiviuCojocar:
I am going to put the checked flag, but this still needs work.
I don't see a good understanding of OOD yet. I recommend reading POODR by Sandi Metz... It helps a lot. :)
:checkered_flag:

PR #33, opened 2015-03-29
Submission

benny-kollek:


tansaku:
Great work getting this in.

I'm a little confused about the feature test, in that the operational one tests whether dishes can be added and removed from the menu, which is great, but it's not one of the features requested in the specification.

I'd very strongly advise getting feature tests wrapped around the features that ARE required first, before working on feature tests for optional extras like adding and deleting menu items, changing menu item names etc.

You have a redundant Customer class - I'd remove that completely - rather than have it hanging around on the off chance that you'll use it.

Your dish has attr_accessor which seems unecessary - please avoid attr_accessor where possible - I think attr_reader would be more than sufficient here.

Your restaurant has account_sid and auth_token hard coded in - please prefer to reference from ENV variables.

I also have the feeling that your code is not properly stubbed to avoid sending SMS messages when you are testing ...

:construction:

PR #34, opened 2015-03-29
Week 2 MA Takeaway Challenge

kanishkwalia:
Attempt at takeaway challenge.

Unable to get certain features of the challenge working such as sum of the total and SMS via Twilio.


spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'menu'
+
+describe Menu do
+	let(:menu) {Menu.new(["Dish1", "Dish2"])}
+	let(:Dish1) {Double :dish, name: "Dish1", price: 10}
```
tansaku: prefer lower case for both :Dish1 and Double here

tansaku:
Some good work here, although we really must get your tabbing issue sorted.  

Unfortunately the absence of any feature tests means that it's difficult for me to see how you plan to have your classes interact.    For example, you have a standalone Customer class that looks a simpler clone of menu - and it's unclear how it will interact with any of the other classes.

You've got a Send_sms module but its not mixed in with any of the other classes.  I'd recommend feature tests to try and architect what the user experience is, and use that to drive how the classes interact.

:construction:

PR #37, opened 2015-03-29
My attempt at the Takeaway challenge

GBouffard:
I struggle with some of the basics. I have quite a few things, but I don't know if I wrote it correctly. Am I making the right tests? I get the idea on how the takeaway can all work together and see it in my mind, but I don't have the programming/ruby language knowledge to make it happen yet. I really think each student should get a personal feedback on their challenges on what they did well, what looks off, what they should do next time, and especially their key mistakes, so he/she can correct them early and improve their skills, especially beginners like me. I still hope that what I've done so far is ok. It's passing the tests I submitted.


tansaku:
Several lines are too long --> break them up by converting the {} to do ... end

No feature test to guide me how to use your app :-(

inventory_spec, del_spec --> could replace let with 'subject'

inventory --> refer to instance variable by their readers

```ruby
  def ready_to_checkout
    @ready_to_checkout = true
  end
```

^^^ potentially confusing --> looks like a reader

checkout - avoid attr_accessor where possible

So it looks like you made some good progress here, but it's difficult for me to see how you intend these classes to interact with each other in the absence of a feature test.  Were you testing away in IRB as you went along?

You've got extra functionality here such as having a certain quantity of certain dishes that can be used up, but that wasn't explicitly asked for in the project specification.  It's great to add extra functionality to experiment, but you really want to make sure that you have completed the functionality that is required by the specification BEFORE you start adding anything else.  That extra functionality might turn out to be really really useful, but it will also get in your way trying to make the REQUIRED functionality working. It's one more thing to go wrong.  Programs get complex really really quickly, so try to only implement the absolute minimum you need in order to get the job done.  Make sure that that works, and only then come back to implement extra functionality (assuming the client wants it).

PR #41, opened 2015-03-30
MVp1 reached

RBGeomaticsRob:


RBGeomaticsRob:
Travis is failing with the message 'Could not find coveralls-0.7.12 in any of the sources', I have tried this on previous commit from last night that were passing then and are now failing, despite no changes being made to them.

RBGeomaticsRob:
Coveralls required me to take the keys.rb file out of the .gitignore to get a pass mark, the keys file is not tested hence the 96%, would be interested in finding out the way we should go about hiding keys from version control platforms.

PR #44, opened 2015-03-30
Takeaway Challenge 

Stacca:


tansaku:
thanks for submitting what you did :-) 

Some general notes:
* menu_spec --> could be using subject, and we're not really testing menu
* menu --> the dishes method is confused about whether it is a setter or a getter, and in general we don't want to lock a class to a particular UI, which you have my using puts here
* order_spec --> could use subject, also you are stubbing out the functionality that you actually want to be testing with allow

Let's have a chat about how you could make more progress.

:construction:

PR #46, opened 2015-03-30
incomplete submission

jaymuk:


tansaku:
Thanks for submitting where you got to, much appreciated. 

Some general comments are:
1. fix all the hound/rubocop issues
2. Got some text being printed out in rspec - that should be stubbed
3. menu --> needs to avoid use of attr_accessor where possible
4. absence of feature tests means I don't know how I'm supposed to using your system

It's difficult to say much more given how far you got ... my main recommendation would be to drive from feature tests that are clearly related to the user stories

:construction:

PR #53, opened 2015-05-09
Takeaway-challenge submission

chidumaga:


chidumaga:
The HoundCI violations are with regards to the verification message (specified in the Readme) being too long.

lib/customer.rb
```diff
@@ -0,0 +1,20 @@
+require_relative '../lib/menu'
+
+class Customer
+
+  attr_accessor :order_total
```
ptolemybarnes: You declare an attribute reader / writer. Use it!!!

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
+class Menu
+
+  attr_reader :display_menu
+
+  def initialize
+    @display_menu = [["Pounded Yam", 15],
```
ptolemybarnes: Good, although I really think a hash would be appropriate here instead of an array as then you wouldn't have those magic numbers in the customer class.

lib/twilio_response.rb
```diff
@@ -0,0 +1,23 @@
+require 'twilio-ruby'
+
+class TwilioResponse
```
ptolemybarnes: Nice use of a wrapper class.

PR #54, opened 2015-05-09
Take-away challenge

dwatson62:


lib/customer.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def initialize
+    @order = []
+    @total_cost = 0
+  end
+
+  def see_all_dishes(restaurant)
+    restaurant.dishes.each do |item, price|
+      # price = sprintf('%.2f', price)
+      p "#{item} - £#{sprintf('%.2f', price)}"
```
ptolemybarnes: By using "p" you're tying the presentation of your system to the terminal. What if I wanted to have this text displayed on a web page?

lib/restaurant.rb
```diff
@@ -0,0 +1,22 @@
...
+      :Rice => 1.50,
+      :Naan => 2.00,
+      :Vindaloo => 5.00
+    }
+  end
+
+  def place_order(customer)
+    fail 'Nothing ordered yet' if customer.order.empty?
+    customer.current_order
+    require './send_sms.rb'
```
ptolemybarnes: What you need here is a wrapper class for the Twilio gem. Think of a wrapper class as like a phone, and the twilio gem as the SIM card.

spec/customer_feature_spec.rb
```diff
@@ -0,0 +1,77 @@
+require 'customer'
+require 'restaurant'
+
+feature 'Customer wants to see' do
```
ptolemybarnes: Tests look great to me!

PR #55, opened 2015-05-09
First user test

curlygirly:


lib/customer.rb
```diff
@@ -0,0 +1,43 @@
...
+    order =  []
+  end
+
+  def checks_menu(restaurant)
+    restaurant.menu.each do |dish, price|
+      puts "#{dish} costs #{price}"
+    end
+  end
+
+  def selects_item(restaurant, dish, quantity)
```
ptolemybarnes: Nice method, although why make that choice = dish variable assignment?

lib/customer.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'restaurant.rb'
+
+class Customer
+
+  attr_accessor :order
+
+  def intialize
+    order =  []
```
ptolemybarnes: Should this be an instance variable?

lib/restaurant.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'customer'
+
+class Restaurant
+  attr_accessor = :menu
+
+  def initialize
+    @menu = { 'Burger' => '£10',
```
ptolemybarnes: Good, although you're hardcoding your list of dishes into the Restaurant class. Can you think of another way of doing this?

lib/customer.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def checks_menu(restaurant)
+    restaurant.menu.each do |dish, price|
+      puts "#{dish} costs #{price}"
```
ptolemybarnes: By using puts your tying the presentation of your takeaway system to terminal. What if I wanted this string to be printed onto a web page?

spec/features/feature_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  let(:customer) { Customer.new }
+  let(:restaurant) { Restaurant.new }
+
+  scenario 'customer can check a menu with prices' do
+    expect(customer.checks_menu(restaurant)).to eq restaurant.menu
+  end
+
+
+  scenario 'customer can select items off the menu' do
+    customer.selects_item(restaurant,'Burger', 1)
```
ptolemybarnes: magic numbers!

ptolemybarnes:
:checkered_flag:

ptolemybarnes:
:construction:

PR #57, opened 2015-05-09
takeaway-challenge

andygout:


lib/checkout.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'order'
+require_relative 'till'
+
+class Checkout
+
+  include Till
```
ptolemybarnes: as we discussed, modules usually have the names of abilities. So "Tilllike", weird though it sounds, might be appropriate here.

lib/checkout.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  attr_reader :order
+
+  def initialize order
+    @order = order
+  end
+
+  def no_dishes_ordered?
+    @order.dishes.empty?
```
ptolemybarnes: Use attribute readers!!!

lib/checkout.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def send_msg
+    if no_dishes_ordered?
+      fail 'No dishes have been added to the order'
+    else
+      require_relative 'twilio.rb'
```
ptolemybarnes: Interesting inline use of require... I have never seen that before. It's better to have a wrapper class for a gem such as twilio (think of the wrapper class as a phone, and the twilio gem is the antennae.)

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
+class Menu
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = { 'Korma': 4.95, 'Red Curry': 5.95, 'Green Curry': 6.95 }
+  end
+
+  def show_menu
+    @dishes.each { |k, v| "#{k}: £#{v}" }
```
ptolemybarnes: Please use descriptive argument names.

lib/order.rb
```diff
@@ -0,0 +1,17 @@
...
+  include Till
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = []
+  end
+
+  def add_dish dish, quantity = 1
+    quantity.times { @dishes << dish }
```
ptolemybarnes: use attribute readers!!!

PR #58, opened 2015-05-10
Domain modelling in README

DanielJohnston:
No actual code yet!


DanielJohnston:
The desired code is all there. The feature tests work on my box, but messaging isn't properly injected, nor is it properly stubbed, so it fails tests elsewhere. We went through this on Friday morning but I haven't had chance to review the video yet. Other than that, tests should work. There's a fair bit of room for refactoring to remove duplication, primarily in the test specs.

lib/message.rb
```diff
@@ -0,0 +1,25 @@
...
+  # Note that I'm using test credentials for sid, auth and from_no
+  # As far as I can tell, there's no way of using live SMS without paying
+
+  def initialize
+    @account_sid = ENV['TWILIO_ACCOUNT_SID']
+    @auth_token = ENV['TWILIO_AUTH_TOKEN']
+    @sender = ENV['MY_FROM_NUMBER']
+  end
+
+  def send(customer, time)
```
silvabox: This could be refactored into 4 methods: this one; one to calculate the time, one to create the Twilio client (maybe a reader?) and one to create the Twilio message

lib/order.rb
```diff
@@ -0,0 +1,50 @@
+class Order
```
silvabox: This is a common problem in OOD - which classes should be responsible for what?  If you end up with responsibilities in the wrong places, you can have spaghetti code and lots of interdependent classes.

Should order also be responsible for calculating the price and placing itself?  Probably not.  You place the order with the takeaway.  The order is just the list of items and quantities.  Where you are using an array or hash to store a collection of related properties, you probably need a class (eg. `OrderLine`

lib/takeaway.rb
```diff
@@ -0,0 +1,23 @@
...
+    # Not enforced, but menu hash structure is {"dish name" => cost}
+    @menu = {}
+    @name = name
+  end
+
+  def add_menu_items(items)
+    @menu.merge!(items)
+  end
+
+  def list_menu
```
silvabox: It's dangerous to expose internal implementations like this - I could hack around with the hash and break the program.  It's less of an argument if you are not sanitizing the input in `add_menu_items`, but it's still something to consider.  You could return a copy of the hash instead to prevent accidental changes.

spec/customer_spec.rb
```diff
@@ -0,0 +1,13 @@
+@tel_no = ENV['MY_MOBILE_NUMBER']
+
+require 'customer'
+
+describe Customer do
+  it 'can create and retrieve details of a customer' do
```
silvabox: This should be 2 tests - 1 for each data item

spec/features/customer_places_order_feature_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+    # Check the menu is what we expect
+    expect(takeaway.list_menu).to eq(default_menu)
+
+  end
+
+  scenario 'selects items and quantity to order' do
+    @tel_no = ENV['MY_MOBILE_NUMBER']
+    # Set up a default menu
+    default_menu = {}
+    default_menu["Pie and chips"] = 450
```
silvabox: Move this set up code into a before block so it's not cluttering your tests

silvabox:
Good work Daniel, although your Travis builds are failing - have you checked why?  Hopefully it's just a rogue test!

Well done. :checkered_flag:

DanielJohnston:
Travis failure is likely because the Twilio API / Message class isn't properly stubbed, so it falls over when it tries to pull in the environmental variables for API keys etc. Thanks for the feedback!

PR #59, opened 2015-05-10
takeaway-challenge

MollieS:


silvabox:
Hi Mollie, thanks for submitting your takeaway challenge!  I can sort of see what you were trying to do and maybe what you were thinking, but I feel like I need to go through it with you properly as I don't think I can explain my feedback properly in comments.  Should we have a 1-2-1?

:construction:

PR #61, opened 2015-05-10
Takeaway challenge (includes interactive menu)

rodcul:
$ ruby lib/menu.rb

--------------------------------
1. Thai tofu............... 9.5
2. Grilled octopus......... 11
3. Chipotle bbq chicken.... 12.5
4. Harvest................. 8
5. Mediterranean spinach... 9

--------------------------------

7. Show order total so far
8. Confirm order, send SMS
9. Cancel order


lib/dish.rb
```diff
@@ -0,0 +1,28 @@
...
+  def initialize
+    @dishes = []
+
+    preload_dishes
+  end
+
+  def print
+    return 'Puts no students at this time' if @dishes.empty?
+    i = 1
+    @dishes.each do |dish|
```
ptolemybarnes: You have declared an attr_reader for dishes. Why not use it here?

lib/dish.rb
```diff
@@ -0,0 +1,28 @@
...
+      "#{i}. #{dish[:name].capitalize.ljust(24, '.')} #{dish[:price]}"
```
ptolemybarnes: What is this #each block supposed to do? Remember that #each returns the array it is called on.

lib/dish.rb
```diff
@@ -0,0 +1,28 @@
...
+      i += 1
+    end
+end
+
+  private
+
+  def preload_dishes
+    @dishes << { name: 'thai tofu', price: 9.5 }
```
ptolemybarnes: Same again, access the @dishes instance variable via its reader.

lib/dish.rb
```diff
@@ -0,0 +1,28 @@
...
+    i = 1
+    @dishes.each do |dish|
+      "#{i}. #{dish[:name].capitalize.ljust(24, '.')} #{dish[:price]}"
+      i += 1
+    end
+end
+
+  private
+
+  def preload_dishes
```
ptolemybarnes: You've hardcoded a set of dishes into this class. Is there another way of loading the menu's dishes?

lib/order.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'dish'
+require_relative 'twilio'
+
+class Order
+  attr_reader :dishes, :order
+
+  def initialize
+    @dishes = Dish.new
```
ptolemybarnes: So are we dealing with plural dishes here, as the name of the instance variable would suggest, or a singular dish, as the classname would suggest?

lib/twilio.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
+
+class TwilioSMS
```
ptolemybarnes: Good job. You wrapped the Twilio object in a wrapper class.

spec/features/takeaway_features_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  scenario 'Add multiple items to an order' do
+    myorder = Order.new
+    myorder.add(1, 2)
+    expect(myorder.order[0][3]).to eq 19.0
+  end
+
+  scenario 'calculate total for multiple items' do
+    myorder = Order.new
+    myorder.add(1, 2)
```
ptolemybarnes: Magic numbers!!!

spec/features/takeaway_features_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    expect(myorder.total).to eq 38.0
+  end
+
+  xscenario 'send SMS when order is complete' do
+    service = double :twilio
+    allow(service).to receive(:new)
+    myorder = Order.new
+    myorder.add(1, 2)
+    expect(myorder.notify(service)).to receive(:new)
```
ptolemybarnes: Remember that the expect(...).to receive syntax is used for declaring expectations about what will happen in the future, i.e as a result of the lines of code following on from the expectation. You've not got any lines of code following on from this expectation so I think you're misusing it a bit.

PR #62, opened 2015-05-11
customer can see menu and place order. twillio not implemented

bagolol:


spec/customer_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  it 'can see the menu' do
+    customer = Customer.new { takeaway }
+    expect(customer.see_menu).to eq({'red curry' => 5, 'green curry' => 5})
+  end
+
+  it 'can place an order' do
+    customer = Customer.new { takeaway }
+    customer.place_order(1, "red curry")
+    allow(takeaway).to receive(:order).and_return ([{'red curry' => 5}])
+    expect(customer.takeaway.order).to eq ([{'red curry' => 5}])
```
silvabox: this is a vacuous test - you are simply testing the stub here - not any behaviour of `Customer`

lib/customer.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'takeaway'
+
+class Customer
+
+  attr_accessor :takeaway
+
+  def initialize
+    @takeaway = yield
```
silvabox: Good attempt to use dependency injection here - but as you have an `attr_accessor` for `takeaway`, it is unnecessary.  If you do want to specify the `takeaway` when `Customer` is created, you don't need to use a block here, just pass `takeaway` as an argument to `initialize`

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'takeaway'
+require 'menu'
+
+describe Takeaway do
+
+  it 'has a menu' do
+    takeaway = Takeaway.new
```
silvabox: you don't need `takeaway = Takeaway.new` - use `subject` instead

spec/customer_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  it 'can receive an order confirmation' do
+    customer = Customer.new { takeaway }
+    time = Time.new
+    customer.place_order(1, "red curry")
+    allow(takeaway).to receive(:register_order)
```
silvabox: I don't understand this test - why all these `allow`s?

silvabox:
Not a bad start Rocco!  I'm a bit confused as to whether the menu is defined by `Menu` or `Takeaway`.  You're probably not far off getting the Twilio stuff implemented.  You've captured it in a file, but think about maybe wrapping that in a class.
:construction:

spec/feature/feature_test_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+feature 'as a customer I want to order from a takeaway' do
+  scenario 'a menu is available' do
+    customer = Customer.new { Takeaway.new }
+    expect(customer.see_menu).to eq ({'red curry' => 5, 'green curry' => 5})
+  end
+
+  scenario 'an order can be placed' do
+    customer = Customer.new { Takeaway.new }
+    customer.see_menu
+    customer.place_order(2, "red curry")
```
ptolemybarnes: magic number. check out: https://robots.thoughtbot.com/ruby-2-keyword-arguments

spec/feature/feature_test_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    expect(customer.takeaway.order).to eq ( [{"red curry"=>5}, {"red curry"=>5}])
+  end
+
+  scenario 'an order is placed and the customer receives the total amount' do
+    customer = Customer.new { Takeaway.new }
+    time = Time.new
+    customer.see_menu
+    customer.place_order(1, "red curry")
```
ptolemybarnes: magic number. check out: https://robots.thoughtbot.com/ruby-2-keyword-arguments

PR #63, opened 2015-05-11
Just started

smarbaf:


spec/features/main_takewaay_feature_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'customer wants to order a meal' do
+  scenario 'customer wants to see a list of dishes and prices' do
+    quiche = Dish.new("food", "price")
```
ptolemybarnes: I can see where you're going with this. But, assuming you want this Dish to be a quiche, why not Dish.new("quiche", 3). The second argument being the price.

spec/features/main_takewaay_feature_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+    soup = Dish.new("food", "price")
+    sandwich = Dish.new("food", "price")
+    fish = Dish.new("food", "price")
+    steak = Dish.new("food", "price")
+    expect(food.count(dish)).to eq(5)
```
ptolemybarnes: Remember that variables should always point towards objects. If there is a variable that doesn't point towards an object, the Ruby interpreter will give an error. 
So there are two variables here, "food" and "dish", that don't seem to point towards any object. When we're writing these high-level feature tests though, we get to decide what objects we want to point them to. What object do you want "food" to point to?

PR #64, opened 2015-05-11
Takeaway challenge

ashleigh090990:
I know there's a lot to be desired for this... But this is what I have so far!  I've not been able to test the SMS part with Twilio. Good luck trying to understand this!


lib/checkout.rb
```diff
@@ -0,0 +1,6 @@
+module Checkout
```
ptolemybarnes: Nice use of a module, though perhaps unnecessary at this stage as it's used in only one class?

lib/order.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'checkout'
+
+class Order
```
ptolemybarnes: Excellent! You read / wrote to instance variables via the class' public interface.

lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  include Checkout
+
+  def initialize
+  	@menu_list = {"Bottle of Water" => 3.08, "Disney Themed Napkins (x5)" => 50, "Carbonara" => 0.4, "Ice cream" => 0.48, "Happiness" => 0}
+  	@order_list = []
+  end
+
+  def print_menu
+  	menu_list.each do |dish_key, dish_price|
```
ptolemybarnes: Indentation >_<

lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+      "#{dish_key}: £#{dish_price}"
```
ptolemybarnes: Not sure what this method is supposed to achieve... Remember that the return value of each is merely the array that it is called on. Should you have used #map?

lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+  	end
+  end
+
+  def add_dish(quantity, dish_key)
+  	quantity.times do
+  	  order_list << menu_list.select do |k|
+  	  	k == dish_key
```
ptolemybarnes: Nice, though might be worth extracting the #select part of this out to another method.

lib/twilio_send_sms.rb
```diff
@@ -0,0 +1,13 @@
+require 'time'
+require 'rubygems'
+require 'twilio-ruby'
+
```
ptolemybarnes: Seems like you're using the Twilio gem as intended, though it would have been good to inject it into a "wrapper" class (I made a "Phone" class when I did this. In that sense, the actual Twilio object (Twilio::REST::Client.new) is like the antennae of the phone class).

spec/feature/feature_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'capybara/rspec'
+require 'order'
+
+feature 'A customer wants to order takeaway' do
+  scenario 'Customer is able to view list of dishes from the menu' do
+  	order = Order.new
+  	expect(order.print_menu).to be_a_kind_of(Hash)
```
ptolemybarnes: Avoid tests that do type-checking. The reason for this is that if, in future, you change the implementation of Order such that the menu is no longer a hash, but it still does the same job, this test will break.

spec/order_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'order'
+
+describe Order do
+  it 'contains a list of all of it\'s dishes and is able to print it' do
+    expect(subject.print_menu).to be_a_kind_of(Hash)
```
ptolemybarnes: Again, avoid type-checking.

spec/order_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  it 'Dishes can be added to the order and viewed' do
+  	subject.add_dish 1, "Bottle of Water"
```
ptolemybarnes: https://www.youtube.com/watch?v=s29S51KSbiQ

spec/order_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  it 'Multiple amounts of the same dish can be added to order' do
+  	subject.add_dish 4, "Bottle of Water"
+  	expect(subject.view_order).to eq [{"Bottle of Water" => 3.08}, {"Bottle of Water" => 3.08}, {"Bottle of Water" => 3.08}, {"Bottle of Water" => 3.08}]
+  end
+
+  it '' do
+  end
+
+end
```
ptolemybarnes: No newline at end of file! (I don't know why that's a problem, but I've been told it is).

PR #66, opened 2015-05-11
Approach 2 - Takeaway challenge

AlexHandy1:


lib/customer.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'takeaway'
+
+class Customer
```
ptolemybarnes: What is the responsibility of the Customer class?

lib/dishes.rb
```diff
@@ -0,0 +1,19 @@
+class Dishes
+
+  attr_accessor :dish_list
+
+  def initialize
```
ptolemybarnes: You've hardcoded your hash of dishes into the Dish class. Is there a better way of making that data available?

lib/dishes.rb
```diff
@@ -0,0 +1,19 @@
...
+    @dish_list = {
+      :FishAndChips => 5,
+      :MeatAndTwoVeg => 7
+    }
+  end
+
+  def new_dish item, price
+    @dish_list[item] = price
```
ptolemybarnes: Access / set instance variables via attribute readers / writers

lib/message_handler.rb
```diff
@@ -0,0 +1,20 @@
+require 'twilio-ruby'
+
+class Message_Handler
```
ptolemybarnes: Nice use of a wrapper class.

lib/takeaway.rb
```diff
@@ -0,0 +1,59 @@
...
+    current_order.delete_at(item_num-1)
+  end
+
+  def confirm_order
+    fail "No items in basket" if order_obj.order_empty?
+    message_handler.send_order_message
+  end
+
+  def deliver_order
+    order_obj.delivered = true
```
ptolemybarnes: You're reaching into an object to set the value of an instance variable, breaking encapsulation by doing so. Consider a method analogous to #report_broken in Boris Bikes

lib/order.rb
```diff
@@ -0,0 +1,30 @@
+require_relative 'menu'
+
+class Order
```
ptolemybarnes: Nice short methods, good. I am a little bewildered that you've declared setters / getters for your instance variables but then accessed those instance variables directly. :-S

lib/takeaway.rb
```diff
@@ -0,0 +1,59 @@
+require_relative 'order'
+require_relative 'menu'
+require_relative 'message_handler'
+
+class Takeaway
```
ptolemybarnes: So this class handles orders? It's always tempting to name classes after some real-world object as you've done with "Takeaway" but it's generally better to go with something more descriptive: "OrderHandler", perhaps.

lib/takeaway.rb
```diff
@@ -0,0 +1,59 @@
...
+      current_order << main_menu.select {|dish, price| dish == item }
+    end
+  end
+
+  def check_order
+    puts "You have ordered..."
+    item_num = 1
+    current_order.each do |dish, price|
+      dish.each do |k,v|
+        puts "Item#{item_num}: #{k} that costs £#{v}"
```
ptolemybarnes: Remember, your user is probably not going to be using terminal. When you use "puts" your tying the presentation of your system to terminal. Imagine we wanted this text to instead be printed on a web page. In that case it would suffice to have the text returned as a single string or something.

lib/welcome.rb
```diff
@@ -0,0 +1,20 @@
+require_relative 'customer'
+require_relative 'takeaway'
+require_relative 'message_handler'
+require_relative 'menu'
+require_relative 'order'
+require_relative 'dishes'
+
+#>> IRB UI will be welcome.customer.check_menu (under hood calls takeaway)
+class Welcome
```
ptolemybarnes: Again, this class is only necessary if we're tying the presentation of the system to terminal. (this will make more sense after this week).

spec/welcome_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+describe Welcome do
+  let (:customer) { double :customer }
+  let (:customerClass) { double :Customer, new: customer}
+  let (:takeaway) { double :takeaway}
+  let (:takeawayClass) { double :Takeaway, new: takeaway}
+  let(:menuClass) {double :Menu, new: menu}
+  let(:orderClass) {double :Order, new: order}
+  let(:message_handlerClass) {double :Message_Handler, new: message_handler}
+
+  subject {Welcome.new customerClass, takeawayClass, menuClass, orderClass, message_handlerClass}
```
ptolemybarnes: Interesting. Does this work?

spec/welcome_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'welcome'
+
+describe Welcome do
```
ptolemybarnes: What you're doing here is creating a "God" class.

spec/dishes_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'dishes'
+
+describe Dishes do
+  it "has a list of a dishes" do
+    expect(subject.dish_list).to eq :FishAndChips => 5, :MeatAndTwoVeg => 7
```
ptolemybarnes: You can use the syntactical sugar for hashes here: "FishAndChips: 5, MeatAndTwoVeg: 7"

spec/order_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'takeaway'
+
+describe Order do
+
+  it "should have an empty container of items" do
+    expect(subject.orders).to eq []
+  end
+
+  it "should have a status of not delivered" do
+    expect(subject.delivered).to be_falsey
```
ptolemybarnes: Nice use of the be_falsey syntax

PR #69, opened 2015-06-21
Partially completed takeaway challenge

KateWilkinson:


silvabox:
Good start again Kate, but your takeaway has too many responsibilities as discussed.    Think again about how the order and the takeaway interact.

:construction:

PR #70, opened 2015-06-21
takeaway-challenge 21 june 2015

katsuraku:


silvabox:
Great work Kirsten - only a few steps away from having a really clean solution and definitely one of the best I've seen.  Your tests are well written and your code is clean.  Just look at whether you need `@check_price`.

:checkered_flag:

PR #71, opened 2015-06-21
Mostly complete - problems testing twilio

kfcrobbie:


lib/customer.rb
```diff
@@ -0,0 +1,68 @@
+require_relative 'menu'
+require 'twilio-ruby'
+require 'rubygems'
+
+class Customer
+  attr_reader :menu, :order
+  def initialize (menu)
```
silvabox: Avoid spaces between method names and the opening parethesis

lib/customer.rb
```diff
@@ -0,0 +1,68 @@
+require_relative 'menu'
+require 'twilio-ruby'
+require 'rubygems'
+
+class Customer
+  attr_reader :menu, :order
```
silvabox: These `attr_reader`s expose too much of the class's internal implementation

lib/customer.rb
```diff
@@ -0,0 +1,68 @@
...
+  def initialize (menu)
+    @order = {}
+    @menu = menu.show_menu
+    @total_price = 0
+  end
+
+  def add_item (item, number=1)
+    raise 'invalid number' if number %1 !=0
+    if pizza_already_in_order?(item)
```
silvabox: Nice refactor - it's good to extract this logic into another method.

lib/customer.rb
```diff
@@ -0,0 +1,68 @@
...
+      current_total = order[item]
+      order[item] = current_total.to_i+ menu[item]*number
```
silvabox: Conversely, this should probably be extracted into another method too!

lib/customer.rb
```diff
@@ -0,0 +1,68 @@
...
+      @order.delete(item)
+    end
+  end
+
+  def pizza_already_in_order? (item)
+    @order.keys.include?(item)
+  end
+
+  def total_price
+    @total_price=0
```
silvabox: Is `@total_price` required?  `total_price` is always calculated on the fly, so there is no need to persist its value in an instance variable

spec/customer_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+                                       barbeque: 8,
+                                       spicy: 9,
+                                       meaty: 11,
+                                       vegetarian: 6,
+                                       hawaiian: 8} }
+
+  subject {Customer.new(menu)}
+
+  it 'allows a customer to check the menu' do
+    expect(subject.menu[:cheese]).to eq 7
```
silvabox: All your really testing here is the behaviour of the `menu` hash.

lib/customer.rb
```diff
@@ -0,0 +1,68 @@
...
+    @total_price
+  end
+
+  def place_order payment_amount
+    raise 'That is not the correct payment amount!' if payment_amount !=total_price
+    print 'Thank you for your order'
+    send_text_confirmation
+  end
+
+  def send_text_confirmation
```
silvabox: Text handling should be extracted into another class

silvabox:
Hi Rob, thanks for submitting your takeaway challenge.  You've written some good code in places, but your objects are too reliant on low-level arrays and hashes.  Your `Customer` class has too many responsibilities - perhaps you need classes for the takeaway and message handling?  Also, your `Menu` class does nothing other than wrap a hash - so you might equally have just used a hash instead.

:construction:

PR #72, opened 2015-06-21
most of a solution

Harxy:


silvabox:
A good attempt Paul, but your code is not quite OO enough and relies too much on passing things around as arrays. 

Think about the separation of concerns and responsibilities between orders and the takeaway.

Also, don't create code that relies on user interaction in the tests!  The whole point of the tests is that they run automatically without intervention.

:construction:

PR #73, opened 2015-06-21
The challenge is mostly finished, the test for textsender is not working

MassimilianoMura:


silvabox:
Hi Massi, thanks for submitting your takeaway challenge!  You've successfully separated the responsibilities of customer, shop and text sender, but there you've exposed quite a bit of internal implementation using `attr_reader` to return arrays and hashes.  You should define new methods on the containing class to control access to these objects instead.
You've written some nice tests, but you haven't used doubles.

:construction:

PR #74, opened 2015-06-21
Submitting week 2 weekend challenge

FaisalChoura:


silvabox:
Hi Faisal, thanks for submitting your takeaway challenge!  Your design is perhaps a little too simple - you should think about extracting the takeaway and text handling into separate classes.  You've also created an issue with your `@sum` instance variable, where you have 2 versions of the truth in a sense.  Better to just call `total` whenever you need to calculate the total.

Be careful of exposing implementation details - e.g. returning the menu hash out of `Menu` - there's almost no point in the `Menu` class as it is - you could've just used a hash.  You should add methods the menu interface to query the menu, without allowing the consumer to change it.

:construction:

PR #75, opened 2015-06-21
Takeaway challenge submission Twilio test not working

natstar93:


lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+class Menu
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = {:Noodles=>3.0, :Rice=>2.5}
```
pitchinvasion: As we discussed this could be an array of dish objects instead possibly

lib/order_form.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'takeaway'
+require_relative 'menu'
+
```
pitchinvasion: This object is doing too much, so to refactor we'd split out this class into smaller classes each with a single responsibility - this one would just handle the order (you could maybe just call it order as well)

pitchinvasion:
Just added a couple of comments from our chat yesterday. As I said there, you need one class to handle all the interactions (a sort of controller class) and you need to look a bit closer at potentially doing some refactoring so that each object only has one responsibility but that's me being picky, generally it's a very clean solution and a really good effort :checkered_flag:

PR #76, opened 2015-06-21
Takeaway challenge attempt

lroliphant:


silvabox:
Hi Lucy, thanks for submitting your takeaway challenge!  Your tests are well written and you've successfully separated the order from the menu and text service.  Watch out for exposing internal implementations such as `menu` returning the hash in `Menu`!
You're injecting the menu into the order which is good - but in your tests, although you've defined a double for menu, you are still passing in `Menu.new` to the subject.

:construction:

PR #78, opened 2015-06-22
Take it away

TheoLeanse:


spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'menu'
+
+describe Menu do
+
+  subject(:menu) { Menu.new }
```
TheoLeanse: When describing a class `subject` is automatically a `.new` instance of the class

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it "can have dishes added to it" do
+    menu.add :burger, 9.99
+    menu.add :chips, 1.99
+    menu.add :coke, 0.99
+    expect(menu.show).to eq "Burger - £9.99 // Chips - £1.99 // Coke - £0.99"
```
TheoLeanse: This tests both `add` and `show` -> figure out a way of showing the menu raw data.

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  it "displays the menu in a pleasant format" do
+    menu.add :burger, 9.99
+    expect(menu.show).to eq "Burger - £9.99" # do not use subject?
```
TheoLeanse: Instead of 'show' which is a little obscure, try something like `pretty_print`

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  it "gives the price of a particular item" do
+    menu.add :burger, 9.99
+    expect(menu.price_of :burger).to eq 9.99
```
TheoLeanse: This is a lovely method.

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  xit "throws an error if that item is not on the menu" do
```
TheoLeanse: This is not required by the user stories but yes good on you for including for l8r

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'order'
+
+describe Order do
+  subject(:order) { Order.new double :menu, price_of: 9.99 }
```
TheoLeanse:
```ruby
subject { Order.new({ double :menu, price_of: 9.99 }) }
```

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it 'adds a dish to its items' do
+    order.add :burger
+    expected_hash = { burger: 9.99 }
```
TheoLeanse: Setting this variable name is probably needlessly class-specific. Use `expected_items` instead --> but `expected` is a great thing to prepend to an expected value!

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    expect(order.items).to eq(expected_hash)
+  end
+
+  xit 'can have two or more of the same item' do
+    order.add :burger
+    order.add :burger
+    order.add :burger
+    order.add :burger
+    expected_hash = { burger: 9.99, burger2: 9.99 }
```
TheoLeanse: I don't think a hash is appropriate here. Either an array of items with their prices, or better: a hash of items and their _quantities_, and since `order` knows the `menu`, it can look up prices from its `@menu` dependency.

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    expect(order.items).to eq(expected_hash)
+  end
+
+  xit 'can remove items from the order' do
```
TheoLeanse: If you use a quantity key in your items hash instead of a price this will be easier to implement.

spec/till_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'till'
+
+describe Till do
+
+  let(:order         ) { double :order, items: { :burger => 9.99 } }
+  let(:notifier      ) { double :notifier, notify: nil }
+  let(:message_centre) { double :message_centre, process: nil }
```
TheoLeanse: Great mocking and stubbing :+1:

spec/till_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  subject(:till)       { Till.new order }
+
+  it "calculates a total" do
+    expect(till.total).to eq 9.99
+  end
+
+  it 'sends the total to receipt when order is finalised' do
+    expect(message_centre).to receive(:process)
+    till.process message_centre, 447785244600
```
TheoLeanse: This is a lovely test.

spec/message_centre_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  it "time-stamps a pre-set message" do
+    expect(message_despatch.generate_message).to eq "Thank you for your order! Your order will be with you by #{message_despatch.time_in_one_hour}."
+  end
+
+  it "sends the message to a messenger" do
+    expect(texter).to receive(:send_message)
+    message_despatch.process 447785244600
+  end
```
TheoLeanse: I really like this set of tests - the method names and variables are extremely legible and I feel comfortable that I could come in and extend this at a moment's notice.

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
+class Menu
+  def initialize
+    @menu = {}
```
TheoLeanse: Avoid setting an instance variable with the same name as the class.

`@items` would make more sense here, or `@item_dictionary` if you want to be super clear that you have prices in there too.

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def add dish, price
+    @menu[dish] = price
+  end
+
+  def show # abstract this to a showable module shared by order and receipt classes? takes an { item: price } hash and lists it on new lines...
```
TheoLeanse: I still think this should be called `print` or `pretty_print` and what would be amazing is if it took a `Printer` that decided to print it however it liked.

TheoLeanse:
```ruby
def print(printer)
  printer.print # this could be an instance of ReceiptPrinter, or HTMLPrinter, or CSVPrinter, or SkyPrinter
end
```

TheoLeanse: To print:
```ruby
Menu.new.add(<items>).print(HTMLPrinter.new)
# => an HTML page with all the items
```

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+    list
+  end
+
+  def price_of dish
+    @menu[dish]
+  end
+
+  private
+  def get_name dish_no
```
TheoLeanse: I would prefer `capitalize_name(dish_number)`

lib/order.rb
```diff
@@ -0,0 +1,25 @@
+class Order
+  attr_reader :items
+  def initialize menu
```
TheoLeanse: I'm not suggesting you do this, but you could do:

```ruby
def initialize options={}
  @menu = options.fetch(:menu, default_menu)
end

private

def default_menu
  { :burger => 9.99 }
end
```

TheoLeanse: Because now you could do:

```ruby
@order = Order.new # and @order.menu would == default_menu
```

lib/order.rb
```diff
@@ -0,0 +1,25 @@
...
+    @items = {}
+  end
+
+  def add dish
+    if items.has_key? dish
+      add "#{dish}*".to_sym
+    else
+      items[dish] = @menu.price_of dish # coupled to menu
+    end
+    "Subtotal: £#{total}."
```
TheoLeanse: I don't like that this method returns a string of the subtotal. This is weird:

```ruby
@subtotal = @order.add(:burger) # whaa
```

lib/order.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def confirm message_centre, customer_details
+    "Total: £#{total}."
```
TheoLeanse: This won't ever print b/c it is not the return value - also it is a weird return value

TheoLeanse:
This is very heartening: extremely clear code, generally great naming conventions and very legible use of `private` and `public` to ensure I can always understand the public interface 'at a glance'. Your tests are neat and well-formed with great use of doubles, mocks, and stubs.

Your dependencies are looking great too: message directions are comprehensible and 'flow' in a sensible way. Avoid odd return values and be aware that any return value may be assigned to a variable, which should make sense with the method naming:

```ruby
variable = method_that_ought_to_define_variable
```

Name instance variables differently to the class they belong to.

To test private methods (if you have to) you can use `subject.send(:method_name)` and this will overcome the private aspect.

PR #81, opened 2015-06-22
Nicola Hughes Week 2 Challenge Submit

DataMinerUK:
Not yet complete


silvabox:
Hi Nicola, thanks for submitting your takeaway challenge!  As we talked about, your design is probably missing an order class which encapsulates the list of dishes and quantities.  Also, be careful of exposing implementation details (e.g. arrays and hashes) using `attr_reader`.

With the dependency on passing arrays, it's not really a complete solution - but I'm not too worried as I don't think you'd have any problems refactoring it.

:construction:

PR #82, opened 2015-06-22
Takeaway challenge submission.

arnottmj:


spec/checkout_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'checkout'
+
+describe Checkout do
+
+  let(:order) {double :order}
```
arnottmj: :+1:

spec/menu_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'menu'
+
+describe Menu do
+
+  it 'should have a list of dishes with prices' do
+    expect(subject.view).to be_instance_of Hash
```
arnottmj: Expectation clauses expecting a specific class is not such a good idea -> you might wanna change this later. Instead, construct the expected object and check us `eq`

```ruby
private

def test_menu_list
  {
    pizza: 2.0,
    pasta: 4.5,
    coke: 2.1
  }
end
```

And then in this test you can do

```ruby
it "lists dishes with prices"
  expect(subject.view_menu).to eq test_menu_list
end
```

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
+require 'order'
+
+describe Order do
+
+  let(:menu) do
+
+    double :menu, view: {pepperoni:  10.0,
```
arnottmj: Groovy stubbing :)

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+                         margharita:  8.0}
+
+  end
+
+  subject {Order.new(menu)}
```
arnottmj: This is totally fine

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  it {expect(subject).to respond_to(:add).with(2).argument}
```
arnottmj: it 'is expected to respond to add with 2 arguments?'

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  it 'should let customers see their order' do
+    expect(subject.view_order).to eq ({})
```
arnottmj: If you use parenthesis (which are optional), don't put a space between them and the statement

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  it 'should let customers see their order total' do
+    expect(subject.view_total).to eq 0
+  end
```
arnottmj: This is all good: wrap it in a context of 'no order is placed'

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  context 'add should' do
```
arnottmj: If you are referencing a method, use this syntax: `context "#add" do`...but that should really be used on `it` statements.

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    it 'add a single pizza to the selection' do
+      subject.add(:pepperoni, 1)
```
arnottmj: Could the item encapsulate its own value? That way you wouldn't have to use `Menu` like a dictionary

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+      expect(subject.view_order).to eq ({pepperoni: [1, 10 ]})
+    end
+
+    it 'update the order total when adding single pizza' do
+      subject.add(:pepperoni, 1)
+      expect(subject.view_total).to eq 10
```
arnottmj: This is really nice.

spec/order_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+    it 'raise error if the quanity isnt an integer' do
+      expect{subject.add(:pepperoni, 1.5)}.to raise_error 'invalid number'
+    end
+
+    it 'raise error if the order has been completed' do
+      subject.complete = true
+      expect{subject.add(:pepperoni, 1.5)}.to raise_error 'order has already been completed'
+    end
+
+    it 'raise error if pizza is in order but quantity is negative resulting in overall negative quantity' do
```
arnottmj: Writing a test quite like this is a good indication that maybe your system is a bit overly-complex

spec/checkout_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  subject {Checkout.new(order)}
+
+  it {expect(subject).to respond_to :submit}
+
+  context 'submit should' do
+
+    it 'raise error if the order total is zero' do
+      allow(subject).to receive(:calc_total) {0}
+      expect{subject.submit}.to raise_error 'your order is empty'
```
arnottmj: :+1:

spec/checkout_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+
+    it 'raise error if it has already been successful' do
+      allow(subject).to receive(:calc_total) {10}
+      allow(subject).to receive(:complete) {true}
+      expect{subject.submit}.to raise_error 'checkout already complete'
```
arnottmj: :

lib/menu.rb
```diff
@@ -0,0 +1,21 @@
+class Menu
+
+  def initialize
+    @menu = {pepperoni:  10.0,
```
arnottmj: Avoid instance variables with the same name as the class

lib/menu.rb
```diff
@@ -0,0 +1,21 @@
...
+             vegetarian:  8.0,
+             tandoori:   11.0,
+             meat:       12.0}
+
+  end
+
+
+  def view
+
+    @menu.clone
```
arnottmj: This is probably unnecessary -> @menu will do

lib/order.rb
```diff
@@ -0,0 +1,76 @@
+class Order
+
+  attr_accessor :complete
+
+  def initialize (menu)
+    @menu = menu.view
```
arnottmj: Good dependency injection --> I think you could probably just do 

```ruby
def initialize(menu) # menu is expected to be an object that responds
  @menu = menu
end
```

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+    @order = {}
+    @total = 0.0
+    @complete = false
```
arnottmj: Good way of setting up state :+1:

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  def view_order
+    @order.clone
+
+  end
+
+  def view_total
+    copy = @total.to_s
+    copy.to_f
```
arnottmj: This will return `@total.to_s.to_f`

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  end
+
+  def add pizza, number
+    raise 'order has already been completed' if complete
+    raise 'item not on menu' if !@menu.include?(pizza)
+    raise 'invalid number' if number % 1 != 0
```
arnottmj: These are really nice 'guard conditions' -> they stop any problematic input from progressing further in the program.

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    if pizza_already_selected? pizza
```
arnottmj: This is a lovely method -> very clear what it does

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+      @order[pizza] = updated_values_for_add(pizza, number)
```
arnottmj: Maybe `add_item_to_order(pizza, number)`?

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+    else
+      raise 'invalid number' if number <= 0
+      temp_total = @menu[pizza]*number
+      @order[pizza] = [number, temp_total]
+    end
+
+    remove_zero_quantity_pizzas
```
arnottmj: This would be a lot simpler if you had used an array instead of a hash BUT if you are going to do methods that deliberately alter state, and the return value is less important, use a bang (!) at the end
```ruby
remove_zero_quantity_pizzas!
```

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+    @total = calc_total
+
+  end
+
+  def remove pizza, number
+    add(pizza, -1 * number)
```
arnottmj: This is cool but a bit obscure

lib/order.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  end
+
+  def pizza_already_selected? pizza
+    @order.keys.include?(pizza)
+
+  end
+
+  def updated_values_for_add pizza, number
+    temp_number = number + @order[pizza][0]
```
arnottmj: `temp_number` is a bit obscure --> if you want to be a bit clearer, `temporary_number` if you have to

lib/checkout.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    account_sid = ENV['TWILIO_SID']
+    auth_token = ENV['TWILIO_AUTH']
+    @client = Twilio::REST::Client.new account_sid, auth_token
+
+  end
+
+  def submit
+
+    raise 'your order is empty' if calc_total == 0
```
arnottmj: `if calc_total.zero?`

lib/checkout.rb
```diff
@@ -0,0 +1,46 @@
...
+    raise 'checkout already complete' if complete
+
+    delivery_time = Time.new + 1800
+
+    message = @client.account.messages.create(
+      :body => "Thank you! Your order was placed and will be delivered before #{delivery_time.strftime('%H:%M')}",
+      :to => "+447766767031",
+      :from => "+441279702199")
+    puts message.to
```
arnottmj: `.to` is not a method

lib/checkout.rb
```diff
@@ -0,0 +1,46 @@
...
+    @total = calc_total
+    @order.complete = true
+    @complete = true
+
+
+  end
+
+  private
+
+  attr_accessor :complete
```
arnottmj: Nicely hidden away

lib/checkout.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def submit
+
+    raise 'your order is empty' if calc_total == 0
+    raise 'checkout already complete' if complete
+
+    delivery_time = Time.new + 1800
+
+    message = @client.account.messages.create(
```
arnottmj: Stub this with an 'OK' response

arnottmj:
This is all looking good -> if you had used an array of `MenuItems` you would have saved yourself complexity in dealing with hashes -->

```ruby
@menu = Menu.new(items: MenuItem.new("Pepperoni", 4.0), MenuItem.new("Cheese", 3.2))
# #items returns a list of the items
@menu.items # => [<#MenuItem asdfdsjflsdjfa>, <#MenuItem sdjflasdjflsadfa>]
# #item_names returns a list of the item names
@menu.item_names # => ["Pepperoni", "Cheese"]
# #item_prices returns a list of item prices
@menu.item_prices # => [4.0, 3.2]
# #total returns item_prices.reduce(&:+)
@menu.total # => 7.2
```

arnottmj:
Let's chat Twilio stubbing l8r

arnottmj:
:construction:

PR #84, opened 2015-06-22
Take-away challenge (not good)

Triffanys101:


spec/text_sender_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'text_sender'
+require 'twilio-ruby'
+
+describe TextSender do
+	
+	let(:dummy_messages) { double :dummy_messages }
+  let(:client) do
+		double :client, messages: dummy_messages
```
Triffanys101: Great use of stubbing/mocking

Triffanys101:
I'd advise coming back to this later. Give it a go during Lab Week (or before, if you get time) - draw some diagrams/maps to think about the way that objects are going to be related. Like, an Order should contain some FoodItems, FoodItems should have a price attribute, so Order can sum all the prices of all the FoodItems and get a total, that kinda thing.

PR #92, opened 2015-07-31
My takeaway challenge

Yorkshireman:


spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'order'
+
+describe "Order" do 
+
+  let (:menu){double :menu, dishes: {biriyani: 5.00, uk_curry: 6.00, cheeseburger: 4.00, pizza: 5.00, side_of_salmonella: 1.95}}
```
sjmog: I would extract the value of `dishes` to a private method or other `let` statement

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  
+  before :each do
+    @subject = Order.new menu
```
sjmog: `subject` is not an instance variable. It works like `let`:

```ruby
subject { Order.new menu }
```

sjmog: Use `described_class` instead of referencing the described class by name:

```ruby
# These do the same things
describe Order do
  Order.new
end
describe Order do
  described_class.new
end
```

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    allow(@subject).to receive(:send_sms).and_return true
```

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  context 'when created' do 
+    it 'has a selection hash' do
+      expect(@subject.selection).to be_a Hash
+    end
```
sjmog: Don't test for explicit types. If you need to change this type later on, this test will fail.

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    it 'the selection hash is empty' do 
+      expect(@subject.selection).to be_empty
+    end
+
+    it 'has a menu to choose from' do
+      expect(@subject.menu).to be menu
+    end
```

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  describe '#add_item' do 
+    it 'can add a dish from a menu into its selection hash' do
+      @subject.add_item :biriyani, 1
+      expect(@subject.selection).to include :biriyani
+    end
```

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+
+    it 'will have the appropriate quantity as its value' do 
+      @subject.add_item :biriyani, 1
+      expect(@subject.selection[:biriyani]).to eq 1
+    end
+
+    it 'returns an error message when dish does not exist on the menu' do
+      expect(@subject.add_item :hamburger, 1).to eq "That item is not on the menu"
+    end
```
sjmog: You should use `context` here because all these tests refer to the same scenario.

A good indication you should use a `context` block is when all tests start with the same kind of setup (you could extract these lines to a `before` block within the `context`).

spec/order_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+    describe '#on_menu?' do
```

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      it 'returns false if an item is not on the menu' do 
+        expect(@subject.on_menu? :hamburger).to be false
+      end
+    end
+  end
+
+  describe '#place_order' do
+    before :each do 
+      allow(@subject).to receive(:selection).and_return(biriyani: 1, cheeseburger: 2)
+    end
```
sjmog: Lovely setup :+1:

spec/order_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      it 'it returns a message saying the order was successfully placed' do
+        expect(@subject.place_order 13).to eq "Thankyou! Your order was successfully placed and will be delivered within 1 hour."
+      end
+    end
+
+    context 'when given amount is incorrect' do
+      it 'it returns a message saying the given amount is incorrect and suggests a calculated total' do 
+        expect(@subject.place_order 10).to eq "Order unsuccessful. Total should equal 13.0"
+      end
+    end
```
sjmog: Great use of `context`

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'menu'
+
+describe Menu do
+
+  let(:biriyani){double :biriyani, title: "Biriyani", price: 5.00}
+
+  context 'when created' do 
+    it 'has a default set of dishes' do
+      expect(subject.dishes).to eq Menu::DEFAULT_MENU
```
sjmog: :+1: use of namespaced constant

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    xit 'displays itself'
+      # Don't know how to test for this
```
sjmog: Try adding a `print` method to `Menu` (that you'll probably want to extract to a `Printer` class) and test that method

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  it 'can have dishes added to it' do
+    subject.add_dish(biriyani, 5.00)
+    expect(subject.dishes).to include biriyani
+  end
+
+  describe '#show' do 
+    it 'returns an ASCII table representation of the menu using the terminal-table gem' do 
+      # I don't know how to test for this
+      expect(subject).to respond_to :show
```
sjmog: Ahhh, OK. Check the terminal-table docs and construct your own variant example, then test that `show` outputs a table that looks like that.

You might want to extract your own variant example to a private method in this `describe` block.

lib/order.rb
```diff
@@ -0,0 +1,63 @@
...
+    @selection = {}
+    @menu = menu
+  end
+
+  def add_item dish, quantity
+    if !on_menu? dish
+      return "That item is not on the menu"
+    else
+      selection.store dish, quantity
+    end
```
sjmog: Refactor this long-term conditional:

```ruby
if x
  return y
else
  z
end
```

Becomes

```ruby
return y if x
z
```

lib/order.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  def on_menu? dish
+    menu.dishes.include? dish
+  end
```
sjmog: Love this method.

lib/order.rb
```diff
@@ -0,0 +1,63 @@
...
+      send_sms
+      return "Thankyou! Your order was successfully placed and will be delivered within 1 hour."
+    else
+      return "Order unsuccessful. Total should equal #{build_order[selection]}"
+    end
+  end
+
+  def correct_amount? amount
+    amount == build_order[selection]
+  end
```
sjmog: :+1:

lib/order.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def build_order
+    total = 0 
+    selection.each do |item, quantity| 
+      total += quantity * menu.dishes[item]
+    end
```
sjmog: These four lines should be extracted to a private method `total = calculate_total`

This could even be extracted to another class, `OrderCalculator`

lib/order.rb
```diff
@@ -0,0 +1,63 @@
...
+    account_sid = ENV["TWILIO_ACCOUNT_SID"]
+    auth_token = ENV["TWILIO_TOKEN"]
+    @client = Twilio::REST::Client.new account_sid, auth_token
+    @client.messages.create(
+      from: '+441768392015', 
+      to: "#{customer_number}", 
+      body: 'Thankyou! Your order was successfully placed and will be delivered within 1 hour.')
+    
+    puts "SMS order confirmation successfully sent"
+  end
```
sjmog: Great. This method should probably exist in its own class, as right now this current class has too many responsibilities.

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
+require_relative './dish.rb'
+require_relative './order.rb'
+require 'terminal-table'
+
+class Menu
+
+  DEFAULT_MENU = {biriyani: 5.00, uk_curry: 6.00, cheeseburger: 4.00, pizza: 5.00, side_of_salmonella: 1.95}
```
sjmog: :+1:

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = DEFAULT_MENU
```
sjmog: :+1:

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def add_dish name, price
+    dishes[name] = price
+  end
```
sjmog: :+1:

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def show
+    rows = []
+
+    dishes.each do |dish, amount|
+      row = [ dish, ("£" + (sprintf "%.2f", amount)) ]
+      rows << row
+    end
```
sjmog: These lines should be extracted to a private method.

sjmog:
Fantastic solution, Andy. Very few suggestions, and mostly stylistic ones at that.

:checkered_flag:

lib/order.rb
Yorkshireman: I had it like your example to begin with, but changed it to the former for readability. What do you think?

lib/order.rb
Yorkshireman: I agree

PR #93, opened 2015-07-31
Initial commit - Added feature_spec, takeaway_spec & takeaway class, …

RichardCharman:
…with opening tests for menu


README.md
```diff
@@ -1,55 +1,4 @@
...
-
-* A free account on Twilio will only allow you to send texts to "verified" numbers. Use your mobile phone number, don't worry about the customer's mobile phone.
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
-
-
-**Note: We are looking for good OO design and programming! Remember the Single Responsibility and Dependency Injection/Inversion principles!**
-
-[![Build Status](https://travis-ci.org/makersacademy/takeaway-challenge.svg?branch=master)](https://travis-ci.org/makersacademy/takeaway-challenge)
-[![Coverage Status](https://coveralls.io/repos/makersacademy/takeaway-challenge/badge.png)](https://coveralls.io/r/makersacademy/takeaway-challenge)
+A program to allow customers to view and order from a takeaway menu.
```
danldb: Need much more info here - what were the learning objectives, what difficulties did you overcome, what does the final product look like, what gems did you use, etc.

lib/customer.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'takeaway'
+
+class Customer < Takeaway
+
```
danldb: Is Customer a Takeaway?

Inheritance is dangerous - and this is definitely not a good use case.

lib/takeaway.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def order_in
+    # confirmataion text to go here
+  end
+  
+  def whats_cooking
+    to_cook
+  end
+end
```
danldb: I think this has gotten really confused as you have not understood the implications of inheritance. Refactor this so that they are all separate objects.

danldb:
:construction: 

As discussed, this needs a restart - be wary of inheritance, its uses are few and its dangers are many!

PR #94, opened 2015-08-02
Initial takeaway challenge commit

emily-jane:


Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.2.2'
+ruby '2.1.5'
```
sjmog: Why the downgrade?

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'order'
+
+describe Order do
+
+	let (:order){double(:order)}
```
sjmog: Why would you create a double of the class you are describing?

Unit Test doubles exist to _isolate_ tests - so you would double any classes that _interact_ with this class.

sjmog: Also, tabs! Don't forget to make sure your text editor uses spaces instead of tabs.

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+	it 'should take orders with 2 arguments' do
+		expect(subject).to respond_to(:new_order).with(2).argument
+	end
```

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+	it 'should take order and add it to order' do
+		subject.new_order([[2,2],[1,1]],20)
+		expect(subject.order).to eq({:FishandChips=>14, :Burger=>6})
```
sjmog: Symbols should use lowercase and be underscored (`:fish_and_chips` and `:burger` here)

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+	end
+
+	it 'should raise error if incorrect total price' do
+		expect{subject.new_order([[2,2],[1,1]],8)}.to raise_error "Total cost is incorrect"
```
sjmog: Your way of constructing orders is a little complex. Could you maybe think of a clearer way pass arguments to `new_order`?

sjmog: Great use of `raise_error`

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+	end
+
+	it 'should raise error if placing empty order' do
+		expect{subject.place_order}.to raise_error "No order has been entered"
+	end
+
+	it 'can send texts' do
+		expect(subject).to respond_to(:send_text)
```
sjmog: It seems a little odd that an order can send texts. This method should probably exist in its own class, as right now this current class has too many responsibilities.

text.rb
```diff
@@ -0,0 +1,21 @@
+require 'rubygems'
+require 'twilio-ruby'
+
```
sjmog: What is this file for?

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
+class Menu
```
sjmog: I see a `Menu` but no `menu_spec.rb`. If you need a class, make sure you are driving it through Test-Driven Development!

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+
+	attr_reader :menu_list, :menu_keys, :menu_values
+
+	def initialize
+		@menu_list = menu_list
+		@menu_keys = menu_list.keys
+		@menu_values = menu_list.values
```
sjmog: I don't think these need to be split into their own instance variables. You could use a private method to fetch these from `@menu_list`, and then limit the number of instance variables this class has.

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+	end
+
+		def menu_list
+			menu_list = {Burger: 6,
```
sjmog: You don't need to assign this to a variable.

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+							Pizza: 8,
+							Beer: 4,
+							Wine: 5,
+							Juice: 2}
+		end
+
+		def show
+			puts "MENU"
+			count = 1
+			menu_list.each do |key, value|
```
sjmog: Very nice! You could use `each_with_index` here to avoid using a `count`:

```
menu_list.each_with_index do |(key, value), index|
  p "#{index}: #{key} - £#{value}"
end
```

lib/order.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'menu'
+
+class Order < Menu
```
sjmog: I'm not sure an `Order` is a type of `Menu`.

lib/order.rb
```diff
@@ -0,0 +1,41 @@
...
+
+	attr_reader :order, :texted
+
+	attr_accessor :texted
```
sjmog: Nice use of encapsulation :+1:

lib/order.rb
```diff
@@ -0,0 +1,41 @@
...
+
+	def initialize
+		@order = {}
+		@menu_keys = menu_list.keys
+		@menu_values = menu_list.values
```
sjmog: See the comment in `Menu`

lib/order.rb
```diff
@@ -0,0 +1,41 @@
...
+			dish = menu_keys[x - 1]
+			cost = menu_values[x - 1] * y
+			puts "#{(list.index([x,y]))+1}. #{dish}   x #{y}   £#{cost}"
+			order.merge!({dish => cost})
+		end
+		sum = 0
+		order.values.each {|a| sum += a }
+		fail "Total cost is incorrect" if sum != total_cost
+		puts "TOTAL COST - £#{total_cost}" 
+	end
```
sjmog: It's a little hard to follow the logic here, they're not very book-ish. These lines should be extracted to a private method.

lib/order.rb
```diff
@@ -0,0 +1,41 @@
...
+
+	def place_order
+		fail "No order has been entered" if order == {}
```
sjmog: Great use of guard conditions :)

lib/order.rb
```diff
@@ -0,0 +1,41 @@
...
+		send_text
+	end
+
+	def send_text
+		system("ruby text.rb")
```
sjmog: I see now why the `text.rb` file exists! I don't think this is a good way of doing this. You could wrap the code in `text.rb` in a class `Text` and then, in this method:

```ruby
def send_text
  Text.new.deliver
  @texted = true
end
```

sjmog:
@emily-jane there's a lot of good code here, and good understanding. Do check my comments, though, as I'm not sure you've done this TDD-style, or followed Object-Oriented Principles too well.

If you get a chance to re-submit this, do: if you want to see how to get started, check out some earlier commits from [other pull requests on your cohort](https://github.com/makersacademy/takeaway-challenge/pulls).

:construction:

PR #95, opened 2015-08-02
Submission for Takeaway Challenge

katylouise:
Solution so far - need to do some refactoring and have a go at the extension.


lib/takeaway.rb
```diff
@@ -0,0 +1,76 @@
+require 'twilio-ruby'
+
+class TakeAway
+
+  attr_reader :menu
+  attr_accessor :order, :prices
+
+  def initialize
```
zdajani: initialise with a menu, which would remove the need for the next two methods

lib/takeaway.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def add_menu(menu)
+    @menu = menu.dishes
+  end
+
+  def show_menu
+    menu
+  end
+
+  def place_order(items = {})
```
zdajani: separate items from quantity so you can add to the quantity, without having to add the same item twice

lib/takeaway.rb
```diff
@@ -0,0 +1,76 @@
...
+  def total_cost_per_item(price, quantity)
+    price_in_pence = ((price * 100).round * quantity)
+    prices << price_in_pence
+    "%.2f" % (price_in_pence / 100.0)
+  end
+
+  def show_order
+    order_output = print_order_items
+    order_output += "Total: £#{grand_total}"
+    order_output
```
zdajani: you can remove this line

zdajani:
Really good! I would just suggest making a module for the calculations of the order and keep the takeaway class just for ordering. Also think about separating the Twilio part into another module or class.

lib/takeaway.rb
```diff
@@ -0,0 +1,76 @@
...
+      order["#{dish} x #{quantity}"] = total_cost_per_item(menu[dish], quantity)
+    end
+  end
+
+  def grand_total
+    total_in_pence = prices.inject(:+)
+    "%.2f" % (total_in_pence / 100.0)
+  end
+
+  def total_cost_per_item(price, quantity)
```
zdajani: check if you can delete by quantity

stephenlloyd:
:checkered_flag:

PR #101, opened 2015-08-02
Takeaway challenge

ojlamb:


spec/twilio_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'twilio'
+
+describe Twilio do
+  it "Can calculate time" do
+    time = Time.now + (3600)
+    expect(subject.timer).to eq time.to_s.split(' ')[1]
+  end
+end
```
sjmog: Awesome. So you're only testing your additions, which is right :+1:

spec/order_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'order'
+
+describe Order do
+  it "Has a method select dishes with 2 arguments"do
```
sjmog: Use RSpec instance method syntax for this. See [here](http://betterspecs.org/#describe)

spec/order_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    expect(subject).to respond_to(:select_dishes).with(3).arguments
```
sjmog: Also...this is 3 args, not 2

spec/order_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  it "Can select a number dishes off the menu" do
```
sjmog: No need for the 'can'. Be imperative:

```ruby
it 'selects a number of dishes from a menu' do
```

spec/order_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    subject.select_dishes("hamburger", 6.0, 2)
+    expect(subject.ticket).to eq [ { "hamburger" => 6.0 } , { "hamburger" => 6.0 } ]
```
sjmog: Long line due to repeats --> make `{ "hamburger" => 6.0 }` a local variable

spec/order_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  it "can display the order total" do
+    subject.select_dishes("hamburger", 6.0, 2)
+    expect(subject.sum_order).to eq 12.00
+  end
```

spec/menu_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'menu'
+
+describe Menu do
+  it "is a hash" do
+    expect(subject.display).to be_a Hash
+  end
```
sjmog: Don't test for explicit types. If you need to change this type later on, this test will fail.

Better is:

```ruby
example_object = # construct example object
expect(test).to eq example_object
```

spec/menu_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+  it "can display dishes" do
+    expect(subject.display).to include "hamburger"
+  end
```
sjmog: :+1: though clearer would be `it "outputs an array of dishes"`

spec/dish_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'dish'
+
+describe Dish do
+  describe "#price" do
```
sjmog: :+1:

spec/dish_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    it "Has dish hotdog with price of $5.00" do
+      expect((Dish.hotdog).price).to eq 5.00
+    end
+
+    it "Has dish French Fries with price of $3.00" do
+      expect((Dish.fries).price).to eq 3.00
+    end
+  end
```
sjmog: Very nice :)

spec/customer_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'customer'
+
+describe Customer do
+  let(:order) { double :order }
```
sjmog: :+1:

spec/customer_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  it "can place order" do
+    expect(subject).to respond_to(:place_order)
+  end
```
sjmog: Awesome

spec/customer_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    expect(subject).to respond_to(:get_quantity)
+  end
+
+  it "can get price" do
+    expect(subject).to respond_to(:get_price)
+  end
+
+  it "can confirm order" do
+    expect(subject).to respond_to(:get_confirmation)
+  end
```
sjmog: DRY up `can` in each of these by just removing it. Be imperative

lib/twilio.rb
```diff
@@ -0,0 +1,21 @@
+require 'rubygems'
+require_relative 'customer'
+
+class Twilio
+  def timer
+    date = Time.now + (3600)
+    delivery_time = date.to_s.split(' ')[1]
+  end
+end
```
sjmog: Why are you ending the class here? What about the `send_text` method?

lib/twilio.rb
```diff
@@ -0,0 +1,21 @@
...
+
+def send_text(name, phone)
```
sjmog: This method should be in the class, no?

lib/twilio.rb
```diff
@@ -0,0 +1,21 @@
...
+  account_sid = 'AC2f4d6a3fb2fcdd3e058b6930b7f47697'
+  auth_token = '18a596c4b02371cb565004ab54694bcf'
+  @client = Twilio::REST::Client.new account_sid, auth_token
+
+  message = @client.account.messages.create(
```
sjmog: Then these bits could be in the class initializer

lib/order.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'menu'
+class Order
+  attr_accessor :total_cost
+  attr_reader :menu, :ticket
+  def initialize(menu = Menu.new)
+    @ticket = []
+    @total_cost = 0
+    @menu = menu
+  end
```
sjmog: Lovely initializer here.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'menu'
+class Order
+  attr_accessor :total_cost
+  attr_reader :menu, :ticket
```
sjmog: Great use of encapsulation

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+    @ticket = []
+    @total_cost = 0
+    @menu = menu
+  end
+
+  def select_dishes(dish, price, quantity)
+    quantity.times do
+      add_to_ticket(dish, price)
+    end
+  end
```
sjmog: Very clear, love it.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def add_to_ticket(dish, price)
+    selection = { dish => price }
+    @ticket << selection
+    sum_order
+  end
```
sjmog: Could this be a private method?

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def sum_order
+    prices = @ticket.map { |dish| dish.values }.flatten
+    @total_cost = prices.inject(:+)
+  end
```
sjmog: Great way to make your code readable. Love the chunking up here.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def delete_order
+    @ticket = []
+    @total_cost = 0
```
sjmog: Don't you have an `attr_accessor` for this? You should avoid accessing state directly, and either _mutate_ state via an `attr_reader` or set state using an `attr_writer`. For instance:

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'dish'
+class Menu
+  attr_reader :display
+  def initialize
+    @display
```
sjmog: I'm confused why the initializer returns this

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+  
+  def display
+    @display = {
+      "hamburger" => Dish.burger.price,
+      "cheeseburger" => Dish.cheeseburger.price,
+      "double cheeseburger" => Dish.doublecheese.price,
+      "hotdog" => Dish.hotdog.price,
+      "fries" => Dish.fries.price
+    }
```
sjmog: Quite a lot of duplication here. Could this be expressed more elegantly?

lib/dish.rb
```diff
@@ -0,0 +1,26 @@
+class Dish
+  attr_reader :price
+  def initialize price
+    @price = price
+  end
+
+  def self.burger
+    Dish.new 6.00
+  end
```
sjmog: Very nice. You could just do

```ruby
def self.burger
  new 6.00
end
```

lib/customer.rb
```diff
@@ -0,0 +1,74 @@
...
+  attr_accessor :name, :phone_number, :order, :selection, :price, :total
+  def initialize(order = Order.new)
+    @name
+    @phone_number
+    @order = order
+    @selection
+    @quantity
+    @price
+    @total
+  end
```
sjmog: Why are all these out here?

lib/customer.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def place_order
+    get_name
+    @order.select_dishes(@selection, @price, @quantity)
```
sjmog: You have an accessor for these. Don't use the instance variable directly.

lib/customer.rb
```diff
@@ -0,0 +1,74 @@
...
+    puts "Good evening, can I get your name please?"
+    @name = gets.chomp
+    get_phone
+  end
+
+  def get_phone
+    puts "and what is your phone number, for delivery verification?"
+    @phone_number = gets.chomp.to_i
+    see_menu
+  end
```
sjmog: Love these methods :)

lib/customer.rb
```diff
@@ -0,0 +1,74 @@
...
+  def see_menu
+    self.order.menu.display.each { |k,v| puts "#{k}:#{v}" }
+    get_food
+  end
+
+  def get_food
+    puts "What will you be having today?"
+    @selection = gets.chomp.downcase
+    raise "That dish is not on the menu!" if @order.menu.display.has_key?(@selection) == false
+    get_quantity
```
sjmog: I think these methods are very interdependent (they all call one another in succession). That's not really OOP, it's really functional chaining. You don't want to do that in OOP, or in Ruby to be honest.

Make sure all these methods can be called separately, then use a method to call them all one after the other if needed.

lib/customer.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def see_menu
+    self.order.menu.display.each { |k,v| puts "#{k}:#{v}" }
+    get_food
+  end
+
+  def get_food
+    puts "What will you be having today?"
+    @selection = gets.chomp.downcase
+    raise "That dish is not on the menu!" if @order.menu.display.has_key?(@selection) == false
```
sjmog: Neater would be to end this line with `unless @order.menu.display.has_key?(@selection)`

sjmog:
Lovely stuff @ojlamb . Not sure why Twilio's sending is sitting outside the class, or why you're using some very functional ideas by forcing strong dependence, but in general the code is clear and clean, the TDD is good, and the OOP makes sense.

:checkered_flag:

PR #103, opened 2015-08-02
submission

giusepped:


lib/order.rb
```diff
@@ -0,0 +1,50 @@
+
+class Order
+  MENU = { Pizza: 5,
+           Pasta: 4,
+           Steak: 6,
+           Cod: 5 }
+
+  attr_reader :menu, :ordered_dishes
```
silvabox: should these arrays be exposed to the outside world?

lib/twilio_sms.rb
```diff
@@ -0,0 +1,19 @@
+require 'rubygems'
+require 'twilio-ruby'
+
+class TwilioSms
+
+  def send_order_sms number
+    t = Time.new + 3600
+    t = t.strftime("%l:%M")
+    account_sid = 'AC***************************'
```
silvabox: How do you test this with starred out ids?

spec/order_spec.rb
```diff
@@ -0,0 +1,35 @@
+require_relative '../lib/order'
+
+describe Order do
+
+  it 'shows the menu properly' do
+    hash_example = { dish: :price }
+    subject = Order.new hash_example, 'dummy_twilio'
+    expect(subject.show_menu).to eq("#{:dish}, £ #{:price}\n")
```
silvabox: Beware testing presentation strings quite so precisely as they can easily change.  You might instead test that the name is included in the string and separately that the price is included.

spec/order_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      total = subject.calculate_total
+      expect(subject.calculate_total).to eq total
+    end
+
+    it 'shows the quantity of each dish and its price and the total amount' do
+      subject = Order.new 'dummy_twilio'
+      subject.add_order(:Cod, 2)
+      total = subject.calculate_total
+      str = "Cod, £ 5 * 2\nTotal amount: #{total}"
+      expect(subject.verify_total).to eq(str)
```
silvabox: See my note about testing presentation strings

lib/order.rb
```diff
@@ -0,0 +1,50 @@
...
+  def show_menu
+    result = ''
+    @menu.each do |dish, price|
+      result += ("#{dish}, £ #{price}\n")
+    end
+    result
+  end
+
+  def add_order(dish, quantity)
+    @ordered_dishes[dish] = quantity
```
silvabox: Should you be validating that the dish exists?

lib/order.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def calculate_total
+    total = 0
+    ordered_dishes.each do |dish, quantity|
+      total += menu[dish] * quantity
+    end
+    total
+  end
+
+  def verify_total
```
silvabox: 'verify` implies that this method does some sort of validation or checking, when it is only returning a string.

lib/order.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def initialize menu = MENU, twilio_sms
+    @menu = menu
+    @ordered_dishes = {}
+    @total = 0
+    @twilio_sms = twilio_sms
+    @number = '+44**********'
+  end
+
+  def show_menu
```
silvabox: 'show' implies that this somehow creates a view rather than just returning a string.

spec/order_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      subject = Order.new 'dummy_twilio'
+      subject.add_order(:Cod, 2)
+      expect(subject.ordered_dishes[:Cod]).to eq 2
+    end
+
+    it 'calculates the total amount for the order' do
+      subject = Order.new 'dummy_twilio'
+      subject.add_order(:Cod, 2)
+      total = subject.calculate_total
+      expect(subject.calculate_total).to eq total
```
silvabox: This is a vacuous test.  You calculate the total, then compare that the calculated total is equal to the calculated total!

lib/twilio_sms.rb
```diff
@@ -0,0 +1,19 @@
+require 'rubygems'
+require 'twilio-ruby'
+
+class TwilioSms
+
+  def send_order_sms number
+    t = Time.new + 3600
```
silvabox: Could you encapsulate this line in a method?  e.g. `time_in_one_hour`

silvabox:
Hi Giuseppe, thanks for your Takeaway Challenge.  Sorry your code review was missed last week.

I've made a few comments for some general things, but the biggest issue I think is that the `Order` class has too many responsibilities.  These should be split between at least 2 classes - an `Order` and a `Takeaway`.  The order is responsible for keeping track of the list of ordered dishes; the takeaway is responsible for receiving and processing the order, then sending the text message (via another class - as you correctly have it).

:checkered_flag:

PR #110, opened 2015-08-02
Completed

richgeog:
I was unable to get classes to work together so I had to resort to doing it all within 1 class. Not sure where I was going wrong but something was correct.


stephenlloyd:
:checkered_flag:

PR #113, opened 2015-08-03
rushed code

7091lapS:
Been sick most of the weekend and could only work on this for one day. It's not pretty, but it helped make a lot of things clearer. Playing with the twilio api and trying to stub the sms method was good fun.


spec/restaurant_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'restaurant'
+
+describe Restaurant do
+
+  it 'has a menu of dishes with prices' do
+    expect(subject.menu).to eq({margherita: 5, pepperoni: 6, hawaiian: 6, supreme: 7})
```
sjmog: :+1:

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  it 'allows to choose one or more of the same dishes' do
+    expect(subject).to respond_to(:choose).with(2).arguments
+  end
```
sjmog: :+1:

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  it 'stores the choice the right number of times' do
+    subject.choose(2, "margherita")
+    expect(subject.selected).to eq({margherita: 2})
```
sjmog: :+1:

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  it 'stores a total price for the selected dishes' do
+    subject.choose(3, "margherita")
+    subject.choose(2, "pepperoni")
+    expect(subject.total_price).to eq 27
+  end
+
+  it 'sends an sms when the order is correct' do
+    subject.choose(2, "margherita")
+    subject.choose(1, "pepperoni")
+    allow(subject).to receive(:send_sms) {true} # I know this doesn't make sense
```
sjmog: This makes great sense, actually.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    expect(subject.place_order(16)).to eq true # I couldn't figure out how to
```
sjmog: This does stub it properly - Twilio is already tested, so you don't need to stub its innards. Here you are treating it like a 'black box' which is plenty sufficient.

lib/sms.rb
```diff
@@ -0,0 +1,18 @@
+class Sms
+  def send_text
+    require 'twilio-ruby'
+
+    # put your own credentials here
+    account_sid = 'AC33011a40dc7a73068b3aecbc38deb302'
+    auth_token = '2b6b55ca0e757662eae67ee5df0fcac0'
+
+    # set up a client to talk to the Twilio REST API
+    @client = Twilio::REST::Client.new account_sid, auth_token
```
sjmog: You could put this stuff in the initializer for this class

lib/sms.rb
```diff
@@ -0,0 +1,18 @@
+class Sms
+  def send_text
+    require 'twilio-ruby'
+
+    # put your own credentials here
+    account_sid = 'AC33011a40dc7a73068b3aecbc38deb302'
+    auth_token = '2b6b55ca0e757662eae67ee5df0fcac0'
```
sjmog: Maybe try use use `ENV` variables to avoid exposing this stuff in git

lib/sms.rb
```diff
@@ -0,0 +1,18 @@
+class Sms
+  def send_text
+    require 'twilio-ruby'
```
sjmog: I'd expect this to be required at the top of the class

lib/restaurant.rb
```diff
@@ -0,0 +1,49 @@
+#require 'terminal-table'
+require 'twilio-ruby'
+require './lib/sms.rb'
```
sjmog: Maybe `require_relative` here

lib/restaurant.rb
```diff
@@ -0,0 +1,49 @@
...
+
+class Restaurant
+
+  def initialize #smsclass <= suggested by peer
```
sjmog: Yes, very good idea to take an argument to `initialize`. That way you can use any kind of `SmsSender` and it'll 'just work' (i.e. not just Twilio)

lib/restaurant.rb
```diff
@@ -0,0 +1,49 @@
...
+    selected[:pepperoni] ? menu[:pepperoni] * selected[:pepperoni] : 0
+  end
+
+  def hawaiian_subtotal
+    selected[:hawaiian] ? menu[:hawaiian] * selected[:hawaiian] : 0
+  end
+
+  def supreme_subtotal
+    selected[:supreme] ? menu[:supreme] * selected[:supreme] : 0
+  end
```
sjmog: I'm not sure these methods are such good ideas. They hard-code data in to the behaviour of your application.

You _could_ do it. But I advise against. Better would be
```ruby
def dish_subtotal(dish)
  dish = dish.to_sym # just in case
  selected[dish] ? menu[dish] * selected[dish] : 0
end
```

lib/restaurant.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  attr_reader :menu, :selected
```
sjmog: Not sure why these are at the _bottom_ of the class

lib/restaurant.rb
```diff
@@ -0,0 +1,49 @@
...
+class Restaurant
+
+  def initialize #smsclass <= suggested by peer
+    @menu = {margherita: 5, pepperoni: 6, hawaiian: 6, supreme: 7}
+    @selected = {}
+    #@sms = smsclass.new <= suggested by peer review
+  end
+
+  def choose(amount, dish)
+    raise "Your choice is not on the menu" unless menu.include?(dish.to_sym)
```
sjmog: Great guard condition

lib/restaurant.rb
```diff
@@ -0,0 +1,49 @@
...
+    selected.store(dish.to_sym, amount)
+  end
+
+  def total_price
+    total = margherita_subtotal + pepperoni_subtotal + hawaiian_subtotal + supreme_subtotal
+  end
```
sjmog: Using the general-purpose method I used earlier you could just do

```ruby
def total
  dishes.inject(0) { |subtotal, dish| subtotal += dish_subtotal(dish) }
end

private

def dishes
  menu.keys
end
```

sjmog:
There's a lot to love here! Take a peek at my suggested changes, but in general I'm very impressed by your use of TDD and OOP.

:checkered_flag:

PR #114, opened 2015-08-03
takeaway-challenge solution

Edwardyao123:


lib/menu.rb
```diff
@@ -0,0 +1,18 @@
...
+		rows = []
+		rows << ["1. Chicken Fry Rice", 6.59]
+		rows << ["2. Curry Beef Rice", 5.59]
+		rows << ["3. Sweet King Prawns", 7.59]
+		rows << ["4. Mushroom Soup", 4.59]
+		rows << ["5. Hot & Sour Soup", 4.59]
+		table = Terminal::Table.new :title => "Chinese Takeaway", :headings => ['Dish Name','Price(£)'], :rows => rows
+		table.align_column(1,:center)
+		puts table
+	end
```
danldb: use a constant such as LIST instead of the self.list method. You can then refer to the constant in a method, such as `self.list`

lib/send_sms.rb
```diff
@@ -0,0 +1,21 @@
+require 'rubygems'
+require 'twilio-ruby'
+ 
+class SMS
```
danldb: use dependency injection

```
def initialize(client: Twilio::Rest::Client)
  @client = client.new(account_sid, auth_token)
end
```

spec/features_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+describe 'Features' do
+
+	let(:order) {Order.new}
+	
+
+	# As a customer
+	# So that I can check if I want to order something
+	# I would like to see a list of dishes with prices
+
```
danldb: delete commented out code

spec/menu_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'menu'
+
+describe Menu do
+
+	describe '.list' do
+		xit 'show a table with heading of \'Dish Name\' and \'Price\'' do
+		expect(subject.list).to eq Terminal::Table[:rows]
+		end
```
danldb: testing the interface instead of the logic, that is why it is so difficult!

danldb:
:construction: 

Great attempt Ed - glad to see you were able to send an SMS. You did not pass this challenge this time around as you introduced too much interface, and did not separate the interface from yur business logic, so everything became very confusing and difficult to test. 

You must try to make better use of dependency injection to invert your dependencies - this will make stubbing the Twilio client much easier.

After the course, try to write this again but focus on just the business logic - no interface.

PR #115, opened 2015-08-03
Haroon Takeaway

haroonhassan:


spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'customer'
+
+describe Customer do
+  
+  subject { Customer.new 075 }
```
sjmog: *Great* use of implicit `subject`. Love this.

spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  it 'can request to see a menu' do
+    menu = double('menu', menu_items: {'Chicken' => 5.00, "Salted Beef" => 6.50, "Tuna" => 6.00})
```
sjmog: Good use of doubles

spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    expect(subject.request(menu)).to eq (menu.menu_items)
+  end
```
sjmog: :+1: nice test

spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  it 'can choose items' do
+    dbl = double('choices', choices: ['Chicken', 'Tuna'])
```
sjmog: This should have a clearer name than `dbl`

spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    expect(subject.choices=dbl).to eq dbl
+  end
+
+  it 'can estimate a total' do
+    expect(subject.total_estimate= 11).to eq 11
```
sjmog: I think you should separate out these lines. Check out the [Four-Phase Test](https://robots.thoughtbot.com/four-phase-test) methodology.

spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  it 'can calculate real total' do
+    menu = double('menu', menu_items: { 'chicken' => 5.00, 'tuna' => 6.00, 'beef' => 7.00, 'veg' => 6.50 })
```
sjmog: Prefer symbols to strings where possible

spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    subject.request(menu)
+    subject.choices = ['chicken', 'tuna']
+    expect(subject.real_total).to eq 11
```
sjmog: :+1:

spec/customer_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  it 'can validate the order by checking estimate total vs real total' do
+    allow(subject).to receive(:real_total).and_return(11)
+    allow(subject).to receive(:total_estimate).and_return(11)
+    expect(subject.validate_order?).to be true
+  end
```
sjmog: This is a lovely test

spec/menu_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'menu'
+
+describe Menu do
+  it 'will have a menu of dishes and prices that can be read' do
+    expect(subject.menu_items).to be_a(Hash)
```
sjmog: Don't test for explicit types. If you need to change this type later on, this test will fail.

Better is:

```ruby
example_object = # construct example object
expect(test).to eq example_object
```

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+   end
+
+   def send_text
+    @client = Twilio::REST::Client.new account_sid, auth_token
+    @client.messages.create(
+      from: from_number,
+      to: customer_number,
+      body: msg_to_be_sent
+    )
+   end
```
sjmog: This method should probably exist in its own class, as right now this current class has too many responsibilities.

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+    @customer_number = phone_number
+    @from_number = '+441748352040'
+    @time_order_in = Time.new
+    account_sid = 'AC9ecff5bef6676738eeb43f6725108ca7'
+    auth_token = '9845ae3dd289ebf22cba9b1a393ae050'
+
+   end
+
+   def estimated_arrival_time
+    return arrival_time = time_order_in + (60*60)
```
sjmog: You don't need to use `return` here

lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+class Menu
+
+  attr_reader :menu_items
+
+  def initialize
+    @menu_items = { 'chicken' => 5.00, 'tuna' => 6.00, 'beef' => 7.00, 'veg' => 6.50 }
+  end
```
sjmog: :+1:

lib/customer.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+  
+  def real_total
+    choices.inject(0) { |sum, x| sum + menu[x] }
+  end
+
+  def validate_order?
+   return false if real_total != total_estimate
+   true
+  end
```
sjmog: Lovely use of guard conditions

sjmog: Does these need to be a public method? I feel like it could be private without too much fuss

lib/customer.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def create_order order
+    return order if validate_order?
+    raise "Wrong total"
+  end
```
sjmog: Swap these for guard purposes:

```ruby
raise "Wrong total" unless validate_order?
order
```

sjmog:
This is actually really good, @haroonhassan . The code is generally fine, mostly TDD compliant, and it seems to work (as well as demonstrating good understanding of OOD).

Check my comments, and get that gem installed. As a hint, you'll probably want to split the text sending mechanism into an `SmsSender` class that sets itself up in the initializer, and has a `send_text` message. You can then inject this class into your `send_sms` method in `Order` once an order is ready to send.

If you do get a chance to refactor, do so: there's a lot to like here, especially readability, and I'm excited to see you finish it.

:construction:

PR #119, opened 2015-08-03
takeaway-challenge

ajitsy:
Not the best attempt. I'm confident I can finish this just needed some more time.

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,9 @@
+module Menu
+
+  def menu
+    
+    @takeaway_menu.each do | item, price |
+        puts "#{item}: £#{price}"
+    end
+  end
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,27 @@
+require_relative 'menu'
+
+class Takeaway
+  include Menu
+
+  def initialize
+    @final_order = []
+    @hashed_order = {}
+    @takeaway_menu = { "pizza" => 3, "curry" => 4, "chopsuey" => 2 }
+  end
+
+  attr_reader :final_order, :takeaway_menu, :hashed_order
+
+  def order(dish, qty)
+    fail 'That is not on the menu!' unless dish == "curry" or dish == "chopsuey" or dish == "pizza"
+    final_order << [dish, qty]
+    @hashed_order = Hash[*final_order.flatten(1)]
+  end
+
+  def cost
+    hashed_order.my_inject(0) do | total_cost, order |
+      order.each { | item, qty | total_cost = total_cost + @takeaway_menu[item] * qty }
+      total_cost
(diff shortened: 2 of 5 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+module Menu
+
+  def menu
+    
+    @takeaway_menu.each do | item, price |
+        puts "#{item}: £#{price}"
+    end
+  end
+end
```
danldb: Module referring to an instance variable is definitely a code smell - this module can only be included in a class that has a @takeaway_menu  variable. I'm not sure of the advantage of using a module for this. On the other hand, this could be a class that is instantiated and is responsible for collecting menu items.

lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
...
+    @final_order = []
+    @hashed_order = {}
+    @takeaway_menu = { "pizza" => 3, "curry" => 4, "chopsuey" => 2 }
+  end
+
+  attr_reader :final_order, :takeaway_menu, :hashed_order
+
+  def order(dish, qty)
+    fail 'That is not on the menu!' unless dish == "curry" or dish == "chopsuey" or dish == "pizza"
+    final_order << [dish, qty]
```
danldb: You are not currenlty checkint that any of these items are actually on the menu - what happens if the menu changes and no longer has chopsuey? Your fail clause would no longer be accurate.

lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
...
+    @hashed_order = Hash[*final_order.flatten(1)]
+  end
+
+  def cost
+    hashed_order.my_inject(0) do | total_cost, order |
```
danldb: my_inject

lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
...
+      order.each { | item, qty | total_cost = total_cost + @takeaway_menu[item] * qty }
+      total_cost
+    end
```
danldb: refactor  - this is doing too many things and break SRP.

spec/feature_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'takeaway'
+require 'menu'
+
+ describe "Features" do
+   let(:takeaway) { Takeaway.new }
+   let(:x) { {"pizza"=>3, "curry"=>4, "chopsuey"=>2} }
```
danldb: `x`?!?!

do not ever use `x` as a name - names should be descriptive! Why not call it `items` or `menu` or anything at all, just not `x`.

danldb:
:construction: 

Good attempt, but not quite there yet as you ran out of time. 

Fix your indentation, write a README, complete the SMS section and we can have another look at this.

PR #128, opened 2015-09-12
Harriet's Take Away Challenge

harrietc52:


tansaku:
Great effort here Harriet and a very good base to work from.  Great to see you wrestling to fix the hound issues - keep going! 

Some comments and suggestions for moving forward:

* test coverage low ~10% which is not a good sign - but I think it's mainly so low because the build failed trying to load env.rb.  I'd recommend using dotenv instead of env.rb
* when I check locally you've actually got quite good code coverage (~95%), and the only pieces missing are related to the text sending which we talked about in group session on Monday :-)
* the Menu class doesn't have many responsibilities - it could probably take over some of the things that Customer is doing
* You're Delivery class is one of the best I've seen in terms of splitting out the components, although there is still room for improvement with indenting, and the delivery_time method feels like it could be split out as the code smell tool reek suggests:

  [13]:Delivery#delivery_time doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]

* I'm thinking that maybe the Delivery class should perhaps be called something.  Seems like currently it has two responsibilities - calculating the delivery time and texting the customer - perhaps they could be split over two classes.  If you wanted to get advanced the operations relating to time could be duck-punched (or monkey-patched) into the Time class itself! :-)

* you're using `each` in menu.rb and I don't think it's doing anything.  Effectively you have a method here that is simple a reader of the dishes hash

* note that tests like this:

```
    it { is_expected.to respond_to(:place_order).with(2).argument }

```
become somewhat redundant once you are actively testing the method like so:

```
    it "raises an error if chosen dish is not on menu" do
     expect{ subject.place_order("spring rolls", 3) }.to raise_error
     "This dish is not on the menu"
    end
```

note also that breaking the line as you do above creates problem - ruby considers them two separate lines.  Note also that you are only 1 space indented here - you need something like the following:

```
    it "raises an error if chosen dish is not on menu" do    
      message = "This dish is not on the menu"
      expect{ subject.place_order("spring rolls", 3) }.to raise_error message
    end
```

although there are many ways to restructure to reduce line length (you could switch `{ ... }` to `do ... end` or introduce procs) the above is probably the simplest in this instance

* your Customer#place_order is a little long you can do things like initializing your Hash with default zero `Hash.new(0)` to simplify your logic here

* I'm not so keen on the `attr_writer` feels - like you are opening up some access to your customer internals that is not quite necessary

* Ideally you'd update your README with some explanation of how to get set up with the project - what kind of experience one might expect from interacting with irb - a little transcript is a great move - it's important to be thinking about how someone else will be able to get set up to use your code.

* Your Customer class is locked to Delivery - you could consider loosening that up with a little injection of dependencies, but I'd focus on other lower hanging fruit first

You might be tempted to re-start, but I would strongly recommend trying to refactor your existing codebase to improve.  Do so and I can potentially hook you up with sticker :-)

:construction:

PR #129, opened 2015-09-12
Jongmin's takeaway-challenge

jongmin141215:


jongmin141215:
I have many tests checking output (and format of the output), which I think, has to be changed fundamentally. I'll work on a new takeaway challenge in the lab week(?).

tansaku:
all tests passing - that's great!  Some really good work here.  Some notes to help you improve

* you have some additional output spilling into your test output:


```sh
→ rspec -fp
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
.What would you like to order? Please enter a menu item number.
How many of dishes do you want?
.What would you like to order? Please enter a menu item number.
How many of dishes do you want?
.What would you like to order? Please enter a menu item number.
How many of dishes do you want?
..What would you like to order? Please enter a menu item number.
.How many of dishes do you want?
..

Finished in 0.00938 seconds (files took 0.37873 seconds to load)
8 examples, 0 failures
```

* ideally you would ensure that no additional output leaks into the tests.  You have puts statements inside your objects like so:

```
  def show
    @menu.map.with_index do |item, index|
      puts "#{index + 1}. #{item[:dish]}: £#{item[:price]}"
    end
  end
```

it's okay to have a puts statement for debugging, but in general you should always return strings and display any output from a separate runner class.  Menu#show should return a string for the menu - not send it to STDOUT via puts.  You're doing some quite sophisticated testing to handle that:

```
describe Menu do
  it "diplays menu items" do
    expect { subject.show }.
      to output("1. BBQ: £10\n2. Hamburger: £8\n" <<
        "3. Pizza: £9\n4. Pasta: £7\n5. Sushi: £6\n").to_stdout
  end
end
```

and it's great to see you handling this complexity, but you really want to keep your classes free of commitment to a particular output method - each classes instance methods should be returning strings and you should be testing that output directly rather than testing the `output`

* you also have commented code in your Menu class - that needs to be purged before any submission

* you've updated your README and you're using dotenv, which is totally awesome - well done; however the README should show a transcript like the following to make it obvious what to do:

```sh
2.2.2 :001 > m = Menu.new
 => #<Menu:0x007faeb3bd4e30 @menu=[{:dish=>"BBQ", :price=>10}, {:dish=>"Hamburger", :price=>8}, {:dish=>"Pizza", :price=>9}, {:dish=>"Pasta", :price=>7}, {:dish=>"Sushi", :price=>6}]> 
2.2.2 :002 > m.show
1. BBQ: £10
2. Hamburger: £8
3. Pizza: £9
4. Pasta: £7
5. Sushi: £6
 => [nil, nil, nil, nil, nil] 
2.2.2 :003 > o = Order.new
 => #<Order:0x007faeb55df898 @orders=[]> 
2.2.2 :004 > o.choose_dish 1
ArgumentError: wrong number of arguments (1 for 0)
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/order.rb:24:in `choose_dish'
	from (irb):4
	from /Users/tansaku/.rvm/rubies/ruby-2.2.2/bin/irb:11:in `<main>'
2.2.2 :005 > o.choose_dish
What would you like to order? Please enter a menu item number.
1
 => 1 
2.2.2 :006 > o.choose_how_many
How many of dishes do you want?
3
 => 3 
```

see the mistake I made above trying to follow your instructions? If I have a transcript I know exactly what to do and I don't have to guess.

* You've gone for an interactive text interface, which you really don't want in irb, since that is already interactive - this kind of interactive interface is cool, but it should be launched from a takeaway.rb file in the root dir that would start the system up and then manage the interaction with the user.

* you've got a pretty good level of test coverage, except that line 21 or order.rb is not covered, and none of the texting is covered, but we talked about that in group session on Monday

* the code smell checker [Reek](https://github.com/troessner/reek) (`gem install reek ; reek -U`) finds the following issues with your code:

```sh
→ reek -U
lib/order.rb -- 4 warnings:
  [45, 46]:Order#check_orders calls order[:price] 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [44, 46]:Order#check_orders calls order[:quantities] 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [20]:Order#present doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]
  [65, 65]:Order#total_price refers to order more than self (FeatureEnvy) [https://github.com/troessner/reek/blob/master/docs/Feature-Envy.md]
lib/text.rb -- 1 warning:
  [7]:Text#self.send_text_message has approx 8 statements (TooManyStatements) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Statements.md]
5 total warnings
```
I think this is a reflection of Customer having too many responsibilities - feels like you could refactor out another class there - perhaps Order?

:construction:

PR #130, opened 2015-09-12
Takeaway challenge

sarahkristinepedersen:


tansaku:
Some great work here @sarahkristinepedersen a very solid base.  There are however quite a few things you could improve:

* Please do include all the gems you use in your Gemfile. This is an important courtesy to other developers and yourself in the future, so that all the project dependencies can be pulled in whenever the project is checked out on a new machine, e.g. `twilio-ruby` gem

* You shouldn't need to be explicit about paths in RSpec files pulling in the associated application:

```
require "./lib/restaurant"
```

can be replaced with the exception of dish_spec, all of your tests are working with multiple models (chicago or integration)  

```
require "restaurant"
```

* I notice that with the exception of dish_spec you have multiple different objects being created in each test.  That's okay, but note that if you create real objects in your tests other than that which is the subject, then you are  creating an integration test (or using the Chicago style).  In general you want to separate up your unit from your integration/feature tests.  Unit tests can just rest in the root of the spec folder, but features of integration tests should go in a subfolder (spec/features or spec/integration) or even in a separate folder on the root directory to allow them to run completely separately.

If you are using Chicago style unit tests where you are using other real classes apart from the subject under test, you  should be trying to use them as sparingly as possible.

* The subject method is intended to allow variations on the setup of the class under test.  So this is fine:

```ruby 
describe Restaurant do
  subject { Restaurant.new(menu, notification_service) }
end
```

However other setup should ideally not be in the subject:

```ruby
subject do
  dishes = [Dish.new("Burger", 8), Dish.new("Pizza", 10)]
  menu = Menu.new(dishes)
  notification_service = instance_double "Notificationservice"
  allow(notification_service).to receive(:send_notification)
  Restaurant.new(menu, notification_service)
end
```

The above is not an unreasonable start, but ultimately we would like it split out as follows:

```ruby 
describe Restaurant do
  let(:notification_service) { instance_double "Notificationservice" }
  before { allow(notification_service).to receive(:send_notification) }
  let(:dishes) { [Dish.new("Burger", 8), Dish.new("Pizza", 10)] }
  let(:menu) { Menu.new(dishes) }
  subject { Restaurant.new(menu, notification_service) }
end
```

which would allow the other elements to be re-used through other tests

* The [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) suggests that:

The following test shows a process of reaching through a series of related objects.  The warning sign is the multiple periods in `subject.menu.dishes.length`.  Here we are seeing Restaurant is being tested for properties that belong to the menu - effectively they are none of restaurant's business and shouldn't be tested here; and we shouldn't be seeing chains like this in application code either

```ruby
describe Restaurant do
  it "has a menu" do
    expect(subject.menu.dishes.length).to eq(2)
  end
end
```

* Here also for your edification are some issues I got from running the [reek](https://github.com/troessner/reek) (`gem install reek ; reek -U`) code smell tool on it:

>  [10]:Restaurant#submit_order has 4 parameters (LongParameterList) [https://github.com/troessner/reek/blob/master/docs/Long-Parameter-List.md]
  [10]:Restaurant#submit_order has approx 7 statements (TooManyStatements) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Statements.md]
  [16]:Restaurant#submit_order is controlled by argument expected_sum (ControlParameter) [https://github.com/troessner/reek/blob/master/docs/Control-Parameter.md]

* You also have some outstanding hound/rubocop issues, and here's some links to more details on those in the Ruby Style guide:

> lib/restaurant.rb:17 I: Style/LineEndConcatenation: Use `\` instead of `+` or `<<` to concatenate those strings.
spec/order_spec.rb:27 I: Style/DotPosition: Place the . on the next line, together with the method name. https://github.com/bbatsov/ruby-style-guide#consistent-multi-line-chains
spec/restaurant_spec.rb:40 I: Style/DotPosition: Place the . on the next line, together with the method name. https://github.com/bbatsov/ruby-style-guide#consistent-multi-line-chains
spec/restaurant_spec.rb:57 I: Style/DotPosition: Place the . on the next line, together with the method name. https://github.com/bbatsov/ruby-style-guide#consistent-multi-line-chains
spec/restaurant_spec.rb:59 I: Metrics/LineLength: Line is too long. [94/80] (https://github.com/bbatsov/ruby-style-guide#80-character-limits)

Overall some a good foundation here.  Don't be intimidated by the above list, you've got a great base here, but I want to give you all the help I can to take things to the next level :-)  There are a few other things as well, but have a go at fixing up some of the above, we'll do another round of feedback and then and we'll see about getting you a sticker :-)

Please do ask if anything is unclear.

:construction:

PR #133, opened 2015-09-12
Takeaway challenge

lucetzer:


tansaku:
Tests all green - great work! And a really solid base here.

Some comments to help you improve:

* note that you are getting a couple additional outputs in your rspec that you should clean out - you can see them clearly with rspec -fp

```sh
$ rspec -fp
/Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/.env.rb:1: warning: already initialized constant ENV
.WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/spec/customer_spec.rb:9:in `block (2 levels) in <top (required)>'.
..........Hello! Here is our menu!
chicken katsu: £3.50
beef udon: £4.00
crispy fish: £5.50
miso soup: £1.50
...

Finished in 0.01266 seconds (files took 0.28042 seconds to load)
14 examples, 0 failures
```
these are obviously not preventing your tests from passing, but you should be clearing all these out.

* the first output is that you are overriding the ENV hash, which is generally a bad idea since there may be other ENV vars that you need.  I recommend using the dotenv gem to handle the ENV var you need - note also that all your credentials are on github and need to be removed for security purposes https://help.github.com/articles/remove-sensitive-data/

* the second - is that you broke a line in such a way that it became two lines

```ruby
  it "raises error if customer selects a dish that's not on the menu" do
    expect{ subject.make_order("pasta", 1) }.to raise_error
    "Sorry! That dish is not on the menu"
  end
```

ruby treats the above as two separate lines (effectively ignoring the latter) - you should prefer something like

```ruby
  it "raises error if customer selects a dish that's not on the menu" do
    message = "Sorry! That dish is not on the menu"
    expect{ subject.make_order("pasta", 1) }.to raise_error message
  end
```

or

```ruby
  it "raises error if customer selects a dish that's not on the menu" do
    expect do 
      subject.make_order("pasta", 1) 
    end.to raise_error "Sorry! That dish is not on the menu"
  end
```

or extract a proc if you want to get fancy ...

* 3rd you have a puts statement hanging around:

```ruby
  def menu_pricelist
    puts "Hello! Here is our menu!"
    dishes.each { |dish, price| puts "#{dish}: £#{price}0" }
  end
```

which you want to avoid where at all possible.  Return the strings you want displayed and then possibly have a base runner ruby file to do the puts - avoid having puts in any of your classes when you submit your code (fine for on the spot debugging, but clean them out for submission) 

* ideally you'd update your README to explain to anyone who wants to use it how to use it, e.g. show how one would use it in irb, what env vars need to be set up etc.

* for some reason the test coverage tool didn't run on your build (not sure why), but I ran locally and your coverage is pretty good (97%) - only thing not covered is twilio and we talked about how to cover that on Monday :-)

* you've got your dependencies nicely injected, but it feels like customer has too many responsibilities (and conversely menu too few); for example `dish_available?` could be a method in menu rather than customer

* your make_order method could be simplified by creating a hash with a default value `Hash.new(0)`

* you're exposing more public interface than you need `attr_reader :menu, :send_text` could all be private (ensuring you prefer `menu` to `self.menu`)

* lots of nice short methods here; good use of reduce, however your cellphone number is now on github :-/  again see https://help.github.com/articles/remove-sensitive-data/

:construction:

PR #135, opened 2015-09-13
Michael Roger - Takeaway Challenge

Benaud12:


tansaku:
Great base Michael - tests all green - that's great, and 95% test coverage well done - the remaining 5% is the SMS stuff we talked about last week.  Some tips to help you improve:

```
2.2.3 :001 > order = Order.new
 => #<Order:0x007fe3830d27f8 @basket={}> 
2.2.3 :002 > dish = Dish.new('chips',100)
 => #<Dish:0x007fe3868054c8 @name="Chips", @price=100.0> 
2.2.3 :003 > menu = Menu.new
 => #<Menu:0x007fe385811440 @menu={}> 
2.2.3 :004 > menu.make_dish_available(dish)
 => #<Dish:0x007fe3868054c8 @name="Chips", @price=100.0> 
2.2.3 :005 > order.add_to_basket(menu, 1, 2)
 => 2 
2.2.3 :006 > order.summary
2x Chips | £200.00
Total cost: £200.00
 => nil 
2.2.3 :007 > order.submit(200)
ArgumentError: Account SID and auth token are required
	from /Users/tansaku/.rvm/gems/ruby-2.2.3/gems/twilio-ruby-4.3.0/lib/twilio-ruby/rest/base_client.rb:32:in `initialize'
	from /Users/tansaku/.rvm/gems/ruby-2.2.3/gems/twilio-ruby-4.3.0/lib/twilio-ruby/rest/client.rb:125:in `initialize'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/texting.rb:18:in `new'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/texting.rb:18:in `send_text'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/order.rb:47:in `issue_confirmation_text'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/order.rb:33:in `submit'
	from (irb):7
```

The above is really helpful to a user getting setup - also do include notes about environment variables they will need to set up etc.

* I notice that you also included your inject challenge in here - ideally that would be in a separate pull request on the inject challenge repo:

https://github.com/makersacademy/inject-challenge

* As we can see from the IRB interaction we kind of have to assemble lots of objects and pass them in in the correct order to make things work.  It might make it easier on our collaborator objects to set some of those options as defaults in initialize that can be overriden.


* Avoid naming instance variables the same as the class name. Otherwise you risk asking class instances this sort of thing:

```ruby
menu = Menu.new
# Hey menu, what's your...menu?
menu.menu # :( weird
```

in your case you're having Menu wrap around a hash - you could either inherit from Hash, or pass on requests using delegation ...

* The [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) suggests that:

The following test shows a process of reaching through a series of related objects.  The warning sign is the multiple periods in `subject.menu.dishes.length`.  Here we are seeing Menu is being tested for properties that belong to the dish - effectively they are none of menu business and shouldn't be tested here; and we shouldn't be seeing chains like this in application code either

```ruby
describe Menu do
  it 'adds a new dish to the menu' do
    subject.make_dish_available(dish1)
    expect(subject.menu.values.include?(dish1)).to be true
  end
end
```

* I'm very pleased to see how you've broken up the private and public methods (in Menu)

* Note that you should generally prefer to avoid having `puts` in your classes or modules.  If there is output to present on the command line we should prefer to return it from the method and have all output displayed from some ruby code in a base takeaway.rb file or similar, i.e. one that doesn't have classes/modules.  The idea here is 'separation of concerns' - the same reason why we try not to mix different languages (e.g. ruby and html) wherever possible.  We don't want to lock our business logic to a particular output representation, i.e. in this case takeaway and customer are business objects and so they shouldn't do anything that locks them to STDOUT/STDIN i.e. prefer not to use gets and puts inside those classes/modules


* Your Order#add_to_basket method could be simplified by creating a hash with a default value `Hash.new(0)`

* I notice Order locked to Texting - perhaps you could make that relationship a little more relaxed using some Dependency Injection?

* Please do also look into your outstanding hound issues

* Finally for your edification are some issues I got from running the [reek](https://github.com/troessner/reek) (`gem install reek ; reek -U`) code smell tool on it:

> $ reek -U
lib/menu.rb -- 1 warning:
  [15, 22, 38]:Menu tests no_dishes_available? at least 3 times (RepeatedConditional) [https://github.com/troessner/reek/blob/master/docs/Repeated-Conditional.md]
lib/order.rb -- 3 warnings:
  [15, 17]:Order#add_to_basket calls dish.price 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [46]:Order#issue_confirmation_text doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]
  [32]:Order#submit is controlled by argument payment (ControlParameter) [https://github.com/troessner/reek/blob/master/docs/Control-Parameter.md]
lib/texting.rb -- 1 warning:
  [7]:Texting#self.send_text has approx 8 statements (TooManyStatements) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Statements.md]
7 total warnings


Overall some a good foundation here.  Don't be intimidated by the above list, you've got a great base here, but I want to give you all the help I can to take things to the next level :-) Have a go at fixing up some of the above and we'll see about getting you a sticker :-)

:construction:

PR #136, opened 2015-09-13
Emily's takeaway challenge

emilyworrall:


lib/customer.rb
```diff
@@ -0,0 +1,33 @@
+require_relative 'order'
+
+class Customer
+  include Menu
+  include Text
```
sjmog: Oooh, sexy

sjmog: Love the composition here

lib/customer.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def check_menu
+    display_menu
+  end
```
sjmog: this method just calls another, similarly-named one. I think you can lose it

lib/customer.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def verify_and_pay(order, total_price)
+    if order.order_total(total_price) == true
```
sjmog: No need to check `== true` as anything other than `nil` or `false` will evaluate to true.

This includes `0`, bear in mind.

lib/customer.rb
```diff
@@ -0,0 +1,33 @@
...
+      self.order_complete
+    else
+      fail 'Payment failed, please pay correct total.'
```
sjmog: Try reorganising this to use guard conditions (i.e. raise early):

```ruby
def verify_and_pay(order, total_price)
  raise 'Payment failed, please pay correct total.' unless order.order_total(total_price)
  order_complete
end
```

lib/customer.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+  end
+
+  def order_complete
+    send_text(text_content)
```
sjmog: I'm not sure `Customer` should be sending texts. That seems like the job of something else.

You use composition to `include` the `Text` module, but I think you should have a `Text` class and inject it into `order_complete`.

lib/menu.rb
```diff
@@ -0,0 +1,14 @@
...
+      pepperoni_pizza: 5.00,
+      spaghetti_pomodoro: 6.00,
+      spaghetti_carbonara: 7.50
+    }
+
+    dish_list.each do |dish, price|
+      puts "#{dish}: £#{'%.2f' % price}"
+    end
+  end
+end
```
sjmog: I think this should be a class. You can call objects from within other objects!

lib/order.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'text'
+require_relative 'menu'
+
+class Order
+  include Menu
```
sjmog: This is legitimate. Orders should maybe know about their menus in this way, and it's conceivable that an order would also be a menu. Conceivable, but not too good. (I still think Menu should be a class)

lib/order.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  attr_accessor :contents
```
sjmog: :)

lib/order.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+  end
+
+  def order_total(total_price)
+    prices = []
+    order = Hash[@contents.each_slice(2).to_a]
+    order.each do |dish, quantity|
+    prices << display_menu[dish]*quantity
+    end
+    if total_price == prices.inject(:+)
```
sjmog: I'm a little confused as to why you are checking this. Plus, it seems a little odd that `order_total` returns `true` or `false`, when it sounds like it should return a number.

PR #138, opened 2015-09-13
Usman - Takeaway Challenge

UsmanJ:


tansaku:
hi @UsmanJ I notice you have some unsynced changes - can you follow the instructions here https://help.github.com/articles/syncing-a-fork/ to get the build running?

tansaku:
Hey Usman, some good work here, but as we were discussing your build broke due to this error:

```
→ rspec
/Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:4:in `require_relative': cannot load such file -- /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/.env.rb (LoadError)
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:4:in `<top (required)>'

```

* Did we already talk about preferring the dotenv gem to .env.rb?  Either way do check out https://github.com/bkeepers/dotenv  

* By commenting out the `require_relative '../.env.rb'` line in customer.rb I'm able to run your tests, see a single failure:

```sh
  1) Customer sends a message
     Failure/Error: subject.send_message("+447415121250")
     TypeError:
       no implicit conversion of Symbol into String
     # ./lib/customer.rb:42:in `[]'
     # ./lib/customer.rb:42:in `send_message'
     # ./spec/customer_spec.rb:50:in `block (2 levels) in <top (required)>'
```

and then I can see your test coverage is at 94% which is pretty good, and that 6% gap is due to the Twilio stubbing issues we talked about last week; I see you've attempted to address that, but something hasn't worked quite right (or SimpleCov is breaking somehow) - either way you've got funky indenting in both the spec and code for send_message - do get that fixed up and check that area again.

* The Gemfile can lock the project to a particular version of Ruby.  In general you should prefer the latest, currently 2.2.3

* The above is a relatively straightforward thing to do that doesn't involve much programming - I'll often get it done while thinking about other problems in the back of my mind :-)  It's also a great idea to show the full story of how your app is used (from a user's perspective) in the README, e.g. an irb transcript like:

```sh
$ irb
2.2.3 :001 > customer = Customer.new
 => #<Customer:0x007fbc3424b6c8 @menu=#<Menu:0x007fbc3424b6a0 @menu_list={"4x chicken wings"=>1.59, "chicken burger meal"=>4.49, "fries"=>1.29, "pizza"=>4.99}>, @order={}, @order_total=0, @total_items=0> 
2.2.3 :002 > customer.select_dish("chicken burger meal", 2)
 => 8.98 
2.2.3 :003 > customer.order
 => {"chicken burger meal"=>2} 
2.2.3 :004 > customer.total_items
 => 2 
2.2.3 :005 > customer.receipt
 => "2x chicken burger meal - £8.98" 
2.2.3 :006 > customer.display_menu
4x chicken wings - £1.59
chicken burger meal - £4.49
fries - £1.29
pizza - £4.99
 => {"4x chicken wings"=>1.59, "chicken burger meal"=>4.49, "fries"=>1.29, "pizza"=>4.99} 
2.2.3 :007 > customer.order_total
 => 8.98 
2.2.3 :008 > customer.order_total_cost
 => "Total    £8.98" 
2.2.3 :009 > customer.send_message("+447415121250")
TypeError: no implicit conversion of Symbol into String
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:42:in `[]'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:42:in `send_message'
	from (irb):9
```

* note that tests like this:

```ruby
  it {is_expected.to respond_to(:select_dish).with(2).argument}

```
become somewhat redundant once you are actively testing the method like so:

```ruby
  it 'should not be able to add dishes not on the menu' do
    expect{subject.select_dish("cars", 3)}.to raise_error 'Dish not on menu'
  end
```

* Prefer to use the symbol hash format https://github.com/bbatsov/ruby-style-guide#symbols-as-keys (and anything else that hound is telling you)

* You've got an unused attr_reader :menu in Menu - please ensure you remove any code not necessary for the functioning of your system

* Similarly in menu_spec you've got a useless single line call to `subject` - this doesn't break your tests, but it's the kind of sloppiness that could well prevent you from getting through a tech test to a job interview.  Please remove any unneeded codage anywhere in tests or app

* You've got heavy and excessive use of attr_accessor in customer.rb, which you really don't need - do avoid exposing any more object state than is strictly necessary (particularly the ability to manipulate it)

* Nice guard clause `fail 'Dish not on menu' unless menu.menu_list.key?(dish)` but again watch your indenting; must be 2-space https://github.com/bbatsov/ruby-style-guide#spaces-indentation

* Nice use of private methods, but send_message is very long - see your hound/rubocop comments - and also this from the the [reek](https://github.com/troessner/reek) (`gem install reek ; reek -U`) code smell tool:

>   [6]:Customer has at least 5 instance variables (TooManyInstanceVariables) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Instance-Variables.md]

* The biggest overall problem with your code is that you've got too many responsibilities all shoved in to Customer, and it is also locked to the Menu class which is basically a simple container.  Some of Customer's responsibilities, e.g. `display_menu` should be moved over to give Menu control over the right kind of things, and the text messaging should be extracted as a module or similar.  Then both should be passed to the Customer via Dependency Injection (DI)

One way to do this would be to have Customer work like so:

```ruby
require 'menu'

class Customer

  def initialize(klass = Menu)
    @menu = klass.new
  end

  def display_menu
    @menu.display
  end
```

* Your phone number is now on github see https://help.github.com/articles/remove-sensitive-data/ to remove it, and prefer ENV variables and possibly also the dotenv gem in order to avoid this in future

* Note that you should generally prefer to avoid having `puts` in your classes or modules.  If there is output to present on the command line we should prefer to return it from the method and have all output displayed from some ruby code in a base takeaway.rb file or similar, i.e. one that doesn't have classes/modules.  The idea here is 'separation of concerns' - the same reason why we try not to mix different languages (e.g. ruby and html) wherever possible.  We don't want to lock our business logic to a particular output representation, i.e. in this case takeaway and customer are business objects and so they shouldn't do anything that locks them to STDOUT/STDIN i.e. prefer not to use gets and puts inside those classes/modules  - fixing this issue will also remove some messy text output from your rspec runs.

Overall a great foundation here.  Don't be intimidated by the above list, this is a good submission, but I want to give you all the help I can to take things to the next level :-) Have a go at fixing up some of the above and we'll see about getting you a sticker :-)

:construction:

PR #140, opened 2015-09-13
Harry Venables Takeaway Challenge

hvenables:


hvenables:
Code reviewed. Great job!

tansaku:
Great foundation here!

Some comments:

* your phone and sid are on github even if your auth_token is not - I prefer dotenv to env.rb myself
* phone and sid should be in ENV vars
* whatever you are doing you need instructions in your README to explain what someone who downloads your code should do to get setup and run the tests.  I have to create this file:

```
module Env
end
```

in env.rb in order to even run the tests.  Also this set up prevents the CI from running and checking your build.

* you have assorted indenting issues

```
describe Restaurant do
subject { Restaurant.new(Menu, pizza: 5.99, burger: 4.99, chip: 2.99, sandwich: 3.99)}
```

must be

```
describe Restaurant do
  subject { Restaurant.new(Menu, pizza: 5.99, burger: 4.99, chip: 2.99, sandwich: 3.99)}
```

rule of thumb is the internals of every block of code '{ ... }' 'do ... end' 'class ... end' 'def ... end' must be indented two spaces; and it seems trivial, but you need to really start caring about this if you want a job

* overall your test coverage is dangerously low: https://coveralls.io/builds/3559227 but that's probably related to the build not passing - I've checked locally with simplecov and in fact you have 94% coverage which is not too bad
* Restaurant class itself looks pretty good in terms of small public interface, short methods etc. 


* great to see you fixed up some of the hound comments, but you still have some very long lines and some other issues that hound picked up that you need to address

* As we've discussed your module tests, e.g. for Order need to be split out into shared_examples https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples , since at the moment they are bloating your restaurant_spec


* note that tests like this:

```
    it 'should respond to receipt' do
      is_expected.to respond_to(:receipt)
    end
```

```
    it { is_expected.to respond_to(:receipt) }
```

```
    it 'should show customer total cost' do
      subject.make_order(:pizza, 2)
      time = (Time.new + 3600).strftime("%H:%M")
      message = "Thank you! Total cost: £11.98 - pizza: £11.98. It should arrive before #{time}"
      expect(subject.receipt).to eq(message)
    end
```

The code smell analyzer reek says the following about restaurant:

  [28]:Restaurant#arrival_time doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]

indicating that this could also be moved into a module

* I can tell from simplecov output that lines 6 and 7 of order.rb are not being covered by your existing tests, and of course there is very limited test coverage on text.rb, but we talked about that in group session on Monday

Reek also has the following to say about Order

  [7, 10]:Order#make_order calls quantity > 1 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [3]:Order#make_order has approx 7 statements (TooManyStatements) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Statements.md]
lib/restaurant.rb -- 1 warning:

* You've got a great dependency injection set up for your relationship between menu and restaurant, but reeks got an issue with:

  [18, 19]:Menu#list refers to menu more than self (FeatureEnvy) [https://github.com/troessner/reek/blob/master/docs/Feature-Envy.md]

https://github.com/troessner/reek

```
gem install reek
reek -U
```
will give you reek comments locally.

tansaku:
:construction:

PR #141, opened 2015-09-13
Tim Hyson - Takeaway Challenge

timhyson:


tansaku:
Good start Tim.

Some comments about your code and my recommended next steps

* you have hound issues should be fixed up
* you have non-ascii single quotes in your Gemfile that are breaking the build, which means you didn't get a test coverage report.  Fix that up and you'll get a better evaluation of your tests.
* checking the coverage locally I see that you only have 74% test coverage which is kind o f low - lines 21 and 22 of menu.rb are not tested, and the majority of the methods in customer.rb are not tested.  Similarly for twilio.rb, but of course we talked about how to test that as a group on Monday
* you're using each in menu.rb and I don't think it's doing anything.  Effectively you have two methods here that are simply readers for the dishes hash
* ideally you'd be updating your README with some instructions for me and other developers who might want to use your code, e.g. show us how this all works on irb
* you've got some classes that you were intended to use, but didn't like Takeaway.  For submission of a tech test you'd need to ensure these were cleared away
* here's what the code smell analyzer tool reek says about your code

lib/customer.rb -- 1 warning:
  [11]:Customer#display_menu doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]
lib/twilio.rb -- 1 warning:
  [17]:TwilioAPI#delivery_time doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]

this utility function smell is an indicator that you should have these methods in modules rather than classes, and perhaps should be class methods.  An advanced (and possibly more risky approach) would be to monkey patch Time to create Time#delivery_time.  In the case of the Customer#display_menu method you've got the Customer locked to Menu (which you could fix with Dependency Injection), but more importantly you should be handing off the display of the menu to the Menu#display_menu method rather than repeating the operation in Customer#display_menu which is not very DRY.

Overall my intuition is that you need to get clear with what experience you want the end user to have in irb and then focus on supporting that.

You might be tempted to re-start, but I would strongly recommend trying to refactor your existing codebase to improve.  Do so and I can potentially hook you up with sticker :-)

:construction:

PR #142, opened 2015-09-13
Zhivko Siderov: Take Away

zsid:


tansaku:
Tests are green, coverage at 93% - pretty good work.  

Some thoughts: 

* some hound issues to fix
* Please update your README; in particular I'd like to see an irb transcript to make it crystal clear how to use the code
* You're using dotenv which is excellent, but you are using dotenv-rails, which you don't need just yet - stick with simple dotenv for the moment
* You've got your responsibilites distributed across classes and modules which is great, however the Restaurant is hard coded to a particular menu module, which you could loosen up with a bit of dependency injection
* great to see you using private to reduce the number of public methods, improve encapsulation
* the SendText operations could be pulled out into a separate module for further separation  of responsibilities, as could estimated_delivery_time, which could even be monkey-patched into the Time class (if you were feeling brave!)
* I'd love to see separate tests for the menu module using the shared_example format: https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples
* Customer#menu and Customer#submit_order are not being tested (as we can see from the coverage) - are they actually needed?
* most of the restaurant code is not actually being exercised by the tests - of course this relates to our Monday afternoon discussion of how to test/stub the twilio API, so hopefully you should have an idea of how to improve this now

Overall some really good work here.  Don't be intimidated by the above list, you've done really well, but I want to give you all the help I can to take things to the next level :-)


:construction:

PR #148, opened 2015-09-13
takeaway-challenge 

catherinestevenson:


tansaku:
A good base here Catherine.  Here's some tips:

* Please update your README following the [contribution notes](https://github.com/makersacademy/takeaway-challenge/blob/master/CONTRIBUTING.md), i.e.

* Also it's a great idea to show the full story of how your app is used (from a user's perspective) in the README, e.g.

```sh
→ irb
2.2.3 :001 > c = Customer.new
 => #<Customer:0x007fdcdd810300 @order_basket={}, @order_total=0, @total_items=0> 
2.2.3 :002 > c.view_menu
pizza: £10.0
burrito: £6.0
burger: £7.0
chicken: £5.0
chips: £2.0
 => {"pizza"=>10.0, "burrito"=>6.0, "burger"=>7.0, "chicken"=>5.0, "chips"=>2.0} 
2.2.3 :003 > c.select_item("burrito", 3)
pizza: £10.0
burrito: £6.0
burger: £7.0
chicken: £5.0
chips: £2.0
pizza: £10.0
burrito: £6.0
burger: £7.0
chicken: £5.0
chips: £2.0
 => 3 
2.2.3 :004 > c.receipt
 => "Your order has 3 items with a total of £18.0" 
2.2.3 :005 > c.make_payment
18
RuntimeError: Payment does not match total, please make payment.
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:34:in `make_payment'
	from (irb):5
```

* customer spec is a good start (great use of the include matcher) although you want to be careful about indenting (the hash in include should be indented two spaces) and ideally you should be testing more of the Customer's public interface, e.g. `select_item`, `receipt` and `make_payment`

* takeaway spec if failing because you are trying to test a module directly.  It either needs to be mixed in with a dummy class to be tested, or you should use shared examples: https://www.relishapp.com/rspec/rspec-core/v/3-3/docs/example-groups/shared-examples, however there are various alternatives here.  You could make Takeaway a class and then have Customer refer to that via composition, e.g. just `@takeaway = Takeaway.new` in the customer constructor, although we might then want to clean that up with dependency injection (DI) like so:

```
require 'takeaway'

class Customer

  def initialize(klass = Takeaway)
    @takeaway = klass.new
  end

  def view_menu
    @takeaway.menu
  end
```

don't worry if you're confused by all the options here.  Unfortunately there is no definitive right of wrong answer.  There's a series of different techniques that you'll gradually learn the pros and cons of.  Please don't hesitate to ask for more information on any of these topics and we can look at them in the context of your code.

* On the takeaway code, note that you should generally prefer to avoid having `puts` in your classes or modules.  If there is output to present on the command line we should prefer to return it from the method and have all output displayed from some ruby code in a base takeaway.rb file or similar, i.e. one that doesn't have classes/modules.  The idea here is 'separation of concerns' - the same reason why we try not to mix ruby and html wherever possible.  We don't want to lock our business logic to a particular output representation, e.g. in this case takeaway and customer are business objects and so they shouldn't do anything that locks them to STDOUT/STDIN i.e. prefer not to use gets and puts inside those classes/modules

* great to see a nice set of private methods in customer - you've refactored the `select_item` method so that it's much easier to read and you've got a great guard clause at the top:

```ruby
fail "Sorry, that item is not on the menu." unless menu.key?(item)
```

note you could do similar with the error that potentially gets raised by `make_payment` to shorten up the if/else statement there.

* You've also got a few comments from hound that you should fix up - let me know if you need help interpreting any of them - the list below include links to some descriptions of the issues themselves:

> lib/customer.rb:26 I: Metrics/LineLength: Line is too long. [88/80] (https://github.com/bbatsov/ruby-style-guide#80-character-limits)
lib/customer.rb:29 I: Metrics/AbcSize: Assignment Branch Condition size for make_payment is too high. [15.07/15] (http://c2.com/cgi/wiki?AbcMetric)
lib/customer.rb:29 I: Metrics/MethodLength: Method has too many lines. [8/5] (https://github.com/bbatsov/ruby-style-guide#short-methods)
lib/customer.rb:34 I: Style/SignalException: Use `fail` instead of `raise` to signal exceptions. (https://github.com/bbatsov/ruby-style-guide#fail-method)
lib/customer.rb:36 I: Metrics/LineLength: Line is too long. [92/80] (https://github.com/bbatsov/ruby-style-guide#80-character-limits)
lib/customer.rb:41 I: Style/AccessModifierIndentation: Indent access modifiers like `private`. (https://github.com/bbatsov/ruby-style-guide#indent-public-private-protected)
lib/takeaway.rb:3 I: Metrics/MethodLength: Method has too many lines. [8/5] (https://github.com/bbatsov/ruby-style-guide#short-methods)

Overall some really good work here.  Don't be intimidated by the above list, you've got a great base here, but I want to give you all the help I can to take things to the next level :-)  There are a few other things as well, but have a go at fixing up some of the above, we'll do another round of feedback and then and we'll see about getting you a sticker :-)

:construction:

tansaku:
Great work submitting some changes @catherinestevenson - all tests passing and >98% test coverage!

Here's some further feedback:

* I get the following issue running bundle on your project - easy fix there I think:

```sh
Your Gemfile lists the gem coveralls (>= 0) more than once.
You should probably keep only one of them.
While it's not a problem now, it could cause errors if you change the version of just one of them later.
```

* Wonderful new README!  One thing I might add is a list of any shortcomings you can possibly see, or additional features/improvements you'd like to add

* "New things I learnt/improved on" is nice, although I might be tempted to make it a little paragraph rather than a list, highlighting perhaps where in the codebase you used the Single Responsibility principle, Dependency Injection etc.  Particularly since at the moment your Customer class is arguably breaking SRP.  See the following code fragments from Customer:

```ruby
  def view_menu
    @takeaway.menu.each { |key, value| "#{key} - #{value}"}
  end
``

```ruby
@takeaway.menu.key?(item)
```

notice how both of these fragments involve reaching in to the collaborator class (Takeaway) and manipulating it directly, rather than asking it do to things via public instance methods.  This is breaking the Demeter Principle:

The [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) suggests that:

So the fix here is move `@takeaway.menu.each { |key, value| "#{key} - #{value}"}` over to takeaway itself where it would look like this:


```ruby
class Takeaway

  attr_reader :menu

  def initialize
    @menu = { 'pizza' => 10.00,
              'burrito' => 6.00,
              'burger' => 7.00,
              'chicken' => 5.00,
              'chips' => 2.00 }
  end

  def view_menu
    @menu.each { |key, value| "#{key} - #{value}"}
  end
end
```

and then in Customer you can just have:

```ruby
  def view_menu
    @takeaway.view_menu
  end
```

and if you want to get real fancy you can replace the above with the following

```
class Customer

  extend Forwardable
  delegate view_menu: :takeaway

  ...
end
```

which does approximately the same thing, i.e. when you call view_menu on customer it just gets called on the Takeaway.  That's not something we covered, so don't worry about not knowing this yet.  The basic refactoring here is important - shifting a responsibility from one class with many responsibilies to one that has very few.  The Forwardable/delegate thing above is just a little trick for you to try if you feel like it - feel free to ignore that bit :-)

* I get the following warning running your rspec tests:

```sh
WARNING: Using the `raise_error` matcher without providing a specific error or message risks false positives, since `raise_error` will match when Ruby raises a `NoMethodError`, `NameError` or `ArgumentError`, potentially allowing the expectation to pass without even executing the method you are intending to call. Instead consider providing a specific error class or message. This message can be supressed by setting: `RSpec::Expectations.configuration.warn_about_potential_false_positives = false`. Called from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/spec/customer_spec.rb:17:in `block (2 levels) in <top (required)>'.
```

Note that by breaking the line (to go below 80 chars) in:

```ruby
  it 'cannot select an item that is not on the menu' do
    expect{ subject.select_item('sausage', 1)}.to raise_error
    'Sorry, that item is not on the menu.'
  end
```

you have actually created two separate lines that are interpreted separately - the last string 'Sorry, that item is not on the menu.' is effectively discarded .  You should prefer some alternatives like:


```
  it 'cannot select an item that is not on the menu' do
    message = 'Sorry, that item is not on the menu.'
    expect{ subject.select_item('sausage', 1) }.to raise_error message
  end
```

or

```
  it 'cannot select an item that is not on the menu' do
    expect do 
      subject.select_item('sausage', 1) 
    end.to raise_error 'Sorry, that item is not on the menu.'
  end
```

and there are other ways, such as using '\' to break the line, or there are other places that are safe to break, or extract a proc if you want to get fancy ...

* The only line you don't have test covered is `Text.new.send_message` ...

* Watch your indenting throughout.  Hiring partners tell us this is a huge turnoff.  For example in takeaway_spec.rb the it blocks are indented 4 spaces instead of 2.  In customer.rb the private block is indented an extra two spaces, and then so are all the private methods

* You could get to 100% test coverage if you were to write another little test like this:

```
  it 'raises an error when right amount is not paid' do
    allow_any_instance_of(Text).to receive_message_chain(:new, :send_message)
    subject.select_item('pizza', 2)
    expect(subject.pay('£20.00')).to eq \
      'Thank you for your order. You will receive text confirmation shortly.'
  end
```

Note that this receive_message_chain approach (which should generally be reserved for third party APIs i.e. legacy code) could also be used to reduce your dependence on multiple doubles in the takeaway_spec.rb

* I see you were thinking about ENV vars - however note that your phone/sid/auth_token are now public on github :-( see https://help.github.com/articles/remove-sensitive-data/ to remove them, and really do use those ENV variables and also the dotenv gem in order to avoid this in future

* In Text, method send_message is not indented properly and has too many lines.  You don't need any of the attr_readers there or indeed any of the instance variables - so this could all be turned into a module perhaps?

So this is the second round of feedback I mentioned before - fix some of these up and we can definitely get you a sticker.  You've done great work, but there are also lots of opportunities to really excel!  Grab with both hands!

:construction:

PR #149, opened 2015-09-13
Not as complete as I'd like this to be be, will keep working on it throughout the week.

adellanno:


tansaku:
Great base @adellanno  - tests all green - that's great, and >95% test coverage which is pretty good.  Some tips to help you improve:

```
$ irb
2.2.3 :001 > menu = Menu.new
 => #<Menu:0x007fc09e057038 @menu={"hamburger"=>2, "pizza"=>5}, @order=[], @order_total=[]> 
2.2.3 :002 > menu.check_total
 => nil 
2.2.3 :003 > menu.check_menu
 => {"hamburger"=>2, "pizza"=>5} 
2.2.3 :004 > menu.select_dishes('hamburger', 'pizza', 7)
Hamburger has been added to your order
Pizza has been added to your order
Thank you, you will receive a confirmation text shortly
 => <Twilio::REST::Message @path=/2010-04-01/Accounts/ACef87d7dfc16f3cebd55fc9ec9969335a/Messages/SM527ae98a2c76494893a5ad198b0dd744> 
2.2.3 :005 > 
```

* Also you seem to have a phone number, and twilio credentials hard coded into the system.  Did you want it publicly shared on Github?  If not consider removing it: https://help.github.com/articles/remove-sensitive-data/

* You should generally prefer to avoid having `puts` in your classes or modules.  If there is output to present on the command line we should prefer to return it from the method and have all output displayed from some ruby code in a base takeaway.rb file or similar, i.e. one that doesn't have classes/modules.  The idea here is 'separation of concerns' - the same reason why we try not to mix different languages (e.g. ruby and html) wherever possible.  We don't want to lock our business logic to a particular output representation, i.e. in this case takeaway and customer are business objects and so they shouldn't do anything that locks them to STDOUT/STDIN i.e. prefer not to use gets and puts inside those classes/modules
 
  * removing these puts would also clear up the extra text you have in your rspec output

* You are 4-space indented - please be 2-space indented https://github.com/bbatsov/ruby-style-guide#spaces-indentation

* Prefer to use the symbol hash format https://github.com/bbatsov/ruby-style-guide#symbols-as-keys

* Single Responsibility Principle - you've got everything is in a single class - you should look to extract some responsibilities into other classes and modules

* Avoid commented out code - xit is better in spec files

* I don't think you've stubbed out the SMS sending portion of the code (as we discussed last week) so each time the tests are running you are getting a text message?

* You might also like to consider these issues I got from running the [reek](https://github.com/troessner/reek) (`gem install reek ; reek -U`) code smell tool on your code:

>
→ reek -U
lib/menu.rb -- 4 warnings:
  [22, 26]:Menu#select_dishes calls "#{i}".capitalize 2 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [19]:Menu#select_dishes has approx 8 statements (TooManyStatements) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Statements.md]
  [20]:Menu#select_dishes has the variable name 'i' (UncommunicativeVariableName) [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Variable-Name.md]
  [29]:Menu#select_dishes is controlled by argument total (ControlParameter) [https://github.com/troessner/reek/blob/master/docs/Control-Parameter.md]

Overall some a good foundation here.  Don't be intimidated by the above list, you've got a great base here, but I want to give you all the help I can to take things to the next level :-) Have a go at fixing up some of the above and we'll see about getting you a sticker :-)

:construction:

PR #150, opened 2015-09-13
takeaway challenge

Proskurina:


tansaku:
Well done Yana - a great submission with tests passing and good coverage.  

Some notes to help you improve:

```sh
$ irb
2.2.3 :001 > customer = Customer.new
 => #<Customer:0x007fd9f29562e8 @order=#<Order:0x007fd9f29562c0 @dishes={}>, @twilio_text=#<TwilioText:0x007fd9f2956270>> 
2.2.3 :002 > customer.add_dish('pizza',1)
 => 1 
2.2.3 :003 > customer.add_dish('pizza',1)
 => 2 
2.2.3 :004 > customer.place_order({'pizza'=>2}, 35)
RuntimeError: Total is not correct
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:32:in `place_order'
	from (irb):4
2.2.3 :005 > customer.place_order({'pizza'=>2}, 25.98)
 => <Twilio::REST::Message @path=/2010-04-01/Accounts/AC5c7f51a9ef5ab4cae242e91d8e98a940/Messages/SM5875ff4d44444040a0c40a48345bb08f> ```

* Note, your phone/sid/auth_token are on github see https://help.github.com/articles/remove-sensitive-data/ to remove them, and prefer ENV variables and possibly also the dotenv gem in order to avoid this in future

* Did you intend to check in your .irb-history?  Please remove otherwise.

* Ideally you'd dry out your customer tests using subject like so:

```ruby
let(:order){ double :order }
subject { Customer.new(order) }
```

which would allow your tests to look like this:

```ruby
  it "can add dish" do
    expect(order).to receive(:add_dish).with('pizza',1)
    subject.add_dish('pizza',1)
  end
```

* I would avoid tests like this:

```
describe Customer do
  it "can see list of dishes" do
    expect(subject.list_of_dishes).to eq(Menu::MENU)
  end
end
```

Because here you are testing the presence of the Menu::MENU collaborator entity within the customer tests.  You have taken a reasonable stab at testing the Menu module separately by creating a dummy class which `extends` Menu.  But note that your Customer class `includes` Menu (2-space indent the include statements BTW).  So your test of Menu is actually a test of something different than the way it is used in Customer.  When testing modules you should prefer shared_examples (see: https://relishapp.com/rspec/rspec-core/docs/example-groups/shared-examples)  This allows you have a separate set of tests for a module like Menu and mix in those tests to tests of classes like Customer that mix in that module.

* Your Order class and associated tests look pretty good - well done.  Although you are including Menu in order when you don't need to and you could simplify your `add_dish` logic by using Hash.new(0)

* In TwilioText, `time` could be a private method, since it is only accessed internally to TwilioText no?

* Also note that TwilioText has no instance state, other than the @client variable that you create (I think unecessarily) so it could be a module ...?

* TwilioText#send_text is a little long - I don't know that you need the temp vars account_sid and auth_token, and the body string creation could be pulled out into a separate method to reduce the responsibilities of send_text

* You've got a few empty lines hanging around (particularly in menu.rb) - try not to have more than one empty line between any line of code

* I see you've followed one of the approaches I described to stubbing Twilio - can you take it further so that the stubbing checks the correct data is being submitted?

* I'd prefer the newer hash syntax throughout if I were you { pizza: 12.99 }

* Customer has a nice use of dependency injection - well done :-)

Overall a great foundation here.  Don't be intimidated by the above list, this is really a good submission, but I want to give you all the help I can to take things to the next level :-) Have a go at fixing up some of the above and we'll see about getting you a sticker :-)

:construction:

Proskurina:
WOW!!! Thank you for such a detailed feedback!!!
My brain hurts, but I tried to amend everything according to your suggestions. 
I refactored TwilioText and made it a module, but now I am not sure I test it correctly (tried to use VCR)..

tansaku:
Wonderful that you're following up on this Yana.  I pulled out your latest changes and went to ran the tests and got this:

```
$ rspec
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.
[Coveralls] Outside the CI environment, not sending data.
/Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/twilio_text.rb:3:in `require_relative': cannot load such file -- /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/.env.rb (LoadError)
```

If I comment out that line in `twilio_text.rb`:

# require_relative '../.env.rb'

I get this failure

```
  1) Customer receives a message after the order is placed
     Failure/Error: response = subject.send_text
     ArgumentError:
       Account SID and auth token are required
```

I'd recommend the [dotenv](https://github.com/bkeepers/dotenv) gem for managing your env vars rather than .env.rb and really this failure above should be flagged to the user in advance in the README and you should provide a default .env file that looks like this:

```
export TWILIO_ACCOUNT_SID="insert_your_twilio_sid_here"
export TWILIO_AUTH_TOKEN="insert_your_twilio_auth_here"
export TWILIO_PHONE="insert_your_twilio_phone_here"
export TWILIO_DESTINATION_PHONE="insert_your_twilio_destination_phone_here"
```

I see that you removed (commented out) your twilio text spec.  If you are removing something you should remove it, not comment it out.  Note that you could still be testing the Twilio module with shared examples: https://www.relishapp.com/rspec/rspec-core/v/3-3/docs/example-groups/shared-examples, as you are doing with menu_spec (again that has commented code that should be deleted)

Well done for trying to use VCR.  I see don't have a cassette folder checked in, which I think it should be, but it looks like VCR is not configured correctly. I had to do quite a fair amount work to get everything configured everything for you correctly and got vcr working.  What you needed was some additional config like so:

```ruby
TWILIO_VARS = []
TWILIO_VARS << 'TWILIO_ACCOUNT_SID'
TWILIO_VARS << 'TWILIO_AUTH_TOKEN'
TWILIO_VARS << 'TWILIO_PHONE'
TWILIO_VARS << 'TWILIO_DESTINATION_PHONE'

VCR.configure do |config|
  config.cassette_library_dir = "fixtures/vcr_cassettes"
  config.hook_into :webmock # or :fakeweb
  TWILIO_VARS.each do |var|
    config.filter_sensitive_data("<#{var}>") { ENV[var] }
  end
end
```

Note that the main part of the above is described here: https://github.com/vcr/vcr#usage.  You added the gem and the require and updated your spec, but did not include the configuration.  With the above configuration VCR will generate a fixture that consists of the network interaction with Twilio and prevent the text message being sent 

There is the danger of sensitive data being stored in the fixture itself, but that can be taken care off with the filter_sensitive_data setting you also see above.  Read more about that here:

http://stackoverflow.com/questions/9816152/filtering-sensitive-data-with-vcr/9831148#9831148

Note that with all this set up we now expose the following dependency in your test:

```
  1) Customer receives a message after the order is placed
     Failure/Error: expect(response.body).to eq "Thank you! Your order was placed and will be delivered by #{(Time.now + 60*60).strftime('%H:%M')}"
       
       expected: "Thank you! Your order was placed and will be delivered by 08:22"
            got: "Thank you! Your order was placed and will be delivered by 08:16"
       
```

Your test was only currently passing because the test happened to run within the same minute of the application, which is something we shouldn't be relying on.  Now that we have a fixture in place the dynamic time setting of the delivery time is exposed.  With the fixture we can now fix this like so in customer_spec.rb:

```ruby
  it "receives a message after the order is placed" do
    VCR.use_cassette('twilio') do
      response = subject.send_text
      expect(response.body)
        .to eq "Thank you! Your order was placed and will be delivered by 08:25"
    end
  end
```

note that I've also fixed the Rubocop line length issue in this method - you have a number of other rubocop issues you should also fix:

```

lib/customer.rb:25:19: C: Style/OptionalArguments: Optional arguments should appear at the end of the argument list. (https://github.com/bbatsov/ruby-style-guide#optional-arguments)
  def place_order(list=my_list, total)
                  ^^^^^^^^^^^^
lib/twilio_text.rb:7:3: C: Metrics/MethodLength: Method has too many lines. [6/5] (https://github.com/bbatsov/ruby-style-guide#short-methods)
  def send_text
  ^^^
spec/customer_spec.rb:27:81: C: Metrics/LineLength: Line is too long. [81/80] (https://github.com/bbatsov/ruby-style-guide#80-character-limits)
    expect{subject.place_order({dip: 2},3)}.to raise_error "Total is not correct"
                                                                                ^
```

Some other "minor" issues - although these are the kinds of things that will prevent you getting from tech test to interview:

* It looks to me that Customer includes 'menu' unnecessarily, and thus customer_spec includes the menu shared_examples unnecessarily too.

* indenting incorrect on line 32 of customer.rb and unnecessary blank line at start, similarly for menu.rb

Finally you have some conflicts with the upstream that's preventing the full CI from running.  Please also sync your code to the upstream:

https://help.github.com/articles/syncing-a-fork/


So you're making great strides here, but there's still a way to go.  Want to do another round of changes and get some more feedback from me? :-)

PR #151, opened 2015-09-13
Functional TakeAway service - requires testing refactoring, doubling and implementation of time dynamics

MoeSadoon:


README.md
```diff
@@ -1,55 +1 @@
...
-
-* A free account on Twilio will only allow you to send texts to "verified" numbers. Use your mobile phone number, don't worry about the customer's mobile phone.
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
-
-
-**Note: We are looking for good OO design and programming! Remember the Single Responsibility and Dependency Injection/Inversion principles!**
-
-[![Build Status](https://travis-ci.org/makersacademy/takeaway-challenge.svg?branch=master)](https://travis-ci.org/makersacademy/takeaway-challenge)
-[![Coverage Status](https://coveralls.io/repos/makersacademy/takeaway-challenge/badge.png)](https://coveralls.io/r/makersacademy/takeaway-challenge)
+Moe's takeaway challenge
```
giusepped: Remember to update your README.md with some overview of the project, installation instructions and maybe the lessons you have learnt while carrying it out.

lib/take_away.rb
```diff
@@ -0,0 +1,56 @@
+class TakeAway
+ 
+
+  attr_writer :menu, :meal_order
```
giusepped: Always be careful to use `attr_writer` and do not use it if you do not need it (you do not need it in your class because you have defined methods to access the properties of your class). You always want to restrict the access that the outside world has to the property using your own methods. As I have shown you I could do something like 
```
restaurant = TakeAway.new
restaurant.menu = nil
``` 
And this will change your menu to nil. `attr_reader` is less powerful BUT I could still use any method on your properties and do something like `restaurant.meal_order.replace([1, 2, 3])` and this would work

lib/take_away.rb
```diff
@@ -0,0 +1,56 @@
...
+               "cheeseburger" => 4.99,
+               "steak" => 14.99,
+               "salad" => 6.99
+             }
+
+     @meal_order = []
+   end
+
+   def show_menu
+   	@menu.sort{|a,b| a[1]<=>b[1]}.each { |elem| puts " #{elem[0].capitalize}: #{elem[1]}"}
```
giusepped: Just remember you are returning `nil` right now as we discussed.

lib/take_away.rb
```diff
@@ -0,0 +1,56 @@
...
+   end
+ 
+ 
+  def select_dish(dish)
+    fail "I'm sorry, we don't have that dish" unless @menu.include?(dish)
+    @meal_order << dish
+  end
+
+  def order_summary
+    puts "you have ordered #{@meal_order.count} item, consisting of #{@meal_order}"
```
giusepped: What happens if I haven't ordered anything yet?

lib/take_away.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def place_order
+    fail "Please select at least one item" unless !@meal_order.empty?
+    send_message
+      end
+ 
+
+	def send_message
+
+		require 'rubygems' # not necessary with ruby 1.9 but included for completeness
```
giusepped: Remember SRP, the twilio message method should be in a different class!

lib/take_away.rb
```diff
@@ -0,0 +1,56 @@
...
+
+	def send_message
+
+		require 'rubygems' # not necessary with ruby 1.9 but included for completeness 
+		require 'twilio-ruby' 
+		 
+		# put your own credentials here 
+		account_sid = 'ACc883607ece96532c6655b59873591b40' 
+		auth_token = 'd9ed0c3613d2eb80f0c50677c20b2ce2' 
+
```
giusepped: Do not put your credentials inside your class because the world can see them and use them. You have to put them in environment variable and access them with something like `ENV[auth_token]`

spec/take_away_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'take_away'
+ 
+  describe TakeAway do
+
+  	
+
+  	it "customer can enter restaurant" do
+  		expect(TakeAway).to respond_to(:new)
+  	end
+
```
giusepped: This is a bit redundant because the .new method will pretty much always work.

spec/take_away_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  	it "can view menu prices" do
+  		expect(subject).to respond_to(:show_menu)		
+  	end
+
```
giusepped: This is okay but you could also test what the response of the method should be. So your method should actually return a string (again, you probably do not want to return the actual menu object) and then your test should check that the return value is equal to that string.

spec/take_away_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+  		expect(subject).to respond_to(:select_dish).with(1).argument
+  	end
+
+  	it "raises error when dish not available" do
+  		 expect{ subject.select_dish("wrong_dish") }.to raise_error "I'm sorry, we don't have that dish"
+ 		end
+   
+   	it "summarises order" do 
+   		expect(subject.order_summary).to be nil
+   	end
```
giusepped: again, your method should not return `nil`. Make your method return something here (e.g. the dish that has been ordered). So then you can order something in your test and check that the return value is equal to that (for example).

spec/take_away_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+   	it "places order and sends customer a text message" do 
+   		expect(subject).to respond_to(:place_order)
+   	end
+
```
giusepped: Again, maybe be a bit more specific here and put your `send_message` in a separate class and have a double and check that the method has been called on the double when you place the order.

tansaku:
Good base here Moe - nice and green :-)

Some comments:

* you need to specify twilio-ruby in your gem file
* You must fix indenting issues like this:

```
  def place_order
    fail "Please select at least one item" unless !@meal_order.empty?
    send_message
      end
```

it might seem trivial, but its really really important 

* go for single line spacing in places like this:

```
class TakeAway
 

  attr_writer :menu, :meal_order
```

* generally your methods are nice and short, except send_message which could be shortened and as guiseppe mentioned could be factored out - I might put it in a module.  Basically you have one big god class (TakeAway) that is taking all responsibilities.  That's not a terrible place to start for a small project like this, but you really need to go through the process of factoring out the different components into separate classes and modules.

* your test coverage is quite low - do you understand why? if not let's look at this together

https://coveralls.io/builds/3559188/source?filename=lib%2Ftake_away.rb

Try and fix some of the above up (and hound and guiseppe's comments) and we'll see if we can't sort you out a sticker :-)

:construction:

PR #154, opened 2015-09-13
Takeaway challenge

matt-paul:


katylouise:
We discussed:
* It was good that Matt had already decided that it would be advantageous to refactor into two classes.  We discussed how to make this work using dependency injection, and doubles to test it.  
* It would be better to use integers for the money (or a currency gem?).  This avoids the trouble Ruby has with float arithmetic.
* Maybe make the basket a hash to keep track of the quantity and the item ordered.  Then you wouldn't need a separate quantity variable.
* You don't need to make all variables instance vars.  If they are just used within a method, they can be regular local vars.  Or try to avoid creating too many vars within a method.
* Suggested refactoring the order method to separate responsibilities out.
* Good use of environment variables in the Twilio API.
* Good use of inject to sum the prices.
* Good to make not_on_menu? a private method.

tansaku:
Good base here Matt

Some general comments and thoughts:

* in menu_rspec `require_relative '../lib/menu.rb'` could be simplified to `require 'menu'`

* ensure that test descriptions what tests do e.g. 

```
  it 'can show a list of items' do
    expect(subject).to respond_to(:list_of_items)
  end
```

this is a test for responding to the method :list_of_items - not that any items are actually shown - should replace with oneliner and then text is autogenerated

```
  it { is_expected.to respond_to :list_of_items }
```

ultimately these tests are redundant (and can be deleted) once you have testing that actively check the results of calling methods e.g. 

```
  it 'can show a list of items' do
    expect(subject.list_of_items).to eq { :naan => 1.00, :daal => 3.00, :rice => 2.00 }
  end
```

* note that inside `describe '#order'` you must indent two spaces

* your test coverage of this menu class is pretty low (< 64%) - you are actually only testing the presence of a couple of methods, and that one of them raises an error in a guard clause, which means the body of the order method is not tested - you really should test `basket_price` in menu_spec and also test the non-error conditions.

* also note that any ruby method ending in a question mark e.g. `not_on_menu?` is, by convention, one that will return a boolean - you have yours raising an error.  The fix is:

```
  def not_on_menu?(item)
    @list_of_items.has_key?(item)
  end

  def order(item, quantity)
    raise "Item not on menu" unless not_on_menu?(item)
    ...
```

* The code smell analyzer Reek says the following about your code:

```
  [1]:Menu has at least 6 instance variables (TooManyInstanceVariables) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Instance-Variables.md]
  [27]:Menu#not_on_menu? has unused parameter 'quantity' (UnusedParameters) [https://github.com/troessner/reek/blob/master/docs/Unused-Parameters.md]
  [14, 16, 17, 18]:Menu#order calls list_of_items.fetch(item) 4 times (DuplicateMethodCall) [https://github.com/troessner/reek/blob/master/docs/Duplicate-Method-Call.md]
  [11]:Menu#order has approx 7 statements (TooManyStatements) [https://github.com/troessner/reek/blob/master/docs/Too-Many-Statements.md]
```
So although you are getting the job done here, there is a lot to fix up in order to make this all clean code.

* Your basket class looks like something you were thinking you might use and then didn't. That's fine - but you need to remove anything like this from your final submission (it can hang around in git history no problem) - although menu has a lot of responsibilities so many we do want to move them over to basket

* ideally you'd have your twilio code in a module and also have some tests for it (as we discussed as a group on Monday)

* you've got some great explanation of how your code should be used in comments in your code - that should all be in your README explaining to people how to use your system:

```
#
# menu = Menu.new       # => #<Menu:0x007fec0a90f530 @list_of_items={:naan=>1.0, :daal=>3.0, :rice=>2.0}, @basket=[], @subtotal=[]>
# menu.list_of_items    # => {:naan=>1.0, :daal=>3.0, :rice=>2.0}
# menu.order(:naan, 3)  # => [3.0]
# menu.order(:rice, 2)  # => [3.0, 4.0]
# menu.basket           # => [[:naan, 1.0, 3], [:rice, 2.0, 2]]
# menu.basket_price     # => 7.0

# >> Thank you for ordering 3 x naan, costing £1.0 each
# >> Thank you for ordering 2 x rice, costing £2.0 each
```


Want to take a stab at fixing some of the above (and the hound comments) and re-pushing the code so you can get a sticker?

tansaku:
:construction:

PR #156, opened 2015-09-14
update all

TJQKAs:


tansaku:
A good base here Evgenii.  Your tests are all green and ~80% test coverage - a good start. Also very pleased to see you making use of `Hash.new(0)`. Some tips to help you improve:

* Review the [output from coveralls](https://coveralls.io/builds/3559443/source?filename=lib%2Ftakeaway.rb) to see where you are missing tests exercising your application code

* When locking the Gemfile to a particular version of Ruby, you should prefer the latest, currently 2.2.3

* You have a series of empty files in your root called '10,', '15,' etc.  I think you created these accidentally - you need to delete them and ensure they are not checked in

* I see you have comments with `manual feature tests` in some of your files, e.g. in order.rb. I think it's a great idea to show the full story of how your app is used (from a user's perspective) in the README:

```sh
→ irb
2.2.2 :001 > takeaway = Takeaway.new
 => #<Takeaway:0x007ff21219e518> 
2.2.2 :002 > takeaway.order(1)
 => 1 
2.2.2 :003 > takeaway.list_ord
 => {:Pizza=>1} 
2.2.2 :004 > takeaway.order(7)
 => 1 
2.2.2 :005 > takeaway.list_ord
 => {:Pizza=>1, :Cofe=>1} 
2.2.2 :006 > takeaway.t_price
 => 19 
```

* In general please don't shorten method names to things like `t_price`.  I think you mean `total_price` here, in which case go with `total_price`.  You also have the method t_price defined twice - once via the attr_reader and then again in the code - making the attr_reader redundant

* Note that you should generally prefer to avoid having `puts` in your classes or modules.  If there is output to present on the command line we should prefer to return it from the method and have all output displayed from some ruby code in a base takeaway.rb file or similar, i.e. one that doesn't have classes/modules.  The idea here is 'separation of concerns' - the same reason why we try not to mix ruby and html wherever possible.  We don't want to lock our business logic to a particular output representation, e.g. in this case takeaway and customer are business objects and so they shouldn't do anything that locks them to STDOUT/STDIN i.e. prefer not to use gets and puts inside those classes/modules

* You have all your functionality in one class. Please have a go at extracting the menu related functionality into another class to that you have at least two classes interacting here.

* There are a few hound issues to fix up - here's some more detail on them, i.e. links to the Ruby style guide

> lib/takeaway.rb:13 I: Style/HashSyntax: Use the new Ruby 1.9 hash syntax. (https://github.com/bbatsov/ruby-style-guide#hash-literals)
lib/takeaway.rb:29 W: Lint/UnusedBlockArgument: Unused block argument - `key`. If it's necessary, use `_` or `_key` as an argument name to indicate that it won't be used. (https://github.com/bbatsov/ruby-style-guide#underscore-unused-vars)
spec/takeaway_spec.rb:13 I: Style/HashSyntax: Use the new Ruby 1.9 hash syntax. (https://github.com/bbatsov/ruby-style-guide#hash-literals)

Overall some really good work here.  Don't be intimidated by the above list, you've got a good start here, but I want to give you all the help I can to take things to the next level :-)  There are a few other things as well, but have a go at fixing up some of the above, we'll do another round of feedback and then and we'll see about getting you a sticker :-)

:construction:

PR #157, opened 2015-09-14
Chris Weeks takeaway-challenge

chweeks:


lib/order.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'menu'
+
+class Order
+
+  attr_reader :order, :amount
+
+  def initialize
+    @order, @amount = [], []
```
sjmog: Nice, but let's do these separately for now to avoid confusion

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def add_to_order menu, dish, quantity
+      fail 'Item not on menu' if menu.dishes.include?(dish) == false
```
sjmog: The conditional should be a private method here

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+      amount << quantity
+      order << dish
+  end
+
+  def total_price
+    total, count = 0, 0
+    order.each do |item|
```
sjmog: You could use `each_with_index` wherever you find yourself manually tracking a count

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def show_order
+    count = 0
+    output = ''
+    order.each do |dish|
+      quantity = order.count(dish)
+      output += "\n#{dish.show_details} X#{amount[count]}"
+      count += 1
+    end
```
sjmog: You can use `inject` here to DRY this up some.

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+    output
+  end
+
+  def place_order payment
+    fail "Incorrect Payment" if payment != total_price
+    show_order
+    total_price
+    Text.new.send_text
```
sjmog: Very nice. For even better reduction of dependency, you could pass `Text` as a parameter to the `place_order` method and then call `text_sender.call`.

lib/text.rb
```diff
@@ -0,0 +1,25 @@
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load
```
sjmog: Love the use of `Dotenv` here.

lib/text.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def send_text
+    @client = Twilio::REST::Client.new(ENV["ACCOUNT_SID"], ENV["AUTH_TOKEN"])
+    @client.messages.create(
+    body: "Thank you for your order. It will be delivered before #{time}",
+    to: ENV["PHONE_NUMBER"],
+    from: sender)
+    'text sent'
+  end
```
sjmog: This is a really fantastic class.

spec/dish_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'dish'
+
+chicken_chow_mein = Dish.new('Chicken Chow Mein', 5.95)
```
sjmog: Put this inside the class and use `let` to define it.

sjmog: Even better, use the [explicit `subject`](https://www.relishapp.com/rspec/rspec-core/docs/subject/explicit-subject) RSpec pattern.

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'menu'
+
+describe Menu do
+  let(:dish) {double :dish, show_details: 'Coke: £1'}
```
sjmog: Nice mocking.

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  it { is_expected.to respond_to :dishes }
+
+  describe '#add_dish' do
```
sjmog: Lovely use of hash syntax here

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it { is_expected.to respond_to :add_dish }
+
+    it 'should add dish to menu' do
+      subject.add_dish(dish)
+      expect(subject.dishes).to eql [dish]
```
sjmog: better `subject.dishes.first`. Avoid complex types as expectations (make em simple)

spec/order_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    it 'should show order' do
+      subject.add_to_order(menu, dish, 3)
+      expect(subject.show_order).to eql "\nPizza: £2.00 X3"
+    end
+  end
+
+  describe '#total_price' do
+
+    it { is_expected.to respond_to :total_price }
```
sjmog: These individual expectations are not necessary: you test the method each time later on (here, you test this on line 42). You should avoid duplicating test functionality as this will slow tests down without increasing coverage.

spec/text_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+describe Text do
+  it { is_expected.to respond_to :time }
+
+  it { is_expected.to respond_to :send_text}
+
+  describe '#send_text' do
+    it 'should send text when order is placed' do
+     client = double :client
+     allow(client).to receive_message_chain(:messages, :create)
+     expect(Twilio::REST::Client).to receive(:new).and_return(client)
```
sjmog: Could you be even better at stubbing here? You've made a start but this doesn't totally isolate `Text` from `Twilio`.

What about:

1. Passing `Twilio::REST::Client` as a parameter in the initializer of `Text`
2. Use `double_class("Twilio::REST::Client").as_stubbed_const

sjmog:
Great stuff here. Let's shake up those iterators and ensure we have proper isolation for the `Twilio` section. Also, we can remove several tests to help our suite run faster.

Otherwise, amazingly tidy code and a clear understanding of all concepts so far :).

:checkered_flag:

PR #158, opened 2015-09-14
Required features built, pending test on message sending

gavinwcheng:


tansaku:
Well done Gavin - a great submission with tests passing and good coverage.  

Some notes to help you improve:

```sh
irb
2.2.3 :001 > restaurant = Restaurant.new
 => #<Restaurant:0x007fc173284bf0 @menu={"starter"=>3, "main"=>6, "drinks"=>2}, @orders={}> 
2.2.3 :002 > customer = Customer.new
 => #<Customer:0x007fc17327cc48> 
2.2.3 :003 > customer.order restaurant, 'starter', 4
 => 4 
2.2.3 :004 > customer.order restaurant, 'main', 4
 => 4 
2.2.3 :005 > customer.confirm_total restaurant, 25
RuntimeError: Total does not match sum of dish prices
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/restaurant.rb:17:in `confirm_order'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:12:in `confirm_total'
	from (irb):5
2.2.3 :006 > customer.confirm_total restaurant, 36
ArgumentError: wrong number of arguments (0 for 1)
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/restaurant.rb:29:in `send_order_confirmation'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/restaurant.rb:18:in `confirm_order'
	from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/lib/customer.rb:12:in `confirm_total'
	from (irb):6
```




* Prefer to use the symbol hash format https://github.com/bbatsov/ruby-style-guide#symbols-as-keys

* No need for ternary here:

```ruby
  def total_correct? total
    total == self.total_calc ? true : false
  end
```

can just be 

```ruby
  def total_correct? total
    total == self.total_calc
  end
```

* nice guard clauses!  `fail "Total does not match sum of dish prices" unless total_correct? total`


* I see you've followed one of the approaches I described to stubbing Twilio - can you take it further so that the stubbing checks the correct data is being submitted?

* Your TwilioRuby class doesn't really have any state - it could be a module and then it could be tested with shared examples: https://www.relishapp.com/rspec/rspec-core/v/3-3/docs/example-groups/shared-examples

* note that tests like this:

```ruby
    it "responds to order with 3 arguments" do
      expect(subject).to respond_to(:order).with(3).arguments
    end
```

```ruby
    it { is_expected.to respond_to(:order).with(3).arguments }
```

```ruby
    it "calls take_order on restaurant when called" do
      expect(restaurant).to receive(:take_order).with(dish, amount)
      subject.order restaurant, dish, amount
    end
```


* Avoid leaving commented code hanging around - delete it - it will be stored in your previous git commits if you need it again.  Your final submissions should not include any commented code:

```

  # def initialize(restaurantClass = Restaurant.new)
  #   @restaurant = restaurantClass
  # end

```

* Your Customer class has no state - is it actually necessary?

* Your dishes and amounts are strings and numbers, so I don't think they need to be doubles.  You need doubles for objects that you have created, or objects coming from other libraries, not generally for the base ruby classes, particularly the simple ones

* Do fix up your outstanding Rubocop issues (see embedded comments in the pull request) and also you might also like to consider these issues I got from running the [reek](https://github.com/troessner/reek) (`gem install reek ; reek -U`) code smell tool on your code:

> $ reek -U
lib/customer.rb -- 2 warnings:
  [11]:Customer#confirm_total doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]
  [7]:Customer#order doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]
lib/restaurant.rb -- 2 warnings:
  [29]:Restaurant#send_order_confirmation doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]
  [22]:Restaurant#total_correct? is controlled by argument total (ControlParameter) [https://github.com/troessner/reek/blob/master/docs/Control-Parameter.md]
lib/twilio_ruby.rb -- 1 warning:
  [14]:TwilioRuby#sms_content doesn't depend on instance state (UtilityFunction) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]


Overall a great foundation here.  Don't be intimidated by the above list, this is really a good submission, but I want to give you all the help I can to take things to the next level :-) Have a go at fixing up some of the above and we'll see about getting you a sticker :-)

:construction:

PR #160, opened 2015-09-14
Kyle's inject

KyleMacPherson:


tansaku:
Good start @KyleMacPherson   

Some tips to help you improve:

* When I first ran bundle on your code I got this:

```

[!] There was an error parsing `Gemfile`: Undefined local variable or method `‘twilio' for Gemfile. Bundler cannot continue.

 #  from /Users/tansaku/Documents/Github/MakersAcademy/takeaway-challenge/Gemfile:9
 #  -------------------------------------------
 #  gem 'capybara'
 >  gem ‘twilio-ruby’
 #  -------------------------------------------
```

caused by using non-ascii characters in the Gemfile.  If I might be so bold, did you perhaps copy and paste these from another document?  It's really important to type these things out by hand ...

* The above issue prevented your 'build' from passing, and so our continuous integration could not check your tests were passing, or your code coverage.  Running SimpleCov locally I could see you have 80% test coverage, not bad, but could be higher.  That 20% missing is all from the SMS component which we talked about how to test last week.

* As you know the Gemfile can lock the project to a particular version of Ruby.  In general you should prefer the latest, currently 2.2.3

```sh
 irb
2.2.2 :001 > t = Takeaway.new
 => #<Takeaway:0x007f9e046b9c80 @menu={"half_chicken"=>10, "butterfly_chicken"=>10, "chips"=>3, "rice"=>3}> 
2.2.2 :002 > t.select('chips', 'rice')
 => "You ordered: chips, rice. The total is 6." 
2.2.2 :003 > 
```

* As you probably know RSpec provides a `subject` method that creates a new instance of the class under test (the one following `describe`), so this:

```
describe Takeaway do
  it 'shows menu when requested' do
    t = Takeaway.new
    expect(t.view_menu).to eq(t.menu)
  end
end
```

can become

```
describe Takeaway do
  it 'shows menu when requested' do
    expect(subject.view_menu).to eq(subject.menu)
  end
end
```

* You only have two classes and they don't interact, so you don't find yourself in a situation needing doubles to stub out interaction between the two - you should probably pull out a menu class and then test TakeAway and Menu independently, so their responsibilities can be separated according to the Single Responsibility Principle (SRP)

* The Takeaway#verify method is not being accessed by the test or any collaborator object so it could be private

* You've made interesting use of the splat argument in TakeAway#select, but note that you've declared instance variables that you don't need, that can be removed to reduce the overall size of the method (which could also use inject rather than the each loop in order to be neatened up)

* Your phone/sid/auth_token are on github see https://help.github.com/articles/remove-sensitive-data/ to remove them, and prefer ENV variables and possibly also the dotenv gem in order to avoid this in future

* you also have some hound issues you should look into fixing up

Overall some a good foundation here.  Don't be intimidated by the above list, you've got a great base here, but I want to give you all the help I can to take things to the next level :-)  There are a few other things as well, but have a go at fixing up some of the above, we'll do another round of feedback and then and we'll see about getting you a sticker :-)

Please don't hesitate to ask if anything is unclear

:construction:

PR #162, opened 2015-09-14
Monday morning, from number is invalid

AlexPerson:


lib/takeaway_class.rb
```diff
@@ -0,0 +1,38 @@
...
+
+	attr_reader :menu, :order, :actualcost
+
+	def initialize
+		@menu = {"Cheeseburger" => 1,
+			"Pizza" => 2,
+			"Fries" => 3
+		}
+		@actualcost = 0
+	end
```
ojlamb: You might want to separate responsibilities of your menu and order into different classes.

ojlamb: Ruby convention is 2 space indentation. Clean and readable code is really important moving forward. Try adjusting the settings in your text editor .

lib/test.rb
```diff
@@ -0,0 +1,20 @@
+require 'twilio-ruby'
+
+account_sid = 'ACf8addc5058d1be8f246fd75fa4ce9013'
+auth_token = 'bb5aa7da4e982b37aac058a31d962549'
```
ojlamb: use .gitignore or an environment variable to hide tokens before pushing to git.

tansaku:
Hey @AlexPerson  some good work here - some tips to help you improve

* The Gemfile can lock the project to a particular version of Ruby.  In general you should prefer the latest, currently 2.2.3

* Please do include all the gems you use in your Gemfile. This is an important courtesy to other developers and yourself in the future, so that all the project dependencies can be pulled in whenever the project is checked out on a new machine, e.g. `twilio-ruby` gem

```sh
$ irb
Error : 'load /Users/tansaku/.irbrc' : The From phone number +441908410626 is not a valid, SMS-capable inbound phone number or short code for your account.
2.2.3 :001 > takeaway = Takeaway.new
 => #<Takeaway:0x007fac93bfadc8 @menu={"Cheeseburger"=>1, "Pizza"=>2, "Fries"=>3}, @actualcost=0> 
2.2.3 :002 > takeaway.takeorder({"Cheeseburger" => 1}, 1)
 => {"Cheeseburger"=>1} 
2.2.3 :003 > takeaway.actualcost
 => 1 
2.2.3 :004 > 
```

* You should be able to use just `require` in your specs - no need for `require_relative`

* note that tests like this:

```ruby
	it 'can place orders when passed a dish, quantity of meals, and total cost' do
		expect(subject).to respond_to(:takeorder).with(2).arguments
	end
```

```ruby
    it { is_expected.to respond_to(:takeorder).with(2).arguments }
```

```ruby
	it 'remembers orders' do
		subject.takeorder({"Cheeseburger" => 1}, 1)
		expect(subject.order).to eq ({"Cheeseburger" => 1})
	end
```

* Otherwise for the time beingI'd just backup the points made by @ojlamb and hound/rubocop

Overall this is a good base.  Don't be intimidated by the above list, you've done really well, but I want to give you all the help I can to take things to the next level :-)  There are a few other things as well, but have a go at fixing up some of the above, we'll do another round of feedback and then and we'll see about getting you a sticker :-)

:construction:

PR #165, opened 2015-09-14
Dom Batten Takeaway Challenge (Reuploaded)

dbatten4:


spec/order_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    it 'should be able to remove an item from the basket' do
+      subject.add_item("Duck Pancakes")
+      subject.remove_item("Duck Pancakes")
+      expect(subject.basket.empty?).to eq(true)
+    end
+
+    it 'should not be able to remove an item not in the basket' do
+      subject.add_item("Duck Pancakes")
+      expect{ subject.remove_item("Egg Fried Rice") }.to raise_error 
+        'Item was not in the basket'
```
roidriscoll: When you add a line break to put the error message on the next line, you're dissociating the error message you are wanting to compare against. Your test still passes, but only because ANY kind of error is raised. Did you notice a kind of 'false positive' warning when you ran the tests? I suspect this was an attempt to stop hound barking about long line lengths but in this case the workaround breaks the test slightly.

lib/order.rb
```diff
@@ -0,0 +1,84 @@
+require 'twilio-ruby'
+require './.env.rb'
+
+class Order 
+
+  attr_reader :total, :menu, :basket
```
roidriscoll: `total` should be an `attr_accessor` since you are writing to the instance variable.

blogpost.txt
```diff
@@ -0,0 +1,40 @@
...
+
+gem "maruku"
+gem "aws-s3"
+
+Then running the bundle command installs all the gems in the Gemfile.
+
+
+What are some dangers to using a gem?
+
+When you use a gem, you are relying on someone else's code rather than your own. This can cause compatibility issues between your own code and that in the gem.
```
roidriscoll: Also if development on that gem ceases, you run the risk of your project being dependent on outdated and possibly broken code.

blogpost.txt
```diff
@@ -0,0 +1,40 @@
+What is a ruby gem?
```
roidriscoll: As a suggestion, it would be great practice to try using Markdown for articles like this. You will use it again and again when creating README.md files and it's fairly essential for a dev to know, so may as well get as much practice in as possible :grin:

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+class Menu
+
+  attr_reader :dishes_available
+
+  def initialize 
+    @dishes_available = { "Duck Pancakes" => 4.99,
+      "Chicken Chow Mein" => 3.85,
```
roidriscoll: So I might be tempted to set this hash to a constant (such as `CURRENT_MENU`), then initialize `@dishes_available` to that constant. Why? Because menu is data, and we should always be trying to separate data from behaviour. 

A `CURRENT_MENU` constant could be linked to an external file for example, that a non-technical user updates when prices change etc. This facility would allow editing of the data without interfering with the logic in our program. In this case it's such a simple class that the benefits of this aren't immediately obvious, but as our program grows in complexity so does our need for separation of concerns.

spec/menu_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'menu'
+
+describe Menu do 
+
+  it 'should have a list of available dishes' do
+   expect(subject).to respond_to(:dishes_available)
+  end
+
+  it 'should have available dishes' do
+    expect(subject.dishes_available).to be_truthy
```
roidriscoll: I'd argue that all this is testing, is that `dishes_available` exists, not that there's anything useful in it...

lib/order.rb
```diff
@@ -0,0 +1,84 @@
...
+  def place_order
+    fail 'Basket is empty' if basket.empty?
+    fail 'The total price has been miscalculated' unless total_price_verified?
+    send_message
+    "Thank you for your order. You will receive text confirmation shortly"
+  end
+
+  private 
+
+  def dish_check_and_add(dish, quantity)
```
roidriscoll: This might seem nitpicky, but readable code is what we strive for as Rubyists. With that in mind, let's look at this method name: `dish_check_and_add(dish, quantity)`. If that were a regular sentence in English, the fact that `dish` is mentioned twice would be a bit irksome, no? So as with English, we apply the same rules to Ruby. How about: `check_and_add(dish, quantity)`? Or even better: `check_basket_and_add(dish, quantity)`

lib/order.rb
```diff
@@ -0,0 +1,84 @@
+require 'twilio-ruby'
+require './.env.rb'
+
+class Order
```
roidriscoll: This class has too many responsibilities.

lib/order.rb
```diff
@@ -0,0 +1,84 @@
+require 'twilio-ruby'
+require './.env.rb'
+
+class Order 
+
+  attr_reader :total, :menu, :basket
+
+  def initialize(menu = Menu.new)
```
roidriscoll: Dependency Injection :+1:

lib/order.rb
```diff
@@ -0,0 +1,84 @@
...
+    basket_total = 0.0
+    basket.each do |dish, quantity|
+      basket_total += quantity * menu.dishes_available[dish].round(2)
+    end
+    basket_total.round(2) == total ? true : false
+  end
+
+  def place_order
+    fail 'Basket is empty' if basket.empty?
+    fail 'The total price has been miscalculated' unless total_price_verified?
```
roidriscoll: Nice use of guard conditions!

lib/order.rb
```diff
@@ -0,0 +1,84 @@
...
+    send_message
+    "Thank you for your order. You will receive text confirmation shortly"
+  end
+
+  private
```
roidriscoll: It's great that you are conscious of separating some of your interface into private methods!

lib/order.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def quantity_check_and_remove(dish, quantity)
+    basket[dish] == quantity ? basket.delete(dish) : basket[dish] -= quantity
+  end
+
+  def send_message
+    @account_sid = ENV[:account_sid]
+    @auth_token = ENV[:auth_token]
+    @client = Twilio::REST::Client.new @account_sid, @auth_token
+    @client.messages.create(
```
roidriscoll: It seems unnecessary to use instance variables here - they are not used outside the method so why give them more scope than needed?

lib/order.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def dish_check_and_add(dish, quantity)
+    basket.key?(dish) ? basket[dish] += quantity : basket[dish] = quantity
+  end
+
+  def quantity_check_and_remove(dish, quantity)
+    basket[dish] == quantity ? basket.delete(dish) : basket[dish] -= quantity
+  end
+
+  def send_message
```
roidriscoll: This could benefit from being its own class, then `send_message` could be a public method, which uses a bunch of private methods with which to construct itself- then everything would be lovely!

roidriscoll:
Hi Dom, apologies for the late review - it has been a hectic week. Please heed the pointers I've left for you, and have a crack at stubbing out that API! :)

:construction:

dbatten4:
Hi Roi

You’ve actually already given me a review a couple of days ago, good to have two though!

Dom

tansaku:
:construction:

PR #166, opened 2015-09-17
Amended takeway challenge post feedback

lucetzer:


tansaku:
Great work following up on this @lucetzer - you fixed all the rspec output issues, README is updated (but still not showing us the whole shebang, i.e. what methods do we call on Customer after we've created one); you've got 100% test coverage (awesome), but used my rough code as is without extending its rigor or going to [VCR](https://github.com/vcr/vcr) which would be my recommendation ... 

You didn't move dish_available? into menu, or use Hash.new(0) to simplify your method logic ... this kind of refactoring is great to practise if you have a spare moment ... :-) but don't feel you have to do this - I know that there's lots on your plate.

You have cleaned up the customer public interface, and got your phone number off github, but there are reek issues that you could look into.

So I could give you a sticker, but do you want to take a stab at these others first?  You've done a lot of good work and certainly deserve a sticker.  Either way find me on Monday and we can hook you up with a sticker - if you just want to concentrate on the new challenge this weekend I totally understand :-)

lucetzer:
Thanks Sam for all your comments. I've refactored dish_available? and I will look into VCR and dotenv when I get some spare time.

tansaku:
great work @lucetzer see me to get a sticker today :-)

:checkered_flag:

PR #169, opened 2015-10-23
testing code review process

tansaku:


lib/dish.rb
```diff
@@ -0,0 +1,7 @@
+class Dish
+  attr_accessor :name, :price, :description
```
bat020: should these properties even be mutable? perhaps just an attr_reader here.

lib/dish.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  def initialize name = 'No Name', price = 'No Price', description = 'No Descr.'
```
bat020: surely we'd never want to initialise a dish with no name, price etc. so these defaults should be taken out.

PR #175, opened 2015-10-24
Pull Request for takeaway-challenge

DeathRay1977:


README.md
```diff
@@ -1,8 +1,86 @@
...
+ 
+ * print_menu - Print the menu to the screen.
+ * pick_item - Pick an item from the menu.
+ * add_to_order - Add the item to the order.
+ * verify_total - Verify the total against estimate.
+ * notifiy - notify customer delivery in an hour.
+
+
 * Feel free to use google, your notes, books, etc but work on your own
 * You must submit a pull request to this repo with your code by 9am Monday morning
```
danldb: Good readme

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+                 order_klass:,
+                 order_total_klass:,
+                 notifier_klass:,
+                 config_loader_klass:,
+                 delivery_estimator_klass:)
+    @menu = menu_klass.new(menu_loader_klass: menu_loader_klass)
+    @order = order_klass.new(order_total_klass: order_total_klass)
+    @notifier = notifier_klass.new(config_loader_klass: config_loader_klass)
+    @estimator = delivery_estimator_klass.new
+  end
```
danldb: You were absolutely correct in the standup - many instance variables and an overly-long parameter list. I feel like this is a symptom of pushing all the dependencies up to the top of the tree.

lib/delivery_estimator.rb
```diff
@@ -0,0 +1,5 @@
+class DeliveryEstimator
+  def eta
+    Time.now + 60 * 60
+  end
+end
```
danldb: Over-engineered: Here you have a class that has a single responsibility, but it is a vacuous responsibility.

lib/config_loader.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+  def account_sid
+    fail "Account SID Missing" unless @config.key?('Account SID')
+    @config['Account SID']
+  end
+  def auth_token
+    fail "Auth Token Missing" unless  @config.key?('Auth Token')
+    @config['Auth Token']
+  end
+end
```
danldb: This wasn't really needed, but quite clever. Avoid 'er' at the end of class names - if it is a service object it would end in a verb.

lib/menu_loader.rb
```diff
@@ -0,0 +1,16 @@
...
+    $/="\n\n"
+    File.open(file, "r").each do |object|
+      @objects << YAML.load(object)
+    end
+    objects
+  end
+  def objects
+    @objects.dup
+  end
+end
```
danldb: I don;t see the need for a menu loader - this is again a symptom of going beyond the brief - there was no mention of a persistent data model i.e. yaml file.

lib/notifier.rb
```diff
@@ -0,0 +1,17 @@
+require 'twilio-ruby'
+class Notifier
+  def initialize(config_loader_klass:)
```
danldb: no reason to use klass pattern here - just pass an instance. Why is Notifier responsible for building new instances?

lib/notifier.rb
```diff
@@ -0,0 +1,17 @@
...
+    config_loader = config_loader_klass.new('./support/config.yaml')
+    account_sid = config_loader.account_sid
+    auth_token = config_loader.auth_token
+    @client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: _this_ is what should be extracted out and replaced with a double in your testing.

lib/notifier.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+  def notify(estimated_time)
+    time = estimated_time.strftime("%H:%S")
+    b = "Thank you! Your order was placed and will be delivered before #{time}"
```
danldb: b is completely meaningless - think about your naming

lib/order.rb
```diff
@@ -0,0 +1,19 @@
...
+    @order_totaller = order_total_klass.new
+    @order_lines = Hash.new(0)
+  end
+
+  def add(dish, qty)
+    @order_lines[dish] += qty
+  end
+
+  def order_lines
+    @order_lines.dup
```
danldb: good

lib/order_total.rb
```diff
@@ -0,0 +1,9 @@
+require 'bigdecimal'
+
+class OrderTotal
+  def calculate(order_lines)
+    total = BigDecimal.new(0)
+    order_lines.each {|dish, quantity| total += (dish.price * quantity)}
+    total.to_f
+  end
+end
```
danldb: Again, I don't see the need for this class.

spec/config_loader_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    it 'Loads the Account SID' do
+      expect(config_loader.account_sid).not_to be_empty
+    end
+
+    it 'loads the Auth Token' do
+      expect(config_loader.auth_token).not_to be_empty
+    end
+  end
```
danldb: beyond the brief, but I concede it is nicely done.

spec/delivery_estimator_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  let(:now) {Time.new(2015, 10, 25, 0, 0, 0, "+00:00") }
+  subject(:delivery_estimator) {described_class.new}
+
+  context 'Estimate time of delivery' do
+    it {is_expected.to respond_to(:eta)}
+
+    it 'Returns a time an hour into the future' do
+      allow(Time).to receive(:now).and_return now
+      expect(delivery_estimator.eta).to eq(now + 60 * 60)
+    end
```
danldb: good stubbing

danldb:
:checkered_flag: 

So much better than last week - well done! 

You are still going beyond the brief, so please do try to cut down on that. I still find that your submissions are over-complicated. As the challenges get more difficult, you will see that the trick to solving them is to simplify and simplify again until you find the real core of the problem. Here are some rules for simple design:

1. Passes the tests
2. Reveals intention
3. No duplication
4. Fewest elements

In that order of priority.

In this case, I think the 4th rule was the killer :)

Amazing to see such a huge improvement and a good submission, thanks Norm :)

PR #194, opened 2015-10-26
Week 2. Takeaway Challenge

yevdyko:


spec/menu_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'menu'
+
+describe Menu do
+  subject(:menu) { described_class.new }
+
+  context '#initialize' do
```
DeathRay1977: Vacuous test?

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
...
+             'vegetable tempura' => 7.00,
+             'beef soup ramen' => 6.50 }
+
+  def initialize(dishes = DISHES)
+    @dishes = dishes
+  end
+
+  def open
+    dishes.map { |item, price| "#{item}: £#{sprintf('%.2f', price)}" }
+  end
```
danldb: great

lib/messenger.rb
```diff
@@ -0,0 +1,16 @@
+require 'twilio-ruby'
+
+class Messenger
+  account_sid = "PN2ba53d6e4525039c3c3c17ebb2bc2a15"
+  auth_token = "d8bfb8790ec316cd64d6d254a4bcc6c6"
+
+  @client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: this could be injected to allow for easier testing

lib/order.rb
```diff
@@ -0,0 +1,21 @@
+class Order
+  attr_reader :basket, :check, :all
+
+  def initialize
+    @basket = Hash.new(0)
+  end
+
+  def add_basket(item, num)
+    basket[item] += num
```
danldb: what happens if the item doesn't exist previously?

lib/order.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def basket_summary(menu)
+    @all = menu.dishes[item]*num
```
danldb: what does `item` refer to here?

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
+class Takeaway
+  attr_reader :menu, :order, :messenger, :result
+
+  ERROR = "Total cost does not match the sum of the dishes in your order!"
```
danldb: make a new error class rather than make a constant error.

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize(menu = Menu.new, order = Order.new, messenger = Messenger.new)
```
danldb: do you need a new order every time a takeaway is initialized?

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+    @messenger = messenger
+  end
+
+  def open_menu
+    menu.open
+  end
+
+  def add_to_order(item, num = 1)
+    order.add_basket(item, num)
+    "#{num}x #{item}(s) added to your basket."
```
danldb: don't return a string

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+    "Total: £#{@result}"
+  end
+
+  def checkout(amount = 0)
+    is_correct_amount?(amount) ? (send_sms) : (fail ERROR)
+  end
+
+  private
+
+  def empty?
```
danldb: Don't check the state of another object.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it 'reports items being added to basket' do
+      expect{takeaway.order('kimchi', 2)}.to change{order.basket.length}.by(1)
+    end
+  end
+
+  describe '#order_summary' do
+    it 'creates an order summary' do
+      takeaway.order('kimchi', 2)
+      takeaway.order('salmon maki', 3)
+      expect(takeaway.order_summary).to eq("kimchi x2 = £6.00, salmon maki x3 = £16.50")
```
danldb: you test the return value (which is meaningless in this case as it returns a string that will never be seen), but does not test the action - which is that the variable @all should be modified. If you had tested that, I suspect you would have spotted an error you had made in your  code. This would have been picked up if you had written your test first.

danldb:
:construction: 

CI failed (automatic construction flag).

Focus on testing _first_. Some of your tests did not test the action of the method, instead they only tested the return value. A reasonable attempt, but it is really important that you nail the TDD workflow. Keep practicing :)

PR #196, opened 2015-10-26
take-away challenge

jrose111284:


lib/menu.rb
```diff
@@ -0,0 +1,16 @@
...
+          "cheesebuger"=>1.49,
+          "Hamburger"=>0.99,
+          "Fries"=>0.89, }
+
+    def initialize(menu = MENU)
+      @menu = menu
+    end
+
+    def read_menu
+      MENU
```
danldb: should be `@menu` instead of `MENU`. However this entire method is unnecessary as you have an attr_reader.

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
+class Menu
+  attr_reader :MENU
```
danldb: should be `attr_reader :menu`. Constants have global access anyway.

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  MENU = { "hotdog"=>1.99,
+          "cheesebuger"=>1.49,
+          "Hamburger"=>0.99,
+          "Fries"=>0.89, }
```
danldb: should add `.freeze` on a constant to prevent it from being changed.

lib/sms_text.rb
```diff
@@ -0,0 +1,17 @@
+require 'rubygems'
+class Sms_Text
+
+	def send_text
+		account_sid = 'AC1cffe35eb134133af61d8ab716f40588'
+		auth_token = '48e47a677c14dd9162f479523ed7fb3c'
```
danldb: these shouldn't be on github - they give anyone access to your account. USe environment variables or dotenv instead.

lib/sms_text.rb
```diff
@@ -0,0 +1,17 @@
...
+
+		@client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: This should be injected in to make the SMS class reusable - what if you change SMS provider? Currently you would have to rewrite this entire class.

lib/sms_text.rb
```diff
@@ -0,0 +1,17 @@
...
+
+		@client.account.messages.create({
+		:from => '+442033225791',
+		:to => '+4477837673624',
+		body: "Your order has been accepted!  #{view_current_order} will arrive by #{(Time.now+3600).strftime('%H:%M')}"
+		})
+  end
+
+end
```
danldb: Pay attention to hound - it is telling you good things here!

lib/take_away.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'menu'
+class Takeaway
+
+  attr_reader :my_order, :check_out
+  
+  MENU = { "hotdog"=>1.99,
```
danldb: Duplication - you already have a default menu.

lib/take_away.rb
```diff
@@ -0,0 +1,47 @@
...
+          "cheesebuger"=>1.49,
+          "Hamburger"=>0.99,
+          "Fries"=>0.89, }
+
+  def initialize(menu_klass = Menu.new)
```
danldb: `_klass` convention refers to an actual class, here you are passing an instance, not a class. It should just be `menu: Menu.new`

lib/take_away.rb
```diff
@@ -0,0 +1,47 @@
...
+    @menu = menu_klass
+    @my_order = {}
+    @total = 0.0
+  end
+
+  def read_menu
+    MENU
```
danldb: makes no sense to expose a constant through a method

lib/take_away.rb
```diff
@@ -0,0 +1,47 @@
...
+      my_order[menu_item] = my_order[menu_item] + 1
+    else
+      my_order[menu_item] = 1
+    end
+  end
+
+  def calculate
+    my_order.each do |item, quantity|
+    item_price = MENU[item]
+    @total += item_price * quantity
```
danldb: indentation!

spec/take_away_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      subject.order('hotdog')
+      subject.order('cheesebuger')
+      subject.order('Fries')
+      subject.order('Fries')
+      expect(subject.calculate).to eq 7.25
+    end
+  end
+
+  context '#basket' do
+    it { is_expected.to respond_to :basket }
```
danldb: this doesn't really test the basket method

spec/sms_text_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'sms_text'
+
+  describe Sms_Text do
+    subject(:sms_text) { described_class.new }
+
+    before do
+      allow(subject).to receive(:send_text)
+    end
+
```
danldb: no tests for this class?

spec/take_away_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+  end
+end
```
danldb: no communication between takeaway and SMS?

danldb:
:construction: 

Generally good. You have totally isolated your SMS class from the Takeaway - which means the takeaway is not currently able to send SMS. You should be injecting your client into the class so that you can later replace it with a double and thus avoid stubbing subject.I would recommend revisiting this during lab week.

PR #197, opened 2015-10-26
Dan's Takeaway Challenge

d9nny:


spec/text_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe Text do
+	subject(:text) {described_class.new}
+	let(:number) {double "07966793842"}
+	let(:client) {double :client}
+	let(:messages) {double :messages}
+
+	context '#message' do
+		it 'sends a message' do	
+
```
DeathRay1977: Vacuous test.

spec/text_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'text'
+
+describe Text do
```
DeathRay1977: No stubbing of Twilio API

lib/confirmation.rb
```diff
@@ -0,0 +1,20 @@
+class Confirmation
```
DeathRay1977: It does two things. Verifies user cost and totals up the real cost. Maybe pull one of the responsibilities into another class?

lib/text.rb
```diff
@@ -0,0 +1,19 @@
+require 'rubygems' 
+require 'twilio-ruby' 
+
+class Text 
+
+	def initialize
+		account_sid = 'AC11aeed4f312018cfab3572efe4aeb2ee'
```
DeathRay1977: Contains personal info.

lib/menu.rb
```diff
@@ -0,0 +1 @@
+{fish_and_chips: 5.99, curry_and_rice: 7.99, wun_ton_soup: 5.99, sweet_and_sour_pork: 6.50, prawns_chop_suey: 5.50}
```
danldb: no class here? If you are intending to hold an external state, you could use YAML, JSON, CSV or even plain text. I would expect a .rb file to be executable.

lib/takeaway.rb
```diff
@@ -0,0 +1,31 @@
...
+		menu.each { |meal,price| print "#{meal}		 £#{price}\n"}
+	end
+
+	def place_order(dish, qty)
+		 @checkout[dish] = [menu[dish],qty] 
+	end
+
+	def confirm(user_total, confirmation = Confirmation.new(checkout))
+		confirmation.user_cost(user_total)
+	end
```
danldb: Not sure if this needs a separate class - I am in two minds. On the 1 hand it does help to simplify this class, but on the other hand I'm not sure I see a clear responsibility (the confirmation could be handled by a `fail` at the point of confirmation.). At risk of over-engineering.

lib/takeaway.rb
```diff
@@ -0,0 +1,31 @@
...
+
+	def proceed(text = Text.new)
+		text.message
```
danldb: Nice - very simple, and good use of DI.

lib/text.rb
```diff
@@ -0,0 +1,19 @@
+require 'rubygems' 
+require 'twilio-ruby' 
+
+class Text 
+
+	def initialize
+		account_sid = 'AC11aeed4f312018cfab3572efe4aeb2ee' 
+		auth_token = '7d7dd325948ef9ecc02f99e797d6cce0' 
+		@client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: I would use injection for the client to allow easy stubbing.

danldb:
:checkered_flag: 

Well done - generally good, I can see that you are really thinking about OOD. There were a few areas where you could improve, that you may want to refactor later on - see inline notes.

PR #238, opened 2015-12-07
Take away challenge

Adzz:
Still unfinished, major need of refactor, and currently text send on programme start, need to change this.
README also needs updating.


tansaku:
You had your takeaway_challenge reviewed by **Alaan Rahim**.
### The good points are:
* All public interface elements are tested
* The solution correctly prefers classes to modules
* Business and presentation logic has been cleanly separated
* The solution makes good use of the Ruby language, e.g., reduce/inject is used to aggregate (e.g. sum) over a collection

* README should be updated
* README should include instructions on how to use application
* Tests should test real behaviours not stubs
* Stubbing the Twilio API
* Unit vs Integration tests
* Appropriate use of Dependency Injection
* Personal details and tokens on GitHub
* Open Closed Principle

### Additional comments:
- Ensure class names in the describe section of your rspec files are not strings.
- Think about modelling the task in terms of the real world, i.e. you cannot place an order through a menu; a customer would need to place the order using a menu.
- Try not to over-complicate methods - having too many lines in a method may be a sign of doing this (5 lines or over is typically too much).
- Use Hash.new(0) to be able to easily push new key/value pairs into a hash, while using the inject/reduce Ruby methods to add to a menu item's quantity.
- Export ENV variables to your .zshrc file and reference them in your ruby code through: ENV['NAME_OF_VARIABLE']


see https://github.com/makersacademy/takeaway_challenge/blob/master/docs/review.md for more details on the items above

see also this video exemplar on how a coach would approach this challenge: https://www.youtube.com/watch?v=mgiJKdH9x8c

PR #273, opened 2016-01-31
Takeaway Challenge - Week 2

MarcoCode:


mechacoach:
You had your takeaway-challenge reviewed by **Godfrey Chiu**

### The good points are:

* The README has been updated
* The README has instructions
* The Twilio API is stubbed appropriately
* The solution correctly prefers classes to modules
* The Twilio dependency can be injected appropriately
* Business and presentation logic has been cleanly separated
* ENV vars have been used in place of sensitive data, The dotenv gem has been used to ensure others can get set up easily with the necessary environment variables
* The solution makes good use of the Ruby language, e.g.
* The open closed principle has been respected, e.g. food items are not hard coded into restaurant

* Tests should test real behaviours not stubs
* Unit vs Integration tests
* Explicit tests for every element of the public interface

### Additional comments:
Investigate London vs Chicago style unit tests using doubles

Possibly inject your arguments from feature spec (and manually in irb) rather than doing it directly in takeaway.rb. That way, when you unit test takeaway, you are forced to create doubles and stub methods.

Need to create doubles/stubs in your takeaway_spec.rb

Investigate setter methods vs self.status = [], vs @status = []. Which one is better syntatically, which is better Ruby practice?

see https://github.com/makersacademy/takeaway-challenge/blob/master/docs/review.md for more details

PR #283, opened 2016-01-31
Takeaway Weekend Challenge

katie210:


lib/sms.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
+
+class SMS
+
+  def send_text(message)
```
hedudelgado: The Readme is missing, remember to fill it up when you have a chance ^__^

hedudelgado: Be careful, the sensitive data might be exposed :) (private)
Some methods may be moved to private as there is no need to call them to make the program work.

hedudelgado: Remember to remove comments.

mechacoach:
You had your takeaway-challenge reviewed by **Eduardo**

### The good points are:

* All tests test real behaviour and not stubs
* The Twilio dependency can be injected appropriately
* reduce/inject is used to aggregate (e.g. sum) over a collection

* README should be updated
* README should include instructions on how to use application
* Stubbing the Twilio API
* Unit vs Integration tests
* Explicit tests for every element of the public interface
* Use of modules
* Separation of Concerns
* Personal details and tokens on GitHub
* Open Closed Principle

### Additional comments:
The comments have been posted on my partner's repo. 

see https://github.com/makersacademy/takeaway-challenge/blob/master/docs/review.md for more details

PR #305, opened 2016-02-03
An example solution

silvabox:


lib/menu.rb
```diff
@@ -0,0 +1,44 @@
...
+  include Enumerable
+  extend Forwardable
+
+  def_delegators :items, :each, :include?, :member?, :empty?
+
+  def self.from_hash(hash)
+    menu = Menu.new
+    hash.each { |name, price| menu.add(name, price) }
+    menu
+  end
```
ptolemybarnes: Could be:
```
def self.from_hash(hash)
   hash.each_with_object(Menu.new) do |(name, price), menu|
      menu.add(name, price)
   end
 end
```

If ya wanna be cool

lib/menu.rb
```diff
@@ -0,0 +1,44 @@
...
+  def add(name, price)
+    items[name] = price
+  end
+
+  def price_of(name)
+    items[name]
+  end
+
+  def to_s
+    string = ""
```
pitchinvasion: `each_with_object` works here too

spec/menu_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'spec_helper'
+require 'menu'
+
+describe Menu do
+  subject(:menu) { Menu.new }
+  let(:items) { { :'Pizza' => 9.25, :'Chips' => 2.00 } }
```
sjmog: Does it not make more sense to use lowercase symbols and some sort of `humanize` method in a view layer?

silvabox: Actually, in production they wouldn't be symbols at all - I should probably just use strings here.

lib/menu.rb
silvabox: Like!  I didn't know that!

PR #308, opened 2016-03-12
Week 2 weekend challenge - Takeaway-challenge

hkp108:


silvabox:
Hi Harsheet, thanks for submitting your takeaway challenge.  You've done a pretty good job - the code is clean and well structured.  Arguably, you can simplify `Menu` by adding `Dish` instances directly rather than creating them inside `add_dish`, then you don't need the class dependency injection.  Good work though.

silvabox:
Sorry that comment was only posted today - it's been sat in a window on my machine since Friday!

PR #314, opened 2016-03-12
Takeaway challenge

junyuanxue:


silvabox:
Hi Junyuan, thanks for submitting your takeaway challenge.  You code is clean and your solution is nicely designed.  You are breaking encapsulation a bit by exposing you menu and baskets so have a think about how you might fix that.

PR #316, opened 2016-03-12
Takeaway Challenge

MuriloDalRi:


silvabox:
Nice work Murilo, this is a very thorough solution!  You could improve it by removing the responsibility from Takeaway for creating orders and just pass a constructed order into the takeaway for completion.

PR #318, opened 2016-03-12
Functionality done, experimenting on heroku branch

paulalexrees:


silvabox:
Hi Paul, thanks for submitting your takeaway challenge.  I'm inclined to say your solution is incomplete since you are relying on passing hashes around, which is definitely not satisfactory.  Also, there is no validation of the order against the menu.  I think you got distracted by trying to deploy it and lost sight of what the challenge was actually about.

PR #319, opened 2016-03-13
Takeaway Challenge - Anne

AnnemarieKohler:


silvabox:
Hi Anne, thanks for submitting your takeaway challenge.  Have a look at your design - you can simplify some of you dependency injection by injecting an instance rather than a class.  For example:
```
@order_confirmation_class = order_confirmation_class.new
```
Just inject an instance of `OrderConfirmation` from the off.  Note also you've suffixed your variable name with `_class` but set it to an instance.

PR #321, opened 2016-03-13
Takeaway Challenge Pull Request

eripheebs:


silvabox:
Hi Erika, thanks for submitting your takeaway challenge.  Your code is simple and clean which is good.  There is a weird naming issue: `@baske` instead of (I presume) `@basket`?

Passing all of those parameters to `Takeaway.new` is a red flag and can easily be avoided: you were so close!  There's no need to inject all of those dependencies and "`_klass`" dependencies.  Have a look at the solution I posted to see what I mean

PR #322, opened 2016-03-13
New beginning

hawksdoves:


silvabox:
Hi Phoebe, thanks for submitting your takeaway challenge.  Your solution has some structural issues and breaks encapsulation in places.  You need to work on your naming of things too as it makes for confusing reading.  Try looking at it again in a few days time without bias towards your thought process the first time around.

PR #327, opened 2016-03-13
Takeway challenge

ccfz:


silvabox:
Hi Caspar, thanks for submitting your takeaway challenge.  There are issues with your design: your order class has too many responsibilities and you are overly reliant on arrays.  You should also address some of your naming, which is slightly inconsistent.  Have a look at the example solution I posted for some guidance.

PR #329, opened 2016-03-13
weekend challenge

sachinkaria:
challenge completed without twilio


silvabox:
Hi Sachin,

A reasonable start, but your encapsulation is poor and there are some problems with your naming.  `order.myorder` and `order.place_order` kind of suggest a problem with the design.

PR #336, opened 2016-03-15
Challenge pull request

kylebuttner:


silvabox:
Hi Kyle,
Thanks for submitting your takeaway challenge.  Your classes should be in `lib` rather than `docs`.  You should not expose the `@view` array outside of `Order` and is `view` the right name for it?  There is also a logic error with your quantity setting in dish: what happens if you add a dish more than once?

PR #338, opened 2016-04-22
Takeaway Challenge

GeekG1rl:


silvabox:
You've got the right sorts of class names, but you are breaking encapsulation by reaching into the hash stored inside `Menu`.  Your order tests seem to switch between `order` and `test_order` - it's not clear why.  You might consider only calculating the total price when it's required, rather than keeping track of it as state in the `Order` class.

PR #339, opened 2016-04-22
Solution for takeaway_challenge

omajul85:


lib/feature_test
```diff
@@ -0,0 +1,3 @@
+dishes = {"Chicken"=>4.5, "Pork"=>5.5, "Steak"=>9.3}
+m = Menu.new dishes
+t = Takeaway.new m, Order, Notifier
```
danldb: should this have been committed?

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
+  
+  attr_reader :dishes
+  
+  def initialize(dishes)
+    @dishes = dishes
+  end
+  
+  def read_menu
+    return "The menu is empty" if dishes.empty?
```
danldb: presentation logic

lib/order.rb
```diff
@@ -0,0 +1,51 @@
...
+  
+  def initialize(menu)
+    @dishes = {}
+    @total = 0
+    @menu = menu
+  end
+  
+  def add(dish, qty = 1)
+    fail ERR_INVALID_ORDER unless menu.contains? dish
+    dishes[dish] = qty
```
danldb: what happens if you want to add more than 1 quantity of the same dish?

```ruby
order.add(:pizza, 2)
order.add(:pizza, 1)
```
I would expect to receive 3 pizzas.

lib/order.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+  
+  alias update add
```
danldb: ?

lib/order.rb
```diff
@@ -0,0 +1,51 @@
...
+  
+  def remove(dish)
+    dishes.delete dish if contains?(dish)
+  end
+  
+  def total
```
danldb: you already have an attire_reader :total, that is being overwritten by this method

lib/order.rb
```diff
@@ -0,0 +1,51 @@
+class Order
+  
+  ERR_INVALID_ORDER = "The selected dish is not on the menu".freeze
+  ERR_EMPTY_ORDER = "The order is empty".freeze
+  
+  attr_reader :dishes, :total
+  
+  def initialize(menu)
+    @dishes = {}
+    @total = 0
```
danldb: this instance variable never gets used

danldb:
:eyes:

classes have large public interfaces - do they need all these public methods? 

worth drawing a domain diagram before you start coding - identify the messages (i.e. dependencies) that will be passed between classes.

Your methods are generally concise and readable - you obviously spend time thinking about method naming.

PR #341, opened 2016-04-23
Takeaway Challenge

lexiht:


config.yml
```diff
@@ -0,0 +1,7 @@
+{
+  account_sid: 'AC08b3a906870b047455fd47f4edf71587',
+  auth_token: '2065af8a307ed31b5cea7ff031f769bb',
+  from: "+447481343709",
+  to: "+841649129008",
+  body: "Thank you! Your order was placed and will be delivered before %s"
+}
```
danldb: this file should be in your .gitignore if you wanted to hide your secret keys and phone numbers

lib/dish.rb
```diff
@@ -0,0 +1,10 @@
+class Dish
+
+  attr_accessor :quantity
+  attr_reader :args
+
```
danldb: public attr_accessors break encapsulation -a void them wherever possible.

What is this class actually doing? What is its responsibility?

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+    @dishes = []
+  end
+
+  def print_menu
+    @dishes << @dish
+    @dishes.map do |dish|
+      "#{dish.args[:name]}: £#{dish.args[:price]}"
+    end.join(", ")
+  end
+
```
danldb: Every menu only has one dish

lib/order.rb
```diff
@@ -0,0 +1,19 @@
...
+    @menu = menu
+    @selected_dishes = []
+  end
+
+  def total
+    @selected_dishes.map do |dish| 
+      dish.quantity * dish.args[:price]
+    end.inject(:+)
+  end
+
```
danldb: over reliant on the attr_accessor to perform any functionality.

lib/sms.rb
```diff
@@ -0,0 +1,33 @@
+require 'twilio-ruby'
+require 'pry'
+require 'yaml'
+
+class SMS
+
+  TIME_FORMAT = "%H:%M"
+
+  def initialize(config, client: nil)
+    @client = client || Twilio::REST::Client.new(config[:account_sid], config[:auth_token])
```
danldb: good use of Dependency Injection

lib/sms.rb
```diff
@@ -0,0 +1,33 @@
...
+    @config = config || YAML.load_file('config.yml')
+  end
+
+  def deliver
+    client.account.messages.create(args)
+  end
+
+  private
+
+    attr_reader :config, :client
```
danldb: good use of private attr_reader

lib/sms.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def args
```
danldb: method name does not communicate what it returns.

lib/takeaway.rb
```diff
@@ -0,0 +1,37 @@
+require 'pry'
```
danldb: don't need to require pry here

spec/order_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+describe Order do     #less let
+
+  subject(:order) { described_class.new(menu) }
+  let(:takeaway) { Takeaway.new(menu: menu, order: order, sms: sms, config: :config) }
+  let(:menu) { double :menu, dishes: [dish, dish2] }
+  let(:dish) { Dish.new(name: "Pizza", price: 7.50) }
+  let(:dish2) { Dish.new(name: "Soup", price: 3.20) }
+  let(:dish1) { double :dish }
+  let(:sms) { instance_double "SMS", deliver: nil }
```
danldb: Mixing real objects with doubles. The only real object in this spec file should be from the Order class.

danldb:
:eyes:
Over-reliance on attr_accessors to get and set instance variables - your classes become overly coupled as a result.

attr_accessors are usually a code smell - they should basically never be on the public interface.

PR #343, opened 2016-04-23
Ran bundle, files created

iammatthewward:


silvabox:
This is a pretty good attempt, but there are some simple improvements you can make.  You don't need to create a new instance of `Message` for every customer, you could just inject an instance into `Customer`, rather than the class.  You are breaking encapsulation by reaching into the hash inside `Menu`; instances of `Menu` should provide methods to make this unnecessary.

In `delivery_time`, use the `Time` object's built in formatting methods (i.e. `strftime`) rather than interpolating 2 new time objects.

In your `takeaway_spec` you have a lot of unnecessary `let` statements.  Use `let`s to set up objects that are used across multiple tests, otherwise just define the object inside the `it`.

PR #344, opened 2016-04-23
Takeaway challenge

OMGDuke:


silvabox:
You have a nice structure, but there are some naming issues which are east to fix.  `del_time` should be `delivery_time`; which may seem trivial, but remember you want to optimize for readability.  When seeing `del_time` out of context, there is an unnecessary cognitive load to realize it means `delivery_time`.  When writing predicate methods avoid names like `not_enough?`, otherwise you risk introducing double negatives.

In the code:
```ruby
def confirm(pay)
  pay_err = "You have not paid enough, "\
  "The total is £#{format('%.2f', order_cost)}"
  fail pay_err if not_enough?(pay)
  @message_class.new(del_time).send
end
```
you create the pay_err string before the guard clause, so for all calls to this method where `pay` is correct, that string is created unnecessarily.

This code breaks encapsulation:
```ruby
def on_menu?(item)
  @menu.dishes.any? {|dish| dish.name == item}
end
```
By exposing `dishes`, you've created feature envy.  The iteration should be pushed down into `Menu`, so you might have:
```ruby
def on_menu?(item)
  @menu.include?(item)
end
```

PR #345, opened 2016-04-23
Started takeaway challenge

llcclarke:


silvabox:
You've made a reasonable start and it looks like you were planning to extract the menu and order classes.  Did you run out of time?

PR #346, opened 2016-04-23
first commit takeaway-challange

elibar-uk:


silvabox:
Hi Elia, thanks for submitting your Takeaway Challenge.  It seems like maybe you got stuck?  Sometimes it helps to just 'spike' the code first (write some code without tests to get a feel for what might work, then _throw it away_).

You've got `Menu` and `Customer` classes, which is a good start, but it's not clear what their responsibilities are; did you do any design before starting to code?

elibar-uk:
Hi Ben,
My whole idea on design was based on not having to hardcode the menu items
into the menu class. However, it wasn't possible  to implement (in terms of
RSpec) as I couldn't find the way to mock the 'hash behaviour'.
The next stage from where it is now would be to extract the order class out
of the takeaway (the one you called Customer).

PR #347, opened 2016-04-23
Michael Harrison Takeaway Challenge

harrim91:
Week 2 weekend challenge


silvabox:
On the whole, this is pretty good.  You've made good choices for your object design and the responsibilities are well defined.

You could tidy up your tests a bit by reducing the number of `let`s and repeated random assignments and/or reorganizing them so they're only defined for the tests that need them.

Beware assigning dynamic content in a constant:
```
CONFIRMATION = "Thanks! Your order will be delivered before"+ "#{(Time.new + 3600).strftime('%H:%M')}".freeze
```
The time portion is only evaluated once - when the file is loaded - so all orders will be sent with the same message; having a time 1 hour after the program started.

PR #348, opened 2016-04-23
One maybe two user stories finished

nmrichards:


lib/messages.rb
```diff
@@ -0,0 +1,21 @@
+class Messages
+  require 'twilio-ruby'
+  SID = ENV["SID"]
+  TOKEN = ENV["TOKEN"]
+  TWI = ENV["TWILIO_NUMBER"]
+  TO = ENV["PHONE"]
```
danldb: TO should probably not be a constant - the customer's number can change

lib/messages.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def initialize(from = TWI, to = TO, account_sid = SID, auth_token = TOKEN)
+    @client = Twilio::REST::Client.new account_sid, auth_token
+    @from =  from
+    @to = to
+    current = Time.now
+    @t = format('%d:%02d', current.hour + 1, current.min)
```
danldb: Naming - don't use single letter abbreviations - you will forget what they stand for.

lib/messages.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def send_message
+    m = "Thank you! Your order was placed and will be delivered before #{@t}"
+    @client.account.messages.create(from: @from, to: @to, body: m)
+    "Sent message to Nick"
```
danldb: your method name is mixing a verb and a noun, which is confusing - should there be another class called message? Should the method be returning a message? A better name here would be `#send`, allowing you to call `message.send`. Also, when using a verb, you would not expect a response - the string is unnecessary and belongs in a presentation layer.

lib/order.rb
```diff
@@ -0,0 +1,28 @@
+class Order
+  attr_reader :order
+  def initialize(menu)
+    @menu = menu
+    @order = {}
+  end
+
+  def order_food(food, quantity)
```
danldb: again `order.order_food` sounds weird - avoid mixing nouns and verbs in method names

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def order_summary
+    @order.order_summary
+  end
+
+  def confirm_order(total_confirmation, text_message = Messages.new)
+    fail "Incorrect total" unless total_confirmation == @order.total
+    @text_message = text_message
+    messages
+    @order = Order.new(menu)
```
danldb: strong coupling here - you cannot change Order without changing Takeaway

danldb:
:eyes: 

Pay attention to method naming - it's the second hardest problem in software development. 

Did you draw a domain diagram at the beginning? Spend more time considering the messages you are passing between your classes.

PR #349, opened 2016-04-23
Take Away Challenge

festinalent3:


.env
```diff
@@ -0,0 +1,4 @@
+TWILIO_ACCOUNT_SID = 'YOUR_SID'
```
danldb: this whole file should be in your `.gitignore`

lib/menu.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  def initialize
+  self.store(:pizza_deluxe, 15.00)
+  self.store(:salad, 10.00)
+  self.store(:hamburger, 12.00)
+  self.store(:chicken_tikka_masala, 15.50)
+  self.store(:curry, 14.50)
+  end
+
+end
```
danldb: no need for a class here - this is just a Hash.

lib/message.rb
```diff
@@ -0,0 +1,26 @@
...
+  private
+
+  def twilio_send(message_body)
+    begin
+      @client = Twilio::REST::Client.new ENV['TWILIO_ACCOUNT_SID'], ENV['TWILIO_AUTH_TOKEN']
+      message = @client.account.messages.create(:body => message_body,
+      :to => ENV['CUSTOMER_NR'],
+      :from => ENV['TWILIO_NR'])
+    rescue Twilio::REST::RequestError => e
+      puts e.message
```
danldb: avoid `puts` - this is presentation logic, it has no place in the /lib.

lib/order.rb
```diff
@@ -0,0 +1,25 @@
+class Order
+
+  def initialize
+    @list = {}
+  end
+
+  def add(dish, quantity)
+    @list.store(dish.to_sym, quantity)
```
danldb: what happens if they then add another of the same dish?
```ruby
order.add(:fish, 2)
order.add(:fish, 1)
```
I would expect to see 3 fish now.

spec/menu_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe Menu do
+  subject(:menu) { described_class.new }
+
+  it 'presents a list of dishes with prices' do
+    expect(menu).to include :pizza_deluxe
+    expect(menu[:pizza_deluxe]).to eq 15
+
+    expect(menu).to include :hamburger
+    expect(menu[:hamburger]).to eq 12
```
danldb: In unit tests, only 1 expectation per test.

danldb:
I like how clean and concise your methods are.

Your code is very clear and readable - this is a really tough skill to develop!

Keep an eye out for edge cases.

Use DIP to reduce coupling between your classes.

Your should always be able to easily describe the one job of a class in a single sentence - try doing that in a comment at the top of each class.

PR #352, opened 2016-04-23
Takeaway Challenge

ollieh-m:


lib/checker.rb
```diff
@@ -0,0 +1,36 @@
+class Checker
+
+	def check_items(order,menu)
```
danldb: indentation

lib/checker.rb
```diff
@@ -0,0 +1,36 @@
...
+		order.keys.each do |order_item|
+			unless menu.keys.include?(order_item)
+				raise "Sorry, #{order_item} is not on the menu. Please re-order. (View the menu using .view_menu)"
+			end
+		end
+	end
+
+	def check_price(total,order,menu)
+		price = calculate_price(order,menu)
+		if total > price
```
danldb: avoid if / else conditionals - use guard clauses, or extract functionality away. In this case your method has too many responsibilities.

avoid presentation strings inside your error messages

lib/menu.rb
```diff
@@ -0,0 +1,33 @@
...
+
+	def initialize(file)
+		@list = reformat(file)
+	end
+
+	def list
+		@list.dup
+	end
+
+	def list=(new_file)
```
danldb: public attr_accessor totally breaks encapsulation

lib/texter.rb
```diff
@@ -0,0 +1,48 @@
+require 'twilio-ruby'
+
+class Texter
+
+	attr_reader :processed_orders
+
+	def initialize(client=Twilio::REST::Client.new(ENV['TWILIO_ACCOUNT_SID'],ENV['TWILIO_AUTH_TOKEN']))
```
danldb: You could use a constant to set the default - it would be a bit tidier than this long line.

danldb:
:eyes:

Carefully consider the messages that will be passed between your objects - did you draw out a domain diagram, or write down the responsibilities and collaborators for each class?

It may help you to write down the purpose / responsibility of each class as a comment at the top of the file. That way you can be sure your class has one clearly defined job.

Your Checker class is overcomplicated, and I'm not sure it is necessary.

Your menu class consists of basically a public reader and writer - I like the idea of using an external text file, but I'm not sure on the implementation.

Make sure to allow yourself time to refactor after each feature is implemented - it appears as though you moved onto the advanced challenge without first refactoring.

ollieh-m:
Thanks for the feedback.

Unfortunately the answer is yes, I did draw out an overall plan before writing any tests or code - here it is, though the yellow bits I added later, on Monday. So it must be more that I'm not getting how classes should be interacting than that I'm not trying to think it through... 

My thinking was the **Menu** is where the dishes and prices get put in a useable form; the **Establishment** is where you view the menu and place an order; the **Texter** is what sends the text; and the **Checker** is what checks the order when it's placed. I had originally thought that the checking logic could be in the Establishment, but then that was looking like it had too much to do so I thought it should be extracted to a different class...

I'm not sure if the problem is that one or more of these classes is doing too much, or if there are too many classes making things too complicated. I hope these judgments will become clearer for me over the coming weeks, but if you can give me any more of a pointer that'd be great.

[takeaway_challenge_overview.docx](https://github.com/makersacademy/takeaway-challenge/files/243579/takeaway_challenge_overview.docx)

PR #353, opened 2016-04-23
adding menu class

hnamitha1:


lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
+    
+  attr_reader :dishes  
+  def initialize
+   	@dishes = {
+   		"pizza"=>1,
+   		"burger"=>2,
+      "fries"=>1.5
```
danldb: indentation!!

lib/message.rb
```diff
@@ -0,0 +1,25 @@
...
+class Message
+
+  def complete_order(total_price)
+    send_text("Thank you for your order: £#{total_price}. Expect your food by #{delivery_time} ")
+  end
+
+  def send_text(message)
+  	account_sid = 'ACa4cf08b22ebab768c156ad454eaccb71' 
+	  auth_token = '3321406ad1dfa603ec0195d363a5fbef' 
+		@client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: use dependency injection to inject the client to reduce dependency and allow easier testing

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
+class TakeAway
+
+  attr_reader :menu, :basket
```
danldb: why are these attire_readers public?

danldb:
:eyes: 

If you had injected the client into your Message class, you could have replaced it with a double, thus avoiding having to stub the subject under test.

Concentrate on improving your indentation.

PR #355, opened 2016-04-24
Adam Skuse - Takeaway Challenge

AdamSkuse:


lib/confirmation.rb
```diff
@@ -0,0 +1,22 @@
+require 'rubygems'
+require 'twilio-ruby'
+
+
+class Confirmation
+
+  def initialize(phone_number)
+    @account_sid = "AC195e376d56f18ac7d3ed5ca4b7c34c91"
+    @auth_token = "baf8789ae86bff0c9f44bbd73d88c6b9"
+    @client = Twilio::REST::Client.new @account_sid, @auth_token
```
danldb: inject this to reduce dependency

lib/customer.rb
```diff
@@ -0,0 +1,41 @@
...
+
+class Customer
+
+attr_reader :menu, :basket, :phone_number
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @basket = []
+    @running_total = 0
+    @phone_number = "+447880521694" # change this to an arg for customer
```
danldb: don't leave `# todo:` comments! Just do it at the time to avoid forgetting.

lib/menu.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  attr_reader :list
+
+def initialize
+   @list = {
+      a: 10,
+      b: 5,
+      c: 3,
+      d: 4
+    }
```
danldb: wat?

spec/customer_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    it 'has a menu on initialize' do
+      expect(customer.menu).to be_a_kind_of(Menu)
+    end
+  end
+
+  describe '#read_menu' do
+
+    it 'can read the menu' do
+      expect(customer.read_menu).to be_an_instance_of(Hash)
```
danldb: don't type check - prevents future use of polymorphism

spec/Confirmation_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+describe Confirmation do
+  subject(:confirmation) { described_class.new("+447880521694") }
+
+  before do
+    allow(confirmation).to receive(:send)
+  end
+
+  it "sends a confirmation SMS on checkout" do
+    expect(confirmation).to receive(:send)
+    confirmation.send
```
danldb: this test will never fail - are you writing test-first? Always watch the test fail before you make it green.

spec/Confirmation_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'Confirmation'
+
+
+describe Confirmation do
+  subject(:confirmation) { described_class.new("+447880521694") }
```
danldb: if you had injected the client, you could have replaced it with a double here to avoid sending real sms during testing.

danldb:
:eyes:

PR #356, opened 2016-04-24
creating pull request

chriselevate:


danldb:
:eyes:

Not much to add here - you need to start again, test driven, as you mention. I like that you are using DI in the code. I would not focus on this during this week - focus on the current content above all else.

PR #357, opened 2016-04-24
first 2 user stories completed

RichardWatkins1:


lib/app.rb
```diff
@@ -0,0 +1,18 @@
+require 'rubygems'
+require 'twilio-ruby'
+ 
+account_sid = "AC328ae31fa3ea52b0e959f91cb768852b"
```
danldb: No encapsulation - this is in the global namespace.

danldb:
:eyes:

Your build is failing because you didn't run `bundle install` locally before you committed. Your Gemfile.lock is out of sync with your Gemfile.

Please fix your indentation - if this was a tech test I would not give you an interview because your code is hard to read. Indentation in Ruby, by convention, should be 2 spaces.

Great use of DI. 

Avoid polluting the global namespace - encapsulate your SMS client in a class.

PR #358, opened 2016-04-24
First commit - restaurant has inbuilt menu hash and ability to add to…

flimflamjim:
… an order hash.


silvabox:
It's a good start.  You are breaking encapsulation still in your `Order` and `Restaurant` classes though by reaching into their internal hashes.  Although these are cloned, so safe, they will still encourage behaviour to be put in the wrong place.

PR #360, opened 2016-04-24
Takeaway challenge

MaxProvin:


lib/messenger.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
+
+class Messenger
+
+  CLIENT = Twilio::REST::Client.new('ACf630b9f6345f5dd86bfb1705041c59fe', '409a60d7be1227c7598832a89d0a32d2')
```
danldb: hide your secret keys.

lib/messenger.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def send_text
+    CLIENT.account.messages.create({
+      from:'+441143031215',
+      to:'+447929842337',
+      body:"Thank you! Your order was placed and will be delivered before #{delivery_time}"
+    })
+    puts "Sent message to somebody out there"
```
danldb: stop using puts

danldb:
:eyes:

Be careful with presentation logic in your methods. 

I like the small methods and adherence to SRP - definitely keep focusing on that! :)

PR #362, opened 2016-04-24
Take-away Challenge

wrumble:


silvabox:
It's not a bad start - you've maybe hard-coded too much in `Menu` and probably shouldn't be exposing the a hash in `send_menu` (it's a duplicate, so safe, but why do you need a hash?  Are there methods you want from the hash that could be methods on your menu object?).  There are some naming issues, for example `send_menu`, which doesn't really reveal any intention.  Also you end up with `order.order` which is often a clear indication of a naming problem.

PR #365, opened 2016-04-24
Takeaway challenge

RobinHeathcote:


lib/messenger.rb
```diff
@@ -0,0 +1,34 @@
...
+    send_message
+    body
+  end
+
+  private
+
+  FROM = "+441143031281".freeze
+  TO = "+447508420659".freeze
+
+    ACCOUNT_SID = "ACd876ef2e031b1244ad45dbdd3b450231".freeze
```
danldb: you just published your account details and phone number on github

lib/user.rb
```diff
@@ -0,0 +1,3 @@
+require_relative 'menu.rb'
+
+class User
```
danldb: don't commit unfinished code stubs like this.

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
+require_relative 'dish.rb'
+class Menu
+  attr_reader :dishes
+
+  def initialize(dishes:)
+    @dishes = [dishes]
```
danldb: This is a bit strange. I would expect `dishes:` to be a list of dishes, but you are initialising `@dishes = [dishes]` - are you putting one array inside another array?

danldb:
👀

PR #367, opened 2016-04-24
End of challenge week2

Wil0:
Take away challenge. All tests pass


silvabox:
Your class design is OK but you are breaking encapsulation by exposing internal objects.  There is one major logic issue, which is that the `has_dish?` method in `Menu` does not take a parameter.  This should be failing at least one test somewhere as your code simply cannot work in IRB.  The right tests would also reveal that this method:
```ruby
def dish_included?
  @list_of_dishes.each {|dish, price| @list_of_dishes.has_key?(dish)}
end
```
is _always_ truthy.

PR #370, opened 2016-04-25
TakeAway Challenge Complete

DXTimer:


lib/order.rb
```diff
@@ -0,0 +1,41 @@
...
+		p dish
+		p quantity
+		matches?(dish) ? @orders[dish] += quantity : @orders[dish] = quantity
+		puts "#{quantity} x #{dish.name} added to your order."
+	end
+
+	def summary
+		return puts "There are no dishes in your order." if empty?
+		@orders.each do |dish, quantity|
+			@total += (dish.money * quantity).cents
```
danldb: this should be extracted into a separate method

lib/order.rb
```diff
@@ -0,0 +1,41 @@
...
+	def empty?
+		@orders.empty?
+	end
+
+	def matches?(dish)
+		@orders[dish]
+	end
+
+	def print_total
+		Money.new(total, "GBP").format(:symbol => true)
```
danldb: Use Dependency Injection

danldb:
:eyes: 

Failing tests - that is why the build is failing.

Fix your indentation to 2 spaces - it is still set to 4.

Too much presentation logic in your business classes. Avoid puts.

PR #373, opened 2016-04-25
weekend challenge 2

naridas:
previous pull had my number... so had to remove that commit and do forced push

### DIFF
--- a/lib/sms.rb
+++ b/lib/sms.rb
@@ -1,0 +1,39 @@
+require "twilio-ruby"
+
+class SMS
+
+	TIME_FORMAT = "%H:%M"
+
+	def initialize(config, client:Twilio::REST::Client.new(config[:account_sid], config[:auth_token]))
+		@client = client
+		@config = config
+	end
+	def send #changed due to sam
+		client.messages.create(message_args)
+	end
+
+	private 
+
+	attr_reader :client, :config
+
+	def message_args #added cause of sam since makes tons of sense
+		{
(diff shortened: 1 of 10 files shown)


lib/sms.rb
```diff
@@ -0,0 +1,39 @@
+require "twilio-ruby"
+
+class SMS
+
+	TIME_FORMAT = "%H:%M"
+
+	def initialize(config, client:Twilio::REST::Client.new(config[:account_sid], config[:auth_token]))
```
danldb: I would probably store the account_sid & the auth_token in environment variables.

spec/sms_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+	let(:config) do 
+		{
+			account_sid: "C3462f5dec",
+			auth_token: "01d58",
+			from: "+440750",
+			to: "+4407501",
+			body: "Thank you! Your order was placed and will be delivered before %s"
+		}
+	end
```
danldb: This config looks pretty cumbersome to use. Ideally, the only input I would need to send a message would be the telephone number and the body.

danldb:
:eyes:

Be wary of putting personal details into github - you realised in time and were able to cover up the error.

Fix indentation - should be 2 spaces, not 4.

You relied on the config hash quite heavily - SMS and Takeaway had in-depth knowledge of the contents of that hash, leading to more coupling than necessary.

PR #375, opened 2016-04-25
Attempt at solution for the Takeaway challenge

rahulrama:


README.md
```diff
@@ -1,90 +1,71 @@
...
+[![Build Status](https://travis-ci.org/makersacademy/takeaway-challenge.svg?branch=master)](https://travis-ci.org/makersacademy/takeaway-challenge)
 
-Build Badge Example
+Modules and classes used
+------------------------
+The solution makes use of the following classes and modules, each stored in a file of the same name:
+- Takeaway (class)
+  -  (public method)
+  -  (public method)
+- Plane (class)
```
danldb: plane??

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+  LINEWIDTH = 40
+  attr_reader :items
+
+  def initialize
+    @items = []
+    menu1 = File.open('./lib/menu.txt').each_line do |line|
+      menu_line = {}
+      menu_line[:dish], menu_line[:price] = line.chomp.downcase.split(" £")
+      @items << menu_line
+    end
```
danldb: I love that you are storing the menu as a .txt file.

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+    puts "----".ljust(LINEWIDTH/2) + "-----".ljust((LINEWIDTH-34)/2) +
+      "--------".ljust(LINEWIDTH) + "\n"
+    items.each do |x|
+      puts x[:dish].capitalize.ljust(LINEWIDTH/2) +
+        ("£"+x[:price]).ljust((LINEWIDTH-34)/2) +
+        ("x#{x[:quantity]}").ljust(LINEWIDTH)
+    end
+    puts "-"*(LINEWIDTH-9)
+  end
+
```
danldb: Divergent responsibility - what is the single responsibility of this object? You are including presentation logic with logic calculations and other things.

lib/sms.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  CONFIG =
+    { account_sid: "AC1a757e2dd73b62efcbb52def2d72432d",
+      auth_token: "e83f862f205340368bfb1e7a17c3e4c1",
+      from: "+441443606449",
+      to: "+447783980181",
+      body: "Thank you! Your order was placed and will be delivered before %s"
+    }
+
+  def initialize(config = CONFIG, client: nil)
```
danldb: mixing different types of arguments here is confusing. In this case I would use either keyword arguments for both or not use keyword arguments at all.

danldb:
:eyes:

I would revisit this - do you _need_ a presentation layer? Using `puts` to present information to a client made your tests quite difficult and brittle (change one tiny thing in the copy, and your tests will break). 

I like your use of DI. There is a case to be made that the twilio client could also be extracted into a separate class.

PR #470, opened 2016-08-26
Takeaway Challenge

peter-miklos:


lib/dish.rb
```diff
@@ -0,0 +1,12 @@
+#understands the attributes of a dish
```
danldb: good class purpose

lib/menu.csv
```diff
@@ -0,0 +1,15 @@
+1020,Roasted Tomatoes,1.9911
```
danldb: nice - love that you used a CSV for this!

lib/menu.rb
```diff
@@ -0,0 +1,30 @@
...
+
+class Menu
+  FILENAME = './lib/menu.csv'
+  def initialize(filename = FILENAME)
+    @filename = filename
+    @menu_content = Hash.new(0)
+  end
+
+  def menu
+    menu_content.dup #returns a copy of the menu
```
danldb: sweet

lib/order.rb
```diff
@@ -0,0 +1,53 @@
+#understands an order
```
danldb: circular reference - Order understand Order?? There is no meaning here.

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+
+class Order
+
+  def initialize
+    @order_content = Hash.new(0)
+    @checkout_value = nil
+  end
+
+  #add dish with quantity and sum to the order
```
danldb: too many comments - it becomes hard to manage the comments as you are coding - methods change, and their purpose changes, it's easy to leave misleading comments. Let your tests be your documentation.

lib/takeaway.rb
```diff
@@ -0,0 +1,79 @@
...
+  end
+
+  def total #returns the total value of the current order
+    "Total: £#{order_log_class.total}"
+  end
+
+  def checkout(amount) #check out current order and send sms if completed
+    order_log_class.checkout_order(amount)
+    send_text_message(compose_text_body)
+    "Checkout completed. Confirmation has been sent in a text message"
```
danldb: no need to return this string

PR #471, opened 2016-08-27
Initial Pull Request

TomStuart92:


lib/adapter.rb
```diff
@@ -0,0 +1,43 @@
...
+    messages.each{|message| to_redact << message.sid}
+    to_redact.each do |sid|
+      message = client.account.messages.get(sid)
+      message.redact
+    end
+  end
+
+  private
+
+  attr_reader :twilio_number, :phone_number, :client
```
danldb: this class feels like it is over-engineered, and unnecessary.

lib/interface.rb
```diff
@@ -0,0 +1,88 @@
+require_relative 'menu'
+require_relative 'order'
+require_relative 'adapter'
+
+# Understands how to interact with the user
```
danldb: great that you separated this functionality into a separate class, but again this UI is way more than was asked for.

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
+# Understands the pricing of food items
+class Menu
+  def initialize
+    @menu = { 'spring roll' => 0.99, 'peking duck' => 7.99 }
+  end
+
+  def to_s
```
danldb: I would expect this to return a string, but instead is sends messages to STDOUT via puts. Not useful if you want to use the strings elsewhere in your code, as this method returns nil.

lib/order.rb
```diff
@@ -0,0 +1,39 @@
+require_relative 'menu'
+
+# Understands how to aggregate food items from a menu
```
danldb: I like the class responsibilities - not easy to do, and they are well implemented.

danldb:
A little over engineered - beware YAGNI. 

Classes where the name ends in 'er' or 'or' are often code smells - they tend to work on functionality that should be handled elsewhere. Not always, but it's worth keeping in mind. In this case, I felt your Adapter class had a bit of SMS and a bit of Restaurant, but no real functionality of it's own.

PR #472, opened 2016-08-27
Takeaway Challenge

ewansheldon:


lib/confirmation_system.rb
```diff
@@ -0,0 +1,27 @@
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load
```
danldb: nice

lib/confirmation_system.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class ConfirmationSystem
+  attr_reader :msg
+
+  def initialize(client = Twilio::REST::Client.new(ENV['SID'], ENV['TOKEN']))
```
danldb: good use of DI

lib/interface.rb
```diff
@@ -0,0 +1,81 @@
+require_relative 'menu'
+require_relative 'order'
+
+class Interface
+  def initialize(menu = Menu.new, order = Order.new)
+    @menu = menu
+    @order = order
+    puts "Welcome to Ewan's pizzeria!"
```
danldb: no - do not put presentation strings in your business logic, especially in the initialize method.

lib/interface.rb
```diff
@@ -0,0 +1,81 @@
+require_relative 'menu'
```
danldb: great taht you have separated the the ui into a separate class - but the client didn't ask for any of this in the brief.

danldb:
Generally good, but beware building anything that isn't specifically asked for. This is a common issue when doing tech tests - people fail because they build something complex when they only needed to submit something simple.

PR #473, opened 2016-08-27
Initial pull request 

lsewilson:


lib/sms_confirmation.rb
```diff
@@ -0,0 +1,26 @@
+require 'twilio-ruby'
+require 'dotenv'
+
+Dotenv.load
```
danldb: nice use of dotenv

lib/sms_confirmation.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class SMSConfirmation
+
+  def initialize(client = Twilio::REST::Client.new(ENV['SID'],ENV['TOKEN']))
```
danldb: good use of dependency injection, but looks a little weird having all that in the argument. You could also do something like:
```ruby
def initialize(client = nil)
  @client ||= Twilio::REST::Client.new(ENV['SID'],ENV['TOKEN'])
end
```

danldb:
Great submission, thanks!

PR #474, opened 2016-08-27
initial commit

andrea6859:


lib/menu.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  attr_reader :menu
+
+  def initialize
+    @menu =
+    {"Spring Rolls" => 3,
+      "Prawn Crackers" => 1,
+      "Chicken Satay" => 7,
+      "Thai Green Curry" => 8,
+       "Rice" => 2,
```
danldb: indentation - if this was a tech tests, you just failed! ;-)

lib/restaurant.rb
```diff
@@ -0,0 +1,65 @@
...
+def confirm_order(item, quantity)
+"Thank you for your order: #{quantity} => #{item}."
+end
+
+def delivery_time
+ (Time.now + (60*60)).strftime("%H:%M")
+end
+
+def confirm
+  "Thank you for your order.  The total price is £#{total_price}. It will be delivered by #{delivery_time}"
```
danldb: lots of presentation strings that were never asked for in the brief.

lib/text_message.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
+
+
+class TextMessage
+
+def send_confirmation(confirm)
+
+account_sid = ENV['TWILIO_ACCOUNT_SID']
+auth_token = ENV['TWILIO_AUTH_TOKEN']
```
danldb: nice but should be constants rather than variables in the method

lib/text_message.rb
```diff
@@ -0,0 +1,21 @@
...
+
+
+@client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: use Dependency injection to inject the client

PR #476, opened 2016-08-28
First commit: created menu class and menu_spec

littlethao:


lib/menu.rb
```diff
@@ -0,0 +1,10 @@
+# Understands the menu items and their values
```
danldb: nice

lib/messenger.rb
```diff
@@ -0,0 +1,30 @@
+# Understands the delivery of confirmation message.
+
+require 'rubygems' # not necessary with ruby 1.9 but included for completeness
```
danldb: if it's not needed in the ruby version you are using, don't leave it in your code.

lib/messenger.rb
```diff
@@ -0,0 +1,30 @@
...
+    # set up a client to talk to the Twilio REST API
+    @client = Twilio::REST::Client.new @account_sid, @auth_token
+  end
+
+  def confirm_order
+    send_sms("Thank you for your order! Delivery will be before #{generate_delivery_time}")
+  end
+
+  def send_sms(message)
+    @client.account.messages.create({
```
danldb: use an attr_reader in the private section of your code to avoid using @.

PR #478, opened 2016-08-28
Almost there. (Twilio tests work in IRB but not yet in RSpec)

nephast:


lib/menu.rb
```diff
@@ -0,0 +1,18 @@
...
+attr_reader :list
+
+
+  def initialize
+    @list = { wings: 12,
+              chicken: 9,
+              special: 15}
+  end
+
+  def print_menu
```
danldb: YAGNI - no presentation layer was asked for here, and this prsumes that a UI would want a string in this format.

lib/order.rb
```diff
@@ -0,0 +1,44 @@
+require "./lib/menu.rb"
+class Order
+
+  attr_reader :basket, :item, :menu, :quantity
+
+  def initialize
+    @menu = Menu.new
```
danldb: USe Dependency injection here

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    @basket[@item] = quantity
+    @basket
+  end
+
+  def check_item(item)
+    fail "Sorry, that item is not on the menu" unless menu.list.key?(item)
+  end
+
+  def total_price
+  calcul_total.inject(:+)
```
danldb: Indentation! If this was a tech test, you just failed ;-)

lib/text.rb
```diff
@@ -0,0 +1,35 @@
+require 'twilio-ruby'
+
+class Text
+
+  def initialize
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token  = ENV['TWILIO_AUTH_TOKEN']
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
```
danldb: use DI here.

lib/text.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def eta
+    time = Time.now + (60*60)
+    "#{time.hour}:#{time.min}"
+  end
+
+  private
+
+  def twilio_number
```
danldb: should be a constant

spec/text_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'text'
+
+describe Text do
+
+let(:twilio_ruby) {double :twilio_ruby, new: twilio}
+  subject(:text) {described_class.new}
+
+  let(:twilio) {double :twilio, :messages => messages}
+  let(:messages) {double :messages, :create => "Thank you! Your delivery should arrive by 21:10"}
```
danldb: well done negotiating a big mess of doubles, it's not easy!

PR #481, opened 2016-08-29
Takeaways challenge

supasuma:
struggled a lot with this


.env
```diff
@@ -0,0 +1,5 @@
+SID = "AC4f30143cf5eccea4c0f4cb53848159b0"
+TOKEN = "dafdda535cb4838d43f2fd45e9988591"
+
+MY_NUMBER = "+447761719320",
```
danldb: hmm your .env file shouldn't be in github! Ooops! You should probably remove this.

lib/menu.rb
```diff
@@ -0,0 +1,27 @@
...
+    @meals_list = {
+      "nachos"=>5.00,
+      "ribs"=>6.00,
+      "chilli"=>4.00,
+      "burger"=>4.00,
+      "sundae"=>3.00
+    }
+  end
+
+  def print_menu
```
danldb: YAGNI - presentation logic in your business layer.

lib/message.rb
```diff
@@ -0,0 +1,25 @@
...
+  def delivery_time
+    time = Time.now + 60 * 60
+    time.strftime("%H:%M")
+  end
+
+  def send_text
+    account_sid = ENV['SID']
+    auth_token = ENV['TOKEN']
+
+    @client = Twilio::REST::Client.new account_sid, auth_token
```
danldb: use dependency injection for the client

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+        it 'raises an error if no order placed' do
+          allow(order).to receive(:basket_empty?).and_return(true)
+          message = 'No orders have been added'
+          expect{ takeaway.confirm_order }.to raise_error message
+        end
+
+        it 'sends a message if order is correct' do
+          expect(message).to receive(:send_text)
+          message.send_text("test")
```
danldb: doesn't this send a real SMS?

danldb:
avoid making real API calls during your test process.

PR #484, opened 2016-08-29
takeaway challenge - murtz

mrmurtz:


lib/restaurant.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def add_order(item, quantity=1)
+    @menu.dish_on_menu(item)
+    @basket[item] = quantity
+    @total += @menu.price(item)*quantity
+  end
+
+  def check_order
+    puts "Here's a summary of your order:".center(25)
```
danldb: too many responsibilities - is it checking the order, or compiling a message for some user?

lib/restaurant.rb
```diff
@@ -0,0 +1,48 @@
...
+    basket.each do |dish, quantity|
+      puts "#{quantity}x #{dish.ljust(10)} (£#{@menu.price(dish)*quantity})"
+    end
+    puts "Total: £#{@total}"
+    end
+
+  def complete_order
+    time = Time.new
+    add_hour = time + 1*60*60
+    delivery_time = add_hour.strftime("%H:%M")
```
danldb: looks like another private method here to calculate the delivery time.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+
+    describe "total" do
+    it "adds the price of the dish added to the order (and calculates quantity) " do
+      restaurant.add_order("Burger", 2)
+      expect(restaurant.total).to eq 5
+    end
+    end
+
+    # describe "complete_order" do
```
danldb: don't leave commented code in your commits

PR #486, opened 2016-08-29
My terrible work

lawrenced87:
I don't know why but I went down the complete wrong track with this. Its unfinished due to spending a long time trying to double user input, and therefore ruining all my methods and making them impossible to double, and most importantly not seeing the direct comparisons between this challenge and the oystercard challenge.


danldb:
OK - you can revisit this at a later stage.

PR #487, opened 2016-08-30
first pull request

Matty79:


lib/takeaway.rb
```diff
@@ -0,0 +1,61 @@
+require_relative 'menu'
+require_relative 'text'
+
+class Takeaway
+
+attr_reader :menu, :basket, :subtotal
+
+  def initialize(menu = Menu.new, text_class = Text)
```
danldb: nice DI

lib/takeaway.rb
```diff
@@ -0,0 +1,61 @@
...
+    total = 0
+    basket.each do |item, quantity|
+      total += (quantity * (menu.check_price(item)))
+    end
+    total.round(2)
+  end
+
+  def complete_order
+    confirm_bill
+    @text_class.new.send_message
```
danldb: Probably would have been easier to inject Text.new at the top instead.

PR #488, opened 2016-08-30
Takeaway challenge (week 2 weekend challenge)

BenRoss92:
Unfortunately only managed to write 1 test and pass it - was an usher at my brother's wedding; all time (apart from Monday evening) taken up helping prepare everything for wedding and meeting/doing activities with bride's family. Will try to do more of challenge this week.


danldb:
Good start at least - your build is passing!

PR #493, opened 2016-08-30
:(

tam-borine:


lib/dispatch.rb
```diff
@@ -0,0 +1,24 @@
...
+class Dispatch
+Dotenv.load
+
+  def send_sms
+
+  id = ENV["AUTH_ID"]
+  pw = ENV["AUTH_TOKEN"]
+  customers = { ENV["NUMBER1"] => "insert_name_here"}
+
+  @client = Twilio::REST::Client.new id, pw
```
danldb: Would have been better to use dependency injection here, to no couple your Dispatch class to the Twilio client

lib/price_calc.rb
```diff
@@ -0,0 +1,30 @@
...
+    total
+  end
+
+  def receipt
+    p "The total charge was #{calculate}"
+  end
+
+  def price_expectations_checker
+    diff = @expected_price - calculate
+    p "Your expected price was different from our calculate price by £#{diff}"
```
danldb: avoid presentation in your business logic

lib/restaraunt.rb
```diff
@@ -0,0 +1,21 @@
+#understands customer expectations
+require_relative 'menu'
+require_relative 'order'
+
+class Restaraunt
+  include Menu
```
danldb: dependency injection would have reduced the dependency here

spec/dispatch_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'dispatch'
+
+describe Dispatch do
+
```
danldb: no tests? :(

danldb:
Not bad, but too much coupling and not enough tests.

PR #497, opened 2016-10-07
Glenn B's submission for takeaway-challenge

glynester:
This is my attempt at the takeaway-challenge


lib/menu.rb
```diff
@@ -0,0 +1,54 @@
+require 'csv'
+
+
+class Menu
+  
+  attr_reader :menu
```
silvabox: Don't expose the internal hash of objects.  And don't call it `menu` as it's now unclear what a menu is.  Is it an internal hash of items, or is it the object that contains an internal hash of items?

lib/menu.rb
```diff
@@ -0,0 +1,54 @@
...
+      !!menu[item]
+    end
+  
+    def get_details(item)           
+      menu.select { |k, v| k[item] }    
+    end
+  
+  private
+  
+  def load_items(item,price)
```
silvabox: `load_item`

lib/messager.rb
```diff
@@ -0,0 +1,23 @@
...
+    @client.account.messages.create({
+      :from => '+441143032569', 
+      :to => '+447832366992', 
+      :body => txt_message,
+    })
+  end  
+  
+  private
+  def config
+    account_sid = 'ACa742df3682aa80d6d524a052cf5b0013'            # My credentials are saved as environment variables.
```
silvabox: You should change your API keys as these ones are now publicly accessible on GitHub

lib/restaurant.rb
```diff
@@ -0,0 +1,19 @@
+require_relative 'takeaway'
+
+class Restaurant < TakeAway
```
silvabox: This class is superfluous as it adds no additional behaviour.  Just use `Takeaway`

lib/takeaway.rb
```diff
@@ -0,0 +1,92 @@
+require_relative 'messager'
+require_relative 'menu'
+
+class TakeAway
+  
+  attr_reader :read_menu, :menu, :orders, :messager
```
silvabox: `read_menu` is redundant here as you override it further down the file.

lib/takeaway.rb
```diff
@@ -0,0 +1,92 @@
...
+
+class TakeAway
+  
+  attr_reader :read_menu, :menu, :orders, :messager      
+  
+  def initialize(menu = Menu.new, messager = Messager.new)
+    @menu_obj = menu
+    @messager = messager
+    @orders = []
+    @menu =  @menu_obj.menu
```
silvabox: The class shouldn't have access to `@menu_obj.menu` here.  The `menu` attribute now points to a Hash, but I would expect it to point the the `Menu` object I passed in if I'd overridden the default.

lib/takeaway.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+  
+  def change_menu(file)
+    @menu_obj = Menu.new(file)
```
silvabox: what about `@menu`?

lib/takeaway.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def read_menu
+    @menu_obj.read_menu
+  end
+  
+  def order(item, qty=1)
+    print :"That item is not on the menu.\n" unless @menu_obj.contains?(item) 
+    (qty.times {add_item(item)}) if @menu_obj.contains?(item)
```
silvabox: Why keep repeating the conditional `if @menu_obj.contains?(item)`.  You've already tested it once in the method above, so just prevent the method executing any further

lib/takeaway.rb
```diff
@@ -0,0 +1,92 @@
...
+      counts.each{|key,val| 
+      	key.each{|k,v|
+      	output_arr << "\"#{k}\" x #{val} = £#{(key[k] * val).round(2)}"
+      	}
+      }	
+      print output_arr.join(", ")
+  end
+  
+  def total
+    @tot = 0
```
silvabox: this should be a local variable, not an instance variable.

glynester:
Thanks for reviewing Ben

Regarding the comment "However, your objects are all just dumping everything to stdout. They should be communicating meaningful values to one another and not be concerned with user interface.", I'm not sure how to handle this. If I want messages to appear on screen when a user does something, how do I do this in accordance with what you have said? Can you point me to a link or someone's code who you know has done this properly.

PR #498, opened 2016-10-07
Initial pull request

darrenf54321:


lib/menu.rb
```diff
@@ -0,0 +1,16 @@
+require_relative 'takeaway'
+#require_relative 'order'
+
+module Menu
+
+  MENU = {"Chicken" => 5.00, "Madras" => 6.00, "Nan Bread" => 2.50, "Cobra" => 4.00}
+
+  def show_menu
```
silvabox: This method doesn't really do anything

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
...
+    MENU.each { |item, price| "#{item}, #{price}"}
+  end
+
+  def hash_menu
```
silvabox: The responsibility of `Menu` should be to provide an interface for listing dishes and prices without exposing its internal implementation (i.e. it should not return the internal hash)

lib/order.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'menu'
+
+module Order
```
silvabox: Again, this should be a class, not a module.  One way to determine this is to ask, 'Does an order get created?'.  Clearly it does, so it must be a class.

lib/order.rb
```diff
@@ -0,0 +1,17 @@
...
+
+include Menu
```
silvabox: An `Order` is not a `Menu`.  It might _collaborate_ with a menu though.

lib/order.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def total_price
+    menu_prices
+    @overall_price = @unit_prices.inject(0) {|sum, i|  sum + i }
+  end
+
+  def menu_prices
+    @unit_prices = []
```
silvabox: you do not need `@unit_prices` to be an instance variable.  Avoid unnecessary state in your objects - just return the value from the method and let the calling code decide what to do with it.

lib/order.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'menu'
+
+module Order
+
+include Menu
+
+  def total_price
+    menu_prices
+    @overall_price = @unit_prices.inject(0) {|sum, i|  sum + i }
```
silvabox: `@overall_price` is not needed here.  The following is equivalent:
```ruby
def total_price
  menu_prices.inject(0) {|sum, i|  sum + i }
end
```

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+require_relative 'order'
+require_relative 'text'
+require_relative 'menu'
+
+require 'rubygems'
+require 'twilio-ruby'
+
+class Takeaway
+
+include Text, Order, Menu
```
silvabox: These should be objects that `Takeaway` collaborates with, not objects it inherits behaviour from.

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def initialize
+    @menu_choice = {}
+    @quantity_total = []
+    @total_count = 0
+  end
+
+  def select_item
+    while @item != "finish" do
+      puts "Please type in your order or 'finish' to complete your order."
```
silvabox: don't build a command-line interface into your objects - that is a separate responsibility.   Your `Takeaway` object should be responsible for receiving orders and confirming them.  Nothing else.

PR #499, opened 2016-10-07
First commit

B-raw:


lib/order.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def confirm_order
+    @messenger.send(confirmation)
+  end
+
+private
+
+  def total_bill
+    @bill = 0
```
silvabox: You don't need an instance variable here. `@bill` is never called outside of this method and does not need to be remembered

lib/list_module.rb
```diff
@@ -0,0 +1,7 @@
+module List
```
silvabox: `List` is too generic a name for this module.  It's often difficult to find a satisfactory name for this kind of thing but the one method it provides outputs a formatted list of dishes.

lib/list_module.rb
```diff
@@ -0,0 +1,7 @@
...
+  def list_dishes(data)
+    data.each_with_index do |dish, i|
+      puts "#{i+1}: #{dish[:name]} costs #{dish[:price]}"
```
silvabox: `puts` should only be used for debugging.  Construct and return a string here instead.  If some other piece of code want to `puts` that string, it can.

PR #500, opened 2016-10-07
Takeaway challenge

groundberry:


lib/fake_twilio.rb
```diff
@@ -0,0 +1,20 @@
+class FakeTwilio
```
maryrosecook: Love that you mocked Twilio to avoid sending texts! It's definitely worthwhile using irb or a throwaway Ruby file to figure out how this code works for your own learning.

lib/restaurant.rb
```diff
@@ -0,0 +1,39 @@
+require_relative 'order'
+require_relative 'menu'
+
+class Restaurant
+  attr_reader :order
+
+  def initialize(menu, messenger = nil)
+    @menu = menu
+    @menu.read_menu
```
maryrosecook: Maybe have `Menu` call `read_menu` on itself, rather than relying on `Restaurant` to do it.  This will improve the encapsulation of `Menu`

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+    @items = Hash.new(0)
+    @menu = menu
+  end
+
+  def add_to_order(dish, quantity = 1)
+    @items[dish] += quantity
+  end
+
+  def order_total
+    total = 0
```
maryrosecook: Can you do this with `inject`? It'll be a bit shorter and clearer.

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+class Order
+  attr_reader :items
+
+  def initialize(menu)
+    @items = Hash.new(0)
+    @menu = menu
+  end
+
+  def add_to_order(dish, quantity = 1)
+    @items[dish] += quantity
```
maryrosecook: Nice! Love the simplicity of this.

lib/restaurant.rb
```diff
@@ -0,0 +1,39 @@
...
+    @messenger = messenger
+  end
+
+  def see_menu
+    # TODO: how do I test this?
+    puts @menu
+  end
+
+  def add_to_order(dish, quantity = 1)
+    if @order.nil?
```
maryrosecook: Could `@order` be created in `initialize` to avoid having to write this extra code?

lib/restaurant.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def initialize(menu, messenger = nil)
+    @menu = menu
+    @menu.read_menu
+    @messenger = messenger
+  end
+
+  def see_menu
+    # TODO: how do I test this?
+    puts @menu
```
maryrosecook: Try googling replacing STDOUT with StringIO.

PR #501, opened 2016-10-07
Initial pull request. Gemfile updated + some stuff done

crispinandrews:


lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+               'Beer': 3.00}
+  end
+
+  def show
+    @dishes.each do |food, price|
+      puts "#{food}: " + ("£%.2f" % [price])
+    end
+  end
+
+  def do_we_have(item)
```
silvabox: `has_dish?` would be a better name for this method

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
...
+    @order = []
+  end
+
+  def menu_list
+    @menu.show
+  end
+
+#Move methods to order class
+  def add(dish, amount)
+    fail 'Thats not on the menu!' if menu.do_we_have(dish) == false
```
silvabox: Consider how this would read as:
```ruby
fail 'Thats not on the menu!' unless menu.has_dish?(dish)
```

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
...
+    @order << {item: dish, qty: amount, total: (amount * menu.price(dish))}
+  end
+
+  def add_dish(dish, amount)
```
silvabox: Why `add` _and_ `add_dish`?

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'menu'
+
+describe Menu do
+
+  let(:takeaway) { double :takeaway }
```
silvabox: This is unused - it should be deleted

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:menu) { double(:menu, show: dishes_list) }
+  let(:dishes_list) { "item: £2.00" }
+  let(:@dishes) do
```
silvabox: This is unused and should be deleted

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'menu'
+
+describe Menu do
+
+  let(:takeaway) { double :takeaway }
+  let(:menu) { double(:menu, show: dishes_list) }
```
silvabox: Oops - you're supposed to be testing the `Menu` class.  Now you're just testing a double.

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  it {is_expected.to respond_to(:do_we_have).with(1).argument}
+
+  it 'has a menu list' do
+    expect(menu.show).to eq(dishes_list)
+  end
+
+  it 'should know what is on the menu' do
+    allow(menu).to receive(:do_we_have).and_return(true)
+    expect(menu.do_we_have('item')).to eq(true)
```
silvabox: Even if `menu` was an instance of `Menu` and not a double, this is still a pointless test as you've mocked `do_we_have` on the previous line so it will always return true.

PR #502, opened 2016-10-08
Initial pull request

chrmsan:


lib/order.rb
```diff
@@ -0,0 +1,12 @@
+class Order
```
silvabox: This object doesn't do anything at the moment

lib/takeaway_shop.rb
```diff
@@ -0,0 +1,41 @@
...
+
+private
+
+	def index_items
+		@menu_items.map.with_index(1) do |menu_item, i|
+			@menu << "#{i}		#{menu_item[:item]}		#{menu_item[:price]}			#{menu_item[:description]}\n"
+		end
+	end
+
+	def print_menu
```
silvabox: You have 3 methods here that all include `menu` in the name.  Isn't that a clue that they belong in `Menu`?

spec/order_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+describe Order do
+	subject(:order) {described_class.new}
+	let(:item_no) {double(:item_no)}
+	let(:quantity) {double(:quantity)}
+
+	it 'should store a numbered order and its quantity and store them as a in  an array' do
+		order.request
+		expect(order.order_array).to match_array([{item_no: item_no, quantity: quantity}])
```
silvabox: How does anything ever get into `order_array`?

PR #503, opened 2016-10-08
Take a way..

tadasmajeris:


lib/takeaway.rb
```diff
@@ -0,0 +1,79 @@
...
+    else
+      fail "File '#{filename}' not found!"
+    end
+  end
+
+  def menu
+    puts dish_list
+  end
+
+  def add_dish(id, qty=1)
```
silvabox: This responsibility belongs in `Order`.  What else is `Order` for if not validating dishes and their quantities?

lib/takeaway.rb
```diff
@@ -0,0 +1,79 @@
...
+    fail_if_wrong_qty(qty)
+    dish = dish_list.select_dish(id)
+    order.remove_item(dish, qty)
+  end
+
+  def order_total
+    order.total
+  end
+
+  def place_order
```
silvabox: The name of this method suggests that I should pass the order as a parameter (and indeed you should).  The taking of the order is handled by `Order` and `Menu`.  The takeaway should process the finalized order.

lib/takeaway.rb
```diff
@@ -0,0 +1,79 @@
...
+    confirmation_text
+  end
+
+  def placed_orders
+    @placed_orders.dup
+  end
+
+  def confirmation_text
+    h = Time.now.hour
+    delivery_time = "#{h==23 ? 0 : h+1}:#{Time.now.min}"
```
silvabox: `Time` already provides this functionality.  `Time.now + 3600` will add 3600 seconds (i.e. 1 hour)

PR #505, opened 2016-10-08
Malinna's Tasty Takeaway

MalinnaLeach:


lib/interface.rb
```diff
@@ -0,0 +1,96 @@
+class Interface
+
+  attr_reader :menu_array, :order
+
+  def initialize(filepath)
+    @menu = Menu.new(filepath)
+    @order = Order.new
+    @order_array = []
```
maryrosecook: It's idiomatic in Ruby to not describe the type of a variable unless it's absolutely necessary for clarity. `@orders` would be clear enough.

lib/interface.rb
```diff
@@ -0,0 +1,96 @@
...
+    @menu.build_menu_display
+  end
+
+  def user_interaction
```
maryrosecook: Maybe call this `listen_for_user_input`?

lib/interface.rb
```diff
@@ -0,0 +1,96 @@
+class Interface
+
+  attr_reader :menu_array, :order
+
+  def initialize(filepath)
+    @menu = Menu.new(filepath)
```
maryrosecook: Can you inject the menu instance so that Interface doesn't have to worry about `filepath`, which is a var it doesn't care about?

lib/interface.rb
```diff
@@ -0,0 +1,96 @@
...
+    @order = Order.new
+    @order_array = []
+    @menu.build_menu_display
```
maryrosecook: Could this method be called by Menu.initialize to increase Menu's encapsulation?

lib/interface.rb
```diff
@@ -0,0 +1,96 @@
...
+    order_confirm
+  end
+
+  def welcome
+    file_reader('/Users/malinnaleach/Programs/takeaway-challenge/lib/welcome.txt')
+    display(@menu.menu_array)
+  end
+
+  def file_reader(filepath)
+    File.open(filepath) {|f| puts f.read}
```
maryrosecook: Have a look at doing this without a block, to make this line more succinct.  Also, try to make function names verbs.

lib/interface.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  def display(array)
+    array.each {|string| puts string}
+  end
+
+  def build_order(menu, order)
```
maryrosecook: Could this method be refactored to use private helper methods to make it shorter and clearer?

lib/interface.rb
```diff
@@ -0,0 +1,96 @@
...
+        send_text
+        break
+      elsif input == "end"
+        puts "No worries, please come again.\n"
+        exit
+      end
+    end
+  end
+
+  def send_text
```
maryrosecook: Could some methods be made private to increase encapsulation?

PR #506, opened 2016-10-08
Takeaway challenge

fmlharrison:


lib/message.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  attr_reader :to, :from
+
+  def initialize(to = '+447540139449', from = '+441202286041')
+    @to = to
+    @from = from
+    @client = Twilio::REST::Client.new ACCT_SID, AUTH_TOKEN
+  end
+
+  def send_message
```
maryrosecook: You can use the dependency injection technique with other people's code as well as your own.  Imagine one of your classes, A, depends on another of your classes, B.  Imagine you're testing A and to do that you need to stub B.  You can write A so that it gets B injected.  This will let you stub B.  You can use the same technique when B is not one of your classes (B) but someone else's class (Twilio).  Does that make sense?

PR #512, opened 2016-10-09
Initial commit for menu class and rspec test

james1968:


lib/make_call.rb
```diff
@@ -0,0 +1,19 @@
+require 'twilio-ruby'
+require_relative 'order'
+class Checkout
+
+  def initialize
+    account_sid = 'ACdb0c001d98d842c9465935ffde45fb6b'
+    auth_token = 'aa648095dec33b70637f15de4c6c9964'
```
silvabox: You should never commit auth tokens to a public repository.  You ought to change your token on Twilio for safety.

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+class Menu
+
+attr_reader :print_menu, :todays_menu
+
+    def initialize
```
silvabox: Your indentation has gone off here.  Ensure your text editor is configured to convert tabs to 2 spaces.

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+      @todays_menu = []
+    end
+
+    def todays_menu
+      @todays_menu = [{:pizza => :Margherite, :price => 1.25},
+                      {:pizza => :Pepperoni, :price => 1.75},
+                      {:pizza => :Spicy, :price => 2.75}]
+                   end
+
+    def print_menu
```
silvabox: This overrides the `attr_reader :print_menu`, which is not needed; neither is `@print_menu`

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+class Menu
+
+attr_reader :print_menu, :todays_menu
+
+    def initialize
+      @print_menu = []
+      @todays_menu = []
+    end
+
+    def todays_menu
```
silvabox: This overrides the `attr_reader :todays_menu`, which is therefore not required.  Note though that this method exposes the internal array to the public, which breaks encapsulation.

lib/order.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'menu'
+#require_relative 'make_call'
+
+
+class Order
+
+attr_reader :order, :phone_number, :order, :cust_order
+
+    def initialize
+        @menu = Menu.new
```
silvabox: The menu could be passed in as a parameter to remove this dependency

lib/order.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'menu'
+#require_relative 'make_call'
+
+
+class Order
+
+attr_reader :order, :phone_number, :order, :cust_order
```
silvabox: You've defined `:order` twice here and it isn't used anywhere.  Nevertheless, it would lead to `order.order` which doesn't make sense.  A method with the same name as the class is an indication of a design problem.

lib/order.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    def view_menu
+        puts "Todays menu is as follows..."
+        @menu.todays_menu
+        puts "#{@menu.print_menu}"
+    end
+
+    def place_order
+          puts "... please select your pizza and press enter"
```
silvabox: User interface logic should not be in this class - hopefully now you've been introduced to Separation of Concerns, you can see why.  The `Order` class should only be responsible for representing the selected dishes and their quantities (and probably the total).

lib/make_call.rb
```diff
@@ -0,0 +1,19 @@
...
+    @client = Twilio::REST::Client.new account_sid, auth_token
+    @order = Order.new
```
silvabox: This does not belong here - the Checkout is not responsible for creating an order.

lib/make_call.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def confirm_order
```
silvabox: pass the order object as a parameter here so that you can query the phone number.

PR #513, opened 2016-10-09
Initial pull request.

hanrattyjen:


lib/confirmation_text_sender.rb
```diff
@@ -0,0 +1,24 @@
...
+require 'twilio-ruby'
+require 'time'
+
+class ConfirmationTextSender
+
+  def send_message(message)
+
+    account_sid = 'AC9418dc2fc0bb5cbb33e505c227d3a3c1'
+    auth_token = '5490266efdf7640c6e1d59a7da563967'
+    @client = Twilio::REST::Client.new account_sid, auth_token
```
silvabox: you're reassigning `@client` on every call to this method, so it shouldn't be an instance variable.  A better alternative would be to lazy instantiate this in a separate method.

lib/confirmation_text_sender.rb
```diff
@@ -0,0 +1,24 @@
+require 'rubygems'
+require 'twilio-ruby'
+require 'time'
+
+class ConfirmationTextSender
+
+  def send_message(message)
+
+    account_sid = 'AC9418dc2fc0bb5cbb33e505c227d3a3c1'
+    auth_token = '5490266efdf7640c6e1d59a7da563967'
```
silvabox: Don't commit auth_tokens and secrets to public repos.  You ought to regenerate this in Twilio to be on the safe side.

lib/confirmation_text_sender.rb
```diff
@@ -0,0 +1,24 @@
...
+    @message = @client.messages.create(
+      to: "+447984599636",
+      from: "+441173253687",
+      body: "Hi from FOOD! Your order has cost £#{message} and will arrive at #{calculate_time}."
+    )
+  end
+
+  def calculate_time
+    delivery_time = Time.now + 30*60
+    format_time = delivery_time.strftime("%I:%M%p")
```
silvabox: the local variable assignment here is redundant.

lib/menu.rb
```diff
@@ -0,0 +1,11 @@
+require "csv"
+
+class Menu
+
+  attr_reader :menu
```
silvabox: This class is the menu, so it shouldn't have an attribute called `menu` as that leads us to `menu.menu`.  This class should provide methods to query the menu data, not just expose the array.

lib/customer.rb
```diff
@@ -0,0 +1,34 @@
...
+require 'csv'
+
+class Customer
+
+  attr_reader :menu
+  attr_accessor :order_created, :pending_order, :selection
+
+  def initialize()
+    @pending_order = []
+    @menu = Menu.new
```
silvabox: The menu should be passed in as a parameter

lib/customer.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'order'
+require_relative 'menu'
+require_relative 'confirmation_text_sender'
+require 'csv'
+
+class Customer
+
+  attr_reader :menu
+  attr_accessor :order_created, :pending_order, :selection
```
silvabox: The `selection` attribute does not appear to be used anywhere

lib/order.rb
```diff
@@ -0,0 +1,34 @@
+require 'confirmation_text_sender'
+
+class Order
+
+  attr_reader :order, :total
+
+  def initialize(order, text_confirm_order = ConfirmationTextSender.new)
```
silvabox: I'm confused - what is the order?  I seem to have to initialize and instance of `Order` by passing it an `order` parameter.

lib/customer.rb
```diff
@@ -0,0 +1,34 @@
...
+    @selection = selection
+  end
+
+  def select_dish
+    puts "Please enter a menu choice:"
+    selection = gets.chomp.to_i
+    add_to_order(selection)
+  end
+
+  def confirm_order
```
silvabox: This does not feel like the customer's responsibility

lib/order.rb
```diff
@@ -0,0 +1,34 @@
...
+    @order = order
+    @total = 0
+    @text_confirm_order = text_confirm_order
+  end
+
+  def confirm_order
```
silvabox: You have three methods here that all have names starting `confirm_` but they have very different behaviour.  This makes the interface confusing.

lib/customer.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_accessor :order_created, :pending_order, :selection
+
+  def initialize()
+    @pending_order = []
+    @menu = Menu.new
+    @order_created = nil
+    @selection = selection
+  end
+
+  def select_dish
```
silvabox: The `Customer` class should not be responsible for the user interface.  The UI can be added as a layer later on.

PR #514, opened 2016-10-09
Takeaway Challenge 

lcbeh:


lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
+  attr_reader :list
+
+  def initialize
+    @list = VEGETARIAN_MENU
```
maryrosecook: Does `@list` have to exist? Could print_menu just use `VEGETARIAN_MENU`?

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+    list.each do | dish, price |
+      dish_list = (dish.to_s).ljust(25)
+      price_list = "£#{price.to_s}".rjust(25)
+      puts dish_list + price_list
+    end
+  end
+
+private
+
+  VEGETARIAN_MENU = { Thai_green_curry: 9, Aubergine_teriyaki: 9, Mushroom_risotto: 8, Butternut_rotolo: 9, Thali: 11, Smoky_bean_burger: 7 }
```
maryrosecook: Nice! I like that this is private and a constant.  Nice encapsulation.

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize
+    @list = VEGETARIAN_MENU
+  end
+
+  def print_menu
+    list.each do | dish, price |
+      dish_list = (dish.to_s).ljust(25)
+      price_list = "£#{price.to_s}".rjust(25)
+      puts dish_list + price_list
```
maryrosecook: It's more idiomatic to use string interpolation `"#{}"` than concatenation `a + b`.

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_reader :list
+
+  def initialize
+    @list = VEGETARIAN_MENU
+  end
+
+  def print_menu
+    list.each do | dish, price |
+      dish_list = (dish.to_s).ljust(25)
+      price_list = "£#{price.to_s}".rjust(25)
```
maryrosecook: Could you make an extra method that can generate a single line of the menu to make this code clearer?

lib/order.rb
```diff
@@ -0,0 +1,39 @@
+require_relative 'menu'
+
+class Order
+  attr_reader :sum, :basket, :menu
+
+  def initialize(order_menu)
+    @menu = order_menu.list
```
maryrosecook: It feels like it's breaking encapsulation a bit to have order reach into `order_menu` to grab `list`.  Could `Menu` have some methods on it that examine `list` to answer questions like "is this item on the menu?"

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+    @basket = Hash.new(0)
+    @sum = 0
+  end
+
+  def add(dish, quantity)
+    if @menu.has_key?(dish.to_sym)
```
maryrosecook: See comment above about adding methods to `Menu` to deal with these types of requests.

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+      basket[dish.to_sym] += quantity
+      "#{quantity} x #{dish} added to basket"
+    else
+      "Sorry, #{dish} is not on our menu"
+    end
+  end
+
+  def print_order
+    basket.each do | dish, quantity |
```
maryrosecook: Maybe break `print_order` into more functions to make it more readable?

lib/order.rb
```diff
@@ -0,0 +1,39 @@
+require_relative 'menu'
+
+class Order
+  attr_reader :sum, :basket, :menu
+
+  def initialize(order_menu)
+    @menu = order_menu.list
+    @basket = Hash.new(0)
+    @sum = 0
```
maryrosecook: Can this be calculated every time it's asked for, rather than stored on the order object?

lib/takeaway.rb
```diff
@@ -0,0 +1,43 @@
...
+
+class TakeAway
+  attr_reader :menu, :order, :text
+
+  def initialize (menu = Menu.new, order = Order.new(menu), text = Text.new)
+    @menu = menu
+    @order = order
+    @text = text
+  end
+
```
maryrosecook: I love the delegation in view_menu, `add` and `review`.  There is no logic in them and the message is completely delegated to the other classes.

lib/text.rb
```diff
@@ -0,0 +1,25 @@
+require 'sinatra'
+require 'twilio-ruby'
+
+class Text
+  attr_reader :client
+
+  def initialize
+    @client = Twilio::REST::Client.new ACCOUNT_SID, AUTH_TOKEN
```
maryrosecook: Could you inject the Twilio dependency to allow you to stub it in the tests?

PR #515, opened 2016-10-09
Weekend2_challenge

Juli0GT:


lib/takeaway.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'menu'
+require_relative 'order'
+require_relative 'sms'
+
+class Takeaway
+  attr_reader :menu
+
+  def initialize
+    @new_order
```
maryrosecook: It feels like this line (and therefore this initialize method) isn't necessary.

lib/order.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'menu'
+class Order
+  attr_reader :order, :total_prize
+
+  def initialize
+    @menu = Menu.new
+    @order = []
```
maryrosecook: Could this be renamed to `items` to make it a little clearer what it contains?

lib/takeaway.rb
```diff
@@ -0,0 +1,46 @@
...
+    @menu.dishes_list
+  end
+
+  def start_order
+    @new_order = Order.new
+    @new_order.order
+  end
+
+  def add_chicken
+    @new_order = Order.new if @new_order == nil
```
maryrosecook: It feels like having to check if `@new_order` has been created on `add_chicken`, `add_lamb` and `add_beef` is repetitive.  Could you refactor this so that you only have to write that line once? Perhaps by moving that code to another function? Or even by making sure `@new_order` is always created when Takeaway is instantiated?

lib/order.rb
```diff
@@ -0,0 +1,31 @@
...
+    @total_prize = 0
+  end
+
+  def add_chicken
```
maryrosecook: It feels like `add_chicken`, `add_lamb` and `add_beef` could be combined into one method? Can you look at the differences between each one and abstract that difference? Happy to talk about this in person if you want more guidance!

lib/order.rb
```diff
@@ -0,0 +1,31 @@
...
+    @total_prize += @menu.dishes_list[1][:prize]
+  end
+
+  def add_beef
+    @order << @menu.dishes_list[2]
+    @total_prize += @menu.dishes_list[2][:prize]
+  end
+
+  def check_order
+    @order << {total: "£#{@total_prize}"}
```
maryrosecook: Could `total_prize` be calculated on the fly?  So remove the `@total_prize` attribute that is updated when new things are added to the order.  Replace it with a `total_prize` method that goes through the items in `@order` and counts up the total price?

PR #522, opened 2016-10-09
Takeaway challenge

wirsindpapst:
A mess


lib/menu.rb
```diff
@@ -0,0 +1,33 @@
+class Menu
+
+  require 'ap'
```
silvabox: your requires should be at the top of the file really, not inside the class definition

PR #523, opened 2016-10-09
Initial pull request

rgollancz:


lib/basket.rb
```diff
@@ -0,0 +1,23 @@
+class Basket
+  attr_reader :summary
+  attr_accessor :items
+
+  def initialize
+    @items = []
+  end
+
+  def summary(items = unique_items)
```
maryrosecook: Really nice job with making the code provide a sophisticated view of the basket (items shown with their quantities).  To make methods like this easier to write, try to break them up into sub methods.  One way to do this is to make a list of the things the method needs to do (get a unique list of dishes, get the number of each dish ordered, turn the dish name and number to a nicely formatted string) etc and then write methods that do each thing.  You've started this nicely with `unique_items` - just carry on down this path.

lib/basket.rb
```diff
@@ -0,0 +1,23 @@
...
+    summary = []
+    count = 0
+    items.each do |item|
+      @items.each { |dish| count += 1 if dish == item }
+      summary.push("#{item.keys[0]} x#{count} = £#{item[item.keys[0]] * count}")
+    end
+    @summary = summary
```
maryrosecook: No need to store `@summary`.  You can make `summary` return it.  Any code that needs the summary can call `summary`.

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    @current_order = Order.new
+  end
+
+  def view_menu
+    @menu.dishes
+  end
+
+  def add_to_order(item,quantity)
+    order = current_order
+    dish = @menu.dishes.select { |k,v| k == item }
```
maryrosecook: This feels like it's breaking the encapsulation of `@menu` a bit because it's reaching in to get `dishes` and then making assumptions about what `dishes` is and how it should be used.  Try moving any code that works on `dishes` inside `Menu`.  eg you could refactor this line to call something like `@menu.dish(item)`.

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    order.add_dish(dish,quantity)
+  end
+
+  def view_order
+    @current_order.view_basket
+  end
+
+  def view_cost
+    @current_order.view_cost
```
maryrosecook: Love this delegation!

PR #524, opened 2016-10-09
takeaway challenge

antonyCastineiras:


lib/interface.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'order'
+require_relative 'menu'
+require_relative 'dish'
+require 'twilio-ruby'
+require 'envyable'
+Envyable.load('./config/env.yml','development')
+
+class Interface
+
+	attr_reader :menu_class, :menu, :order_class
```
silvabox: Why make `menu_class` and `order_class` public attributes?

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+class Menu
+
+	DISHES = [
+		{:name => "Pizza", :price => 9.99},
+		{:name => "Kebab", :price => 4.99},
+		{:name => "Chicken Wings", :price => 3.49},
+		{:name => "Chips", :price => 2.99},
+		{:name => "Ice-cream", :price => 4.99}]
+
+	attr_reader :dishes, :dish_class
```
silvabox: why make `dish_class` an attribute?

You shouldn't expose `dishes` here either - it breaks encapsulation.  The `Menu` class should provide its own interface to query the dishes.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
+require_relative 'dish'
+
+class Order
+
+	attr_reader :current_order
```
silvabox: It doesn't make sense for and `order` object to have a `current_order` attribute.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+
+	def initialize
+		@current_order = []
+	end
+
+	def update_order(item)
```
silvabox: This should be called `add_dish`

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+		@current_order << item
+	end
+
+	def confirm_order
+		total_price = @current_order.inject(0) {|sum,item| sum += item.price }.round(2)
```
silvabox: This should be in a separate method.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+		puts "Is the following order correct? y/n"
+		@current_order.each { |item| puts "#{item.name} #{item.price}" }
+		puts "£#{total_price}"
+		input = STDIN.gets.chomp
```
silvabox: There shouldn't be any command line interactions in this object - that's a separate concern.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+		input.downcase == "y" ? send_text : input.downcase == "n" ? puts("Sorry. Please review your order.") : puts("please enter y or n")
+	end
+
+	def send_text
```
silvabox: The order shouldn't be responsible for sending text messages.  Do you have a missing object?

spec/dish_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'dish'
+
+describe Dish do
+
+	subject { described_class.new("Pizza",9.99) }
+	
+	describe '#initialize' do
+
+		it 'has a name' do
+			expect(subject.name).to be_a(String)
```
silvabox: What, any string?  You should be testing that the name is 'Pizza'

spec/dish_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+		end
+
+		it 'has a price' do
+			expect(subject.price).to be_a(Float)
```
silvabox: As above.  Furthermore, in general you should test behaviour and not types; i.e. test what an object _does_ and not what an object _is_. (duck typing)

PR #528, opened 2016-10-09
Takeaway challenge submission from Frances Maxwell

francesmx:


lib/menu.rb
```diff
@@ -0,0 +1,33 @@
+require_relative 'dish'
+
+class Menu
+
+  attr_reader :list
```
silvabox: Don't expose the internal array publicly.  Create an interface on `Menu` for the menu to be queried.

lib/order.rb
```diff
@@ -0,0 +1,70 @@
+require_relative 'menu'
+require 'twilio-ruby'
+
+class Order
+
+  attr_reader :basket, :menu
```
silvabox: Don't expose the basket array publicly.

lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @basket = []
+    @total = 0
+    @confirmed = false
+    time = (Time.now+(60*60))
+    @confirmation_text = "Thank you! Your order was placed and will be delivered before #{time.strftime("%H")}:#{time.strftime("%M")}. You ordered:"
```
silvabox: There's no need to initialize this here; just construct the whole message at the end.

lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def confirm
+    confirm_and_text
+  end
+
+  private
+
+  def check_exists(item)
+    item_found = false
+    @menu.list.each do |x|
```
silvabox: Because this line violates Demeter and iterates over an object inside `@menu`, that's a strong indication that this functionality belongs in `Menu`.  Hence you might have:
```ruby
def check_exists(item_name)
  fail "Item on not on the menu" unless @menu.has_dish?(item_name)
end
```

lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+      if x.name == item
+        item_found = true
+      end
+    end
+    fail "Item on not on the menu" unless item_found
+  end
+
+  def retrieve_price(item)
+    item_to_find = @menu.list.select { |y| y.name == item }
+    @item_cost = item_to_find[0].price
```
silvabox: It's dangerous to introduce arbitrary state into your objects.  `@item_cost` does not make sense as an instance variable of `Order`.  Instead, return relevant methods from your methods and allow the caller to decide what to do with the result.

lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+    @total = 0
+    @confirmed = false
+    time = (Time.now+(60*60))
+    @confirmation_text = "Thank you! Your order was placed and will be delivered before #{time.strftime("%H")}:#{time.strftime("%M")}. You ordered:"
+  end
+
+  def add_item(item, quantity = 1)
+    check_exists(item)
+    @confirmation_text += " #{quantity} portion(s) of #{item},"
+    retrieve_price(item)
```
silvabox: This and the following line are unnecessary.  Just calculate the total on the fly when it is requested.

lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+    recalculate_total(quantity)
+    basket << {item: item, quantity: quantity}
```
silvabox: You've lost precision again here.  Both `check_exists` and `retrieve_price` fetch the dish object from the menu, but then you revert back to storing only the name.  If you store a reference to the dish here, your basket has access to everything it needs - names, prices and quantities.

lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  def recalculate_total(quantity)
+    @total += @item_cost * quantity
+    @item_cost = 0
+  end
+
+  def confirm_and_text
+    account_sid = "AC9260c0d2e30d8e0ed7b71f2206744564"
+    auth_token = "1a3ecd829703c63b70836734592ce643"
```
silvabox: Do not commit auth tokens to public repos.  You should regenerate this on Twilio to be on the safe side.

spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'dish'
+
+describe Dish do
+
+  dish_1 = "Chicken Teriyaki"
```
silvabox: don't define local variables outside of `it` blocks.  use `let` and `subject`.

spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  price_1 = 6.95
+  subject = described_class.new(dish_1, price_1)
```
silvabox: Note, this is not the correct syntax for setting the subject.  it should be:
```ruby
subject { described_class.new(dish_1, price_1) }
```

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'menu'
+require 'order'
+
+describe Menu do
```
silvabox: Do not test against the return value _type_.  We are interested in what an object _does_ not what it _is_.

PR #531, opened 2016-10-09
Not everything complete...

twp88:


lib/interface.rb
```diff
@@ -0,0 +1,32 @@
+class Interface
+  attr_reader :order, :order_total
+
+  def initialize
+    @menu = Menu.new
+    @send = SendMessage.new
+    @order = {}
```
maryrosecook: It feels like order should be its own class.  "Interface" has two responsibilities: providing an interface for the user and dealing with orders.

lib/interface.rb
```diff
@@ -0,0 +1,32 @@
...
+    @order
+  end
+
+  def add_to_order(food)
+    @order[food] = @menu.menu_list[food]
+  end
+
+  def calculate_order
+    @order.each {|k,v| @order_total += v}
+    "So far you have ordered #{@order}. The total cost of this order is #{@order_total}"
```
maryrosecook: Could this method be broken down into more methods.  At the moment, it has two responsibilities.

lib/interface.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def place_order
+    @send.send_the_message
```
maryrosecook: It feels like `place_order` is not quite the right name to describe what this method does?

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+class Menu
+
+attr_reader :menu_list
+
+  def initialize
+    @menu_list = { "pizza" => 4.50,
```
maryrosecook: Nice - super clear and good data design.

lib/send_message.rb
```diff
@@ -0,0 +1,17 @@
+require 'twilio-ruby'
+class SendMessage
+
+  def send_the_message
```
maryrosecook: Could this method be broken down into more methods.  It feels like it has two responsibilities.

PR #533, opened 2016-10-09
Weekend Challenge attempt

LawrenceHunt:


lib/item.rb
```diff
@@ -0,0 +1,12 @@
+require_relative 'menu'
+
+class Item
+
+attr_reader :name, :price
```
silvabox: Indentation

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'restaurant'
+require_relative 'item'
+require 'yaml'
+
+class Menu
+
+attr_accessor :items
```
silvabox: never expose internal implementation with `attr_accessor`

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+
+def initialize(filename = 'menu.yml')
+  @items = YAML.load_file(filename)
```
silvabox: this initialization logic should be in a separate method.

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+  @items.map! { |hash| Item.new(hash) }
+end
+
+def select_by_name(name)
+  @items.select{ |item| item.name.downcase == name.downcase }.first
+end
+
+def check
```
silvabox: `check` does not really describe the intention of this interface.

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+
+attr_accessor :items
+
+def initialize(filename = 'menu.yml')
+  @items = YAML.load_file(filename)
+  @items.map! { |hash| Item.new(hash) }
+end
+
+def select_by_name(name)
+  @items.select{ |item| item.name.downcase == name.downcase }.first
```
silvabox: try `@items.find`

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+  @basket = Hash.new(0)
+  @total = 0
+end
+
+def check_menu
+  @restaurant.menu.check
+end
+
+def add_item(dish, quantity = 1)
+  chosen_dish = @restaurant.menu.select_by_name(dish)
```
silvabox: instead of reaching through `@restaurant` to get to the menu, you could initialize the object with a menu instead of a restaurant.

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+  @basket[chosen_dish] += quantity
+end
+
+def remove_item(dish, quantity = 1)
+  chosen_dish = @restaurant.menu.select_by_name(dish)
+  @basket[chosen_dish] -= quantity
+end
+
+def check_basket
```
silvabox: this is just duplicating `attr_reader :basket`

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+def checkout(total)
+  fail 'you put the wrong total in!' if total != calculate_total
+  @restaurant.confirm_order(total)
+end
+
+private
+
+def calculate_total
+    @basket.each { |item, quantity| @total += (item.price * quantity) }
+    @total
```
silvabox: my memoizing `@total`, you are introducing potential inconsistencies when new dishes are added.  Just call this method `total` and recalculate every time.

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+
+def checkout(total)
+  fail 'you put the wrong total in!' if total != calculate_total
+  @restaurant.confirm_order(total)
+end
+
+private
+
+def calculate_total
+    @basket.each { |item, quantity| @total += (item.price * quantity) }
```
silvabox: Use `reduce`

lib/restaurant.rb
```diff
@@ -0,0 +1,19 @@
...
+class Restaurant
+
+attr_reader :menu, :messager
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @messager = Messager.new
+  end
+
+  def confirm_order(total)
```
silvabox: pass the order object, rather than the total.  Then you have a richer interface

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+describe Menu do
+
+  let(:pasta) {double(:item, name: "Pasta", price: 5)}
+  let(:pizza) { double(:item, name: "Pizza", price: 7) }
+
+  subject {described_class.new()}
+
+
+  it 'can return list of items and prices' do
+    expect(subject.check).to eq subject.items
```
silvabox: This test is fairly vacuous, it just confirms that `check` and `items` return the same thing: which could be something, or could be`nil`.

PR #540, opened 2016-10-10
Work in progress. YAML files created for ease of changing later.

stejgregory:


lib/menu.rb
```diff
@@ -0,0 +1,28 @@
+require 'yaml'
+
+class Menu
+
+  def initialize
+    @chippy_menu
+    @chinese_menu
+  end
+
+  def load_chippy_menu
```
silvabox: Think about the Open/Closed principle.  What happens when I want to add a pizza menu?

PR #542, opened 2016-10-10
Take away challenge

Katy600:


lib/order.rb
```diff
@@ -0,0 +1,37 @@
+# require 'take_away_menu'
+#
+class Order
+
+  attr_reader :add_item, :list_of_items, :menu, :item
+
+  def initialize
+    @list_of_items = []
```
maryrosecook: Can this variable name be more specific about what it contains?

lib/take_away_menu.rb
```diff
@@ -0,0 +1,32 @@
+class TakeAwayMenu
+attr_reader :items, :list_dishes
+
+def items
+  @items = @items = {'pasta' => 5, 'chips' => 2, 'salad' => 4, 'pizza' => 6, 'soup' => 5}
```
maryrosecook: The double `@items =` isn't necessary :)

lib/take_away_menu.rb
```diff
@@ -0,0 +1,32 @@
...
+end
+
+    def list_dishes
+       puts "1. pasta: £5.00"
```
maryrosecook: Could this method use the `@items` hash, rather than hardcoding the dishes and prices?

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+    @menu = TakeAwayMenu.new
+  end
+
+  def add_item(item)
+    selected_items = @menu.items.assoc(item)
+    @list_of_items << Hash[*selected_items]
```
maryrosecook: As I understand it, this code gets the item name and price for the item to add and puts them as a key and value in a hash.  It then adds that hash to the `@list_of_items` array.  There's a simpler way to do this using operations like this: `hash[item]` and `hash[key] = value` and `array << hash`.  Happy to talk about this if this comment isn't enough guidance!

spec/order_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  it 'has a list of dishes' do
+      @menu.list_dishes
+      expect(@menu).to respond_to(:list_dishes)
+    end
+
+  it 'has a method that allows you to order' do
+    order.add_item(item)
+    expect(order).to respond_to :add_item
+  end
```
maryrosecook: Watch out for whitespace formatting.

spec/order_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  it 'can access the menu to take an order' do
+      order.add_item(item)
+      expect(@menu.items['pasta']).to eq 5
```
maryrosecook: Your rspec syntax is solid.  Were there any particular bits you were struggling with?

Katy600:
Thank you Mary. This was really helpful. I shall have a research into
Hash's properly to try and understand them.

maryrosecook:
No problem at all!

PR #543, opened 2016-10-10
Takeaway challenge

DavidAMAnderson:


lib/menu.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'dish'
+require 'csv'
+
+class Menu
+
+  attr_reader :dishes
```
silvabox: don't expose the dishes array publicly

lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def display
+    @dishes.each do |dish|
+    puts "#{dish.name}: £#{dish.price}"
+    end
+  end
+
+# Allows adding of menu items in bulk
+  def add_menu(filename = 'menu.csv')
```
silvabox: `menu.add_menu(filename)` doesn't seem right to me.  Maybe `menu.load(filename)` would be better.

lib/order.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'menu'
+require_relative 'dish'
+require_relative 'sms'
+
+class Order
+
+attr_reader :selection
```
silvabox: Don't expose the selection array publicly

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    quantity.times {@selection << dish}
+  end
+
+  def remove_dish(dish)
+    fail 'That dish is not in the order' unless @selection.include?(dish)
+    @selection.delete(dish)
+  end
+
+  def total_price
+    @bill = 0
```
silvabox: You don't need an instance variable here

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    @selection.each do |dish|
```
silvabox: Use `inject` or `reduce` instead

lib/sms.rb
```diff
@@ -0,0 +1,20 @@
+require 'rubygems'
+require 'twilio-ruby'
+
+class SMSConfirmation
+
+  def initialize
+    account_sid = 'AC6e70d6dcc83c097a6e913c3afbd424a3'
+    auth_token = '901c21df552c5cfcf478c47b4aa310e3'
```
silvabox: Don't commit auth tokens to public repos.  You ought to regenerate this in Twilio to be on the safe side.

PR #545, opened 2016-10-10
Submit Takeaway challenge

elizabethvenner:


lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+class Menu
+
+  attr_reader :dishes
+
+  def initialize(filename)
+    @dishes = {}
+    load_menu(filename)
+  end
+
+  def read_menu
```
maryrosecook: Love this domain-specific name that wraps internal logic.

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+    @dishes
+  end
+
+  private
+
+  def load_menu(filename)
```
maryrosecook: This is a lovely neat method.  Could the code use inject to build the @dishes? This would make the iteration more functional eg:

```
@dishes = file.readlines.inject({})...
```

lib/message.rb
```diff
@@ -0,0 +1,30 @@
...
+Dotenv.load
+
+class Message
+
+  def initialize
+    @client = Twilio::REST::Client.new ENV['ACCOUNT_SID'], ENV['AUTH_TOKEN']
+  end
+
+  def send_text(to)
+    @client.messages.create(
```
maryrosecook: Could this formatting be a little tidier? Have a look at a Ruby style guide for guidance on calling a method with a lot of args.

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+  def basket_contents
+    receipt = ""
+    @basket.each {|k, v| receipt << "#{v} x #{k} = £#{'%0.2f' % (v * menu.dishes[k].to_f)}, "}
+    receipt.slice!(-2..-1)
+    receipt
+  end
+
+  def total
+    total = 0
+    @basket.each {|k, v| total += v * menu.dishes[k].to_f}
```
maryrosecook: Your pre-review comment about this was good.  Ruby can inject (reduce) over a hash.

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    '%0.2f' % total.to_s
```
maryrosecook: This price formatting is nice.  Perhaps the code could be in a separate method?

lib/takeaway.rb
```diff
@@ -0,0 +1,43 @@
...
+    @menu = Menu.new(filename)
+  end
+
+  def print_menu
+    @menu.read_menu
+  end
+
+  def please_may_i_order
+    @order = Order.new(@menu_file)
+    "Sure, go right ahead!"
```
maryrosecook: Could the logic for talking to the customer be separate from the logic for making the order?

PR #546, opened 2016-10-10
Takeaway pull request

michkles:
Not completed

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,23 @@
+class Menu
+
+attr_reader :dishes
+
+  def initialize
+
+    @dishes ={
+      onion_rings:   2,
+      garlic_bread:  2,
+      chicken_wings: 3,
+      chicken_salad: 5,
+      mexicano:      7,
+      pepperoni:     7,
+      hawaii:        9,
+      coke:          2,
+      juice:         3,
+      cider:         3,
+      beer:          4
+    }
+  end
+
+
+end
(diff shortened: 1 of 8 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+class Menu
+
+attr_reader :dishes
```
silvabox: Don't expose the dishes hash publicly.  Write methods on `Menu` to allow other objects to query the dishes.

lib/order.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'menu'
+
+class Order
+
+attr_reader :basket, :menu, :summary
```
silvabox: Don't expose the basket publicly.  You've defined a `summary` method further down, so the `summary` attribute is obsolete

lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'order'
+
+class Takeaway
+
+attr_reader :order, :selected
```
silvabox: what is `selected`?  There are no tests for it and it isn't used anywhere

lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
...
+  def initialize(order = Order.new)
+    @order = order
+  end
+
+  def list_menu
+     order.menu.dishes
+  end
+
+
+  def add_item(dish, qty = 1)
```
silvabox: There's no need to duplicate this functionality.  Construct the order inside an `Order` instance then pass it to the takeaway for completion.

PR #548, opened 2016-10-10
Still need to add UI and formating

James-SteelX:


lib/menu.rb
```diff
@@ -0,0 +1,32 @@
+class Menu
+
+  attr_accessor :menu
```
silvabox: Never expose internal raw data with `attr_accessor`.  You've added methods to this class to interact with the data, so it's unnecessary.

lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+              {name: 'Nann',     price: 3.00},
+              {name: 'Tikka',    price: 5.00},
+              {name: 'Korma',    price: 5.00},
+              {name: 'Jalfrezi', price: 6.00},
+              {name: 'Bahji',    price: 3.00}
+            ]
+  end
+
+  def add_dish(dish, price)
+    hash = Hash.new
```
silvabox: This could be simplified to
```ruby
  @menu.push({ name: dish, price: price })
```

lib/restaurant.rb
```diff
@@ -0,0 +1,63 @@
+require_relative 'menu'
+require 'twilio-ruby'
+#require 'yaml'
+
+class Restaurant
+
+  attr_accessor :order, :menu
+
+  def initialize(menu_class = Menu)
```
silvabox: There's no need to pass the menu class, just pass the actual menu.

lib/restaurant.rb
```diff
@@ -0,0 +1,63 @@
+require_relative 'menu'
+require 'twilio-ruby'
+#require 'yaml'
+
+class Restaurant
+
+  attr_accessor :order, :menu
```
silvabox: don't expose the internal order array publicly.

lib/restaurant.rb
```diff
@@ -0,0 +1,63 @@
...
+     add(dish, quantity)
+  end
+
+  def cost
+    @cost = @order.inject(0) {|sum, hash| sum + hash[:price]}
+    "This order will cost £#{@cost}0"
+  end
+
+  def confirm_order(cost)
+   fail 'Please enter the correct amount' unless @cost == cost
```
silvabox: This method relies on a previous call to `cost` to initialize `@cost` which will cause inconsistent behaviour.

lib/restaurant.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def add_to_order(dish, quantity = 1)
+    dish = dish.capitalize
+    fail "#{dish} isn't available" if dish_check(dish)
+     add(dish, quantity)
+  end
+
+  def cost
+    @cost = @order.inject(0) {|sum, hash| sum + hash[:price]}
+    "This order will cost £#{@cost}0"
```
silvabox: I expect this method to return a numeric cost - not a presentation string.

lib/restaurant.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def display_menu
+  @menu.show_menu.each do |k|
+     "#{k}"
+   end
+  end
+
+  def add_to_order(dish, quantity = 1)
+    dish = dish.capitalize
+    fail "#{dish} isn't available" if dish_check(dish)
```
silvabox: it's not intuitive that `dish_check` should return true if the check fails and false if it succeeds.  This should be:
```ruby
fail "#{dish} isn't available" unless dish_check(dish)
```

lib/restaurant.rb
```diff
@@ -0,0 +1,63 @@
...
+    @client = Twilio::REST::Client.new account_sid, auth_token
+    @client.messages.create(
+    from: '+441202286016',
+    to:   '*************', #phone number to send to
+    body: "Thankyou, your order will be with you by #{Time.now + 3600}"
+    )
+    "Thankyou for your order, our chefs are hard at work creating your delicious food!"
+  end
+
+  def dish_check(dish)
```
silvabox: name boolean methods with a `?` suffix for better readability. e.g. `valid_dish?`

lib/restaurant.rb
```diff
@@ -0,0 +1,63 @@
...
+   !display_menu.any? { |h| h[:name] == dish }
+  end
+
+  def add(dish, quantity)
+    quantity.times{@menu.show_menu.each do |k, v| if k[:name] == dish
```
silvabox: Do not use `{` and `}` for multiline blocks.  use `do..end`

PR #550, opened 2016-10-10
Takemeaway

souljuse:
Unfinished but tested!

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,12 @@
+class Menu
+
+  attr_reader :menu
+
+  def initialize
+    @menu = []
+  end
+
+  def entry(dish = Dish.new("Patate lesse", 1))
+    @menu << {dish: dish.name ,price: dish.price}
+  end
+end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -6,10 +6,7 @@ class Menu
     @menu = [{dish: "cianfrusaglie alla rinfusa", price: 3}, {dish: "raudi accesi", price: 3}, {dish: "fragole alla carbonara", price: 14}]
   end
 
-  def to_s
-    'the menu is empty'
+  def entry(dish = Dish.new)
+    @menu << {dish: dish.name, price: dish.price}
   end
-  # def entry(dish = Dish.new)
-  #   @menu << {dish: dish.name, price: dish.price}
-  # end
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -3,10 +3,13 @@ class Menu
   attr_reader :menu
 
   def initialize
-    @menu = []
+    @menu = [{dish: "cianfrusaglie alla rinfusa", price: 3}, {dish: "raudi accesi", price: 3}, {dish: "fragole alla carbonara", price: 14}]
   end
 
-  def entry(dish = Dish.new("Patate lesse", 1))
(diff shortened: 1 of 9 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,12 @@
+class Menu
+
+  attr_reader :menu
```
silvabox: Don't expose this object as a public attribute

lib/menu.rb
```diff
@@ -0,0 +1,12 @@
...
+
+  def initialize
+    @menu = [{dish: "cianfrusaglie alla rinfusa", price: 3}, {dish: "raudi accesi", price: 3}, {dish: "fragole alla carbonara", price: 14}]
```
silvabox: Why not an array of `Dish` instances?  What is the point of `Dish` otherwise?

lib/menu.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  def entry(dish = Dish.new)
```
silvabox: The default here will cause an argument error as `Dish.new` requires 2 parameters.

lib/order.rb
```diff
@@ -0,0 +1,15 @@
+class Order
+  attr_reader :dish, :price, :order, :quantity
```
silvabox: It doesn't make sense for `dish` and `quantity` to be attributes as an order is a collection of these.

`order` should not be exposed as an attribute.  This class is the order, so it doesn't make sense for it to have an order property anyway.

lib/order.rb
```diff
@@ -0,0 +1,15 @@
...
+  def initialize()
+    @order = []
+    @price
+    @dish
+  end
+
+  def add(dish, quantity = 1)
+    @quantity = quantity
+    @price = dish.price
+    @order << {dish: dish.name ,price: dish.price, quantity: @quantity}
```
silvabox: just add the dish to the hash - why split it into name and price?

lib/takeaway.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_reader :order, :total, :menu
+
+  def initialize(order = Order.new, menu = Menu.new )
+    @order = order
+    @menu = menu
+    @total = 0
+  end
+
+  def see_menu
+    menu.menu
```
silvabox: This is one of the problems with naming an attribute the same as its class: what would `menu.menu` mean in plain English?

lib/takeaway.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def add(dish, quantity = 1)
```
silvabox: The order already has this functionality - there's no need to repeat it in Takeaway

lib/takeaway.rb
```diff
@@ -0,0 +1,38 @@
...
+    order.add(dish, quantity)
+    charge
```
silvabox: This is unnecessary - the order should know how to total itself.

lib/takeaway.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def complete_order
+    send_text("Thank you for your order: £#{@total}")
+  end
+
+  private
+
+  def charge
+    @total += (order.price) * order.quantity
```
silvabox: This is feature envy.  This method makes a calculation based on two attributes it asks from order.  This is clear indication that the calculation belongs in order.

PR #554, opened 2016-10-15
Takeaway-Challenge

amaalali:
Second pull request incorporating better OO design.


lib/cart.rb
```diff
@@ -0,0 +1,29 @@
+class Cart
+  def initialize
+    @current_order = Array.new
+  end
+
+  def in_cart?(dish)
+    @current_order.dup.include?(dish)
```
silvabox: no need to `dup` the array here - you're not exposing it

lib/cart.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def add(dish)
+    @current_order << dish
+  end
+
+  def check
```
silvabox: this method name does not convey its behaviour.

lib/cart.rb
```diff
@@ -0,0 +1,29 @@
...
+    @current_order.dup
+  end
+
+  def is_empty?
```
silvabox: avoid the `is_` prefix.  Just `empty?` is more Ruby.

lib/cart.rb
```diff
@@ -0,0 +1,29 @@
...
+    @current_order.empty?
+  end
+
+  def total
+    sum = 0
+    @current_order.each do |item|
```
silvabox: Use `inject` or `reduce` here.

lib/cart.rb
```diff
@@ -0,0 +1,29 @@
...
+      sum += item[:quantity] * item[:price]
+    end
+    "Your total is £#{'%.2f' % sum}"
```
silvabox: Don't return a string here - you're losing valuable data accuracy.  Return the numeric total and allow the caller to decide what to do with it.

lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+  def print
+    k=1
+    @menu.each do |item|
+      puts "#{k}. #{item[:dish]} ... £#{'%.2f' % item[:price]} "
+      k+=1
+    end
+  end
+
+  def dishes
+    @menu
```
silvabox: You're exposing the internal array here.  At the very least use `dup`.  Otherwise add the relevant methods to `Menu` so I never need access to the array.

lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def number_of_dishes
+    @menu.dup.length
```
silvabox: Don't need `dup` here

lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def select(item_number, amt)
+    {dish: @menu[item_number-1][:dish], price: @menu[item_number-1][:price], quantity: amt}
```
silvabox: Is this the menu's responsibility?

PR #617, opened 2017-01-15
Week2 weekend challenge: Takeaway.

kateloschinina:
This is a test pull request. More functionality to be implemented and code refactored further.


lib/take_away.rb
```diff
@@ -0,0 +1,95 @@
...
+  def read_menu(filename = @filename)
+    load_menu_from_the_file(filename)
+    @menu.print_menu
+  end
+
+  def add_to_order(menu_number, quantity)
+    is_dish_on_the_menu?(menu_number)
+    update_order(menu_number, quantity)
+    puts "#{quantity}x #{@menu_hash[menu_number].name}(s) added to your busket."
+  end
```
neoeno: When you start to see nouns like `order` appear in your method names again and again, that's sometimes a sign that the Single Responsibility Principle is being breached somehow.

I notice that `TakeAway` takes responsibility for the contents of an order — does that seem right to you?

.env
```diff
@@ -0,0 +1,7 @@
+ACCOUNT_SID=ACcd05917f7ea7e479a69506039b1d0898
+AUTH_TOKEN=ccf501f3df91ed70d4d426fcf2c6553d
+TEST_ACCOUNT_SID=ACcd05917f7ea7e479a69506039b1d0898
+TEST_AUTH_TOKEN=ccf501f3df91ed70d4d426fcf2c6553d
+PHONE_NUMBER=+447766444001
+FROM_PHONE=+441133206201
+#JJ2VEWD3MPiSYmb652NxmhmmTv88JkgmEcliu7oO
```
kateloschinina: Yes, good point! I was struggling to run Travis CI as when env file wasn't uploaded it didn't pass rspec tests, raising errors where it needed env data. Is there a way to go around it?

lib/menu.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  def file_not_exist?(filename)
+    message = "Cannot open file: there is no file with this name. Please try another filename."
+    raise message if !File.exists?(filename)
+  end
+
+  def print_header(output_width)
+    puts "Momofuku Noodle Bar, NY".center(output_width),
+    "-----------------".center(output_width),
+    ""
```
kateloschinina: Is it that comma is missing? =) I guess we can replace it with \n - will do this in my prog

lib/message.rb
```diff
@@ -0,0 +1,41 @@
...
+    triger_twilio_sms
+    prints_error_if_any
+  end
+
+  private
+
+  def prints_error_if_any
+    rescue Twilio::REST::RequestError => e
+    puts e.message
+  end
```
kateloschinina: Actually it does work! =) (just checked it by changing from_number, and got Twilio::REST::RequestError: The From phone number +4411332062011 is not a valid, SMS-capable inbound phone number or short code for your account.)

menu.csv
```diff
@@ -0,0 +1,12 @@
...
+Smothered Katsu Chicken,"pickled green beans, shiitake gravy",24,Daily
+Brussels Sprouts,"miso, chili, pine nuts",14,Daily
+Shrimp,"spicy mayo, pickled red onion, iceberg",13,Buns
+Shiitake,"hoisin, scallion, cucumber",13,Buns
+Chicken Meatball,"jalapeño, iceberg, paprika mayo",13,Buns
+Brisket,"horseradish, pickled red onion, cucumber",13,Buns
+Swiss Chard,"Asian pear, sesame, kimchi vinaigrette",8,Small Plates
+Spicy Cucumber,"scallion, togarashi, almonds",8,Small Plates
+Smoked Chicken Wings,"pickled chili, garlic, scallion",15,Small Plates
+BBQ Pork Belly Ssäm,"lettuce cups, kimchi",17,Small Plates
```
kateloschinina: Thanks! Had a glance on Wikipedia - looking forward to learn how to work with it =)

lib/menu.rb
```diff
@@ -0,0 +1,57 @@
...
+    print_body(output_width)
+  end
+
+  private
+
+  def load_data_into_hash(filename)
+    @dishes = Hash.new
+    CSV.foreach(filename) do |row|
+      dish = Dish.new
+      dish.name, dish.ingredients, dish.price, dish.type  = row[0], row[1], row[2].to_i, row[3]
```
kateloschinina: A little bit confused how to use keyword arguments in this case. Two options that work for me are 
(right now I am creating Dish as Dish = Struct.new(:name, :ingredients, :price, :type))
(1) dish = Dish.new(row[0], row[1], row[2].to_i, row[3]) - don't like this one as it makes it unclear which row corresponds to what
and (2)  dish = Dish.new
      dish.name, dish.ingredients, dish.price, dish.type  = row[0], row[1], row[2].to_i, row[3]

lib/menu.rb
```diff
@@ -0,0 +1,57 @@
...
+      @dishes.store(@dishes.length+1,dish)
```
kateloschinina: Tried to do it with an array, but I found it inconvenient to iterate through it. Why using hashes is worse than using arrays?

lib/menu.rb
kateloschinina: Tried to replace with \n  - and by some reason .center(output_width) doesn't work right. ((

spec/take_away_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    let(:order_hash) { double :order_hash }
+    before do
+      allow(take_away).to receive(:finalize_order)
+    end
+    it { is_expected.to respond_to(:place_order).with(1).argument }
+    it 'saves an order' do
+      # Figure out how to test!
+      # (take_away.order).stub(:total) { "total" }
+      # expect(customer.order).to eq("total")
+    end
```
kateloschinina: Indeed, I have moved some take_away functionality to order, and it seems to work and be tested better, and look neater!

PR #623, opened 2017-01-15
Challenge completed (except for placing orders via text message)

sliute:


sliute:
Hi Kay,


Thanks for the feedback! I love it because it's detailed and thoughtful and
gives me concrete starting points I can use to improve my thinking about
code. I'll go through it item by item and I plan to use the results asap
(i.e. from this weekend forward).


Arigato (with ceremonial bow :-) ),

PR #646, opened 2017-01-16
First pull request

rossbenzie:


lib/order.rb
```diff
@@ -0,0 +1,60 @@
...
+    "The requested item is not on the menu"
+  end
+  def not_in_basket_error
+    "You have not ordered this item"
+  end
+  def confirmation_msg
+    "Thank you. Your order was placed and will be delivered before"
+  end
+  def delivery_time
+    "#{(Time.now.hour + 1).modulo(24)}:#{Time.now.min}"
```
neoeno: Nice — good catch on the modulo thing. This would have been a good opportunity to test `#buy` to make sure it behaves well at 11pm, midnight, 1am etc — and thereby a good opportunity to use mocking to ensure you didn't have to stay up until midnight just to run your tests.

lib/order.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+  def display_header
+    print "\nYOUR ORDER\n\n"
+    printf("%-20s %5s %8s\n","ITEM","QTY","PRICE")
+  end
+  def display_basket
+    basket.each_item do |item, qty|
+      printf("%-20s %5d %8.2f \n", item, qty, menu.price(item))
```
neoeno: I've not seen `printf` used a lot in ruby — but I think this is a neat piece of code. I'd personally spend quite a while trying to figure out how to make it more readable, though. How could I make it clearer what's going on here?

PR #663, opened 2017-02-26
Takeaway challenge!

FloraHarvey:


spec/dish_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  before :each do
+    allow(menu).to receive(:list).and_return([{:number => 3, :name => "Margherita", :price => 8.95}])
+  end
+
+  describe "#get_details" do
+    it "gets the correct details" do
+      expect(dish.get_details(3, menu)).to eq({:number => 3, :name => "Margherita", :price => 8.95})
+    end
+  end
+
```
maryrosecook: Wasn't sure where to put this comment! Yes, expecting other classes to receive messages is a good way to test boundaries between classes.  Have a look at the POODR testing chapter for details on an approach to this.

spec/order_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+    #   it "returns a hash with dish details" do
+    #     order.add_dish(1, menu)
+    #     expect(order.get_total).to eq(21.90)
+    #   end
+    # end
+
+    describe "#place_order" do
+      it "confirms order has been placed and send a text message" do
+        order.add_dish(3, menu)
+        allow(order).to receive(:send_text).and_return(message)
```
maryrosecook: This test is a perfect candidate for using `expect(...).to receive(...)`. You could use it on the send_text method of an SMS instance.

lib/menu.rb
```diff
@@ -0,0 +1,53 @@
...
+  require 'csv'
+
+  attr_reader :current_order, :list
+
+  def initialize
+    @list = []
+    load_menu
+  end
+
+  def view_menu
```
maryrosecook: Have a think about whether the details of menu printing code belong in the Order class.

lib/menu.rb
```diff
@@ -0,0 +1,53 @@
...
+    sections = get_dish_type
```
maryrosecook: Could this logic be refactored so it's clearer?

PR #677, opened 2017-02-27
Weekend Challenge 2: Takeaway Challenge

ClemCB:


README.md
```diff
@@ -47,45 +35,30 @@ As a customer
...
-This repo works with [Coveralls](https://coveralls.io/) to calculate test coverage statistics on each pull request.
-
-Build Badge Example
-------------------
+## Limitations/Improvements
 
-[![Build Status](https://travis-ci.org/makersacademy/takeaway-challenge.svg?branch=master)](https://travis-ci.org/makersacademy/takeaway-challenge)
-[![Coverage Status](https://coveralls.io/repos/makersacademy/takeaway-challenge/badge.png)](https://coveralls.io/r/makersacademy/takeaway-challenge)
+- I would implement greater testing of behaviours, rather than states.
+- I would make improvements to checking the bill against running total - currently there is an error raised if the bill does not equal total price of items. However, there is no way for the customer to check the bill against their own calculations.
```
maryrosecook: I feel this user story is a little ambiguous.  I think fine to just have a method that reports the total for the order.  Then the user checks it against a manual total they may have come to in their head.

lib/menu.rb
```diff
@@ -0,0 +1,11 @@
+class Menu
+
+  attr_reader :food_options
+
+  def food_options
+    dishes = { 'Spring rolls' => 2.50,
```
maryrosecook: Does the code need to do `dishes = ` here?

maryrosecook: Could the dishes be extracted to a constant so that it can be changed without changing code?

lib/order.rb
```diff
@@ -0,0 +1,65 @@
+class Order
+
+  attr_reader :current_order, :bill, :string_summary
+
+  def initialize
+    @menu = Menu.new
+    @current_order = []
+    @bill = 0
+    @string_summary = ''
```
maryrosecook: Is this required? Could the string_summary be derived from the `@current_order` when it was needed?

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+  def save_order(food_order, quantity)
+    array_save_order(food_order, quantity)
+    string_save_order(food_order, quantity)
+  end
+
+  def array_save_order(food_order, quantity)
+    @menu.food_options.each do | dish, cost |
+      if dish == food_order
+          cost = cost * quantity
+          @bill += cost #needs to be purely functional - is saving to an instance variable and will cause problems in the future!
```
maryrosecook: Totally agree!

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+    @menu.food_options.has_key?(food_order)
+  end
+
+  def save_order(food_order, quantity)
+    array_save_order(food_order, quantity)
+    string_save_order(food_order, quantity)
+  end
+
+  def array_save_order(food_order, quantity)
+    @menu.food_options.each do | dish, cost |
```
maryrosecook: Is it possible to rewrite this code so that, instead of iterating through the food options to find the price, it just does a hash lookup to find the item price using the item name as a hash key?

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def order_received(food_order, quantity = 1 )
+    raise "There is no #{food_order} available on the menu" unless on_menu?(food_order)
+    save_order(food_order, quantity)
+    "Order received for #{quantity} #{food_order}."
+  end
+
+  def check_total
+    raise "There is an error - your bill is not the same as the item cost." unless bill_correct?
+    to_two_decimals = sprintf('%.2f', @bill)
```
maryrosecook: Is it possible to extract the code that converts a number to 2 decimal places to a separate method for clarity?

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  def initialize
+    @menu = Menu.new
+    @current_order = []
+    @bill = 0
+    @string_summary = ''
+  end
+
+  def order_received(food_order, quantity = 1 )
+    raise "There is no #{food_order} available on the menu" unless on_menu?(food_order)
```
maryrosecook: Having an `on_menu?` helper method makes the code nice and clear! Does the method belong on the order?

lib/takeaway.rb
```diff
@@ -0,0 +1,39 @@
...
+    @menu = menu
+    @order = order
+    @text_confirmation = text_confirmation
+  end
+
+  def read_menu
+    @menu.food_options
+  end
+
+  def make_selection(food_order, quantity = 1)
```
maryrosecook: It's great that Takeaway has methods like `order_summary`, `make_selection`, `read_menu`. This makes it like a user interface class for the user to interact with.  But it's still very focused because it delegates the actual work to other classes.

lib/text_confirmation.rb
```diff
@@ -0,0 +1,34 @@
+require 'dotenv'
+require 'twilio-ruby'
+
+class TextConfirmation
```
maryrosecook: It feels like TextConfirmation has two responsibilities: sending texts and composing texts about takeaway orders.  Could these responsibilities be split up?

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    it 'raises an error if bill does not have same value as combined selection price', :focus => true do
+      allow(order).to receive(:bill_correct?).and_return false
+      expect{order.check_total}.to raise_error "There is an error - your bill is not the same as the item cost."
+    end
+  end
+
+  describe 'is saved and can be requested by the customer' do
+
```
maryrosecook: Try and be consistent with your whitespace usage.  There's an empty line after this describe line, but not after the other describe lines in the same project.

spec/text_confirmation_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'text_confirmation'
+
+describe TextConfirmation do
+
+  subject(:text_confirmation) { described_class.new }
+  let (:order) { double :order }
+
+  it 'should create and send a text message' do
+    expect(text_confirmation).to receive(:order_finished).with(order)
```
maryrosecook: It feels like this test isn't testing very much.  The test calls `order_finished` and then checks it was called.  Have a think about what you actually want to check.  Happy to talk this through - grab me if you want!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  describe '#place_order' do
+
+    before :each do
+      allow(takeaway).to receive(:send_text)
+    end
+
+    it 'should send text message' do
+      expect(takeaway).to receive(:send_text).with("Thank you for your order: £0")
```
maryrosecook: Have a look at this test.  I'm worried it only passes because `takeaway` is `allow`ed to receive `send_text` (which is a private method on the actual class).  Happy to talk it through!

spec/menu_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'menu'
+
+describe Menu do
+
+  subject(:menu) { described_class.new }
+
+  it { is_expected.to respond_to(:food_options) }
+
```
maryrosecook: Could this test file also check that the expected items are on the menu?

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    before :each do
+      order.order_received('Spring rolls', 1)
+    end
+
+    it 'is saved to an array' do
+      expect(order.current_order[0]).to include 'Spring rolls'
```
maryrosecook: Yes, this feels like an example where state is being tested rather than behaviour.  Is there are related behaviour of the order that you can check happens when an item has been added to the order?

PR #678, opened 2017-02-27
Takeaway Challenge, Week 2

JohnChangUK:


spec/confirmation_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe Confirmation do
+
+  subject(:confirmation) { described_class.new }
+
+  before do
+    allow(confirmation).to receive(:send_text) {"Your order has been placed and will be delivered around 10:00pm"}
+  end
+
+  # it 'should show the time an hour ahead in hours and minutes' do
+  #   expect(confirmation.time).not_to eq(Time.now)
```
maryrosecook: It feels like this test would test state.  Is there a way to test behaviour?

maryrosecook: Try having a read of the testing chapter in POODR.  It talks about ways to test APIs by testing that the correct messages get sent to the API library with the correct arguments. eg check that `create` gets called on client.messages with the right message body.

lib/order.rb
```diff
@@ -0,0 +1,68 @@
...
+require_relative 'confirmation'
+require 'dotenv'
+Dotenv.load
+
+class Order
+
+  attr_reader :selection, :menu, :prices, :total, :orders, :confirmation
+
+  def initialize(menu = Menu.new)
+  @menu = menu
```
maryrosecook: Try and format your code in accordance with a Ruby style guide.  It really helps readability, and also makes it look fully professional.

lib/order.rb
```diff
@@ -0,0 +1,68 @@
...
+  @prices = []
+  end
+
+  def start
+    puts @menu.introduction
+    puts @menu.show_menu
+  end
+
+  def select_order(item, amount)
+    @selection = { item.to_sym => amount}
```
maryrosecook: Could the code for adding an item to the order be clarified? Is it necessary to put the order into an instance variable and then call another method to add that instance variable to the selection?

lib/order.rb
```diff
@@ -0,0 +1,68 @@
...
+    add_to_orders
+    return "#{amount} x #{item} added to your order!"
+  end
+
+  def get_price
+    @orders.each do |order|
+    order.each do |item, amount|
+    @price = menu.dishes[item]
+    puts "#{amount} x #{item.to_sym} at £#{@price} each = £#{@price *= amount}"
+    @prices << @price
```
maryrosecook: Yes, get_price seems like it could be a good method to change so it doesn't have any side effects (that is, it doesn't rely on changing `@prices`).  Is `get_price` required? Would just having `total_price` be enough? Then total_price could just calculate and return the price and there would be no need for `@prices`.

spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  subject(:menu) { described_class.new }
+
+  describe '#show_menu' do
+    it { is_expected.to respond_to(:show_menu) }
+  end
+
+  it 'should introduce the user' do
+    expect(menu.introduction).to eq("Welcome to MA Takeaway. Please order from our tasty selection of dishes!")
+  end
+
```
maryrosecook: Could menu have some more tests that check that it has some dishes on it?

lib/order.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def calculate_price
+    total_price
+    puts display_total_price
+    puts confirm_alert
+  end
+
+  def total_price
+    @total = @prices.inject { |accum_value, element| accum_value += element }
```
maryrosecook: This line is really easy to understand because of it uses Ruby's built-in inject, rather than using custom code to do the summing.

lib/order.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def start
+    puts @menu.introduction
+    puts @menu.show_menu
+  end
+
+  def select_order(item, amount)
+    @selection = { item.to_sym => amount}
+    add_to_orders
+    return "#{amount} x #{item} added to your order!"
```
maryrosecook: No need to use an explicit `return`.  Interesting question: should you ever use `return` in Ruby?

JohnChangUK: I know that the Ruby style would generally only use an explicit returns for an early return out of the method, e.g. if the method has two conditions and it meets the first. Ruby is more focused on code minimalism.

maryrosecook:
Yes!

PR #706, opened 2017-04-22
Completed Week 2 challenge

hyper0009:
will add complete checkout system and gem tomorrow.


hyper0009:
Completes Weekend challenge - Week2

PR #710, opened 2017-04-23
Takeaway challenge

bruxelles86:


spec/neworder_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'neworder'
+require 'menu'
+require 'timestamp'
+require 'sms'
+
+describe NewOrder do
+  let(:neworder) { described_class.new }
```
neoeno: We might usually phrase this like so:

```
subject(:new_order) { described_class.new }
```

See this:

> A named subject improves on the explicit subject by assigning it a contextually
semantic name. Since a named subject is an explicit subject, it still defines the value that is returned by the subject method in the example scope. However, it defines an additional helper method with the provided name. This helper method is memoized. The value is cached across multiple calls in the same example but not across examples.
https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/subject/explicit-subject

By convention we'd make the name `new_order` rather than `NewOrder` — but hard to know that unless you see code that does it I suppose!

spec/neworder_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    neworder.add_dish(:fried_chicken, 2)
+    expect(neworder.total).to eq 17
+  end
+
+  it 'takes total and raises an error if incorrect' do
+    neworder.add_dish(:pizza, 2)
+    expect { neworder.place_order(19.00) }.to raise_error 'Incorrect total given'
+  end
+
+  # not 100% if this is really testing the .send method is called, but it passes...
```
neoeno: In this case, you could deliberately break the implementation code (maybe by deleting some of it, or changing it a little, or in a few ways) and see if the test fails.

spec/timestamp_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'timestamp'
+
+describe Timestamp do
+
+  it 'returns the time without seconds' do
+    expect(Timestamp.new.hour_from_now.to_s.gsub(/[1234567890]/, '*')).to eq '**:**'
```
neoeno: Interesting approach! You might get a better test if you used dependency injection here to inject `Time` into your Timestamp class. At the moment, this could fail if the method always returns `00:00` — probably not an acceptable response.

lib/sms.rb
```diff
@@ -0,0 +1,19 @@
...
+    @sms_template = "Your order was placed and will be delivered before #{@timestamp}"
+  end
+
+  def send
+    client = Twilio::REST::Client.new('ACa34600616f043f317628e4c853f37b86',
+    '7ea827db27a726a87e117cb7787c4bfa'
+    )
+    client.messages.create :to => "+447413404379", :from => "+441582380312",
+    :body => @sms_template
+  end
```
neoeno: Probably a good opportunity to use the ENV variables, like in the review rubric.

lib/neworder.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def add_dish(dish, number)
+    number.times { @order << { dish => @menu.items[dish] } }
+  end
+
+  def total
+    @amount = 0.00
+    @order.each { |item| @amount += item.values[0] }
+    @amount
```
neoeno: You could also write this as:

```ruby
def total
  @order.map { |item| item.values[0] }.sum
end
```

Also — no need for `@amount` to be an instance variable right? Since it's not used elsewhere?

lib/neworder.rb
```diff
@@ -0,0 +1,30 @@
...
+  attr_reader :menu, :order, :sms
+
+  def initialize
+    @order = []
+    @menu = Menu.new
+    @sms = SMS.new
+  end
+
+  def add_dish(dish, number)
+    number.times { @order << { dish => @menu.items[dish] } }
```
neoeno: This looks a little tricky. I'd probably break it out like:

```ruby
number.times do
  @order << { dish => @menu.items[dish] }
end
```

Just makes it easier to scan by clarifying that the `#times` is iteration and the other braces are for a hash.

lib/neworder.rb
```diff
@@ -0,0 +1,30 @@
+require_relative './menu.rb'
+require_relative './timestamp.rb'
+require_relative './sms.rb'
+
+class NewOrder
```
neoeno: What do you think about this as class name? I'm not sure about the `New`

PR #726, opened 2017-04-24
Partial Solution up to User Story 3. No text message funcionality

BertZZ:


lib/checkout.rb
```diff
@@ -0,0 +1,16 @@
+class Checkout
+  def summarise(food, prices)
+    summary_string = ""
+    food.each do |key, value|
+      cost = prices[key] * value
+      summary_string << "#{value}x #{key} = £#{cost}, "
+    end
+    summary_string.chomp(", ")
+  end
```
neoeno: Here's another way of writing this:

```ruby
  def summarise(food, prices)
    items = food.map do |key, value|
      cost = prices[key] * value
      "#{value}x #{key} = £#{cost}"
    end
    items.join(", ")
  end
```

PR #766, opened 2017-05-22
Pull Request

lubosmichalic:


README.md
```diff
@@ -1,10 +1,11 @@
...
+## Takeaway Challenge
+
 ```
                             _________
-              r==           |       |
-           _  //            |  M.A. |   ))))
-          |_)//(''''':      |       |
+              r==           | Beef  |
+           _  //            |   &   |   ))))
+          |_)//(''''':      |  Beer |
```
sjmog: 🍺

lib/restaurant.rb
```diff
@@ -0,0 +1,10 @@
+class Restaurant
+
+  attr_reader :menu, :takeaway
+
+  def initialize(name, takeaway)
+    @name = name
+    @menu = Menu.new
+    @takeaway = takeaway
+  end
+end
```
sjmog: I don't really see the point of this class. Try killing it?

spec/cart_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'cart'
+
+describe Cart do
+
+end
```
sjmog: 👍

sjmog: But more seriously, `Cart` has two methods: `list`, which is responsible for `reading the list`, and `select_item`, which is responsible for `adding an item to the list`. I think the second method is unnecessary, as it just replicates the functionality of `<<`.

So you could do something like this:

```ruby
describe Cart do
  describe '#list' do
    it 'starts empty' do
      expect(Cart.new.list).to be_empty
    end

    it 'can have items added' do
      cart = Cart.new
      # not sure if I have this syntax right but it's something like this
      expect { cart.list << :item }.to change { cart.list.count }.by(1)
    end
  end
end
```

sjmog: Of course, another argument is: you don't need a `Cart` class because by now it's basically an `Array` instance.

spec/item_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  let(:item) { described_class.new('Burger', 3) }
+
+  it 'has a name' do
+    expect(item.name).to eq 'Burger'
+  end
+
+  it 'has a price' do
+    expect(item.price).to eq 3
+  end
+end
```
sjmog: Take a look at `Struct` for this. There's no 'behaviour' here so a Struct is a better choice.

spec/printer_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+  end
+
+  describe '#input_quatity' do
+    it 'prints a request for quantity' do
+      expect{ printer.input_quantity }.to output("Please enter the quantity\n").to_stdout
+    end
+  end
+
+end
```
sjmog: Great use of RSpec-specific STDOUT testing commands. I forgot about those. Can you post them into Slack too?

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'restaurant'
+
+describe Restaurant do
+  let(:restaurant) { described_class.new('Beef & Beer', 'Takeaway') }
+
+  it 'has a menu upon initialization' do
+    expect(restaurant.menu).to be_an_instance_of Menu
+  end
+
+end
```
sjmog: Again – I'm not sure this class needs to exist really. It doesn't seem to do anything other than wrap `Menu`.

PR #769, opened 2017-06-18
Takeaway challenge

Hempy49:


lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+      confirmation
+      basket.clear
+    else
+      return nil
+    end
+  end
+
+private
+
+  def confirmation
```
maryrosecook: When is it best to use nouns for method names, and when is it best to use verbs?

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+    sms.send_sms("Thankyou! Your order was placed and will be delivered before #{delivery_time}")
+  end
+
+  def delivery_time
```
maryrosecook: Love these small helper methods!

lib/sms.rb
```diff
@@ -0,0 +1,22 @@
+require 'dotenv'
+Dotenv.load
+require 'twilio-ruby'
+require "./lib/order.rb"
+
+class SMS
```
maryrosecook: Great that you broke this responsibility out!

lib/sms.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def send_sms(message)
```
maryrosecook: Could this method be broken down to make it more readable? To understand it at the moment, you need to read the steps and put them together in your head, rather than reading it as a set of higher level things that are being done.

spec/order_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      order.add_dish("galbi", 1)
+      order.add_dish("kimbap", 1)
+      expect(order.basket_summary).to eq ["1x galbi(s) = £10", "1x kimbap(s) = £4"]
+    end
+  end
+
+  describe "#checkout" do
+
+    before do
+      $stdin = StringIO.new("No")
```
maryrosecook: Nice use of StringIO!

spec/order_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "order"
+
+describe Order do
+
+  subject(:order) { described_class.new }
+
+  before do
+    allow(order).to receive(:send_sms)
```
maryrosecook: Watch out for this.  We're mocking the unit under test, which is dangerous, because it means we're not actually testing an accurate version of the real thing.  To avoid this, you could isolate the SMS dependency by injecting it.

PR #772, opened 2017-06-18
Takeaway Challenge

mihobo:
Takeaway Challenge

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,22 +13,15 @@ Takeaway Challenge
        ':..:'                ':..:'
 
  ```
-
 Instructions
--------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+------------
+* Fork this repo
+* Run the command 'bundle' in the project directory to ensure you have all the gems
+* Write a Takeaway program
 
 Task
 -----
-
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+Our task was to create a Takeaway program with the following user stories:
 
 ```
 As a customer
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -48,32 +41,123 @@ So that I am reassured that my order will be delivered on time
...
+  **Placing the order:**  
+  Once the customer is satisfied with their choices, they can place the order which sends them a text that thanks them for their order, with their name and what time it will be delivered (I set it to deliver an hour after the order was placed).
+
+  [I used a Twilio number and my phone number to test the text(s) deliveries could be received.]
+
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+How I plan to finish the challenge
+----------------------------------
+Although, I have met the requirements from the user stories, I would like to incorporate some doubles, mocks and stubs into my program. I found it difficult to add these without the tests breaking, even though the code was working correctly.
```
maryrosecook: Definitely worth lasering into why you found this tricky. Was it because there were too many responsibilities in one class? Was it understanding the errors? Was it debugging your way through several errors to get the code working? Was it understanding how to use doubles? Or was it something else?

README.md
```diff
@@ -48,32 +41,123 @@ So that I am reassured that my order will be delivered on time
...
 
-In code review we'll be hoping to see:
+```ruby
+describe Menu do
+  describe '#initialize' do
+    it { is_expected.to respond_to(:initialize) }
+  end
+```
+
+I expected the errors to show that I didn't have a 'Menu' class or an 'initialize' method. So, then proceeded to write the code for the tests to pass.
```
maryrosecook: These process analyses are really great! It's worth trying to critique your process.  Are they any approaches you took that you realise you could do better next time?

lib/menu.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  def create_menu
+    @@dishes = { 'Kobe Slider' => 7,
+                 'Shrimp Tempura' => 4,
+                 'Salmon Teriyaki' => 5,
+                 'Chicken Katsu' => 8,
+                 'Edamame Pot' => 2 }
+  end
+
+  def read_menu
```
maryrosecook: It feels like this method has two responsibilities.  Could they be separated?

lib/menu.rb
```diff
@@ -0,0 +1,21 @@
+class Menu
+  attr_reader :dishes
+
+  def initialize
+    create_menu
+  end
+
+  def create_menu
+    @@dishes = { 'Kobe Slider' => 7,
```
maryrosecook: Love that you've data-driven the menu!

lib/order.rb
```diff
@@ -0,0 +1,87 @@
+require_relative '../lib/menu.rb'
+require 'date'
+require 'twilio-ruby'
+
+class Order < Menu
```
maryrosecook: Have a think about why you made `Order` inherit from `Menu`.  Is inheritance used for this purpose? Is the `Order` and `Menu` relationship an inheritance one?

lib/order.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def combine_items
+    @basket.group_by(&:itself).each { |dish, quantity| @updated_basket[dish] = quantity.length }
+  end
+
+  def create_basket_total
+    total = 0
+    @basket.each { |dish| dish.each { |_name, price| total += price } }
+    return total
```
maryrosecook: Is the `return` keyword necessary here?

lib/order.rb
```diff
@@ -0,0 +1,87 @@
...
+    puts quantity.to_s + ' ' + dish + '(s) added to your basket.'
+    while quantity > 0
+      @basket << menu_item
+      quantity -= 1
+    end
+
+  end
+
+  def lookup_item(dish)
+    fail 'Please enter an item from the menu.' unless @@dishes.key?(dish)
```
maryrosecook: Is it `Order`'s responsibility to check if an item is on the menu?

lib/order.rb
```diff
@@ -0,0 +1,87 @@
...
+    total = create_basket_total
+    puts 'Total to pay: £' + total.to_s
+  end
+
+  def show_basket
+    show_items
+    show_total
+  end
+
+  def place_order
```
maryrosecook: Could this method be any shorter? Often, blank lines in a method are a sign it could be broken up.

maryrosecook: Does this method really place the order? Could its name be a closer description of what it does?

lib/order.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  def lookup_item(dish)
+    fail 'Please enter an item from the menu.' unless @@dishes.key?(dish)
+    chosen_dish = @@dishes.select do |item, _price|
+      item == dish
+    end
+    chosen_dish
+  end
+
+  def create_basket_items
```
maryrosecook: Could the encapsulation of the `Order` class be improved? One way to improve encapsulation is to try to make as many methods as possible private and then look to see if there are any responsibilities in the collective private methods that could be broken out.

lib/order.rb
```diff
@@ -0,0 +1,87 @@
...
+    combine_items
+    list = ''
+    @updated_basket.each do |dish, quantity|
```
maryrosecook: Does `updated_basket` need to be an instance variable that persists for the life of the `Order` object?

lib/order.rb
maryrosecook: Could the message sending responsibility be delegated?

PR #778, opened 2017-06-19
weekend challenge attempt

alistairkung:


spec/order_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      it "should print the sum of the basket" do
+        expect{subject.review_order}.to output{"Your total is £6."}.to_stdout
+      end
+    end
+  end
+
+  describe "#place_order" do
+    it "should confirm the order has been placed" do
+       expect {subject.place_order}.to output("Order confirmed\n").to_stdout
```
maryrosecook: Make sure to adhere to a Ruby style guide.  (The problem is that this line is indented 3 spaces, rather than 2.)  This style guide is good: https://github.com/bbatsov/ruby-style-guide

maryrosecook: Run `rubocop` and it will point out lots of style violations for you!

Gemfile
```diff
@@ -8,3 +8,4 @@ gem 'rspec'
 gem 'rubocop'
```
maryrosecook: Nice job with the 100% test coverage!

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+    @menu = menu
+  end
+
+  def add(quantity = 1, item)
+    selection = search(item)
+    add_to_basket(quantity, selection)
+    puts quantity.to_s + ' ' + item + ' added to basket!'
+  end
+
+  def review_order
```
maryrosecook: In a way, this method has one responsibility - showing a view of the current order.  However, it also has multiple responsibilities because it handles the details of some of the parts of its main responsibility (checking the basket isn't empty and printing the order total).  Could you delegate the details of those responsibilities?

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def place_order
+    puts "Order confirmed"
+    sms_notifier
+  end
+
+  private
+
+  def search(item)
```
maryrosecook: Is it `Order`'s responsibility to check if an item is on the menu?

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+    fail "Sorry, item not on menu!" unless menu.food_list.has_key?(item)
+    menu.food_list.select { |meal, price| meal == item }
```
maryrosecook: I can't quite tell why this line is here.

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def add_to_basket(quantity, selection)
+    while quantity > 0 do
```
maryrosecook: Is there a more Ruby-like way of doing something a specified number of times?

lib/order.rb
```diff
@@ -0,0 +1,66 @@
+require 'twilio-ruby'
+require_relative 'menu'
+
+class Order
+  attr_reader :basket, :menu
+
+  def initialize(menu = Menu.new)
```
maryrosecook: Love this injection of the Menu!

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def add_to_basket(quantity, selection)
+    while quantity > 0 do
+      basket.push(selection)
+      quantity -= 1
+    end
+  end
+
+  def calculate_total
+    basket.inject(0) { |sum, item| sum + item.values.first }
```
maryrosecook: Really nice!

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def list_basket
```
maryrosecook: The fact that several methods have the same domain concept (basket) in their name is a sign that there's a responsibility that could be broken out.

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+    basket.each { |item| puts "#{item.keys.first} : £#{item.values.first}" }
+  end
+
+  def sms_notifier
```
maryrosecook: Could this responsibility be broken out?

lib/order.rb
```diff
@@ -0,0 +1,66 @@
+require 'twilio-ruby'
+require_relative 'menu'
+
+class Order
+  attr_reader :basket, :menu
+
+  def initialize(menu = Menu.new)
+    @basket = []
```
maryrosecook: Could `@basket` be encapsulated inside `Order`?

spec/order_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  describe "#add" do
+    context "selecting item from menu" do
+
+      before do
+        subject.add("pancakes")
+      end
+
+      it "should add the item to the basket" do
+        expect(subject.basket).to include("pancakes" => 5)
```
maryrosecook: It feels like this is testing the state of `Order`, rather than its behaviour.  This means that if the implementation of basket changes, this test might need to be updated for the new implementation.  Is there a way to better encapsulate `basket` and test the behaviour of `Order` to avoid this?

spec/order_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'order'
+
+describe Order do
```
maryrosecook: Could menu have been doubled to better isolate the `Order` unit tests?

maryrosecook: Is there a way to make these tests not send a text each time they're run?

PR #792, opened 2017-07-16
Reporting on Takeaway Challenge

aballal:
Completed the challenge
Rubocop - No offenses
Test Coverage - 100% (when messaging is not stubbed out)
README explains the project in detail

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,23 +1,14 @@
 class Menu
 
-  def initialize
-    @dishes = load_dishes
+  attr_reader :name
+
+  def initialize(name, dishes)
+    @name = name
+    @dishes = dishes
   end
 
   def dishes
     @dishes.dup
-  end
-
-  private
-
-  def load_dishes
-    {
-      "Cheese & Tomato"   => 6.00,
-      "Mighty Meaty"      => 9.00,
-      "Pepperoni Passion" => 8.00,
-      "Tandoori Hot"      => 9.00,
-      "Vegi Sizzler"      => 8.00
-    }
   end
 
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,23 @@
+class Menu
+
+  def initialize
+    @dishes = load_dishes
+  end
+
+  def dishes
+    @dishes.dup
+  end
+
+  private
+
+  def load_dishes
+    {
+      "Cheese & Tomato"   => 6.00,
+      "Mighty Meaty"      => 9.00,
+      "Pepperoni Passion" => 8.00,
+      "Tandoori Hot"      => 9.00,
+      "Vegi Sizzler"      => 8.00
+    }
+  end
+
+end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -4,7 +4,7 @@ class Menu
 
   def initialize(name, *dishes)
     @name = name
-    @dishes = Hash.new(0)
+    @dishes = Hash.new
     load_dishes(dishes)
   end
 
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -12,6 +12,16 @@ class Menu
     @dishes.dup
   end
 
+  def to_s
+    string = "#{name} Menu\n"
+    @dishes.each do |dish, price|
+      string += "#{dish}".ljust(25)
+      string += "#{'%.02f' % price}".rjust(10)
+      string += "\n"
+    end
+    string
+  end
+
   private
 
   def load_dishes(dishes)
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -2,13 +2,25 @@ class Menu
 
   attr_reader :name
 
-  def initialize(name, dishes)
+  def initialize(name, *dishes)
     @name = name
-    @dishes = dishes
+    @dishes = Hash.new(0)
+    load_dishes(dishes)
   end
 
   def dishes
     @dishes.dup
+  end
+
+  private
+
+  def load_dishes(dishes)
+    fail "No dishes in the menu" if dishes.empty?
+    dishes.each_slice(2) do |name, price|
+      fail "Invalid dish name '#{name}'" unless name.is_a?(String)
+      fail "Invalid dish price '#{price}'" unless price.is_a?(Numeric) && price > 0
+      @dishes[name] = price.round(2)
+    end
   end
 
 end
--- a/lib/messenger.rb
+++ b/lib/messenger.rb
@@ -1,0 +1,18 @@
+require 'dotenv'
+require 'twilio-ruby'
+
+class Messenger
+
+  def initialize
+    @account_sid = ENV['ACCOUNT_SID'] # Account SID from www.twilio.com/console
+    @auth_token = ENV['AUTH_TOKEN'] # Auth Token from www.twilio.com/console
+    @from = ENV['FROM'] # Your Twilio number
+    @to = ENV['TO'] # Your phone number
+    @client = Twilio::REST::Client.new @account_sid, @auth_token
+  end
+
+  def send_message(text)
+    @client.messages.create(body: text, to: @to, from: @from).sid
+    "Success"
+  end
+end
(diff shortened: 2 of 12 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+    string
+  end
+
+  private
+
+  def load_dishes(dishes)
+    fail "No dishes in the menu" if dishes.empty?
+    dishes.each_slice(2) do |name, price|
+      fail "Invalid dish name '#{name}'" unless name.is_a?(String)
```
ddosque: no need to type check - you prevent use of polymorphism in the future

ddosque: Validation is a good thing to think about, but validate for things that are important - you can check an object 'responds_to' any important methods, or you can validate that a necessary object is present, for example.

lib/messenger.rb
```diff
@@ -0,0 +1,18 @@
+require 'dotenv'
+require 'twilio-ruby'
+
+class Messenger
+
+  def initialize
+    @account_sid = ENV['ACCOUNT_SID'] # Account SID from www.twilio.com/console
+    @auth_token = ENV['AUTH_TOKEN'] # Auth Token from www.twilio.com/console
+    @from = ENV['FROM'] # Your Twilio number
+    @to = ENV['TO'] # Your phone number
```
ddosque: This ENV naming is a bit suspect - you want to know what your env variables are for. In this specific case, I would ENV['TWILIO_PHONE_NUMBER'] and then pass an argument to the send_message method with the clients phone number. The clients phone number is sure to change, so storing as a constant seems wrong.

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  def add(dish_name, quantity = 1)
+    return "Cannot add, order already placed" if placed?
+    return "Cannot add, #{dish_name} is not on the menu" unless @menu.dishes[dish_name]
+    update_total(@menu.dishes[dish_name], quantity)
+    update_quantity(dish_name, quantity)
+  end
+
+  def checkout(amount)
+    return "Cannot checkout, order already placed" if placed?
```
ddosque: careful returning strings - they are presentation logic

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def complete_order
+    @placed = true
+  end
+
+  def send_text_message
+    time = Time.new + 60 * 60
+    text = "Thank you! Your order will be delivered before #{time.hour}:#{time.min}"
+    @messenger_class.new.send_message(text)
```
ddosque: honestly I think I would probably make a new SMS class, so this line would look like:
```
@sms.new(text: text, tel: phone_number, time: time).send_message
```

spec/order_spec.rb
```diff
@@ -0,0 +1,107 @@
+require 'order'
+
+describe Order do
+  let(:menu) {
+    double(
+      :menu,
+      name: "Dominos",
+      dishes:
```
ddosque: I would probably separate this out, so you have:
```
let(:dishes) do
  {"cheese & tomato" => 6}
end
let(:menu){ double :menu, name: "dominos", dishes: dishes }
```
Also, keep your fixtures simple - you don't need 5 dishes, just one or 2 will be easier to test and verify your tests are correct.

spec/order_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    it 'has not been placed' do
+      expect(order).not_to be_placed
+    end
+  end
+
+  describe '#read_menu' do
+    it 'can read the menu' do
+      menu = double(:menu, name: "Dominos", dishes: { "Cheese & Tomato" => 6 })
```
ddosque: you already defined menu up top?

spec/order_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        expect(order.checkout(10)).to eq "Incorrect amount"
+      end
+
+      it 'places the order if the amount is correct' do
+        order.checkout(12)
+        expect(order).to be_placed
+      end
+
+      it 'sends a confirmation text' do
+        expect(order.checkout(12)).to eq "Success"
```
ddosque: vacuous, you are only testing line 85 of this spec

spec/order_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+  describe '#checkout' do
+
+    it 'does not checkout if the basket is empty' do
+      expect(order.checkout(0)).to eq "Cannot checkout, basket is empty"
+    end
+
+    context 'basket is not empty' do
+      before do
+        order.add("Cheese & Tomato", 2)
+        allow(messenger).to receive_message_chain(:new, :send_message).and_return("Success")
```
ddosque: It seems critical that order sends this message. I would definitely have a least 1 test that uses expect instead of allow.

lib/messenger.rb
```diff
@@ -0,0 +1,18 @@
+require 'dotenv'
+require 'twilio-ruby'
+
+class Messenger
```
ddosque: What are you making instances of? Messenger? Or messages? 

Beware class names ending in 'er' or 'or', they are quite often a sign that the abstraction might be a bit off. This is a class that builds messages - it should be called Message, TextMessage or SMS.

spec/messenger_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'messenger'
+
+describe Messenger do
+  subject(:messenger) { described_class.new }
+
+  describe '#send_message' do
+    it 'can send a message' do
+      allow(messenger).to receive(:send_message).and_return("Success")
+      message = "Your Dominos pizza is on its way!"
+      expect(messenger.send_message(message)).to eq "Success"
```
ddosque: vacuous - you are just testing the stub on line 8.

What you _actually_ want to test is that the twilio client is initialized properly (e.g. receives the #new message with the correct arguments) and that it receives the message chain messages.create (again with the correct arguments).

You could do that with a chain of doubles, each returning the next.

lib/menu.rb
aballal: Okay, understood. The reason I added type checks for menu instantiation alone was so that incorrect data couldn't get into the menu. One could very easily miss a comma, or enter items in the wrong order while inputting a list of dish name, dish price pairs separated by a comma.

lib/messenger.rb
aballal: Okay, understood.

lib/order.rb
aballal: Okay, understood and that makes a lot more sense.

spec/order_spec.rb
aballal: Okay, understood.

spec/order_spec.rb
aballal: Yes, but I needed a smaller menu for checking STDOUT more easily. After implementing your previous advice of having a simpler menu, this one will not be needed.

spec/order_spec.rb
aballal: Okay, understood. We were asked to stub out messaging in Rspec and only let feature tests send messages (requirements in README). Hence I first checked that Rspec tests were indeed sending messages and then stubbed them out.

spec/order_spec.rb
aballal: Sorry, didn't understand. Why is it vacuous? The test still steps into all the instructions in checkout, complete_order and send_text_message and only line 60 of order.rb is stubbed out by line 85 of order_spec.rb. Moreover, this test case is required when stubbing out the SMS class has to be commented and text sending has to be tested.

spec/messenger_spec.rb
aballal: Agreed, but the stub was only put in place to prevent Rspec from sending text messages (as mentioned in requirements) after having checked that the test passes. I cannot delete the test because I may need to comment the stubbing and test real functionality. I didn't want to give read access to any of the ENV variables hence couldn't test initialize. But after implementing the changes you suggested to SMS class, I can.

aballal:
@ddosque - Thank you very much for the review. That was very helpful.

PR #865, opened 2017-11-02
Takeaway challenge attempt

cazwazacz:
This includes ability to place orders via text message


app.rb
```diff
@@ -0,0 +1,26 @@
+require 'sinatra'
+require './lib/takeaway'
+require './lib/text'
+
+$takeaway = Takeaway.new
```
dearshrewdwit: How would you now think about doing this instead of using the evil evil global variable?

app.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  body = params['Body']
+
+  body.casecmp('menu').zero? ? send_menu : confirmation_text(body)
+
+end
+
+def confirmation_text(body)
+  $takeaway.parse_order(body)
+  total = $takeaway.order[0][:total]
```
dearshrewdwit: if possible define a better interface to the takeaway object with methods to return data you need rather than access data from keys from arrays held as state if that makes sense?

lib/billing.rb
```diff
@@ -0,0 +1,24 @@
+class Billing
+
+  def sum(order)
+    total = 0
+    order.each do |x|
```
dearshrewdwit: be descriptive with your naming!

lib/billing.rb
```diff
@@ -0,0 +1,24 @@
+class Billing
+
+  def sum(order)
```
dearshrewdwit: this method can be refactored to look cleaner:
something akin to:
```ruby
order.map { |item| item[:price] * item[:quantity] if item.key?(:name) }.compact.reduce(:+)
```
Without knowing why you're doing the things here, I would still split this up into three different methods to reduce the number of operations this method is doing: mapping, checking a conditional, looping and summing.

If that makes sense?

lib/billing.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def check_sum(order, order_total = sum(order))
+    total = 0
+    order.each do |x|
+      if x.key?(:name)
+        total += x[:price] * x[:quantity]
+      end
+    end
+    raise "Sum is not correct!" unless order_total == total
+    true
```
dearshrewdwit: Usually if you have to explicitly return true in a method there is a different approach you can take by refactoring.

Focus on the return value of the method and what the method needs to do to get there.

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
+require 'csv'
+
+class Menu
+
+  attr_reader :menu
+
+  def initialize
+    @menu = menu
```
dearshrewdwit: what do you mean by this?

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
+require 'csv'
+
+class Menu
```
dearshrewdwit: This seems mainly a file reading class. I'd consider intializing it with the file path and wrapping some of the reading in methods to reduce the amount that's publicly exposed if that makes sense? maybe at some point creating a csv reading class. 

Essentially, off the top of my head, i'd consider something like this:

```ruby
def initialize
  @menu = []
end

def menu
  create_menu_from_csv
  @menu
end

private

def create_menu_from_csv
  CSV.foreach("menu.csv") do |row|
  name, price = row
  @menu << { name: name, price: price.to_i }
  end
end
```

PR #868, opened 2017-11-05
First version takeaway challenge

mariekerkstoel:


lib/takeaway.rb
```diff
@@ -0,0 +1,81 @@
...
+  def read_menu
+    @menu.read_menu
+  end
+
+  def add_to_basket(dessert)
+    raise 'Please pick a dessert from the menu.' unless in_menu?(dessert)
+    adding_to_total(dessert)
+  end
+
+  def remove_from_basket(dessert)
```
dearshrewdwit: It would be easier to calculate the total only when you need to return it. This way it would help you delete this code here

lib/takeaway.rb
```diff
@@ -0,0 +1,81 @@
...
+    @basket.each { |item| item.each { |name, price| @total -= price if name == dessert } }
+    @basket.select { |item| item.each { |name, _| @basket.delete item if name == dessert } }
+  end
+
+  def empty_basket
+    puts 'Your basket is now empty'
```
dearshrewdwit: I'd like to see a separation between the view and the business logic

lib/takeaway.rb
```diff
@@ -0,0 +1,81 @@
...
+  def finalize_order
+    puts 'Your order has been noted and is now being processed. See you later!'
+    send_text('Thank you for your order! Your order should arrive in less than an hour.')
+  end
+
+  def see_basket
+    @basket.each { |item| item.each { |name, _|
+      puts name
+    } }
+    puts "Your total is: £#{@total}"
```
dearshrewdwit: these two lines indicate a confusion between view and business logic

lib/takeaway.rb
```diff
@@ -0,0 +1,81 @@
...
+    return "total: £#{@total.round(2)}"
+  end
+
+  def send_text(message)
```
dearshrewdwit: I'd like to see this extracted to a separate class that has the behaviour of creating and sending sms's.

spec/dish_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'dish'
+describe Dish do
+  subject(:dish) { described_class.new(name, price, allergy_info) }
+  let(:name) { "spring rolls" }
+  let(:price) { 3.99 }
+  let(:allergy_info) { "No specifications regarding allergies" }
+  describe '#name' do
+    context 'name of the dish'
+    it 'returns name' do
+      # dish = Dish.new("spring rolls")
```
dearshrewdwit: AGHHHHH COMMMENTS IN THE CODE

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,94 @@
+require 'takeaway'
+describe TakeAway do
+  # TakeAway.stub(:gets).and_return(1)
```
dearshrewdwit: 😢

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  let(:take_away) { described_class.new }
+  let(:menu) { double 'fake menu', read_menu: [{ "Chocolate Melt" => 7.99 }, { "Apple Crumble" => 6.99 }, { "Banana Split" => 6.99 }] }
+  # let(:message) { double 'fake message', send_text: 'Thank you for your order! Your order should arrive in less than an hour.' }
```
dearshrewdwit: 😱

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+      take_away.empty_basket
+      expect(take_away.total).to eq 0
+    end
+  end
+  describe '#finalize_order' do
+    before do
+      allow(take_away).to receive(:send_text)
+    end
+    it 'should send me a text that my order has been confirmed' do
+    # this is'nt really testing anything at this point..
```
dearshrewdwit: 😑 🔫

lib/takeaway.rb
```diff
@@ -0,0 +1,81 @@
...
+    item = @menu.dishes.select { |dish| dish.include?(dessert) }.pop
+    @replacement = []
+    @amount.to_i.times { @basket << item && @replacement << item }
+    @replacement.each { |item| item.each { |_, price| @total += price } }
+    response
+  end
+
+  def getting_amount
+    puts 'How many of these would you like?'
+    @amount = gets
```
dearshrewdwit: view/busines logic again

this should just be an argument to the method, which at some point you can create a layer on top of your program that puts/gets to and from the terminal and calls your methods in your program

lib/takeaway.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def response
+    int = @amount.to_i
+    int == 1 ? response = 'Item was added to basket!' : response = 'Items were added to basket!'
```
dearshrewdwit: this seems like you can find a way to do this better...

lib/takeaway.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  def add_to_basket(dessert)
+    raise 'Please pick a dessert from the menu.' unless in_menu?(dessert)
+    adding_to_total(dessert)
+  end
+
+  def remove_from_basket(dessert)
+    @basket.each { |item| item.each { |name, price| @total -= price if name == dessert } }
+    @basket.select { |item| item.each { |name, _| @basket.delete item if name == dessert } }
```
dearshrewdwit: the choice here is to delete all items in the array that are that item. Which causes the frustration of not being able to delete 1 at a time.

Either store dish and quantities using a hash, or if you're particularly wedded to the array perhaps use the index of the item to remove it.

PR #870, opened 2017-11-05
Takeaway Challenge - Lewis Youl

LewisYoul:


lib/takeaway.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'order'
+
+class Takeaway
+  attr_reader :order, :menu
+
+  DISHES = {
```
dearshrewdwit: I'd like this to be read in from elsewhere if possible. as much as I like the family deal, i'm not sure 1.66 sausages equals a pizza. So your pricing needs work.

(Classes should have 1 reason to change) At the moment if i want to change the dishes i have to change this class

lib/order.rb
```diff
@@ -0,0 +1,50 @@
...
+    add_to_cost(dish, quantity)
+  end
+
+  def breakdown
+    fail no_items if @cost.zero?
+    order_breakdown
+  end
+
+  def pay(amount)
+    fail no_items if @cost.zero?
```
dearshrewdwit: If you want to write a custom error try exploring:

```ruby
def your_method
  fail NoItemsError if cost.zero? # which is a bit bizarre that you look at cost to see if there are items, not... the items
...
end
# outside of the class
class NoItemsError < StandardErrror; end
```

https://ruby-doc.org/core-2.2.3/Exception.html

lib/order.rb
```diff
@@ -0,0 +1,50 @@
...
+    fail "Please pay £#{cost} to complete order" if amount != @cost
+    @text.send
+  end
+
+  private
+
+  def update_current(dish, quantity)
+    @current[dish.to_sym] += quantity
+    puts "#{quantity} #{dish} added to your order"
```
dearshrewdwit: This seems to be a bit of confusion between the view layer and the business logic.

lib/text.rb
```diff
@@ -0,0 +1,17 @@
+require 'twilio-ruby'
+
+class Text
+
+  def self.send
+    time = Time.now + 3600
+    account_sid = "ACc62846ea6cd0ccef6201d60d5669b8d3"
```
dearshrewdwit: Hide the deets!!!!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  describe "#basket" do
+    context "when viewing selected items" do
+      it "should return Order#breakdown" do
+        expect(with_order.basket).to eq("anything")
+      end
+    end
+  end
+
+  # describe "#checkout" do
```
dearshrewdwit: COMMENTED CODE 😱

spec/text_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'text'
+
+describe Text do
+
+  let(:defined_time) { described_class.new(4) }
+
+  describe "Instatiation" do
+    context "upon instantiation" do
+      it "should be an instance of class Text" do
+        expect(subject).to be_an_instance_of(Text)
```
dearshrewdwit: let's actually test it? 😂

PR #871, opened 2017-11-05
IPbianco*takeaway-challenge

IPbianco:


lib/takeaway.rb
```diff
@@ -0,0 +1,86 @@
+require_relative 'menu'
+require_relative 'dish'
+require_relative 'order'
+require 'twilio-ruby'
+
+class Takeaway
```
neoeno: Could you list out the responsibilities of `Takeaway` for me?

For instance: one is "present the menu to the user"

lib/takeaway.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  def send_sms
+    time = Time.new + 3600
+    client = Twilio::REST::Client.new ACCOUNT_SID, AUTH_TOKEN
+    message = client.messages.create(
+      body: 'The order will be delivered before ' + time.strftime("%I:%M %p"),
+      to: '+447516988508',
+      from: '+441704325004')
+    puts message.sid
+  end
```
neoeno: This method isn't tested, but should be.

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
...
+  def initialize(dish_class: Dish)
+    @dish_list = [
+      dish_class.new(:Double_trouble, 12.99),
+      dish_class.new(:Kiwilicious, 8.99),
+      dish_class.new(:Smarty_smarties, 8.49),
+      dish_class.new(:Lets_go_nuts, 4.99),
+      dish_class.new(:Chocolate_indulgence, 5.99),
+    ]
+  end
+end
```
neoeno: I like these dish classes and it's nice to see the use of dependency injection here too :)

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
+class Menu
+  attr_accessor :dish_list
```
neoeno: This could be made an `attr_reader`. Why might we prefer that?

lib/order.rb
```diff
@@ -0,0 +1,23 @@
...
+    @total += dish.price * quantity
+  end
+
+  def canceled
+    @status = :canceled
+  end
+
+  def payed
+    @status = :payed
+  end
```
neoeno: I really like the `add` method name, but I don't like `canceled` or `payed` — can you tell me why that is?

lib/menu.rb
IPbianco: I used an attr_accessor because my idea was to be able to create a new dish and push it to the menu dish list, but maybe it is better to use an attr_reader and include a method in the takeaway class, something like "update menu", that receives a dish as a parameter and then adds it, in that way the attribute can't be modified from outside.

lib/order.rb
IPbianco: Both methods are changing the attribute '@status', they represent an action and in consequence their names should start with a verb, so I guess that I should call them 'cancel' and 'pay'

lib/takeaway.rb
IPbianco: - Present the menu to the user
- Take an order from the user
- Present the checkout to the user
- Accept a payment from the user
- Send an sms to the user
- Accept a cancelation from the user

lib/takeaway.rb
IPbianco: I started including an '@client' in the takeaway class so that I could inject a double and then mocked the chain (client.messages.create.sid) but when I was testing in pry, every time I created a new Takeaway I was receiving loads of information of the client class and It was really annoying. 
I will change this, probably adding an SMS class, and I will include a test for it.

lib/menu.rb
neoeno: Absolutely correct — when designing class interfaces, we want to lock down what the user of the class can do as much as possible. The more control you have over what they can do with your state, the fewer possible situations the class can be in, the fewer bugs are possible in your code, and the fewer unit tests you need.

For example: in this case, the user of your class (who might not be you) might get the format of `:dish_list` wrong or set it to nil. By locking it down with an attr_reader and/or an `add_dish` method, you remove that possibility.

lib/takeaway.rb
neoeno: Looks like an accurate assessment to me — now read this: https://en.wikipedia.org/wiki/Single_responsibility_principle

lib/takeaway.rb
neoeno: :+1: Good decision

PR #873, opened 2017-11-05
Week2 Challenge Completed!

tabrza:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:fake_menu) { double(:fake_menu, list: [{name: "Chicken", quantity: 1, price: 4.5}]) }
+  let(:menu_class) { double(:menu_class, new: fake_menu) }
+  let(:fake_sum) { double(:fake_sum) }
+  let(:sum_class) { double(:sum_class, new: fake_sum) }
+  let(:fake_text) { double(:fake_text, message: true) }
+  let(:text_class) { double(:text_class, new: fake_text) }
+  subject(:takeaway) { described_class.new(menu_class: menu_class, sum_class: sum_class, text_class: text_class) }
+
+    describe '#initialize' do
+      it 'shows the menu' do
```
mdsomerfield: These are testing for state rather than behaviour - if you focus on what the class can do rather than what the instance variables are, then you will implicitly test the instance variables themselves.

For example, rather than testing that takeway_menu is equal to fake_menu, test that adding an item to the order gets the item from the menu using allow/expect.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+          expect(fake_menu).to receive(:list)
+          takeaway.menu_list
+        end
+      end
+    end
+
+    describe '#order' do
+      context 'when adding a dish to the basket' do
+        it 'adds a dish' do
+          expect { takeaway.order('Chicken') }.to change { takeaway.basket.count }
```
mdsomerfield: This is better - testing behaviour rather than state. Your assertion isn't quite testing the expected outcome however. When we add 'Chicken', we expect the dish in the basket to be 'Chicken'. We should test that the basket contains 'Chicken', not merely that the count has changed.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    describe '#total' do
+      it 'returns total' do
+        expect(fake_sum).to receive(:total)
+        takeaway.total
+      end
+    end
+
+    describe '#check_out' do
+      it 'empties basket' do
+        takeaway.order('Chicken')
```
mdsomerfield: This is a good test.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'takeaway'
+
+describe Takeaway do
+  
+  let(:fake_menu) { double(:fake_menu, list: [{name: "Chicken", quantity: 1, price: 4.5}]) }
```
mdsomerfield: Good use of fakes

PR #875, opened 2017-11-05
Gabriela's solution to takeaway-challenge

gabrielabud:


lib/dish.rb
```diff
@@ -0,0 +1,10 @@
+class Dish
+  
+  attr_accessor :name, :price
+
+  def initialize(name = nil, price = nil)
```
dearshrewdwit: It's a bit odd to default to nil? What was the idea here?

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+class Menu
+
+  attr_accessor :dish_list
+
+  def initialize(dish_list = [])
+    @dish_list = dish_list
+    load_file_with_dishes
+  end
+
+  def load_file_with_dishes
```
dearshrewdwit: At some point it would good to consider using a class that deals with parsing CSV data and returning ruby objects that you want, so that your menu class can use the behaviour but not be responsible for file reading and writing

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+    file = File.open("./lib/menu.csv", "r")
```
dearshrewdwit: eg: 

```ruby
def create_dishes
  dish_parser = DishParser.new
  @dish_list = dish_parser.process
end
```

lib/order.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'dish'
+
+class Order
+
+  attr_accessor :basket, :total, :status
+
+  def initialize
+    @basket = Hash.new { |hash, key| hash[key] = 0 }
```
dearshrewdwit: Is this the same as ?
```ruby
Hash.new(0)
```

lib/takeaway.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def checkout
+    @order.basket_summary
+    puts ""
+    puts "Total amount to pay #{@order.total}"
+  end
+
+  def complete(amount)
+    raise RuntimeError, 'Amount to pay does not match total' unless amount == @order.total
```
dearshrewdwit: raise your own custom error messages if you want to specify an error type (the default is RunTimeError i believe
eg:
```ruby
...
  raise TotalNotMatchingError if <condition>
...
class TotalNotMatchingError < StandardError; end
```

lib/takeaway.rb
```diff
@@ -0,0 +1,50 @@
...
+    @order.complete_order
+    @total_orders << @order
+    send_message
+  end
+
+  private
+  def send_message
```
dearshrewdwit: Let's extract this to its own class, hide the auth details with environment variables and test it.

test.rb
```diff
@@ -0,0 +1,20 @@
+require './lib/dish.rb'
```
dearshrewdwit: what's this?!

PR #877, opened 2017-11-06
Takeaway challenge - All users stories implemented

antoniobelmar:


lib/order.rb
```diff
@@ -0,0 +1,85 @@
...
+    @@orders.push(self)
+    @restaurant = restaurant
+    @total = 0
+    @items = []
+  end
+
+  def add_item(item, quantity = 1)
+    raise "This item doesn't exist at this restaurant." if get_item(item).nil?
+    increase_items(item, quantity)
+    increase_total(quantity)
```
mdsomerfield: I think there are two problems with this:

1) It's not clear what it does. The name mentions total but you pass it a quantity - it suggests that total is the total number of items, not the order total.

2) When I examine the code, I see that it is tightly coupled with increase_items i.e. it only works as expected if increase_items has been called immediately before.

lib/order.rb
```diff
@@ -0,0 +1,85 @@
...
+
+  def send_notification
+    SMS.new.send_sms
+  end
+
+  private
+
+  def complete_order(amount)
+    puts "Your order from #{@restaurant.name} is complete for £#{@total}."
+    puts "Your change is £#{amount - @total}." if check_change(amount)
```
mdsomerfield: This looks like presentation logic in a business logic class. I see this in a few other places as well (e.g. wherever 'puts' is used). Think about separation of concerns.

lib/order.rb
```diff
@@ -0,0 +1,85 @@
...
+    raise "Sorry, you are missing £#{@total - amount}." unless check_bill(amount)
+    complete_order(amount)
+    send_notification
+  end
+
+  def send_notification
+    SMS.new.send_sms
+  end
+
+  private
```
mdsomerfield: I like the fact that you have well named private methods encapsulating much of the internal logic. It makes it easy to understand what is going on.

lib/sms.rb
```diff
@@ -0,0 +1,22 @@
+require "twilio-ruby"
+require_relative "authentication"
```
mdsomerfield: I think this was checked in by mistake, I had to remove it to get the tests to run (good test coverage though).

spec/order_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    it "ensures a new order gets initialized with an id number" do
+      expect(subject.id).to be_kind_of(Numeric)
+    end
+
+    order1 = Order.new("restaurant1")
+    order2 = Order.new("restaurant2")
+    order3 = Order.new("restaurant3")
+    order4 = Order.new("restaurant4")
+
+    [order2.id - order1.id, order3.id - order2.id, order4.id - order3.id].each do |id_difference|
```
mdsomerfield: This is a little hard to read. Generally avoid doing things which are too 'neat' in a test - the emphasis should be on readability.

lib/order.rb
```diff
@@ -0,0 +1,85 @@
...
+class Order
+  attr_reader :id, :restaurant, :total, :items
+
+  @@orders = []
+
+  def initialize(restaurant)
+    @id = 1 + @@orders.count
+    @@orders.push(self)
+    @restaurant = restaurant
+    @total = 0
```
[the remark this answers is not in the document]
antoniobelmar: If you mean the @total variable then I do understand that. I didn't think of it at all but it is kind of useless.

lib/order.rb
antoniobelmar: 1 - Is this a case of renaming the variable you think? Or is there something deeper that makes this hard to understand?

2 - Is that a problem? Considering they are both private methods and I am calling them successively? In any case this would be removed by your awesome point of removing the @ total variable!

lib/order.rb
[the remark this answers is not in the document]
antoniobelmar: wow awesome! I was afraid there were too many private methods!

lib/sms.rb
antoniobelmar: Yes, this was my own file...I am not sure how to tell someone that they need to create their own file and require it here for the sms sending to work....what approach would you suggest?

PR #882, opened 2017-11-06
Submission to weekend challenge

somemarsupials:


lib/dialogue.rb
```diff
@@ -0,0 +1,39 @@
+# Dialogue provides formatted messages for responses to clients.
```
dearshrewdwit: Great idea, usually the pattern for message strings is to hold them in a yml file and require them in, something akin to this: https://www.dafyddcrosby.com/using-i18n-in-sinatra/

according to locales you can then start internationalising apps

app.rb
```diff
@@ -0,0 +1,28 @@
...
+  def takeaway(takeaway_class = TakeAway)
+    takeaway_class.new
+  end
+
+  def responder(responder_class = Twilio::TwiML::MessagingResponse)
+    responder_class
+  end
+
+  get '/' do
+    takeaway.print_menu.gsub("\n", "<br>")
```
dearshrewdwit: what response are you sending back to the client and why?

app.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'lib/takeaway'
+require 'sinatra/base'
+require 'twilio-ruby'
+
+class App < Sinatra::Base
+  def takeaway(takeaway_class = TakeAway)
```
dearshrewdwit: there's no need to create methods that have injected dependencies inside a sinatra controller, but not in an action. Potentially you could investigate using a before filter: http://www.sinatrarb.com/intro.html if you want to.

Instantiate classes when needed in the controller action.

app.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def responder(responder_class = Twilio::TwiML::MessagingResponse)
+    responder_class
+  end
+
+  get '/' do
+    takeaway.print_menu.gsub("\n", "<br>")
+  end
+
+  post '/sms' do
```
dearshrewdwit: I'd look at refactoring this action to keep it down to the slimmest possible.

- what is the point of it? keep only what is essential and hide away params fetching and perhaps even thinking of a different way to instantiate the responder.

lib/formats/itemisedformat.rb
```diff
@@ -0,0 +1,32 @@
+require './lib/formats/formatter'
+
+# ItemisedFormat is a formatting class. It is usually
+# passed to Order #format to get pretty printed orders.
+class ItemisedFormat
+  DEFAULT_LENGTH = 20
+  DEFAULT_COUNT_LENGTH = 3
+  
+  def initialize(length: DEFAULT_LENGTH, count_length: DEFAULT_COUNT_LENGTH)
+    @item = @count = nil
```
dearshrewdwit: assigning to nil, means you don't need to create these variables yet

lib/menu.rb
```diff
@@ -0,0 +1,33 @@
...
+require './lib/menuitem'
+require './lib/formats/menuformat'
+
+# Menu is a collections of MenuItems.
+class Menu
+  attr_reader :items
+  START = 100
+
+  class << self
+    def from_csv(path, menuitem_class = MenuItem, reader = CSV)
```
dearshrewdwit: a lot is going on here. What can you clarify or move around?

lib/menu.rb
```diff
@@ -0,0 +1,33 @@
...
+  def initialize(items)
+    @items = items
+  end
+
+  def get(index)
+    @items[index - START]
+  end
+
+  def format(formatter = MenuFormat.new, *args)
+    items.each_with_index.map do |item, index|
```
dearshrewdwit: `map.with_index`

PR #890, opened 2017-12-03
Takeaway challenge by Xin Wang

Xin00163:


Xin00163:
This is the Week2 takeaway challenge.

PR #909, opened 2018-01-14
Takeaway Challenge

GiadaSimonetti:


EdwardAndress:
Hi Giada, good effort.  I'm going to leave a few in-line comments.  Please feel free to drop me a line if you have any questions.

lib/sms.rb
```diff
@@ -0,0 +1,15 @@
...
+  auth_token = ENV['TWILIO_AUTH_TOKEN']
+
+  @client = Twilio::REST::Client.new(account_sid, auth_token)
+
+  @client.messages.create(
+    to: ENV['MY_PHONE_NUMBER'],
+    from: ENV['TWILIO_PHONE_NUMBER'],
+    body: "Thank you! Your order was placed and will be delivered within one hour."
+  )
+end
```
EdwardAndress: Using the Twilio API was the tricky part of this challenge.  I'd recommend that you take a look at the exemplar for this file: https://github.com/makersacademy/takeaway-exemplar/blob/master/lib/sms.rb

lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
...
+    @order = {}
+  end
+
+  def place_order(*order, total)
+    raise "The total is not correct and the order has not been placed!" if total != get_total(order)
+    @order = format_order(order)
+  end
+
+  def get_total(order)
+    format_order(order).map { |item, quantity| quantity * menu.dishes[item] }.inject(0, :+)
```
EdwardAndress: I find this a little hard to read because there's quite a lot going on in one line.  What do you think of this alternative?

```ruby
def get_total(order)
  formatted_order = format(order)
  sub_totals = formatted_order.map { |dish, quantity| quantity * price_for(dish) }
  sub_totals.sum
end

def price_for(dish)
  menu.dishes[dish]
end
```

What do you think?  Is it easier to read?  Can you see any way of making it even more readable?

lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  private
+
+  def format_order(order)
+    order.each_slice(2).to_a.to_h
```
EdwardAndress: Having to do this makes me think that maybe the order is not quite in the 'right' format to start with.  Maybe you could even have used an `Order` class.

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      "chop suey" => 4
+    }
+  end
+
+  describe "attributes" do
+
+    it "shows dishes and prices" do
+      expect(menu.dishes).to eq dishes
+
+    end
```
EdwardAndress: I understand that you're trying to ensure there are some default dishes but does it really matter what those dishes are?  Would you want this test top fail if / when the default dishes are changed?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(takeaway.menu).to eq menu
+    end
+  end
+
+  describe "#place_order" do
+
+    context "when correct total given" do
+      it "adds a single item to the order" do
+        allow(menu).to receive(:dishes).and_return({ "spring rolls" => 2 })
+        takeaway.place_order("spring rolls", 3, 6)
```
EdwardAndress: Refactoring how an order is represented would also make this test more readable (potentially).  At the moment there are two numbers being passed to `Takeaway#place_order` and I have to refer back to the code in order to work out what those numbers mean.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      end
+
+      it "adds multiple items to the order" do
+        allow(menu).to receive(:dishes).and_return({ "spring rolls" => 2, "dumplings" => 3  })
+        takeaway.place_order("spring rolls", 3, "dumplings", 1, 9)
+        expect(takeaway.order).to eq({ "spring rolls" => 3, "dumplings" => 1 })
+      end
+    end
+
+    context "when incorrect total given and does not place the order" do
```
EdwardAndress: Good use of `context` 👍

lib/sms.rb
```diff
@@ -0,0 +1,15 @@
+require "bundler"
+Bundler.require()
```
EdwardAndress: I'm not familiar with what's going on here 🤔 .  What does it achieve?

EdwardAndress:
I've focussed on things to improve.  Hopefully it's helpful :)

PR #910, opened 2018-01-14
Takeaway Challenge PR

lewmoore:
PR for Takeaway Challenge

### DIFF
--- a/lib/SMS.rb
+++ b/lib/SMS.rb
@@ -17,7 +17,7 @@ class SMS
     @client.messages.create(
     from: ENV['TWILIO_PHONE'],
     to: ENV['TWILIO_DESTINATION_PHONE'],
-    body: (msg))
+    body: msg)
   end
 
   def time_of_delivery
--- a/lib/SMS.rb
+++ b/lib/SMS.rb
@@ -1,0 +1,27 @@
+require 'twilio-ruby'
+
+class SMS
+
+  def deliver
+    initiate_msg(time_of_delivery)
+  end
+
+  private
+
+  def initiate_msg(msg)
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
+
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+
+    @client.messages.create(
+    from: ENV['TWILIO_PHONE'],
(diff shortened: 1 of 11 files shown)


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+describe Takeaway do
+  let(:menu) { double :menu }
+  let(:order) { double :order }
+  let(:sms) { double :sms, deliver: nil}
+  subject(:takeaway) { described_class.new(menu) }
+
+  it 'knows an order total' do
+    takeaway.order('Lobster', 2)
+    expect(takeaway.bill).to eq(22184)
```
mdsomerfield: Are we testing that the order can calculate the total, or are we testing that the price of two lobster is 22184? I think it's the former, because the second is state/configuration and we don't want to test that.

However, this test will fail if either the order fails to calculate the bill correctly or the price of lobster changes. That makes it a brittle test. We want to mock our dependencies so that we don't need to worry about what the price of lobster really is, just that the bill is calculated correctly.

(You've done this well in other tests)

lib/menu.rb
```diff
@@ -0,0 +1,26 @@
+class Menu
+
```
mdsomerfield: It feels like this class is taking responsibility for two different things - managing the menu and displaying the menu. This is a bit nitpicky, but I'm going to pick nits because actually you've done a good job with the rest of the object model. 

Your method 'list' introduces some display logic around name and price, but a cleaner pattern here would be to have a MenuView or MenuFormatter class that did this for you and have Menu just responsible for the business logic. Something like:

```
class MenuView
  def initialise(menu)
    @menu = menu
  end

  def list
    @menu.getDishes.map { |name, price| "#{name}: £#{price}" }
  end
end
```

It's probably not evident here what advantages that has - I imagine in feels like just another class and a fairly pointless distinction - but as you start to deal with more complex interfaces you'll start to see the benefit of treating your view rendering logic as a separate concern to your data. 

(This is a tip to remember when you start the bowling challenge after week 5).

lib/SMS.rb
```diff
@@ -0,0 +1,27 @@
+require 'twilio-ruby'
+
+class SMS
```
mdsomerfield: I think sending the message and defining the contents of the message are two different responsibilities. You've broken them into different methods (good), but I think that the second one doesn't really even belong in this class.

PR #917, opened 2018-01-14
Submitting partial solution

MissDove:


lib/order.rb
```diff
@@ -0,0 +1,34 @@
+class Order
+
+  attr_reader :basket, :current_total, :dish_list
+
+  def initialize(dish_list)
+    @dish_list = dish_list
+    @basket = Hash.new(0)
```
sjmog: Is it necessary to use a Hash here? I get that you can represent quantity nicely like this, but it feels simpler to use an array of dishes.

It might simplify the add to order and total price bits too.

lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def confirm_order
+    fail 'No orders have been added' if @order.nil?
+    @message.send_text if price_is_correct?
+  end
+
+  private
+
+  def price_is_correct?
+    @order.current_total == @order.total_price_of_basket
```
sjmog: I'm not sure I like this. It feels like you have two sources of truth for the basket price: one is calculated on-the-fly (`#total_price_of_basket`) and the other is stored on the object (`current_total`). What are the cases where they might not match?

spec/message_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+describe Message do
+
+  let(:message) { described_class.new }
+
+  describe '#send_text' do
+    it 'sends a text to the customer' do
+      allow(message).to receive(:send_text)
+    end
+  end
```
sjmog: I feel like you could pass in a double of `Twilio` here and do this in a more sophisticated way.

PR #929, opened 2018-02-04
Takeaway Weekend Challenge - Twilio Gem

Jestfer:


lib/take_away.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'order'
+
+class TakeAway
+  attr_reader :selection, :current_amount, :menu
+
+  def initialize
+    @menu = {
```
samjones1001: Setting the menu in this way isn't necessarily a problem, but it does make it more difficult to alter - if the takeaway wants to change it's menu, or if you want multiple restaurants which sell different items to use your app it could be an issue.

If you were to create a separate menu class and hold this hash of items in there,  you could inject it in the initialiser and make things a bit more flexible.

Something like:

     @menu = menu
end

spec/take_away_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    'Scrambled eggs' => 4.99,
+    'Roasted aubergine & tomato curry' => 5.60,
+    'Jacket Potato with whipped feta' => 8.25,
+    'Mushroom, ricotta & rocket tart' => 9.20,
+    'Cod with beans & artichokes' => 11.40,
+  }
+
+  describe "#menu output" do
+    it 'should output a list with dishes and prices' do
+      expect(subject.list).to eq default_menu
```
samjones1001: This test doesn't feel quite correct to me - looking at the method in the Takeaway class, it seems like it should be printing out a formatted version on the menu, whereas this test seems to be checking that it returns a hash.

You might want to check up on how to use rspec test for something being printed to the console.

spec/order_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'order'
+require 'take_away'
+
+describe Order do
+
+  let(:take_away) { TakeAway.new }
```
samjones1001: It's best to avoid relying on actual instances of classes other than the one under test and try to use doubles instead.

As it stands, if you were to continue working on this code base and accidentally introduced a bug into your takeaway class, it could prevent your tests here in the order class from running correctly, which could make bug fixing more difficult.

PR #939, opened 2018-02-04
Serena Hathi - takeaway challenge (week 2)

serenahathi:


lib/menu.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'takeaway'
+
+class Menu
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = { "Prawn Crackers" => 1.51, "Prawn Gyoza" => 3.51,
```
EdwardAndress: As you mentioned in your own review, it would be nice to read these in from a file.  You could also consider creating a MenuItem class (and injecting the dependency).  One advantage would be that you could interact with menu items in a more readable way.  For example...

```ruby
# you could have a method like this on your order class to get the total price
def total
  items.sum(&:price)
end
```

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+       "Chicken Gyoza" => 3.11, "Spring Rolls" => 3.15 }
+  end
+
+  def print
```
EdwardAndress: This would also be more readable if, by virtue of having a Dish / MenuItem class you could do things like:

```
@dishes.map do |dish|
  "#{dish.name} - #{dish.price}"
end
```

lib/order.rb
```diff
@@ -0,0 +1,65 @@
+require_relative 'menu'
+
+class Order
+
+  attr_reader :basket
+
+  def initialize
+    @basket = Hash.new(0)
```
EdwardAndress: I'd also assign @menu here and inject that dependency.

```ruby
  @menu = menu
end
```

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+    fail "This dish is not in your order" unless @basket.include?(dish)
+    fail "Please enter the correct number of dishes" if quantity > basket[dish]
+  end
+
+  def dish_quantity_nil?(dish)
+    @basket.delete(dish) if basket[dish].zero?
+  end
+
+  def dish_quantity_check(quantity)
+    fail "Please enter the correct number of dishes" if quantity <= 0
```
EdwardAndress: Nice idea to validate the quantity here. How about this as an alternative error message?  Does it clarify what the error is?

"Quantity must be greater than 0"

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+class Order
+
+  attr_reader :basket
+
+  def initialize
+    @basket = Hash.new(0)
+  end
+
+  def add(dish, quantity)
+    dish_quantity_check(quantity)
```
EdwardAndress: These helper methods (`dish_quantity_check` and `dish_in_menu?`) clean up the code base and communicate your intention nicely.

I also like the fact that you can increment the quantity of an item already in the order.

How about this as an alternative pattern?  It makes the `#add` method feel less procedural.

```ruby
def add(dish, quantity)
  @basket[dish] += quantity if valid_addition?(dish, quantity)
end

def valid_addition?(dish, quantity)
  valid_quantity?(quantity) && valid_dish?(dish)
end

def valid_quantity?(quantity)
  quantity > 0 ? true : fail 'Quantity must be greater than 0'
end

def valid_dish(dish)
  # assuming a menu of key-value pairs
  menu.keys.include?(dish) ? true : fail 'That dish is not on the menu'
end
```

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+    @menu = Menu.new
+    dish_in_menu?(dish)
+    @basket[dish] += quantity
+  end
+
```
EdwardAndress: Same for this method (`#remove`)

lib/text_message.rb
```diff
@@ -0,0 +1,30 @@
...
+class TextMessage
+
+  def send
+    send_text
+  end
+
+  private
+
+  def send_text
+    account_sid = ENV["TWILIO_ACCOUNT_SID"]
```
EdwardAndress: Nice that you avoided putting these sensitive details in GitHub!

lib/text_message.rb
```diff
@@ -0,0 +1,30 @@
...
+    auth_token = ENV["TWILIO_AUTH_TOKEN"]
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+
+    @client.messages.create(
+      from: ENV["TWILIO_NUMBER"],
+      to: ENV["MY_NUMBER"],
```
EdwardAndress: This will be the number of whoever places the order so, perhaps we need to be collecting their number when the order is created.

lib/text_message.rb
```diff
@@ -0,0 +1,30 @@
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load("data.env")
+
+class TextMessage
```
EdwardAndress: Good decision to make a separate class for this behaviour.

EdwardAndress:
I kept getting distracted by other things so I'll need to come back and finish this off tomorrow.

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  subject(:menu) { described_class.new }
+  let(:dishes) do
+    { "Prawn Crackers" => 1.51,
+      "Prawn Gyoza" => 3.51,
+      "Chicken Gyoza" => 3.11,
+      "Spring Rolls" => 3.15 }
+  end
+
+  let(:printed_menu) { "Prawn Crackers - £1.51, Prawn Gyoza - £3.51, Chicken Gyoza - £3.11, Spring Rolls - £3.15" }
+
```
EdwardAndress: This is not quite the right use of `context`.  An place where it would work in this challenge would be something like:

```ruby
describe '#dish_on_menu?' do
  context 'when the dish is not on the menu' do
    it 'raises an error' do
      <test>
    end
  end
end
```

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  context "when initialized" do
+    it "stores a menu with dishes and prices in a hash" do
+      expect(menu.dishes).to eq(dishes)
+    end
+  end
+
+  describe "#print_menu" do
```
EdwardAndress: V. minor comment: this should be `#print`, as per the method name.

spec/text_message_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'text_message'
```
EdwardAndress: It would be worth having another go at testing this at some point.  The practice will come in handy.

EdwardAndress:
Hi again! Generally this is a great effort.  You've separated out the code into classes into a sensible way and achieved good test coverage.  So, well done for those things and keep up the good work.

The main points to keep in mind from this are:
1. Dependency injection.
2. Practice stubbing the API call.
3. Consider separating out behaviour into more classes.  For example, a MenuItem class would have been helpful here.  Even though it would have been tiny.

PR #1045, opened 2018-08-19
takeaway challenge

leekarho:


lib/orders.rb
```diff
@@ -0,0 +1,34 @@
...
+
+class Orders
+
+  attr_reader :cust_orders, :menu
+
+  def initialize
+    @cust_orders = {}
+  end
+
+  def choose_menu(menu = Menu.new)
```
samjones1001: good use of dependency injection 👍

lib/sms.rb
```diff
@@ -0,0 +1,22 @@
+require 'twilio-ruby'
+
+class Sms
+
+  def initialize
+    account_sid = 'twilio acc sid'
+    auth_token = 'twilio auth token'
+    @client = Twilio::REST::Client.new account_sid, auth_token
```
samjones1001: A call to .new within a method like this is a bit of a red flag.  It means that you have a very hard dependency  on another class.  This is a problem for at least two reasons:

1) Your class can only ever work with the class it is depending on.  You have no flexibility and it would be very difficult to reuse your code in a different situation.
2) It's very difficult to properly isolate your tests.

The 2nd point is particularly important here - if your tests aren't isolated from using a real version of the Twilio client you'll be getting a text message every time you run RSpec.

spec/bill_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'bill'
+
+describe Bill do
+
+  # let(:menu) { double :menu, :menu_items[:pie] => 4.5, :menu_items[:cranberry_juice] => 2.50 }
```
samjones1001: This is a small thing, but try not to leave commented out code in anything you push to github.

spec/bill_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    # get_orders = { :pie => 2, :cranberry_juice => 2 }
+    cust_bill = Bill.new(get_orders, menu)
+    expect(cust_bill.bill).to eq 14
+  end
+
+  it 'settles the bill' do
+    # get_orders = { :pie => 2, :cranberry_juice => 2 }
+    cust_bill = Bill.new(get_orders, menu)
+    cust_bill.bill
+    expect { cust_bill.conf_message(0, conf) }.to raise_error "please settle the bill"
```
samjones1001: This test is a little unclear to me.  It looks as though your raising an error because there's nothing on the bill, but thats not what the error message says.

The naming of the test is also a little confusing - it doesn't seem to match up to what the test is actually doing.

spec/orders_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'orders'
+# require 'sms'
+
+describe Orders do
+
+  let(:menu) { double :menu, :menu_items => { :pie => 4.5, :chips => 2.50 } }
+  let(:mockbill) { double :mockbill, :bill => 14 }
+  let(:bill_klass) { double :bill_klass, :new => mockbill }
```
samjones1001: Some good use of mocking here 👍

spec/orders_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  it 'chooses the menu' do
+    expect(subject.choose_menu(menu)).to eq menu.menu_items
+  end
+
+  it 'responds to 2 arguments' do
+    expect(subject).to respond_to(:cust_order).with(2).arguments
+  end
+
+  it "can place multiple orders" do
+    subject.choose_menu
```
samjones1001: you haven't made use of your mock menu object here, so the test is not isolated - it relies on a real instance of the Menu class. Should be a quick and easy fix!

spec/orders_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  let(:menu) { double :menu, :menu_items => { :pie => 4.5, :chips => 2.50 } }
+  let(:mockbill) { double :mockbill, :bill => 14 }
+  let(:bill_klass) { double :bill_klass, :new => mockbill }
+
+  it 'chooses the menu' do
+    expect(subject.choose_menu(menu)).to eq menu.menu_items
+  end
+
+  it 'responds to 2 arguments' do
+    expect(subject).to respond_to(:cust_order).with(2).arguments
```
samjones1001: This style of `respond_to` test are a great way to get started with the TDD process, but one you write more in depth tests they become redundant and can be deleted.

spec/orders_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    subject.cust_order("pie", 2)
+    subject.cust_order("chips", 2)
+    expect(subject.order_conf).to eq({ :pie => 2, :chips => 2 })
+  end
+
+  it "returns the bill" do
+    subject.choose_menu
+    subject.cust_order("pie", 2)
+    subject.cust_order("chips", 2)
+    expect(subject.bill(bill_klass)).to eq 14
```
samjones1001: This test is probably a little more in depth that it actually needs to be.

Returning the total price is the responsibility of the `bill` method in the `Bill` class, the `bill` method in the `Order` class is just responsible for calling that - it doesn't know or care (and it shouldn't know or care) about what it actually does.

A better test here would be something like 
```ruby
it 'returns the bill' do
  expect(mockbill).to receive(:bill)
  subject.boll(bill_klass)
end
```

PR #1046, opened 2018-08-19
Dave Lawes - takeaway challenge

DaveLawes:


lib/order_checker.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def check_if_correct_total
+    given_total = @order_array[1]
+    compare_totals(given_total, calculate_total)
+  end
+
+  def calculate_total
+    total = 0
+    @order_array[0].each do |order|
+      total += @menu[order[0] - 1][:price] * order[1]
```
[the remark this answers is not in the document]
mdsomerfield: Also, [order[0] - 1] is confusing. I think there are two problems here:

1) You've offset the index of the order item from it's index in the array by one. If this has been done to make it easier for the user when placing the order, fine, but you should deal with the offset at the entry point for order. So if I ask for 10 of item 1, then in the place_order method I should translate that to 10 of item index 0 before passing the data on to other classes - user interface and business logic are two separate responsibilities and your business logic shouldn't be more opaque because of decisions that were made for user interaction. 

2) It's not clear from reading this code what order[0] even is in the first place. When making the switch above, why not just go the whole distance and substitute the input code for the corresponding items? So something like:

```
def place_order(order_input)
  order = build_order(order_input)
  checker = @order_checker.new(@menu, order)
  .. etc
end

def build_order(order_input)
  @order_array[0].map do {
    menu_index = order_line[0]
    item_quantity = order_line[1]
    [ :item => menu[menu_index], :quantity => item_quantity ]
  }.to_h
end
```

lib/order_manager.rb
```diff
@@ -0,0 +1,34 @@
...
+    @order_checker = order_checker
+    set_menu
+  end
+
+  def see_menu
+    @printer.to_string(@menu)
+  end
+
+  def place_order(order_array)
+    checker = @order_checker.new(@menu, order_array)
```
mdsomerfield: Conventionally, validation of a model is not a separate responsibility, because you are validating state that should be encapsulated - I would expect to generate an Order, and that order would have methods for validating it's own state.

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  context 'Allows a customer to place an order' do
+  # As a customer
+  # So that I can order the meal I want
+  # I would like to be able to select some number of several available dishes
+    context 'Allows customer to check the order total' do
+    # As a customer
+    # So that I can verify that my order is correct
+    # I would like to check that the total I have been given matches the sum of the various dishes in my order
```
mdsomerfield: No need to include the user story comments in the test - the context/describe should detail the feature being tested.

spec/order_checker_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'order_checker'
+
+describe OrderChecker do
+
+  menu = [
+      { dish: "Sea salt & caramel",      price: 3.5 },
+      { dish: "Chilli & chocolate",      price: 3.5 },
```
mdsomerfield: You probably don't need this much test data - unit tests should have just enough data to test the code.

spec/order_checker_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    context 'invalid order' do
+      it 'raises error if invalid order item is included in array' do
+        expect { invalid_order.check_order }.to raise_error("Invalid dish selection!")
+      end
+    end
+
+    context 'invalid total' do
+      it 'raises error if invalid total is provided' do
+        expect { invalid_total.check_order }.to raise_error("Incorrect order total provided!")
```
mdsomerfield: It's hard to understand these tests because your test setup is dissociated from your test. I think this is because you are injecting the order data in initialise, but it doesn't actually feel like that is state intrinsic to the class - the class takes the data, tells you whether it's valid or not, then stops caring, so it feels like you should be passing that into the method instead.

lib/order_checker.rb
```diff
@@ -0,0 +1,46 @@
...
+class OrderChecker
+
+  def initialize(menu, order_array, send_text = SendText.new)
+    @menu = menu
+    @order_array = order_array
+    @send_text = send_text
+  end
+
+  def check_order
+    check_if_valid
```
mdsomerfield: This is fine, but better still is 'validate_selection'/'validate_total' or 'assert_selection_exists'/'assert_total_valid'. The difference is that 'check' doesn't tell me what the consequence of a failed check is, whereas by convention 'validate' and 'assert' generally indicate that an exception will be raised.

PR #1047, opened 2018-08-19
Esam Al-Dabagh - Takeaway Challenge

EsamAl-Dabagh:


spec/orders_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it "will input dish and quantity to #current hash" do
+      order.choose_dish(dish, qty)
+      expect(order.current).to eq([[dish, qty, mockMenu.MENU[dish]]])
+    end
+  end
+
+  describe ".view_order" do 
+    it "will display current order along with total" do 
+      order.choose_dish(dish, qty)
+      expect { order.view_order }.to output("DISH || QUANTITY || PRICE\nDonner Kebab || 2 || 8\nTOTAL = £16\n").to_stdout
```
samjones1001: It's worth thinking about what the responsibilities of your view_order method actually are.  This feels more like a test for the .create method it calls in your bill class which is responsible for actually formatting and printing the order.

I would argue that the only responsibility of the view_order method is to call this .create method - your order class shouldn't know (and shouldn't care) what happens next.

A better test here would be something like 

```ruby
it 'tells the bill to print' do
  expect(bill).to receive(:create)
  order.view_order
end
```

In order to get this to work however, you'll need to think about isolating your tests properly so that you can directly access the bill in this way.

spec/send_sms_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'send-sms'
+
+describe SMS do
+
+  describe ".send_sms" do
+
+    it "will send a SMS to the cusomter's phone number" do
+      allow(subject).to receive(:send_sms).and_return("SMS sent")
```
samjones1001: Again, this isn't testing anything without an expectation.  I'm guessing it's also sending you a message every time you run your tests?

Getting this test to work properly is probably the most difficult part of this challenge - take a look at figuring out how you might do it if you want to put your dependency injection and mocking skills to the test

lib/send-sms.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class SMS
+
+  def initialize(calc_time = DeliveryTimeCalc.new)
+    @calc_time = calc_time
+  end
+
+  def send_sms
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
```
samjones1001: Good job on hiding your details 🙂

PR #1053, opened 2018-08-19
Makers week 2 Takeaway challenge pull request

Lucx14:
submitting my weekend attempt at the takeaway challenge

thanks


lib/menu.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  attr_reader :dishes
+
+
+  def initialize(dishes = { "pepperoni" => 7.50, "hawaiian" => 6.50, "vegetarian" => 5.50 })
+    @dishes = dishes
+  end
+
+
+
```
samjones1001: This is a very minor gripe, but you probably should't be leaving this much white space between methods - a single line will do 🙂

lib/sms.rb
```diff
@@ -0,0 +1,7 @@
+class SMS
```
samjones1001: I think your sense that this should be it's own class was a good one.  It would cut down on the responsibilities currently held in the Takeaway class, and have the added effect of making properly testing the functionality of sending an sms easier.

lib/takeaway.rb
```diff
@@ -0,0 +1,64 @@
...
+require_relative 'menu'
+require_relative 'sms'
+
+class Takeaway
+
+
+  attr_reader :menu, :order
+
+  def initialize(menu = Menu.new)
+    @menu = menu
```
samjones1001: You mention struggling with dependency injection in your reflection, but this is a perfect example of the pattern.

Essentially what you're doing here is saying 'when I create a Takeaway object, it should take one parameter called menu.  Whatever I pass in here, will then be stored in an instance variable called menu.  If I don't pass anything in, this parameter should default to a newly initialised Menu object.'

This gives you greater flexibility in what you pass into your Menu object, and also makes isolating your unit tests possible.

lib/takeaway.rb
```diff
@@ -0,0 +1,64 @@
...
+#    @sms = sms
+  end
+
+  def view_menu
+    menu.print_menu
+  end
+
+
+  # need to think about what place order will do at the moment it is responsible for sending the text!!!
+  def place_order
```
samjones1001: I actually think this method is probably ok.  If we wanted to develop the application further, we'd probably want to implement a way of sending on the completed order to the restaurant or something similar, but, for now, the spec only tells us to send a confirmation to the user when an order is placed which is exactly what you're doing.

lib/takeaway.rb
```diff
@@ -0,0 +1,64 @@
...
+  def order_totals
+    order.map do |dish, quantity|
+      menu.dishes[dish.downcase] * quantity
+    end
+  end
+
+  def account_id
+    account_sid = "ACe16a4eddf4b5e19a3331831d7df88b9b"
+    auth_token = "f6964d00315fe133190e213d9cc775f1"
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
```
samjones1001: As I said above, storing the logic for sending messages buried here within the Takeaway class is going to make testing it correctly very difficult.

This line in particular is a bit of a red flag - within a method you have a call to .new on a different class in this case a Twilio Client object.

Creating new instances of objects like this is going to prevent you from properly isolating your unit tests. A better approach would be to make use of dependency injection as you have above.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'takeaway'
+
+describe Takeaway do
+
+  subject(:takeaway) { described_class.new(menu) }
+  let(:menu) { double(:menu, dishes: { "pepperoni" => 7.50, "hawaiian" => 6.50, "vegetarian" => 5.50 }) }
```
samjones1001: Again, you mentioned isolating tests as a difficulty in your review, but this is a really nice implementation.

Because you used dependency injection in your production code, you have the option of passing in this menu double when you instantiate the instance of the takeaway class you use in your tests.  This removes any dependency on a real instance of the Menu.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# think about extracting Text/sms to its own class
+#  let(:sms) { double(:sms, deliver_sms: nil) }
+
+
+  describe "#view_menu" do
+    it "shows a shows a list of dishes with prices" do
+      allow(menu).to receive(:print_menu).and_return("Pepperoni £7.50")
+      expect(takeaway.view_menu).to eq("Pepperoni £7.50")
+    end
```
samjones1001: I would argue that this test goes a little deeper than it needs to.

Your view_menu method looks like this:
```ruby
def view_menu
  menu.print_menu
end
```

It's only responsibility is to send a print_menu message to whatever is held in the @menu instance variable.  It doesn't know or care what the will happen next - nor should it.

A better test here would be something like:
```ruby
it 'tells the menu to print' do
  expect(menu).to receive(:print_menu)
  takeaway.view_menu
end
```

Testing the actual output of doing this belongs in a test for print_menu within the Menu class.

PR #1054, opened 2018-08-19
Takeaway Challenge

rebeccasedgwick:


lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+require_relative "menu_item"
```
samjones1001: Right now, this class has multiple responsibilities:
- it creates a list of menu items by scanning a csv
- it formats those items for display

You might want to consider extracting a Printer class or something similar

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+  attr_reader :m_items
+
+  def initialize
+    @m_items = {}
+  end
+
+  def load_from_csv
+    @m_items = {}
+    CSV.foreach(DEFAULT_FILENAME, headers: true) do |row|
+      @m_items[row["number"]] = MenuItem.new(number: row["number"], dish: row["dish"], price: row["price"])
```
samjones1001: Seeing a call to .new within a method is a bit of a red flag - you're creating a very hard dependency on your MenuItem class.

What this means is that the Menu class can only ever operate in conjunction with MenuItem, and therefore probably knows more than it should.  You could use dependency injection here to loosen this coupling.

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+  DEFAULT_FILENAME = "./lib/takeaway_menu.csv"
+  attr_reader :m_items
+
+  def initialize
+    @m_items = {}
+  end
+
+  def load_from_csv
+    @m_items = {}
+    CSV.foreach(DEFAULT_FILENAME, headers: true) do |row|
```
samjones1001: passing DEFAULT_FILENAME in here has made this method trickier to test than it needs to be.

I'd consider passing it into the method instead:
``` ruby
def load_from_csv(file = DEFAULT_FILENAME)
  #do some stuff
end
```

lib/messager.rb
```diff
@@ -0,0 +1,18 @@
+require "dotenv/load"
```
samjones1001: No tests for this class! 😱 

Testing Twilio properly is probably the hardest part of this whole challenge - it really tests your skills with mocking and stubbing.  I think Sophie went over a possible implementation on Monday?  It might be worth going back to this now and seeing if you can get it working if you're interested?

lib/messager.rb
```diff
@@ -0,0 +1,18 @@
...
+require "twilio-ruby"
+
+class Messager
+  def self.send_text(text_message)
+    account_sid = ENV["TWILIO_ACCOUNT_SID"]
+    auth_token = ENV["TWILIO_AUTH_TOKEN"]
+    twilio_phone = ENV["TWILIO_PHONE_NUMBER"]
+    customer_phone = ENV["CUSTOMER_PHONE_NUMBER"]
```
samjones1001: good job on hiding your details!

lib/order.rb
```diff
@@ -0,0 +1,25 @@
...
+
+class Order
+  attr_reader :o_items
+
+  def initialize
+    @o_items = []
+  end
+
+  def create_order_item(menu_item:, quantity:)
+    @o_items << OrderItem.new(menu_item.number, menu_item.dish, menu_item.price, quantity)
```
samjones1001: Again, this inline call to .new creates a hard dependency, and will make it very difficult to properly isolate your unit tests.

PR #1058, opened 2018-08-20
Takeaway challenge

SecretSurfSpot:
incomplete takeaway challenge, ready for code review

### DIFF
--- a/lib/confirm_order.rb
+++ b/lib/confirm_order.rb
@@ -1,0 +1,20 @@
+require 'twilio-ruby'
+
+class ConfirmOrder
+  
+  def confirm_order
+    # Your Account SID from www.twilio.com/console
+    account_sid = "AC0669ddc0d4125dde1e1f39d69ae96c44"
+    # Your Auth Token from www.twilio.com/console
+    auth_token = "48bf7fd23a639672b3693db1233ba5c4"
+    time_order_will_arrive = Time.now + 2700
+
+    @client = Twilio::REST::Client.new account_sid, auth_token
+    message = @client.messages.create(
+        body: "Thank you! Your order was placed and will be delivered before
+        #{time_order_will_arrive.hour}:#{time_order_will_arrive.min}",
+        to: "+447787172322",    # Replace with your phone number
+        from: "+447449763947")  # Replace with your Twilio number
+  end
+end
+#puts message.sid
--- a/lib/feature_test.rb
+++ b/lib/feature_test.rb
@@ -1,0 +1,11 @@
+#require_relative "menu"
+#require_relative "show_menu"
+require_relative "order"
+require_relative "confirm_order"
+
+#m = Menu.new
+o = Order.new
+o.take_order
+o.display_order
+co = ConfirmOrder.new
+co.confirm_order
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,28 @@
+require_relative "show_menu"
+
+class Order
+
+  attr_reader :choice_array
+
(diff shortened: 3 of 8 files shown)


lib/confirm_order.rb
```diff
@@ -0,0 +1,20 @@
+require 'twilio-ruby'
+
+class ConfirmOrder
+  
+  def confirm_order
+    # Your Account SID from www.twilio.com/console
+    account_sid = "AC0669ddc0d4125dde1e1f39d69ae96c44"
```
mdsomerfield: This should be a constant, defined outside the method, either as a class variable or as a public variable in a separate 'config' class.

lib/feature_test.rb
```diff
@@ -0,0 +1,11 @@
+#require_relative "menu"
+#require_relative "show_menu"
+require_relative "order"
+require_relative "confirm_order"
+
+#m = Menu.new
+o = Order.new
```
mdsomerfield: A feature test should be run the same way unit tests are - through rspec and with assertions. You're not asserting anything here so the only thing it's really testing is that the program runs without errors.

With feature tests, you should have one test per user story and an assertion that matches the outcome of the user story. For example, for the story:
```
As a customer
```
I would expect to see a test like:

```
describe "Feature" do
  describe "Show menu"
    ... setup
    expect(app.show_menu).to eq("... expected menu output ...")
  end
end
```

lib/show_menu.rb
```diff
@@ -0,0 +1,11 @@
+require_relative "menu"
+
+class ShowMenu
```
mdsomerfield: I'm confused as to why this is a class, and why you have behaviour (menu.display_menu) in your initialize. Showing the menu is a behaviour so I would expect to see a show_menu method on another class, rather than a ShowMenu class itself. 

You might (and probably should) want to separate the concerns of storing the menu data and displaying the data, in which case you could have a Menu class, which has a list of items and a get_menu method, and a MenuPrinter class, which has a show_menu class, which (a) gets the menu from menu, (b) formats it, and (c) printers it to the terminal.

The key here is that each class name is a noun and each method name is a verb, and the responsibilities are clearly defined - Menu is concerned with the data and MenuPrinter is concerned with display.

lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+    puts "Once you've chosen all your desired dishes, end 'exit'"
+    while true do
+      choice = STDIN.gets.chomp
+        if choice == 'exit'
+          break
+        end
+        @choice_array << choice
+    end
+
+    def display_order
```
mdsomerfield: Why is this defined inside your take_order method?

spec/menu_spec.rb
```diff
@@ -0,0 +1,7 @@
+require "menu"
+
+describe Menu do
+  describe "#displaymenu" do
+    it { should_not be_kind_of(String) }
```
mdsomerfield: This doesn't feel like a useful test. Firstly, it's testing state not behaviour. Secondly, asserting that it's not a string kind of means it could be anything else, which doesn't feel like what we want to test. 

Same goes for other tests below.

spec/order_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "order"
+
+describe Order do
+    describe "#take_order" do
+      it "exists" do
+        expect(subject).to respond_to(:take_order)
```
mdsomerfield: When testing, if we test what take_order does, we implicitly test that take_order is a valid method. So there's no need to write this kind of test.

PR #1061, opened 2018-09-16
Week 2 Weekend Challenge

cristinaocanamanzano:
This is my proposed solution for the week 2 weekend challenge: Takeaway


lib/menu.rb
```diff
@@ -0,0 +1,26 @@
...
+      pasta: 5,
+      pizza: 5,
+      lasagna: 6,
+      salad: 4,
+      coke: 2
+    }
+  end
+
+  def display_menu
+    menu_list.each_pair { |dish, price| p "Dish: #{dish}, price: £#{price}" }
```
samjones1001: Having display logic like this in the same class as other, heavier business logic is generally a good sign that you need to create another object - maybe think about a MenuPrinter?

lib/sms_sender.rb
```diff
@@ -0,0 +1,32 @@
+require 'twilio-ruby'
+require 'time'
+require 'dotenv'
+
+Dotenv.load 'config.env'
+
+class SmsSender
```
samjones1001: I notice that this class doesn't have any tests 😱 

I'm guessing this is because running the tests was sending you an sms every time? Did Sophie discuss how to properly isolate tests for Twilio in your code review session on Monday? If so, you might want to have a go at implementing this.

spec/take_away_spec.rb
```diff
@@ -0,0 +1,106 @@
+require './lib/take_away'
+describe TakeAway do
+  let(:menu) { double :menu, display_menu: nil }
+  let(:dish) { double :dish }
+  let(:dish2) { double :dish2 }
+  let(:price) { double :price }
+  let(:sms_sender) { double :sms_sender, text_confirmation: true }
+  subject(:takeaway) { described_class.new(menu, sms_sender) }
```
samjones1001: Use of doubles here looks accurate 👍

spec/take_away_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    end
+    it 'creates a menu' do
+      expect(takeaway.menu).to eq menu
+    end
+  end
+
+  describe '#display_menu' do
+    it 'calls method display menu on menu instance' do
+      expect(takeaway.menu).to respond_to(:display_menu)
+      takeaway.display_menu
```
samjones1001: this is very close to the correct test, but the expectation probably shouldn't be for the menu to respond_to display_menu - we know this is true because you've set up the double to respond to it up above.

Instead you should think about using to receive:

```ruby
 it 'calls method display menu on menu instance' do
      expect(takeaway.menu).to receive(:display_menu)
      takeaway.display_menu
end
```

PR #1062, opened 2018-09-16
Week 2 Challenge: Takeaway

darciew:


README.md
```diff
@@ -1,3 +1,7 @@
+Week 2 - Weekend Challenge
```
soph-g: Good README - clear how to use the application
It would be good to include some more information about how approach, and the design of the application in a diagram.

lib/confirmation.rb
```diff
@@ -0,0 +1,22 @@
+require 'twilio-ruby'
+
+class Confirmation
+
+  def send_text
+    account_sid = 'AC4bcca295ad97a98cacde7dc983f7d14a'
```
soph-g: Strongly advise removing this data from GitHub, and using environment variables instead to store it locally and access it when running the code.

lib/confirmation.rb
```diff
@@ -0,0 +1,22 @@
+require 'twilio-ruby'
+
+class Confirmation
+
+  def send_text
```
soph-g: Could this method be simplified and refactored into smaller methods each with a single purpose?

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+  attr_reader :basket, :total, :menu, :price_count
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @basket = []
+    @total = 0
+    @price_count = []
+  end
+
+  def add(dish)
```
soph-g: This method is currently doing two things - how could you refactor it to give it only one responsibility?

lib/order.rb
```diff
@@ -0,0 +1,43 @@
+require './lib/menu'
+require './lib/confirmation'
+
+class Order
+
+  attr_reader :basket, :total, :menu, :price_count
+
+  def initialize(menu = Menu.new)
```
soph-g: Good use of dependency injection

lib/order.rb
```diff
@@ -0,0 +1,43 @@
+require './lib/menu'
+require './lib/confirmation'
+
+class Order
+
+  attr_reader :basket, :total, :menu, :price_count
```
soph-g: Does basket need to be publicly accessible, when there is a view basket method?

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+  def view_basket
+    @basket
+  end
+
+  def view_total
+    @total = @price_count.inject(0, :+)
+    "£#{total}"
+  end
+
+  def place_order(confirmation = Confirmation.new)
```
soph-g: Why has the Confirmation object been passed as an argument, rather than injected into the class?

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+    confirmation.send_text
+    "Order has been placed. It will arrive within 1 hour"
+  end
+
+  # private
```
soph-g: private keyword shouldn't be commented out

lib/takeaway.rb
```diff
@@ -0,0 +1,15 @@
+require './lib/menu'
+
+class Takeaway
```
soph-g: What is the responsibility of this class?

spec/confirmation_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'confirmation'
+
+describe Confirmation do
+  subject(:confirmation)	{ described_class.new }
+	let(:send_text)	        { double(:send_text) }
+
+  describe 'confirmation message' do
```
soph-g: The send_text method has been stubbed in this test, so the test isn't testing anything! Instead, think about what calling send_text should do, and focus on testing that behaviour.

spec/order_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  # let(:menu_double)       { double :menu }
+  let(:order)               { described_class.new }
+  let(:dish_choice)         { double :dish }
+  let(:send_text)           { double :send_text }
+  let(:confirmation_mock)   { double :confirmation }
+
+
+#-----------beginning of order-----------#
+
+  describe 'a new order' do
```
soph-g: Grouping tests by the method under test will make them easier to read.

spec/order_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'order'
+
+describe Order do
+  # let(:menu_double)       { double :menu }
```
soph-g: Would be great to make use of the dependency injection you have set up in Order to use the Menu double. At the moment your tests aren't isolated. Please make use of the coaching team if you're not sure how to do this :)

spec/order_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      expect(order.basket).to be_empty
+    end
+
+    it 'starts with a total of zero' do
+      expect(order.total).to eq 0
+    end
+  end
+
+  describe 'creating an order' do
+    # before(:each) do
```
soph-g: Remove commented out code

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'takeaway'
+
+describe Takeaway do
+  let(:menu_class) { double :menu_class, :display_menu => "menu" }
+
+  describe 'view menu' do
+    it 'asks menu class to display the menu' do
+      takeaway = Takeaway.new(menu_class)
```
soph-g: Good use of the double

PR #1064, opened 2018-09-16
Takeaway Challenge

FlorenceHapgood:


README.md
```diff
@@ -14,66 +14,45 @@ Takeaway Challenge
 
  ```
 
-Instructions
+Approach
```
soph-g: Good README, and your commits are generally small with clear commit messages.

README.md
```diff
@@ -14,66 +14,45 @@ Takeaway Challenge
...
-As a customer
-So that I can order the meal I want
-I would like to be able to select some number of several available dishes
+To order:
+--------
 
-As a customer
-So that I can verify that my order is correct
-I would like to check that the total I have been given matches the sum of the various dishes in my order
+In irb:
```
soph-g: Might want to find out more about formatting in markdown, so you can clearly indicate ```code snippets``` using backticks.

lib/calculator.rb
```diff
@@ -0,0 +1,11 @@
+class Calculator
+
+  def calculate_total(something_to_calculate)
+    sum = 0
+    something_to_calculate.each { |hash|
+      sum += hash[:Price].to_i
```
soph-g: Could this method be even simpler?
Why are numbers being stored as Strings rather than Integers?

lib/menu.rb
```diff
@@ -0,0 +1,21 @@
+require_relative "printer"
+require_relative "order"
+
+class Menu
+
+  attr_reader :food_options, :printer
```
soph-g: Do food_options and printer need to be publicly available on the Menu? Think these could be private attributes

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+require_relative "printer"
+require_relative "text"
+require_relative "calculator"
+
+class Order
+
+  attr_reader :complete_order, :food_options, :calculator,
+  :total, :print_class, :text_class
+
+  def initialize(food_options = Menu.new.food_options,
```
soph-g: Lots of variables and classes being passed in - could this be simplified?

lib/printer.rb
```diff
@@ -0,0 +1,18 @@
...
+class Printer
+
+  def format(something_to_print)
+    something_to_print.each { |hash|
+      puts "#{hash[:Dish]}: £#{hash[:Price]}"
+    }
+  end
+
+  def print_bill(complete_order, total_due)
+    complete_order.each { |hash|
```
soph-g: Duplication of the format method here - I think you could have called `format` before then printing the total for the same output

lib/text.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
+
+class Text
+
+  def confirmation
+
+    account_sid = 'ACd7bd438097782e4c79a158fb3500290e'
```
soph-g: No use of ENV (Environment Variables). I would look at removing this information from GitHub and using Environment Variables instead. Try asking your cohort to see if anyone has an example of doing this.

lib/text.rb
```diff
@@ -0,0 +1,21 @@
...
+    auth_token = 'ed0c45f4d04589b5cf5f1f488beeeb08'
+    client = Twilio::REST::Client.new(account_sid, auth_token)
+
+    from = '+447533011143'
+    to = '+447746406608'
+
+    client.messages.create(
+    from: from,
+    to: to,
+    body: "Thank you, your order has been placed and will arrive before 15:30!"
```
soph-g: Time has been hard-coded - try getting this to always show an hour after the order has been placed.

spec/calculator_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'calculator'
+
+describe Calculator do
+  let(:mock_array) { double :mock_array }
```
soph-g: This double isn't being used.

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "menu"
+
+describe Menu do
+
+  let(:printer_class) { double :printer_class, :format => 0 }
```
soph-g: Think about making the values returned by doubles meaningful, so you can check for the expected value to be returned. e.g. format should probably return a string, with menu items and prices.

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+      expect(menu.food_options).to eq([{ Dish: "Duck_Pancakes", Price: "12" },
+      { Dish: "Spring_roll", Price: "8" },
+      { Dish: "Noodles", Price: "5" }])
+    end
+  end
+
+  describe "#print" do
+    it "sends an order to print" do
+      menu = Menu.new(printer_class)
+      expect(menu.print).to eq(menu.printer.format(menu.food_options))
```
soph-g: Not clear what #print should return here.

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "menu"
+
+describe Menu do
```
soph-g: Good set up of doubles using dependency injection, and tests have been clearly ordered

spec/order_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    it "can add more than multiple quantites of an item to an order" do
+      order = Order.new
+      order.new_item("Duck_Pancakes", 2)
+      expect(order.complete_order).to eq([{ Dish: "Duck_Pancakes", Price: "12" }, { Dish: "Duck_Pancakes", Price: "12" }])
+    end
+  end
+
+  describe "#total" do
+
+    let(:food_options) { double :food_options }
```
soph-g: Think about fully isolating your tests by setting up the test object to us doubles at the start of a test, rather than within individual tests.

spec/order_spec.rb
```diff
@@ -0,0 +1,72 @@
+require "order"
+
+describe Order do
+  describe "#new_item" do
+
+    it "finds the ordered item from the menu and saves it into complete_order" do
+      order = Order.new
+      order.new_item("Duck_Pancakes", 1)
```
soph-g: What happens if I order something not on the menu?

spec/order_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  describe "#print_order" do
+
+    let(:food_options) { double :food_options }
+    let(:text_class) { double :text_class, :confirmation => 0 }
+    let(:mock_printer) { double :mock_printer, :print_bill => 0 }
+    let(:mock_calculator) { double :mock_calculator, :calculate_total => "something" }
+
+    it "sends an order and bill to the printer" do
+      order = Order.new(food_options, mock_calculator, mock_printer, text_class)
+      expect(order.print_order).to eq(order.print_class.print_bill(order.complete_order, order.total))
```
soph-g: It's not clear what this method should return - would be better to specify the expected value.

lib/order.rb
FlorenceHapgood: For sure, although the only way that I could see to simplify this would be to break up my order class into two different classes. I guess at the moment it's sort of the main class, which we talked about avoiding last week. Is that what you mean?

spec/menu_spec.rb
FlorenceHapgood: I'm a bit confused about this, because I thought the point was that it doesn't matter what is being returned at this point, I'm just checking that the method is being called. It's then in the printer class that I test what is being returned. Or have I got the wrong end of the stick?

spec/menu_spec.rb
FlorenceHapgood: Same question here as my previous point...

spec/order_spec.rb
FlorenceHapgood: Ah ok, you mean set it up under the very first describe, the one that describes the whole class?

PR #1065, opened 2018-09-16
takeaway challenge

m-rcd:


lib/order.rb
```diff
@@ -0,0 +1,37 @@
+class Order
+  attr_reader :summary, :menu
+
+  def initialize(menu)
+    @summary = Hash.new(0)
+    @menu = menu
```
samjones1001: You could use dependency injection here to improve the user's experience by removing the need to create a menu object separately.

something like
```ruby
  @summary = Hash.new(0)
  @menu = menu
end
```

PR #1066, opened 2018-09-16
Program works to specifications

AmyDe:
Problem I came across was testing in RSpec for text message functionality without actually sending message. Read some documents and VCR seems most appropriate, though after several hours could not figure out how to use VCR. Will find out on Monday at Makers. New README created (README2.md) outlining process. Program works incl. text functionality. Just no test written for this particular feature at present.

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,32 @@
+class Takeaway
+
+  attr_reader :basket, :menu, :total
+
+  def initialize
+    @menu = {
+      :Margherita => 4.99,
+      :Pepperoni => 6.99,
+      :"Four Cheese" => 5.99,
+      :Sausage => 6.99,
+    }
+    @basket = {} # item name and quantity
+    @sum_total = 0
+  end
+
+  def order(food, amount = 1)
+    raise "We do not sell that item" if !@menu.has_key?(food.to_sym)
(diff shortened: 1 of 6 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
+require 'twilio-ruby'
+
+class Takeaway
+
+  attr_reader :basket, :menu
+
```
EdwardAndress: What happens if you want to change the menu?  Or create several Takeaways with different menus?

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    @sum_total = 0
+    @time = (Time.now + 3600).strftime "%H:%M"
+  end
+
+  def order(food, amount = 1)
+    raise "We do not sell that item" unless @menu.has_key?(food.to_sym)
+    @basket[food.to_sym] = amount
+  end
+
+  def total
```
EdwardAndress: Try to avoid using single letter vars unless they're some well known mathematical constants, perhaps.  Here, you could change `k, v` for `item, quantity` or something like that.

EdwardAndress:
This is a good effort.  I think the thing to focus on, when you reflect on this exercise, is how you might have broken out some of the behaviour in `Takeaway` into other classes, and when you might have done that.

PR #1069, opened 2018-09-16
All green!

NathanQuayle:


lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+class Menu
+  attr_reader :items
+
+  def initialize
+    @items = {}
+  end
+
```
EdwardAndress: Since the menu key is the same as the item id, you could do this

```
def add(item)
  @items[item[:id]] = item
end
```

Though it feels a bit duplicative to have the key being the same as the ID.  I guess that as I read more of your code, I will understand why you chose to do this! :)

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+class Menu
+  attr_reader :items
+
```
EdwardAndress: Did you consider including the option to initialize a menu with a collection of items so as to avoid having to add them all individually?

lib/order.rb
```diff
@@ -0,0 +1,33 @@
+require 'twilio'
+
+class Order
+  include Twilio
+  attr_reader :current_order
+
+  def initialize
+    @current_order = Hash.new(0)
```
EdwardAndress: Nice idea to override the default return value.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def calculate
+    receipt = "Receipt"
+    total = 0
+
+    @current_order.each_pair do |item, quantity|
+      receipt << "\n#{item.id} x #{quantity}: £#{item.price * quantity}"
+      total += item.price * quantity
+    end
+    receipt << "\nTotal: £#{total}"
```
EdwardAndress: I think this line (25) will return the entire string, without needing the line underneath (26)

lib/twilio.rb
```diff
@@ -0,0 +1,15 @@
+require 'twilio-ruby'
+require_relative 'secrets'
+
```
EdwardAndress: Good decision to encapsulate this away from the other logic.

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  # As a customer
+  # So that I can check if I want to order something
+  # I would like to see a list of dishes with prices
+  it 'so user can check the menu, list dishes with prices' do
+    menu = Menu.new
+
+    menu.add(:pizza, Dish.new(:pizza, 2))
+
+    expect { menu.print_menu }.to_not raise_error
```
EdwardAndress: Is it sufficient to know that this method doesn't raise an error?

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+  # So that I am reassured that my order will be delivered on time
+  # I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
+  it 'so user can be sure there food is on the way, send text confirmation' do
+    menu = Menu.new
+    order = Order.new
+
+    menu.add(:pizza, Dish.new(:pizza, 2))
+    menu.add(:lasagne, Dish.new(:lasagne, 5))
+    menu.print_menu
+
```
EdwardAndress: Is it possible to order something that is not on the menu?

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
...
+  def initialize
+    @items = {}
+  end
+
+  def add(key, item)
+    @items[key] = item
+  end
+
+  def remove(key)
```
EdwardAndress: Since the key is not printed, in `#print_menu` is it's only use for deleting items?

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+    @current_order = Hash.new(0)
+  end
+
+  def add(order)
+    order.each_pair do |item, quantity|
+      @current_order[item] += quantity
+    end
+  end
+
+  def calculate
```
EdwardAndress: Is there a way you can make this method shorter?  Ideally 5 lines or less (including `def` and `end`).

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    menu = Menu.new
+    order = Order.new 
+
+    menu.add(:pizza, Dish.new(:pizza, 2))
+    menu.add(:lasagne, Dish.new(:lasagne, 5))
+    menu.print_menu
+
+    order.add(menu.items[:lasagne] => 1, menu.items[:pizza] => 2)
+
+    expect { order.calculate }.to_not raise_error
```
EdwardAndress: Is it sufficient to know that this method doesn't raise an error?

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    menu = Menu.new
+    order = Order.new
+
+    menu.add(:pizza, Dish.new(:pizza, 2))
+    menu.add(:lasagne, Dish.new(:lasagne, 5))
+    menu.print_menu
+
+    order.add(menu.items[:lasagne] => 1, menu.items[:pizza] => 2)
+    order.calculate
+    expect { order.deliver(Secrets::PERSONAL_NUMBER) }.to_not raise_error
```
EdwardAndress: Does this end up with a message being sent when you run the tests?

lib/order.rb
EdwardAndress: Although I don't quite understand how it helps here 🤔

EdwardAndress:
This is a good solution with scope for small improvements in various places.  See the comments in line against the code.

PR #1072, opened 2018-09-16
Takeaway Challenge completed

nazzkadri:


lib/menu.rb
```diff
@@ -0,0 +1,32 @@
+class Menu
+  attr_reader :menu
+  attr_accessor :ordered_dishes
+
+  def initialize
```
EdwardAndress: Is it possible to change the menu?

lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+    @menu = [
+      { dish: 'pasta', price: 10.00, category: 'main_nonveg_meal' },
+      { dish: 'chickenwings', price: 20.00, category: 'main_nonveg_meal' },
+      { dish: 'fries', price: 2.00, category: 'main_nonveg_meal' },
+      { dish: 'beefburger', price: 5.00, category: 'main_nonveg_meal' }
+    ]
+  end
+
```
EdwardAndress: What do you think about this as an alternative way to print menu items?  Is it clearer?  Does it still do what you want it to do?

```
@menu.each do |item|
  print "#{item[:dish]}, #{item[:price]}\n"
end
```

lib/takeaway.rb
```diff
@@ -0,0 +1,52 @@
...
+    @customer_order = []
+    @menu = menu
+    @total = 0.0
+    @sms = sms_obj || Sms.new
+  end
+
+  def show_menu
+    @menu.show_menu
+  end
+
```
EdwardAndress: I think there are one or two places where you create instance vars, for things which could be local vars.

For example, @customer_order could be a local var, I think, since you pass it into the `#verify_order` method as an argument.

Do you know how local variables and instance variables behave differently?

lib/takeaway.rb
```diff
@@ -0,0 +1,52 @@
...
+    if is_price_correct?
+      #'send sms to the customer'
+       @sms.send_message("Thank you for your order, it will be delivered before 18:55")
+    else
+      'Sum does not match the total sum of the dishes ordered.'
+    end
+  end
+
+  private
+  def is_price_correct?
```
EdwardAndress: Interesting.  Why is this random?

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      takeaway.show_menu() # menu{dishes => price}
+      expect(menu_obj.show_menu).to eq(menu_obj.menu)
+      takeaway.ready_to_order # ordered_dishes
+      expect(menu_obj).to receive(:select_dishes)
+      menu_obj.select_dishes
+      expect { takeaway.ready_to_order }.to change { takeaway.customer_order }
+      takeaway.customer_order.each { |hash| expect(hash.keys).to contain_exactly(*hash_keys) }
+      expect(takeaway).to receive(:verify_order).with(takeaway.customer_order)
+      takeaway.verify_order(takeaway.customer_order)
+      expect(takeaway.total).to eq(takeaway.customer_order.map { |item| item[:price] * item[:quantity] }.reduce(:+))
```
EdwardAndress: The way that you're selecting dishes means that you have to construct this more complicated test (line 54) where you don't actually know in advance what the total will be.  If you were able to select specific dishes you could do something more like this:
```
takeaway.place_order('pasta', 1)
takeaway.place_order('chicken wings', 2)
expect(takeaway.total).to eq 50
```

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      expect(menu_instance).to receive(:select_dishes)
+      menu_instance.select_dishes
+    end
+  end
+
+  describe '#verify_order' do
+    it 'it verifies the total matches the sum of the various dishes in my order' do
+      allow(sms_obj).to receive(:send_message).with("Thank you for your order, it will be delivered before 18:55")
+      takeaway = Takeaway.new(sms_obj)
+      menu_obj = Menu.new
```
EdwardAndress: Why doesn't this menu_obj get injected into `Takeaway.new`?

EdwardAndress:
This is a nice effort!  One or two things to focus your thoughts on:

1. Your decision to include random dish selection has ended up with some tests being complicated.  Whilst we shouldn't write code purely with the intention of making tests simple, I think that in this case, the complicated tests highlight something about the code which could be improved.

2. Be careful that you're creating and associating objects in the way you intend.  In at least one place (there's a comment there), an instance of `Menu` is created but is not injected into `Takeaway.new`.

PR #1074, opened 2018-09-16
Unable to figure out the mocking for the sms, otherwise all complete

TinyGobby:


lib/order.rb
```diff
@@ -0,0 +1,36 @@
+require 'dotenv/load'
+require_relative 'menu'
+require_relative 'sms'
+
+class Order
+  attr_reader :current_total
+
+  def initialize
```
EdwardAndress: Did you try injecting the Menu and Sms classes here?

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+  def add(list, customer_number)
+    add_up_given(list)
+    fail 'Incorrect total' if incorrect?(list[:total])
+    confirm_order(customer_number)
+  end
+
+  private
+
+  def add_up_given(list)
+    list.each do |dish, amount|
```
EdwardAndress: Have you come across `Enumerable#reduce` or `Enumerable#inject`? Neither are particularly readable (to me, at least!) but either could be used to turn this into a one-liner.

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+      @current_total += @menu.prices[dish].to_f * amount.to_f
+    end
+  end
+
+  def incorrect?(given_total)
```
EdwardAndress: Who calculates `given_total`?

lib/sms.rb
```diff
@@ -0,0 +1,21 @@
+class Sms
```
EdwardAndress: Good decision to create a new class here

lib/takeaway.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'menu'
+require_relative 'order'
+
+class Takeaway  
+  def initialize
```
EdwardAndress: Dependency injection could also be useful here!

lib/takeaway.rb
```diff
@@ -0,0 +1,31 @@
...
+    @menu = Menu.new
```
EdwardAndress: Why do both Takeaway and Order need an instance of Menu?  I assume that it's so Order can check that a given item is on the Menu but... it's another point of coupling in the program.

spec/order_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  let(:failing_order) { { sushi: 2, burger: 1, total: 3 } }
+  let(:number) { double(:customer_phone_number) }
+  
+  describe 'initialize' do
+    it 'creates a new order' do
+      expect(subject.current_total).to eq 0
+    end
+  end
+
+  describe '#add' do
```
EdwardAndress: Are these commented out to prevent texts from getting sent?

EdwardAndress:
Generally a good solution.  How would you deal with multiple Takeaways or changing Menus though?  At the moment you'd need to load the files, write some code to create the objects, then edit the files, load them again and create some more objects.  Dependency injection is the key to avoiding this pattern and making this code easier to change.  It would also allow you to avoid sending texts when you run the tests (because you could use mocking more easily).

lib/order.rb
TinyGobby: I've not been able to figure out `Enumerable#inject` (seems to be the right one as I'd need to pass a block) as I need to be able to filter out one of the key/value pairs passed by the hash. I did however change it to: 
`list.each { |dish, amount| @current_total += @menu.prices[dish].to_f * amount }`

lib/order.rb
TinyGobby: Thanks! I hadn't thought about that before.

PR #1075, opened 2018-09-16
fbl11 submission for takeaway-challenge

fbl11:


lib/sms_sender.rb
```diff
@@ -0,0 +1,32 @@
+require 'bundler'
+require 'dotenv'
+Dotenv.load
+
+Bundler.require()
+
```
EdwardAndress: Good decision to create a new class for this behaviour!

lib/takeaway.rb
```diff
@@ -0,0 +1,50 @@
+
+class Takeaway
+  require_relative './menu.rb'
+  require_relative './order.rb'
+  require_relative './sms_sender.rb'
+
+  attr_reader :menu, :order
+
```
EdwardAndress: Something doesn't feel quite right about the idea of initializing a new Order for each new Takeaway...  Do you need to create a new Takeaway for each order?  I suspect the answer to that question is in your code somewhere, and I'll find it soon!

EdwardAndress: Otherwise, it's good that you're injecting the dependencies here 👍

spec/features/features_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  # using this uses a real instance of SMSSender, sending real texts
+  let(:takeaway) { Takeaway.new(menu, order, sms_sender) }
+
+  # As a customer
+  # So that I can check if I want to order something
+  # I would like to see a list of dishes with prices
+    
+    # heredoc -> to define multiline strings
+    # used to compare against output from initialised menu
+  dummy_menu = <<-HEREDOC
```
EdwardAndress: Nice use of heredoc!

spec/features/features_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+1. Pizza: £8.00
+2. Burger: £9.00
+3. Chips: £4.50
+  HEREDOC
+
```
EdwardAndress: I don't quite get how the dummy menu defined above can be output by your Takeaway, without the dummy meny being injected at some point...

spec/sms_sender_spec.rb
```diff
@@ -0,0 +1,11 @@
+describe SMSSender do
+
```
EdwardAndress: I won't leave any comments here because it looks like you paused work on this (and I'm likely to end up stating the obvious).

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,107 @@
+require 'takeaway'
+
+describe Takeaway do
+  let(:menu) { double :menu }
+  let(:order) { double :order }
+  let(:sms_sender) { double :smssender }
+  subject(:takeaway) { described_class.new(menu, order, sms_sender) }
+ 
+  it 'contains a @menu that is an instance of Menu' do
+    ta = described_class.new()
```
EdwardAndress: I think that you chose not to use a double here because you are trying to test the default behaviour, is that correct?  Perhaps the thing to test is that whatever get's passed in as the first arg is assigned to @menu (if you want to test that assignment).  So you could test that `takeaway.menu` is equal to `menu`.

Having said that, people tend not to test the assignment of instance vars like this.  There will be some behaviour somewhere that relies on these vars, which will also test for their correct assignment.  It's worth doing a tiny bit of reading about this though as you're bound to find alternative opinions!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+  describe '#show_menu' do
+    it 'shows menu using menu.show_menu' do
+      default_return = "SHOULD PRINT THIS"
+      allow(menu).to receive(:show_menu).and_return(default_return)
+      expect(takeaway.read_menu).to eq(default_return)
+    end
+  end
+
+  describe '#order_item' do
```
EdwardAndress: There's some nice doubling here, which obviously required some thought.  I think, however, that you only need to ensure that the instance of Takeaway sends the correct messages (since managing the basket is the responsibility of Order).  The rest will be testing in other places.

EdwardAndress:
Nice solution overall Freya!  The thing to focus on here is 'delegation'.  Takeaway seems to micromanage Order a bit.

EdwardAndress:
Please use the feedback in the corresponding Diode project! :)

PR #1076, opened 2018-09-16
Takeaway challenge User stories completed.

shwetzpatil:


lib/menu.rb
```diff
@@ -0,0 +1,38 @@
...
+  def add_dish(dish) 
+    # dish is instance of dish class
+    # raise "add_dish expects an instance of Dish class" unless dish.is_a?(Dish)
+    @dishes.push(dish)
+  end
+
+  def display
+    @dishes
+  end
+
```
EdwardAndress: Does `#dish?` check that a specific dish exists on the menu?  If so, you'll probably find that there's a simpler way to do this, with just one `Array` method ;)

EdwardAndress: Also, perhaps an alternative name would make your intent clearer.  How about something like `#valid_dish?` or `#dish_present?` As you can see, naming is hard, but putting some time into getting it right is worthwhile.

lib/order.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'menu'
+require 'date'
+require 'twilio-ruby'
+
+class Order
+
+  attr_reader :final_order
+
```
EdwardAndress: Do you now know how to obfuscate these credentials?   If not, it's worth looking at how you can use environment variables.

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    @final_order = []
+  end
+
+  def add_order(dish_name, quantity)
+    fail "Sorry, #{dish_name} is not available"\
+     unless @menu.dish?(dish_name) 
+    dish_obj = @menu.get_dish_obj(dish_name)
+    @final_order.push({ dish: dish_obj, quantity: quantity })
+  end
+
```
EdwardAndress: How about something like this instead of lines 28 to 31?

```
item_totals = @final_order.map { | item | item[:dish].price * item[:quantity] }
grand_total = item_totals.sum
```

I think I would do something like that but I am curious to know whether you think my solution is clearer or not.  There might also be another solution using a single method.  In fact, I just found a way of doing that but it's not very clear:

`grand_total = @final_order.reduce(0) { |sum, item| sum + ( item[:dish].price )}`

What do you think about these options?

spec/menu_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      subject.add_dish(dish)
+      expect(subject.dishes).to include(dish)
+    end
+  end
+
+  describe '#has_dish?' do
+    it 'returns true if dish is available' do
+      allow(dish).to receive(:name).and_return("dish_name")
+      subject.add_dish(dish)
+      expect(subject.dish?("dish_name")).to be_truthy
```
EdwardAndress: What's the difference between being `true` and being `truthy`?

spec/order_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'menu'
+require 'order'
+
+describe Order do
+
+  let(:menu) { double :menu }
+  let(:dish) { double :dish }
+  # let(:menu) { object_double(Menu.new, :has_dish? => true) }
+
+  subject { described_class.new(menu) }
```
EdwardAndress: You get this (`subject`) 'for free' when you refer to the class under test as a constant `describe Order do`.  You only need to create the subject explicitly if you want to initialize it with args.

EdwardAndress:
Nice work, Shweta!  Be careful that you don't put your credentials on github though :)

PR #1080, opened 2018-09-16
ibra's attempt

IvyMic:


lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+    count.times do
+      @basket << {item.name.to_sym => item.price}
+    end
+  end
+  
+  def find_dish(item)
+    available_dishes = get_available_dishes
+    wanted_dish = available_dishes.select{ |dish| dish.name == item.to_sym }
+    raise "Error, dish not available" if wanted_dish.empty?
+    wanted_dish[0]
```
EdwardAndress: Is there another method you can use on line 24 that will return a single item rather than an array?  If so, you would no longer need to do `wanted_dish[0]`, which is a bit misleading.

lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def show_menu
+    menu_array = []
+    @menu.dishes.each { |dish| menu_array << {dish.name => dish.price}}
+    menu_array
+  end
+
+  def order_total
+    total = 0
```
EdwardAndress: I like the fact that you compute the total based on the order.  Is there a cleaner way to do it?  Would `@basket.values.sum` work?

lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+      hash.each { |k, v|
+        total += v
+      }
+    }
+    total
+  end
+
+  def complete_order
+    time = (Time.now + 3600).strftime("%H:%M")
+    @orders << {time: @basket}
```
EdwardAndress: What does is value of `@orders` here?  Is it what you expect / want?

lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+  def reset_basket
+    @basket = []
+  end
+
+  def send_order_message(number)
+    send_message(number)
+  end
+
+
+end
```
EdwardAndress: Why is there a little symbol at the end of this line?  What is it telling you?  Does it matter? 🤔

lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+    time = (Time.now + 3600).strftime("%H:%M")
+    @orders << {time: @basket}
+  end
+
+  def reset_basket
+    @basket = []
+  end
+
+  def send_order_message(number)
+    send_message(number)
```
EdwardAndress: Did I miss something?  Where is `send_message` defined?

EdwardAndress: Ah!  It's inside your module!

lib/twilio.rb
```diff
@@ -0,0 +1,16 @@
+require 'twilio-ruby'
+require_relative 'secrets'
+
+module Twilio
```
EdwardAndress: Purely out of curiosity: Why did you choose to use a module rather than a class?

spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "dish.rb"
+
+describe Dish do
+  
+  let(:name) { double(:name) }
```
EdwardAndress: Why are you using doubles here for name and price?

spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  it 'initializes a name with the argument given' do
+    expect(subject.name).to eql(name)
+  end
+  it 'initializes a price with the argument given' do
+    expect(subject.price).to eql(price)
+  end
+  it 'is initialized with available as true by default' do
+    expect(subject.available).to be true
+  end
+end
```
EdwardAndress: For next level testing, try to think about testing 'behaviour' rather than 'state'.

spec/menu_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "menu.rb"
+
+describe Menu do
+  let(:dish) { double(:dish) }
+  let(:fish_dish) { double(:fish_dish) }
+  before(:each){
```
EdwardAndress: This works perfectly well but there's a more idiomatic (and concise) way of achieving the same thing.

spec/menu_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    allow(fish_dish).to receive(:price).and_return(4)
+    allow(fish_dish).to receive(:available).and_return(true)
+  }
+  describe '#add_dish' do
+    it 'adds a dish to the dishes array' do
+      subject.add_dish(dish)
+      expect(subject.dishes).to include dish
+    end
+  end
+  # describe '#show_menu' do
```
EdwardAndress: Why is this commented out?  Does it fail?  What error message do you see?

EdwardAndress: Ah, I see the method is also commented out.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+    @dishes << dish
+  end
+
+  def remove_dish(dish)
+    @dishes.delete(dish)
+  end
+
+  # def show_menu
+  #   menu_arr = []
+  #   @dishes.each { |dish| menu_arr << { dish.name.to_sym => dish.price}}
```
EdwardAndress: Do you know the difference between `Array#map` and `Array#each`?

lib/price.rb
```diff
@@ -0,0 +1,6 @@
+class Price
```
EdwardAndress: This is interesting.  What was your thinking behind creating this class?

spec/menu_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  #     subject.add_dish(dish)
+  #     subject.add_dish(fish_dish)
+  #     expect(subject.show_menu).to eql([
+  #       {chicken: 2},
+  #       {fish: 4}])
+  #   end
+  # end
+
+  describe '#remove_dish' do
+    it 'removes a dish from the dishes array' do
```
EdwardAndress: removes a *specific* dish ?

spec/twilio_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "./lib/twilio.rb"
+
+describe Twilio do
+  describe '#send_message' do
+    it 'sends a message to number given' do
+      allow(subject).to receive(:send_message).with("123456").and_return("Your order was successful")
+      expect(subject.send_message("123456")).to eq("Your order was successful")
```
EdwardAndress: Can you make this test fail by changing the Twilio module?

PR #1083, opened 2018-09-17
Completed

Din-s1:
Coverage is just shy of 90% this is the result of private methods which though I can stub, mock and double would literally end up testing that the mock is mocking rather than the method's functionality.
I've completed the user stories and would continue this project by finding a better way around the private method testing, looking into VCR gemfile which may be the avenue to follow and also creating a user friendly and functional menu such as the one we created in the 'Students directory challenge'.
All in all still a lot to learn and some habits to continue moulding.

### DIFF
--- a/lib/basket.rb
+++ b/lib/basket.rb
@@ -20,7 +20,7 @@ class Basket
 
   def order_list
     @order.select do |items|
-      puts "#{items[:dish]}: £#{items[:price]}"
+      puts "#{items[:platter]}: £#{items[:price]}"
     end
     order
   end
--- a/lib/basket.rb
+++ b/lib/basket.rb
@@ -1,0 +1,36 @@
+require_relative 'takeaway'
+require_relative 'confirmation'
+class Basket
+
+  attr_reader :order, :sum_total, :new_order
+  
+  def initialize(new_order = CustOrder.new, text = ConfirmText.new)
+    @new_order = new_order
+    @text = text
+    @order = []
+    @sum_total = 0
+  end
+
+  def select_dish(platter)
+    @new_order.dishlection(platter).each do |item|
+      @order << item
+      @sum_total += item[:price]
+    end
+  end
+
+  def order_list
+    @order.select do |items|
+      puts "#{items[:dish]}: £#{items[:price]}"
+    end
+    order
+  end
+
+  def bill
+    @sum_total
+  end
+
+  def complete_order
+    @text.send_text
+  end
+
+end
--- a/lib/confirmation.rb
+++ b/lib/confirmation.rb
@@ -1,0 +1,28 @@
+require 'twilio-ruby'
+require_relative 'basket'
+
+class ConfirmText
+
+  def send_text
+    initiate_msg(conf_irm)
+  end
+
+  private
+
+  def initiate_msg(msg)
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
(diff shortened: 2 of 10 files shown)


lib/basket.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'takeaway'
+require_relative 'confirmation'
+class Basket
+
+  attr_reader :order, :sum_total, :new_order
+  
+  def initialize(new_order = CustOrder.new, text = ConfirmText.new)
```
EdwardAndress: Nice use of dependency injection.

lib/confirmation.rb
```diff
@@ -0,0 +1,28 @@
...
+class ConfirmText
+
+  def send_text
+    initiate_msg(conf_irm)
+  end
+
+  private
+
+  def initiate_msg(msg)
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
```
EdwardAndress: Nice obfuscation of your secrets details.

lib/confirmation.rb
```diff
@@ -0,0 +1,28 @@
+require 'twilio-ruby'
+require_relative 'basket'
+
+class ConfirmText
```
EdwardAndress: Good decision to create a separate class for this behaviour.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+describe CustOrder do
+  
+  describe '#menulist functions' do
+    it { is_expected.to respond_to(:dishlection) }
+    it { is_expected.to respond_to(:show_menu) }
+    it 'Should show menu' do
+      expect(subject.show_menu).to eq([])
+    end
+    it 'Should select correct dish' do
+      expect(subject.dishlection("Chicken Gyoza 5pcs")).to eq([{ :platter => "Chicken Gyoza 5pcs", :price => 4.5, :quantity => 0 }])
```
EdwardAndress: I'm a bit confused as to why you want this method to return a hash that includes `quantity => 0` which then never seems to get updated.

spec/confirmation_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'confirmation'
+describe ConfirmText do
+  it { is_expected.to respond_to(:send_text) }
+  context '#send_text' do
+    subject(:text) { described_class.new }
+    it "Should send txt" do
+      # expect(text).to receive(:send_text)
+      allow(text).to receive(:initiate_msg).with(anything)
```
EdwardAndress: What is `anything`?  Does this test pass?

lib/confirmation.rb
```diff
@@ -0,0 +1,28 @@
...
+  # set up a client to talk to the twilio REST API
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+  # send an sms
+    @client.messages.create(
+    from: ENV['TWILIO_PHONE'],
+    to: ENV['TWILIO_DESTINATION_PHONE'],
+    body: msg)
+  end
+
+  def conf_irm
```
EdwardAndress: If I understand correctly, your intention is to separate the message content from the action of sending it.  That's great! So you could extent this class to do something like `initiate_message(cancelled)` inside a method called `send_cancellation_notice` perhaps and include the content for that message in a method called `cancelled`.

Can you go one step further than this?

Is there a way you could encapsulate only the message content, somewhere else, and then use that here?  You could then do something like this `intiate_message(MESSAGES[:confirm])`.

What do you think?  What would that encapsulation look like?  Would it be another class?  Could it be something else?

PR #1085, opened 2018-09-17
Takeaway challenge - Alex Dresco

AlexDresco:


lib/dish.rb
```diff
@@ -0,0 +1,12 @@
+class Dish
+
+  attr_reader :name, :price, :item
+  def initialize(name, price)
+    @price = price
+    @name = name
+  end
+
```
EdwardAndress: Whats the difference between the method created using `attr_reader :item` and the one created using `def item`?

Do you need both?

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
+class Menu
+  attr_reader :name, :price
+
+  def initialize
+    @menu = []
+    @dish = Dish.new(@name, @price)
```
EdwardAndress: Is this instance variable used anywhere?

send_sms.rb
```diff
@@ -0,0 +1,15 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
```
EdwardAndress: This is a perfectly good procedural implementation of sending messages and there's definitely a place for procedural programs / scripts.  Could you refactor this to use a new class instead?

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'order'
```
EdwardAndress: Can you prevent people from ordering things that are not on the menu?

lib/dish.rb
AlexDresco: I've removed the attr_reader one as I'm setting up ` @item` to an hash. Then the tests are still passing.

lib/menu.rb
AlexDresco: Thanks for picking that up. It is something I don't fully understand in my code. 

The feature and unit tests are failing if I change it to @dish, but I don't understand how the method `item` works on the method variable 'item' here

```  
def add(dish)
    @menu << dish.item
  end
```

PR #1087, opened 2018-09-17
Takeaway challenge

DamoH:
User stories completed although more work needed - see my README file for further details.
Ran out of time as could have kept working on this!


lib/send_sms.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'Secrets'
+
+class SendSMS
+
+  def sms
+
+    require 'twilio-ruby'
```
EdwardAndress: Normally this line would be at the top of the class definition (not inside a method).  Any thoughts on why that might be?

DamoH: Yes totally... no excuses other than in a rush and not thinking it through (along with many other violations in my code that I recognise).

README.md
```diff
@@ -1,79 +1,66 @@
...
 
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
 
-Task
------
+Additional Notes
```
EdwardAndress: This is super helpful as it helps me to see that you're thinking about code quality. There will be time constraints on any project and I think you've taken a sensible approach here; starting out with one class and (planning to) abstract some behaviour into a new class / classes.  Deciding when exactly to make this abstraction will be influenced by a few things and getting it right will come with experience.

README.md
```diff
@@ -1,79 +1,66 @@
...
-* Write a Takeaway program with the following user stories:
+User Stories
+------------
 
-```
+User story 1
 As a customer
 So that I can check if I want to order something
 I would like to see a list of dishes with prices
+```
```
EdwardAndress: Nice that you're using feature tests! 👍

lib/basket.rb
```diff
@@ -0,0 +1,6 @@
+class Basket
```
EdwardAndress: Is this the start of your abstraction from the 'god class'?

lib/send_sms.rb
EdwardAndress: On a more positive note, it was a good call to abstract this behaviour into a separate class.

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'send_sms'
+
+class TakeAway
+
+  def initialize
+    @menu = { "chips" => 2.5, "kebab" => 5.5, "burger" => 4.75 }
```
EdwardAndress: What if you wanted to create a new restaurant with a different menu?

EdwardAndress: Or perhaps, several different restaurants, each with a different menu?

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
...
+  def display_menu
+    puts @menu
+  end
+
+  def select(food_item, qty = 1)
+    @basket << { food_item => qty }
+    "You have added #{food_item} x#{qty} portions"
+  end
+
+  def summary
```
EdwardAndress: Does the user ever get to see the total they're spending?

DamoH: I have just updated (and pushed) my summary method to show total spent so far. As a result I changed the @basket to a hash, rather than the array it was before. Feels better!

lib/basket.rb
DamoH: it was!

lib/takeaway.rb
DamoH: Yes, I would perhaps have done this a different (better) way with a separate menu class or the like.

PR #1089, opened 2018-09-17
Takeaway Challenge

ChloeHelen:


lib/menuFormatter.rb
```diff
@@ -0,0 +1,8 @@
+class MenuFormatter
```
EdwardAndress: Nice separation of concerns! 👍

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'menuFormatter.rb'
+require_relative 'order.rb'
+
+class Resturant
+
+  attr_reader :menu
+
+  def initialize
```
EdwardAndress: Is this code easy to change? What if you wanted to create a new restaurant with a different menu?  How would you implement that?

spec/order_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'order'
+
+describe Order do
+
+  let(:dish) { double :dish }
+  let(:price) { double :price }
+  let(:resturant) { double :resturant }
+  let(:menu) { [{ :dish => dish, :price => price }] }
+
```
EdwardAndress: Nice use of doubles! 👍

PR #1090, opened 2018-09-17
Week 2 takeaway without texts submission

racldn:


lib/order.rb
```diff
@@ -0,0 +1,38 @@
+require './lib/menu'
+
+
+class Order
+
+  attr_reader :basket
+
+  def initialize(menu = Menu.new)
```
EdwardAndress: Dependency injection! +1

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'order'
```
EdwardAndress: The tests here are not isolated from the Menu class.  I.e. Your `subject` is using a real instance of `Menu` rather than a double.  You're injecting and instance of `Menu` in your 'Order' class so you could, in the tests, inject a double instead.

Fancy making an update to that effect?

docs/user_story.md
```diff
@@ -0,0 +1,45 @@
+As a customer
```
EdwardAndress: Feature tests! 👍

PR #1091, opened 2018-09-21
Dire attempt at takeaway challenge

djstephenson7:


lib/menu.rb
```diff
@@ -0,0 +1,29 @@
+class Menu
+  attr_reader :menu
+
+  def initialize
+    @menu = menu
```
soph-g: Is this line required? Or, in other words, would your code behaviour change if you deleted the initialize method? If you're not sure, try it and see what happens

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+      { item: "Deep Fried Seaweed", price: 4.00 }
+    ]
+  end
+
+  def show_menu
+    @menu.collect.with_index { |k, i| "#{i + 1}. #{k[:item]}: £#{k[:price]}" }
+  end
+
+  def select
+    @menu = @menu[0, 2]
```
soph-g: Would this code let me select Deep Fried Seaweed?

spec/menu_spec.rb
```diff
@@ -0,0 +1,52 @@
+require './lib/menu'
+
+describe Menu do
+
+  describe '#menu' do
+    it "defines the method" do
+      expect(subject).to respond_to(:menu)
```
soph-g: Remove respond_to tests once you have a test for the actual functionality of a method. 
Good grouping of tests by method

spec/menu_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  describe '#show_menu' do
+    it "defines the method" do
+      expect(subject).to respond_to(:show_menu)
+    end
+
+    it "formats the menu" do
+      allow(subject.show_menu).to receive(:show_menu) { "#{i + 1}. #{k[:item]}: £#{k[:price]}" }
```
soph-g: This test doesn't have an expectation, so isn't currently testing anything. Also, try to avoid stubbing methods on the subject currently under test.

spec/menu_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+  end
+
+  describe '#select' do
+    it "defines the method" do
+      expect(subject).to respond_to(:select)
+    end
+
+    it "Choose two items" do
```
soph-g: What if I wanted to choose a different item?

PR #1093, opened 2018-09-24
initial commit

sebgrebe:
messed up first time around :)

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,23 @@
+class Order
+
+  attr_reader :list, :total
+
+  def initialize
+    @list = []
+    @total = total
+  end
+
+  def add(dish,amount)
+    return @list << {name: dish[:name], price: dish[:price], amount: amount} if @list.select{|item| item[:name] == dish[:name]}.empty?
+    @list.each do |item|
+      item[:amount] = amount + item[:amount] if item[:name] == dish[:name]
+    end
+  end
+
+  def total
+    sum = 0
+    @list.each { |dish| sum += dish[:price] * dish[:amount] }
+    sum
+  end
+
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,55 @@
+require_relative 'order'
+require_relative 'text'
+
+class Takeaway
+  attr_reader :order, :menu, :text
+  attr_accessor :order #added for testing of choose method
+
+  MENU_LIST = [{name: 'Bolognese', price: 6}, {name: 'Kotbullar', price: 7.50}, {name: 'Beetroot tart', price: 11.20}]
+
+  def initialize(menu = MENU_LIST, order = Order.new, text = Text.new)
+    @order = order
+    @menu = menu
+    @text = text
+  end
+
+  def see_menu
+    @menu.each do |dish|
+      puts "Dish: #{dish[:name]}, price: #{dish[:price]}"
+    end
+  end
+
+  def choose(dish,amount=1)
+    choice = @menu.select { |item| item[:name] == dish }
+    fail "Sorry, we do not have this dish." if choice.empty?
+    @order.add(choice[0],amount)
+  end
+
+  def check_total
+    puts "You have ordered: "
+    print_order
+    puts "The calculated total is #{@order.total}"
+  end
+
+  def print_order
+    @order.list.each do |dish|
+      puts "Dish: #{dish[:name]}, price: #{dish[:price]}, amount: #{dish[:amount]}"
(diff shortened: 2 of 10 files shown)


lib/order.rb
```diff
@@ -0,0 +1,23 @@
+class Order
+
+  attr_reader :list, :total
+
+  def initialize
+    @list = []
+    @total = total
+  end
+
```
EdwardAndress: `#add` is a bit hard to read.  Can you find a way of making this method simpler?

Perhaps you could abstract some of the behaviour into another method/s.  For example, it's useful to abstract complex conditions so that they can be named.

```
def add(dish, amount)
  return @list << {name: dish[:name], price: dish[:price], amount: amount} if not_on_list?(item)
  <blah blah blah>
end

def not_on_list?(item)
  @list.select{|item| item[:name] == dish[:name]}.empty?
end
```

OR, using `#include?`

```
def not_on_list?(item)
  @list.map(&:name).include?( item[:name] )
end
```

How does `#include?` compare to `#select` + `#empty?` in terms of communicating intent?

lib/takeaway.rb
```diff
@@ -0,0 +1,55 @@
+require_relative 'order'
+require_relative 'text'
+
+class Takeaway
+  attr_reader :order, :menu, :text
+  attr_accessor :order #added for testing of choose method
+
+  MENU_LIST = [{name: 'Bolognese', price: 6}, {name: 'Kotbullar', price: 7.50}, {name: 'Beetroot tart', price: 11.20}]
+
+  def initialize(menu = MENU_LIST, order = Order.new, text = Text.new)
```
EdwardAndress: Nice use of dependency injection.

lib/takeaway.rb
```diff
@@ -0,0 +1,55 @@
...
+    @order = order
```
EdwardAndress: Although you'll have to create a new Takeaway for each new order.

lib/takeaway.rb
```diff
@@ -0,0 +1,55 @@
...
+  def see_menu
+    @menu.each do |dish|
+      puts "Dish: #{dish[:name]}, price: #{dish[:price]}"
+    end
+  end
+
+  def choose(dish,amount=1)
+    choice = @menu.select { |item| item[:name] == dish }
+    fail "Sorry, we do not have this dish." if choice.empty?
+    @order.add(choice[0],amount)
```
EdwardAndress: Have you tried using `#find` instead of `#select` here?

lib/text.rb
```diff
@@ -0,0 +1,23 @@
+require 'twilio-ruby'
+require 'dotenv/load'
+
+class Text
```
EdwardAndress: Good decision to create a new class for this.

lib/text.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def initialize(client = Twilio::REST::Client)
+    @message = "Thank you! Your order was placed. The total was: "
```
EdwardAndress: Nice that you obfuscated your credentials!

spec/text_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  let(:account) { double :account, messages: messages }
+  let(:api) { double :api, account: account }
+  let(:client) { double :client, api: api }
+  let(:client_class) { double :client_class, new: client }
+
+  describe 'send'
+    it 'calls create on messages of client' do
+      text = described_class.new(client_class)
+      allow(messages).to receive(:create)
+      expect(messages).to receive(:create).with(an_instance_of(Hash))
```
EdwardAndress: I think we covered this bit during the group code review.  This test could be more specific; it could specify the args that we expect `#create` to be called with.

lib/text.rb
```diff
@@ -0,0 +1,23 @@
...
+    @account_sid = ENV['ACCOUNT_ID']
+    @auth_token = ENV['AUTH_TOKEN']
+    @from = ENV['TWILIO_PHONE']
+    @to = ENV['MY_PHONE']
+    @client = client.new @account_sid, @auth_token
+  end
+
+  def send(total)
```
EdwardAndress: So the user only gets to see the total in the confirmation?  Would you be happy with this?

EdwardAndress:
This is a good solution overall.  I think the thing to focus on, with regards to the various bits of feedback I've given, is how your choice of methods communicates your intent.  If you want to perform a 'basic' operation on an Array / Hash, it's often worth browsing through Ruby docs to see if there's a method whose name conveys your intent (and which also gets the job done in one neat move).

PR #1094, opened 2018-10-12
Takeaway pull request

Learner5200:
May make a new pull request if I have time to do the bonus challenge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,34 +1,9 @@
-Takeaway Challenge
-==================
-```
-                            _________
-              r==           |       |
-           _  //            |  M.A. |   ))))
-          |_)//(''''':      |       |
-            //  \_____:_____.-------D     )))))
-           //   | ===  |   /        \
-       .:'//.   \ \=|   \ /  .:'':./    )))))
-      :' // ':   \ \ ''..'--:'-.. ':
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
+# Takeaway Challenge
 
- ```
+## Goals
 
-Instructions
--------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
------
-
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+* A weekend project for week 2 of Makers.
+* Goal was to learn more about object oriented programming, and test-driven development, by creating a takeaway program that satisfies the following user stories:
 
 ```
 As a customer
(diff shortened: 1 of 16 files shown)


lib/add_delete.rb
```diff
@@ -0,0 +1,56 @@
+module AddDelete
```
Caitlin-cooling: Like this use of module to hold additional functionality to ensure single responsibility of classes

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def view
+    dishes.uniq.each do |dish|
+      puts summary(dish)
+    end
+    puts "Total = £#{total}"
+  end
+
+  def add(*dishes)
```
Caitlin-cooling: It will allow you to add a string of an item that is not on the menu

lib/restaurant.rb
```diff
@@ -0,0 +1,33 @@
+require_relative "texter"
+
+class Restaurant
+
+  def initialize(texter = Texter.new)
```
Caitlin-cooling: Perhaps dependancy injection could have been used rather than initialising with a new Texter? Would be great to chat about why you decided to do it this way.

lib/menu.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'add_delete'
+
+class Menu
+  include AddDelete
+
+  def initialize(dishes = [])
+    @dishes = dishes
```
Caitlin-cooling: Nice use of dependency injection

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+    delete_all(dishes, @dishes)
+  end
+
+  def total
+    dishes.map(&:price).inject(&:+)
+  end
+
+  private
+
+  def quantity(dish)
```
Caitlin-cooling: Nice use of encapsulation

lib/texter.rb
```diff
@@ -0,0 +1,24 @@
+require 'dotenv/load'
+require 'twilio-ruby'
+
+class Texter
+  SID = ENV['SID']
```
Caitlin-cooling: Like the use of env variables

README.md
```diff
@@ -48,32 +23,53 @@ So that I am reassured that my order will be delivered on time
...
+* Adding and deleting behaviour is defined in a separate module included in the Menu and Order classes, to avoid repetition.
+* The restaurant is responsible for placing the user's order. Restaurants will first check the total provided to ensure it is accurate, before using the texter to send a confirmation text to the user.
+* The texter is respondible for sending texts, using the Twilio API.
 
-* A free account on Twilio will only allow you to send texts to "verified" numbers. Use your mobile phone number, don't worry about the customer's mobile phone.
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
+## How to Use
 
+* Open irb
+* Require all the files in /lib
```
Caitlin-cooling: Nice that you have lots of classes to ensure single responsibility

lib/dish.rb
```diff
@@ -0,0 +1,12 @@
+class Dish
```
Caitlin-cooling: This class helps the code to follow the open close principle which is good!

PR #1095, opened 2018-10-13
Takeaway challenge complete

MHUS25:


MHUS25:
two unit tests on texting functionality failing due to hidden credentials & phone numbers

lib/takeaway.rb
```diff
@@ -0,0 +1,57 @@
+require 'menu'
+require 'order'
+require 'text'
+
+class Takeaway
+
+  attr_reader :menu, :order, :text
+
+  def initialize(menu = Menu.new, order = Order.new, text = Text.new)
```
soph-g: What is the responsibility of this class?
Although this provides a single interface for the user interaction, the class itself doesn't appear to have a responsibility - it only delegates. Be aware of patterns like this as it can make it harder to extend your code in some cases.

lib/text.rb
```diff
@@ -0,0 +1,20 @@
+require 'rubygems'
+require 'twilio-ruby'
+
+class Text
+    # credentials hidden
+    account_sid = 'AC86bdxxxxxxxxxxxxxxxxxxxxxxxxxxxxx'
```
soph-g: Look into use environment variables to store this data - will also mean that anyone can clone your repo and use it with their own Twilio information.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,70 @@
+require 'takeaway'
+require 'pry'
+
+describe Takeaway do
+
+  subject(:takeaway) { described_class.new(menu, order) }
```
soph-g: Text double hasn't been passed in here - so the double won't be used in the tests.

spec/text_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'text'
+
+describe Text do
+
+		it 'sends a confirmation text' do
+				expect { subject.send_sms }.not_to raise_error
```
soph-g: Try to find a way to check that your text object interacts with the Twilio object in the way you expect it to, either by stubbing the methods called on the Twilio object, or by passing in a double of the Twilio object.

PR #1096, opened 2018-10-14
My init Submission

RyanWolfen7:
I've refactored and got it all sorted out. May add more stuff for giggles, but I'd be interested in some feed back for this. IDK if I'm going down the right path with this. It felt off compared to the Airport challenge and I'm less sure with the structure quality.


README.md
```diff
@@ -16,20 +16,32 @@ Takeaway Challenge
...
 -------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
+  * **Initialize a menu class**
```
soph-g: Good README, detailed an explains the process and how to use the application.

lib/customer.rb
```diff
@@ -0,0 +1,53 @@
+# This is the main class of the program
+# It allows a customer to look, decide, and place orders from a menu
+
+class Customer
+
+  attr_reader :order
+
+  def initialize(text = SmsText.new)
+    @order = Order.new  # Creates an order class upon init
```
soph-g: Dependency injection should be used for the Order, as seen with the Text.

lib/customer.rb
```diff
@@ -0,0 +1,53 @@
+# This is the main class of the program
+# It allows a customer to look, decide, and place orders from a menu
+
+class Customer
```
soph-g: I find the name of this class confusing, because as a user of the takeaway application I wouldn't expect to tell a 'customer' to do things. It looks like this class is an interface to other functionality, it's worth asking 'what is the purpose of this class?'. It might not need to exist, or it might need to be renamed to make it's purpose clearer.

lib/dish.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+    @quantity = DEFAULT_TOTAL
+  end
+
+  # Function called by order when customer
+  # selects more then one instance of the dish
+  def up_quantity
```
soph-g: This is an interesting idea for keeping track of dish quantity in an order.

lib/menu.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize
+    @list = []          # holds dishes
+    @loaded_dish = nil  # needed so that find dish returns a dish
+  end
+
+  def add_dish(dish)
+    @list << dish # adds dish to the list array
+  end
+
+  def list_dishes
```
soph-g: Not sure if it's worth having a private method when the public method only calls the private method. In a case like this I would expect the list_dishes method to include the code for listing the dishes. I'd use a private method for something that I didn't want the user to be able to do, needs to be called by the code in a public method.

lib/menu.rb
```diff
@@ -0,0 +1,37 @@
...
+    print_dishes  # prints the menu
+  end
+
+  def find_dish(dish)
+    search_menu(dish) # searches for the dish in the menu
```
soph-g: Again, not sure that extracting the code to a private method adds value here.

lib/order.rb
```diff
@@ -0,0 +1,68 @@
...
+
+class Order
+
+  DEFAULT_TOTAL = 0 # Base cost of order
+
+  attr_accessor :order, :soter
+
+  def initialize
+    @order = []   # Initial array to recieve dishes
+    @soter  = []  # Holds refinned array of dishes and quantities
```
soph-g: Use of comments suggests that this code isn't clear. Are three arrays required for a single order?

lib/order.rb
```diff
@@ -0,0 +1,68 @@
...
+    @order.each do |dish|
+      @soter << dish if dish?(dish) == false
+      dish.up_quantity if dish?(dish) && dish.is_a?(Dish)
+    end
+  end
+
+  # converts the array of classes into an array of strings
+  def load_basket
+    @soter.each do |dish|
+      if dish.is_a?(Dish)
```
soph-g: What happens if it's not a Dish? Is it likely that something that isn't a dish would be passed in?

lib/texting.rb
```diff
@@ -0,0 +1,34 @@
...
+    init_msg(confirm_order)
+  end
+
+  private
+
+  # Links functionality with twilio
+  def init_msg(msg)
+    account_sid = 'ACcb0e21414d3ab3d87192fa33e93fc1c0' # => please change if testing
+    auth_token = 'fc23137ee0ee4262f9c677df35211a4d'    # => please change if testing
+  # set up a client to talk to the twilio REST API
```
soph-g: Please remove these credentials from GitHub - look into using environment variables to protect this data by storing it locally only.

PR #1097, opened 2018-10-14
Completed challenge

jebax:


lib/chinese_menu.rb
```diff
@@ -0,0 +1,32 @@
+class ChineseMenu
```
soph-g: Would expect to see a 'Menu' class, and be able to instantiate new instances of that class called 'chinese_menu', 'italian_menu'  etc, and these instances would have the specific menu information in them. Classes should be re-usable in different contexts.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    { "Quattro Formaggi" => 10 },
+    { "Truffle" => 12.5 },
+    { "Gold Pizza" => 1200 }
+  ])
+  }
+  let(:menu_class) { double(:menu_class, new: menu) }
+  let(:text) { double(:text, send: nil) }
+  let(:text_class) { double(:text_class, new: text) }
+  let(:basket) { double(:basket, items: [{ name: "Margherita", quantity: 2, cost: 17.0 }]) }
+  let(:basket_class) { double(:basket_class, new: basket) }
```
soph-g: Lots of complexity in the doubling here - might suggest that there's an opportunity to simplify the design of the application by changing the way the classes interact. Something to think about :)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:subject) { Takeaway.new(menu_class, basket_class, text_class) }
+
+  describe "#list_dishes" do
+    it 'should be able to list dishes from a menu' do
+      expect(menu).to receive(:list_dishes)
```
soph-g: Good use of doubling

spec/text_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'text'
+
+describe Text do
+  let(:text_send_stub) {
+    stub_request(:post, "https://api.twilio.com/2010-04-01/Accounts/ACeeb78433dcba23619b98cde03e52ad52/Messages.json").
```
soph-g: It should be possible to mock this in a much simpler way - try simplifying the mocking of the Twilio object to make this test easier to read

PR #1101, opened 2018-10-14
Caitlin - Takeaway Challenge

Caitlin-cooling:


spec/feature_tests/user_stories_spec.rb
```diff
@@ -0,0 +1,81 @@
+require 'menu'
+require 'order'
+require 'send_sms'
+
+# As a customer
+# So that I can check if I want to order something
+# I would like to see a list of dishes with prices
+describe Menu do
```
soph-g: These tests look like a mix between feature tests (testing functionality from the users perspective) and unit tests (testing that a particular unit of code e.g. a method on a class, does what is expected.
I recommend splitting these out, so you have a clear division between feature tests and unit tests.

spec/order_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'order'
+
+describe Order do
+  let(:items) { [{ "Margherita" => 8 }, { "Roasted Vegetable" => 9 }, { "Chorizo" => 12 }] }
+  let(:menu) { double(:menu, new: items) }
+  let(:order) { Order.new(menu.new) }
```
soph-g: Yes, doubling has been used correctly here

spec/send_sms_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'send_sms'
+require 'order'
+
+describe SendSMS do
+  let(:confirmation) { Order.new(menu.new) { include SendSMS } }
```
soph-g: There is a hard dependency on Order here, but menu has been doubled. Recommend trying to use doubles for all dependencies.

PR #1102, opened 2018-10-14
JLJ takeaway-challenge

JL-J:


spec/basket_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+  context 'Describe review_order:' do
+    before do
+      subject.basket["Satay Chicken"] = 6
+    end
+    it 'returns the item in the basket and the total cost' do
+      expect(subject.review_order).to eq "Satay Chicken, £6\nTotal: £6"
+    end
+  end
+
+end
```
imtiyazzaman1: As the `total` and `view` methods are public, they should have their own explicit tests

lib/basket.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def total
+    @basket.values.sum
+  end
+
+  def review_order
+    "#{view}\nTotal: £#{total}"
+  end
+
+end
```
imtiyazzaman1: `view` and `total` methods could be stored as private methods as `review_order` makes use of both of them and it looks like they won't need to be called external outside of the class

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    it 'rasies an error message if the total inputted by user does not match basket total' do
+      expect { takeaway.place_order(4) }.to raise_error "Incorrect total"
+    end
+    it 'confirms order has been placed' do
+      allow(sms).to receive(:text)
+      expect(takeaway.place_order(2)).to eq "Your order has been place. You will recieve a confirmation text shortly."
+    end
+  end
+
+end
```
imtiyazzaman1: Great test, its tests the full functionality of the `Takeaway` class.
Test is also very well isolated using doubles

spec/menu_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+describe Menu do
+
+  context 'Describe view:' do
+    it 'prints the full list of the menu with prices' do
+      menu = Menu.new
+      expect(menu.view).to eq "Sweet and Sour Pork, £8\nSatay Chicken, £6\nBoiled Rice, £2"
+    end
+  end
+
+end
```
imtiyazzaman1: Great test, it tests all the functionality of the `Menu` class

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+require_relative '../lib/basket.rb'
+require_relative '../lib/text.rb'
+
+class Takeaway
+
+  def initialize(menu = Menu.new, current_basket = Basket.new, sms = Text.new)
+    @menu = menu
+    @current_basket = current_basket
+    @sms = sms
+  end
```
imtiyazzaman1: great use of dependancy injection

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def menu
+    @menu.view
+  end
+
+  def order(dish)
+    fail "Dish unavailable" unless @menu.list.has_key? dish
+    cost = @menu.list[dish]
+    @current_basket.basket[dish] = cost
```
imtiyazzaman1: The 3 lines above could have been moved the `basket` class into a new `add` method as this is functionality `Basket` should be responsible

lib/text.rb
```diff
@@ -0,0 +1,23 @@
...
+  def text
+    @client = Twilio::REST::Client.new(ACCOUNT_SID, AUTH_TOKEN)
+    @client.messages.create(
+      from: @twilio_phone_number,
+      to: @client_number,
+      body: "Your order was placed. It will be delivered before #{Time.now + 3600}"
+    )
+  end
+
+end
```
imtiyazzaman1: Class successfully sends a text message! 
sensitive information could have been hidden using environmental variables

lib/takeaway.rb
neoeno: Great feedback here :+1:

lib/text.rb
```diff
@@ -0,0 +1,23 @@
+require 'twilio-ruby'
+
+class Text
+
+  ACCOUNT_SID = 'ACc68fbb7c5f4b5a7c5bb1a06faaab901d'
+  AUTH_TOKEN = '4e5bdc58fcc22d2b586f396a43288263'
+
+  def initialize(number = XXX, client_number = XXX, object = Object)
```
neoeno: These injections are a little odd — perhaps you were trying to figure some things out?

If not — what's `XXX`? and what's the purpose of `object`?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      allow(menu).to receive(:view).and_return("menu")
+    end
+    it 'prints the menu' do
+      expect(takeaway.menu).to eq "menu"
+    end
+  end
+
+  context 'Describe #order:' do
+    before do
+      allow(menu).to receive_message_chain(:list, :has_key?).and_return(false)
```
neoeno: I think you're mocking a little too far here — `list` is a method on `Menu` which makes sense to mock`, but `has_key?` starts to be 'knowing too much' about the `#order` method. What if we changed how `#order` achieved its results? We'd have to change this test, even though it all still works :)

A better solution:

```ruby
allow(menu).to receive_message_chain(:list).and_return({"Boiled Rice" => 4})
```

spec/text_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+describe Text do
+
+  # let(:text) { double(Text.new("dummy", "dummy")) }
+  # let(:object_double) { double(:object, time: 17:00:00) }
+  #
+  # context 'Describe #text_body' do
+  #   it 'returns the body text' do
+  #     expect(text.text).to eq "Your order was placed. It will be delivered before 18:00:00"
+  #   end
+  # end
```
neoeno: I suppose you had trouble testing this class?

PR #1109, opened 2018-10-14
Takeaway Challenge Complete

imtiyazzaman1:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  describe "#current_order" do
+    it "should call current on Order object" do
+      expect(order).to receive(:current)
+      subject.current_order
+    end
+  end
+
+  describe "#place_order" do
```
JL-J: could test whether place order send a text/ sends a method to the Message class.

spec/order_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  subject { described_class.new(menu: menu) }
+
+  before {
+    allow(menu).to receive_message_chain(:dishes, :key?).and_return true
+    subject.add("Half Chicken", 2)
+  }
+
+  describe "#add" do
+    it "should add several dishes to the current order" do
+      expect(subject.current).to eq "Your current order:\n2 x Half Chicken"
```
JL-J: this test is really for current as opposed to add

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'takeaway'
```
JL-J: Really good unit tests both here and in your other specs. Nice lay out and easy to follow what's going on.

spec/message_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'message'
+
+describe Message do
+  let(:client) { double :client }
+  subject { described_class.new(client: client) }
+
+  describe "#send" do
+    it "should create a new txt message" do
+      expect(client).to receive_message_chain(:messages, :create)
+      subject.send
```
JL-J: if possible, it would be good to see if the message that was sent (the body) was as expected, and to test that your time method was working correctly.

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  describe "#total" do
+    before { subject.add("Half Chicken", 2) }
+
+    it "should call dishes on Menu obejct" do
+      expect(menu).to receive_message_chain(:dishes, :[], :*)
+      subject.total
+    end
+
+    it "should return the total cost of the order" do
+      allow(menu).to receive_message_chain(:dishes, :[], :*).and_return(16)
```
neoeno: This stubbing is a bit too clever — you're not just mocking `menu`, but also starting to interfere with the way `Order` does its work. I.e., you could change the way `Order` calculates the product and this test would fail despite the fact that `Order` still works fine.

You probably ought to say `allow(menu).to receive(:dishes).and_return([{"Chicken Wing" => 16}])` or something similar, and let `total` do the work.

lib/order.rb
```diff
@@ -0,0 +1,40 @@
...
+    "Your current order:\n" + order
+  end
+
+  def add(dish, quantity)
+    raise("Item not found on menu") unless on_menu?(dish)
+    @order << { dish => quantity }
+  end
+
+  def total
+    @order.map { |item| price_of_item(item) }.flatten.reduce(:+)
```
neoeno: This logic is a little abbreviated. Could it be made more clear to a reader what's going on here? Maybe using some private methods or variable names?

PR #1110, opened 2018-10-14
Takeaway challenge attempt

imogenmisso:


Gemfile
```diff
@@ -8,3 +8,5 @@ gem 'rspec'
 gem 'rubocop', '0.56.0'
 gem 'simplecov', require: false, group: :test
 gem 'simplecov-console', require: false, group: :test
+gem 'pry'
+gem 'twilio-ruby'
```
hnryjmes: Good - Ensure that all gems being used are in Gemfile ✅

README.md
```diff
@@ -77,3 +77,26 @@ Notes on Test Coverage
...
+5. If user requests and item that is not on the menu, an error is raised
+  *  order.add_to_order('tuna sandwich')
+  *  RuntimeError (Item is unavailable)
+6. User can find out total per item
+  *  order.order_total
+  *  *This will return an array*
+7. User can find out the whole bill
+  *  order.sum_order
+8. User can check that the total bill is correct, and if so, a text will be sent
+  *  order.check_total
```
hnryjmes: Good - README updated ✅
I like the integration of user stories and which method calls are used to interact with your code

hnryjmes: Good - Instructions in README ✅
Although you didn't specify a particular interactive environment like IRB or Pry, it seems obvious from the lines you included that this would be the expected output

spec/order_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  it 'sends text' do
+    subject.add_to_order('veggie burger', 4)
+    subject.add_to_order('pizza', 2)
+    subject.sum_order
+    subject.check_total
+    expect(subject.sms).to be_a Message
+  end
+
+
+end
```
hnryjmes: Good - Tests should test real behaviours not stubs ✅
It looks like your tests are good, i.e. you aren't using `expect` to test something that you've already coded to happen in the test. Your tests link to the actual methods.

hnryjmes: It looks like you didn't reach the stage where you were testing classes interacting with each other, so it makes sense that you didn't use doubles

lib/order.rb
```diff
@@ -0,0 +1,43 @@
+require_relative 'message'
+require_relative 'menu'
+
+class Order
+
+attr_reader :basket, :menu, :sms
+
+  def initialize(menu = Menu.new)
+    @menu = menu.items
```
hnryjmes: Good - Appropriate use of Dependency Injection ✅

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+    @basket[name] = quantity
+  end
+
+  def order_total
+    @basket.map do |name, quantity|
+      @menu[name] * quantity
+    end
+  end
+
+  def sum_order
```
hnryjmes: Good - Design for Single Responsibility Principle ✅
I like that your methods are small and easy to read, it makes the code much better

PR #1116, opened 2018-10-14
Take me away challenge :D

rekapap:
My solution to the Take-Away Challenge.


README.md
```diff
@@ -1,79 +1,28 @@
...
-       .:'//.   \ \=|   \ /  .:'':./    )))))
-      :' // ':   \ \ ''..'--:'-.. ':
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
-
- ```
-
-Instructions
--------
+# Take-Away Challenge
```
soph-g: Try to keep your commits small and commit a single feature at a time. It might be a good idea to try writing a test, use TDD to pass the test, and then commit (after refactoring, if needed).

README.md
```diff
@@ -1,79 +1,28 @@
...
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+Take-Away Challenge for Makers
 
-Task
------
+## Approach
```
soph-g: Would be useful to include a diagram of the classes in your code, together with how they interact.

lib/menu_factory.rb
```diff
@@ -0,0 +1,34 @@
...
+  def create_menu(filename)
+    file = File.read(filename)
+    dish_hash = JSON.parse(file)
+    make_objects(dish_hash)
+  end
+
+  def make_objects(hash)
+    categories = hash.map do |key, value|
+      category = @category_class.new(key)
+      value.each do |item|
```
soph-g: Be aware of this nesting interaction, it makes the code hard to understand. Private methods could have been used to make this easier to read.

lib/menu_view.rb
```diff
@@ -0,0 +1,16 @@
...
+class MenuView
+  MENU = 'MENU'
+  def initialize(menu, category_view_class = CategoryView)
+    @menu = menu
+    @category_view_class = category_view_class
+  end
+
+  def display
+    @menu.categories.reduce("#{MENU}\n") do |sum, category|
+      sum + @category_view_class.new(category).display
```
soph-g: Do you need to instantiate a new instance of CategoryView class for each category, or could you have had a single instance of the class with a 'display' method that takes the category as an argument?

PR #1122, opened 2018-10-15
Takeaway challenge 

Edy1988:


lib/takeout.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def show_menu
+    MENU
+  end
+
+  def process(order)
+    raise "Invalid order" unless valid order
+
+    @sender.create(
```
Katerina-codes: Your Takeout class seems to be doing a lot of work. Could you separate the concerns of the Takeout class from your SMS logic? Perhaps the SMS logic needs to live in its own class?

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+    order_total = 0
+
+    @basket.each do |item|
+      order_total += item.values[0]
+    end
+
+    order_total
+  end
+
+  def show_basket
```
Katerina-codes: This show_basket method seems a little long, it’s also doing quite a few things. Could you take out some of this behaviour and extract it into new methods?

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+    formatted_basket = ""
+
+    @basket.each do |item|
+      name = item.keys[0]
+      price = item[name]
+
+      formatted_basket += "'#{name} $#{format("%.2f", price)}', "
+    end
+
+    "Your order is: #{formatted_basket}total is: $#{format("%.2f", total)}"
```
Katerina-codes: show_basket seems to format *and* show the basket. As I mentioned above these seem to be two different methods. Should they be in the Order class? They seem to be doing the work of a Display class.

spec/order_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'order'
+
+describe Order do
+  describe '#add_dish' do
+
+    it 'starts with an empty basket' do
+      order = Order.new
+
+      expect(order.basket).to eq []
```
Katerina-codes: This test is just testing state so you don't need it. We know Order will be initialised with an empty array that's what you're doing every time you create a new order. A test that would test behaviour rather than state would be using add_dish and making sure the array now has a dish in it (which you're doing in the test below).

spec/takeout_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'takeout'
+
+describe TakeOut do
+
+  describe '#show_menu' do
+    it 'has a menu which shows item and price' do
```
Katerina-codes: Please see my previous comment in your order spec regarding testing state :)

PR #1150, opened 2018-11-11
User stories 1-3 done, not enough time for user story 4

RobertsK284:
Hi, this isn't the complete solution, but I managed user stories 1 - 3.


RobertsK284:
Did some work on User Story 4, but it doesn't work yet.

PR #1162, opened 2018-11-12
Takeaway challenge submission

willmcb:


lib/dish.rb
```diff
@@ -0,0 +1,26 @@
+class Dish
+  attr_reader :name, :description, :price
+  def initialize(dish_hash = {name: "", description: "", price: 0})
```
EdwardAndress: You can achieve this in a slightly more concise way
```
# this will still work even if you pass in a hash
def initialize(name: "", description: "", price: 0)
    @name = name
    @description = description
    @price = price
end
```

lib/dish.rb
```diff
@@ -0,0 +1,26 @@
...
+    @name = dish_hash[:name]
+    @description = dish_hash[:description]
+    @price = dish_hash[:price]
+  end
+
+  def set_all(dish_hash)
```
EdwardAndress: When would you use this method?

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'price'
+class Menu
+  attr_accessor :data
+  def initialize(menu_data)
+   @data = menu_data
+  end
+
+  def print
```
EdwardAndress: Did you consider abstracting this behaviour to another class?  Is it really the responsibility of the menu to print itself?

lib/messenger.rb
```diff
@@ -0,0 +1,23 @@
...
+class Messenger
+  def initialize(client, order, to_from)
+    @client = client
+    @order = order
+    @to_from = to_from
+  end
+
+  def message
+    { from: @to_from[:from], to: @to_from[:to],
+      body: "Hi there, your order totalling £#{Price::format(@order.total)} " +
```
EdwardAndress: I like this `Price::format` thing but perhaps it could also generate the '£'.  Then you could have different formatters for different (english speaking) regions of the world.

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+    @items = []
+    @menu = menu
+    @varified = false
+  end
+
+  def varify
+    @varified = true
+  end
+
+  def create_dish(menu_id, dish = Dish.new)
```
EdwardAndress: I find this a bit confusing and I think it's probably a naming issue.  `create_dish` sounds like something that should be used in `Takeaway` or `Menu` but in the context of an `Order` perhaps `add_dish` or `select_dish` would make more sense (to me!).

If this is the only place `Dish` is used, perhaps a better name would be `Item` because orders are 'itemised'.  This is probably more contentious. Creating Dishes in the Order feels wrong though, somehow.

lib/takeaway.rb
```diff
@@ -0,0 +1,116 @@
+require 'twilio-ruby'
+require_relative 'menu'
+require_relative 'customer'
+require_relative 'order'
+require_relative 'messenger'
+
+class Takeaway
+# :nocov:
+  def initialize(creds, client = Twilio::REST::Client, menu = Menu)
```
EdwardAndress: Nice of of Dependency injection again

lib/start.rb
```diff
@@ -0,0 +1,8 @@
+require_relative 'takeaway'
+
+
+# Start app by adding the following environment variables, then running ruby start.rb
+
+creds = { account_sid: ENV['A_SID'], auth_token: ENV['AUTH_TOKEN'] , from_number: ENV['FROM_NUMBER'] }
```
EdwardAndress: well done for obfuscating the credentials

lib/price.rb
```diff
@@ -0,0 +1,24 @@
+class Price
```
EdwardAndress: Would it be possible to simplify this by dividing the pence amounts by a float (`pence / 100.0`) and then using `stprintf` to add the required number of trailing / preceding zeros?

spec/customer_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:order) { double(:order, :build => true, print: "your order", varify: true) }
+  subject(:customer) { Customer.new("Will", "07539204924", order) }
+  it 'can select a number of available dishes' do
+    expect(customer.select_dishes([1, 2])).to eq(true)
+  end
+
+  it 'can varify order' do
+    customer.select_dishes([1, 2])
+    customer.varify_order
+    expect(customer.print_order).to eq("your order")
```
EdwardAndress: the description says that this test is to do with verification but there's also an expectation about printing.

EdwardAndress: Do you really care what is returned by `print_order` or is it only important that the order receives the right message?

EdwardAndress:
Great solution on the whole.  You've gone beyond the requirement in some regards, which is good when you're doing an academic exercise but less good if you're building something for real (assuming that 'going beyond the requirement' means you don't have a clear idea of what the stakeholder wants at that moment).

A few things to consider.
Naming – I've made a few comments about this against specific lines of code but it's probably worth thinking about more broadly.
Testing delegation – if one class delegates to another, you only need to test that the correct message is passed to the correct object.  You can test that the correct things are done in tests for the receiving class.
I'm not sure that you need the customer.
The tests for the Dish and Menu class are a bit state-y but there's not much behaviour in there to be testing.  Perhaps that's an indication that, in it's current state, the application can live without those two classes.  If there's no behaviour associated with that data, the concepts of Dish and Menu _could_ be represented by simpler data structures.

lib/dish.rb
willmcb: Thanks that's good to know.

lib/dish.rb
willmcb: I added this at the time of making the class as I thought i might need it... classic case of YAGNI really, basically forgot to delete it.

lib/menu.rb
willmcb: I did think about having a general printing class - I think it would make sense to have this behaviour extracted. I just didn't get around to implementing it.

lib/messenger.rb
willmcb: That is a very good idea. I extracted this class after realising that I was a repeating myself across multiple classes. I think the addition of the currency symbols is a really good idea as it would make it more useful and abstract.

lib/order.rb
willmcb: I was torn between Item and Dish and landed on Dish - I take your point though this is confusing and is something that I could think about making clearer.

lib/price.rb
willmcb: Yep, that would be a better solution.

spec/customer_spec.rb
willmcb: Hmm that is a good point - perhaps I was not really thinking about SRP in this test. The expectation about printing is pretty unrelated to the point of the test so should probably go.

PR #1181, opened 2018-12-10
Attempted to complete - Twilio not yet working correctly

scass91:


README.md
```diff
@@ -1,3 +1,23 @@
...
+* Build a test for a method I want, let it fail, then write the code and rework until it passes - using the TDD process learned in week 1 at Makers Academy
+* A takeaway, a menu, and food items were present in the stories. I added a final SMS class to represent the ability to send out messages using Twilio - I came into some difficulty here as my Twilio-generated number was unable to send SMS's
+* Used the user stories info to build these classes, for example allowing the user to select more than one item of food
+* Red - Green - Refactor across the board
+* Used effective debugging methods as I had been taught to this week when I found myself blocked and my tests failing
+* Were I to continue with this I would try to refactor it further, I would also try to get a fully functioning Twilio function
+* If I were to continue I also would increase the food offerings - I did have a pizza :pizza: and a lot of chicken wings this weekend :chicken:, hence the current list being dominated by these!
```
rednblack99: Very important README information 👍 Seriously though, good work remembering to draft a README with your process and useful feedback

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+class Menu
+  attr_reader :food
+
+  def initialize(food)
+    @food = food
+  end
+
+  def print_menu
+    x = food.map do |item, price|
+      "%s £%.2f" % [item.to_s.capitalize, price]
```
rednblack99: What does this do? "%s £%.2f" - looks like it shows your price as £x.xx?

README.md
```diff
@@ -1,3 +1,23 @@
...
+
+![Imgur](https://i.imgur.com/J1h5X0g.png)
```
CazaBelle: An example of how this works would be useful ie menu = Menu.new(food) => what is the parameter you're passing into menu at initialization?  I'm not 100% clear when I read it

lib/order.rb
```diff
@@ -0,0 +1,28 @@
+class Order
+  attr_reader :foods
+
+  def initialize(menu)
```
rednblack99: I believe this is initializing with a new instance of the menu class right? If so, you could have used a default argument here so you don't need to pass an argument everytime.

lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+    @foods = {}
+    @menu = menu
+  end
+
+  def add_to_order(food, quantity)
+    fail "That isn't on the menu!" unless menu.has_food?(food)
```
CazaBelle: Good use of errors

lib/sms.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def delivery
+    client.messages.create(message_content)
+  end
+
+  def message_content
+    {
+      from: FROM,
+      to: TO,
+      body: "Thanks for your order, it will arrive at " % delivery_time
```
rednblack99: Good work on not only getting the SMS to work but adding a delivery time 🥇

lib/sms.rb
```diff
@@ -0,0 +1,31 @@
+require "Twilio-ruby"
+
+class SMS
+
+  attr_reader :client
+  #FORMATTED_TIME = "%H:%M"
```
CazaBelle: probs should remove your token and account number for safety

spec/menu_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'menu'
+
+describe Menu do
+
+  subject(:menu) { described_class.new(food) }
+
+  let(:food) do
+    {
+      wings: 5.00,
```
rednblack99: Props for adding the challenge of using floats. You're the only person I've seen so far using accurate pricing rather than just an integer

spec/order_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  before do
+    allow(menu).to receive(:has_food?).with(:wings).and_return(true)
+    allow(menu).to receive(:has_food?).with(:pizza).and_return(true)
+    allow(menu).to receive(:price).with(:wings).and_return(5.00)
+    allow(menu).to receive(:price).with(:pizza).and_return(7.00)
+  end
+
+  describe "#add_to_order" do
+    it "it enables several dishes to be selected from the menu" do
```
rednblack99: Small thing but don't forget if a test fails it says "class (eg. Order)" and then goes into your describe string so you don't need to say "IT enables" because that will come out as "Order it enables"

spec/order_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      dummy_order
```
CazaBelle: What does dummy_order equal?

spec/order_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    it "it calculates a total for an order" do
+      dummy_order
+      total = 74.00
+      expect(order.total).to eq(total)
+    end
+  end
+
+  def dummy_order
```
CazaBelle: This is cool to put a method in your rspec. Could this be done with a context block?

CazaBelle: or a double?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'takeaway'
+
+describe Takeaway do
+
+  subject(:takeaway) { described_class.new(menu: menu, order: order) }
+
+  let(:menu) { double(:menu, print_menu: readable_menu) }
+  let(:order) { double(:order, total: 74.00) }
```
CazaBelle: good use of doubles

README.md
```diff
@@ -1,3 +1,23 @@
...
+
+Helpful resources
+---------
+
+- [How to use time in Ruby](https://www.rubyguides.com/2015/12/ruby-time/ "Ruby Guides time")
+- [SMS and Twilio - (with bonus cats!)](https://www.twilio.com/blog/2018/04/sms-notifications-ruby-on-rails.html "Twilio Blogpost")
+- [Twilio SMS guide](https://www.twilio.com/docs/sms/quickstart/ruby "Twilio SMS guide")
```
AlinaGoaga: I like how you added the link to the Twilio documentation - comes in handy for someone new trying to run this

lib/menu.rb
scass91: Correct! It changes the string format to one which is actual pricing, the %.2f is important as otherwise prices would be returned as a float with 1 decimal place (i.e. £5.5) which isn't technically correct

scass91:
Trying to delete...

PR #1184, opened 2018-12-10
takeaway challenge Weekend project

tomlovesgithub:


lib/menu.csv
```diff
@@ -0,0 +1,5 @@
+pizza_1,1
```
joaoag: I don't really know that much about CSV files so this is sort of for my curiosity  - how come you stored the menu as a CSV rather than a hash?

lib/order.rb
```diff
@@ -0,0 +1,71 @@
+class Order
+
+  require_relative "./menu.rb"
+  require_relative "./menu_displayer.rb"
+
+  attr_reader :total, :cart, :payment_type, :payment
+
+  def initialize(payment_type = "card")
```
joaoag: Snazzy initialised payment type!

lib/menu.csv
```diff
@@ -0,0 +1,5 @@
...
+pizza_2,2
+pizza_3,3
+pizza_4,4
+pizza_5,5
```
Sindex42: Nice use of a csv file to contain the menu!

lib/order.rb
```diff
@@ -0,0 +1,71 @@
...
+    @total = 0
+    @payment_type = payment_type
+    @payment = false
+  end
+
+  def add(item, amount = 1)
+    item_number = 0
+    @cart << { (item_number += 1) => @menu.list[item - 1], :amount => amount }
+    update_total
+  end
```
Sindex42: Do you have to reference the number of the item in the csv file? Consider making this more clear, I got confused with the argument name and attempted to use the name of the item (i.e pizza_1)

lib/menu_displayer.rb
```diff
@@ -0,0 +1,18 @@
+class MenuDisplayer
+
+  require_relative "./menu.rb"
+
+  attr_reader :menu
+
+  def initialize
```
CKKH: Curious if it would be worth injecting menu as an argument here?

lib/menu.csv
tomlovesgithub: i decided to ~~steal from~~ build on functionality we made in the stu_dir, thought that it could be more versatile/less dependant code if different menus could be loaded in

spec/menu_spec.rb
```diff
@@ -0,0 +1,12 @@
+require "menu.rb"
+
+describe Menu do
+
+  let(:menu) { Menu.new }
```
CKKH: A small thing, as you're testing the menu object here, it could be worth using the following instead of let:

` subject(:menu) { described_class.new }`

lib/menu.csv
tomlovesgithub: and what a great menu it is! fancy a Pizza_3?

lib/order.rb
tomlovesgithub: thanks i have realised in reviewing others work that 
-A. not many others did this
-B. you cant change the payment type back to "card" after changing to "cash" so you'd need to start a fresh order... not much fun on an empty stomach

lib/order.rb
tomlovesgithub: yeah, defo, i guess i was imagining a slightly more imaginative restaurant and ideally a front end which controlled the ordering, but i think [_"magic numbers pizza"_](http://e.lvme.me/rkozb5.jpg) has a ring to it

spec/menu_spec.rb
tomlovesgithub: thanks! been a bit scared of subjects, lets, doubles and the like! the always give the worst error msgs when you put them in the wrong place
Ill be sure to experiment in my next project!

spec/order_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+
+end
+    # As a customer
+    # So that I am reassured that my order will be delivered on time
+    # I would like to receive a text such as
+    # "Thank you! Your order was placed and will be delivered before 18:52"
+    # after I have ordered
+
+    context "after order placed" do
```
tomlovesgithub: `expect(food.eat).to` change(me.hungry).from(true) to(false)`

lib/menu_displayer.rb
tomlovesgithub: still a bit hazy on dependencies, what are the merits of initialising it?

CKKH: Same here! In this case one small merit could be refactoring the method so its one line shorter:

`def initialize(menu = Menu.new)
    @menu = menu
  end`

I could be a small win, but a win all the same!

PR #1185, opened 2018-12-10
3 classes created - without SMS capability

subomionanuga:


spec/order_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  it "selects the item and quantity desired" do
+    expect(subject).to respond_to(:add_item).with(2).arguments
+  end
+
+  it "increase the basket value by the selected items" do
+    allow(subject).to receive(:to_basket).and_return(10)
+  end
+
+  it "adds the total of all items in the basket" do
+  expect(subject.order_total).to eq (0)
```
GLMill: might be good to have a before block that adds an item into the basket for it to calculate :)

lib/menu.rb
```diff
@@ -0,0 +1,19 @@
+
+class Menu
+
+  attr_reader :menu_items
+
+  def initialize
+    @menu_items = [
```
GLMill: need to change this into a hash :)

lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+    @basket_total = 0
+  end
+
+  def add_item(item, quantity)
+    @item = item
+    @quantity = quantity
+  end
+
+  def to_basket
+    item_cost = @menu[@item]
```
GLMill: at the moment this class isnt too dependent on MENU class it may work better to pass menu into class as argument

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "menu"
+
+describe Menu do
+
+  let (:menu_items) {double :menu_items}
```
CazaBelle: What is this double being used in?

lib/menu.rb
```diff
@@ -0,0 +1,19 @@
...
+      { "Chicken wings" => 10 },
+      { "Jerk Chicken" => 15 },
+      { "Lemon Chicken" => 20 },
+      { "BBQ Chicken" => 25 },
+      { "Steamed Lobster" => 50 }
```
CazaBelle: Yum

spec/menu_spec.rb
subomionanuga: I thought I needed it but I still don't grasp it well enough to know when it comes in useful so wasn't sure if I should take it out :'(

PR #1186, opened 2018-12-10
takeaway challenge

HuxleyM:
i made two versions, v2 with 3 classes  i ran into difficulty using rspec but much works bar texting and the other version is simpler two classes. Neither finished :/


lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+        @list = [
+            {:dish => 'green curry', :price => 5.99},
+            {:dish => 'panang', :price => 7.5},
+            {:dish => 'veg tempura', :price => 4.5},
+            {:dish => 'rice', :price => 2.5}
+            ]
+        @choices = []
+    end
+
+    def select(dish_string, quantity)
```
stonefarmer9: Is this th responsibility of the menu class or the order class?

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+        @list.each do |details|
+            if details[:dish] == dish_string
+                details.merge!({:quantity => quantity})
+                return @choices.push(details)
```
stonefarmer9: This is come nice code and I think better than the nested each methods I used!

lib/order.rb
```diff
@@ -0,0 +1,20 @@
+class Order 
+    attr_reader :dishes, :total
+
+    def initialize(menuChoice_obj)
```
stonefarmer9: Is menu choice_obj a clear name? I

spec/menu_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+describe Menu do 
+  let(:thai_menu) {Menu.new}
+  let(:green_curry_details) {{:dish => 'green curry', :price => 5.99}}
+  let(:dish) {'green curry'}
+  let(:quantity) {2}
+  
+  context ' initialize' do
+
+    it 'on #display will show list of menu items' do 
+        expect(thai_menu.list).to include(green_curry_details)
```
stonefarmer9: Is testing an expected inclusion as rigourous as testing a full equality exepctatiom?

spec/menu_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    it 'should have a #select method that will ask for two params' do 
+      expect(thai_menu).to respond_to(:select).with(2)
+    end
+
+    it' should return the array for the specified dish from @list' do 
+        expect(thai_menu.select(dish, quantity )).to eql([green_curry_details_and_quantity])
+    end
+
+    it 'the returned array should include a quanitity key pair made from user input' do 
+    end
```
stonefarmer9: What does this test?

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    describe '#place_order' do 
+      it 'expect a param of a customers phone number' do 
+         expect(order).to respond_to(:place_order).with(1)
+      end
+    end 
+
+
+end
```
subomionanuga: Consider adding a README

lib/menu.rb
CelineKaslin: Not sure that the class menu has the responsibility of select the dishes. 
Maybe an other class like order or basket would have been a way to approach the challenge

lib/order.rb
```diff
@@ -0,0 +1,20 @@
...
+        @dishes = menuChoice_obj
+        @total = 0
+    end
+
+    def sum
```
CelineKaslin: maybe a separate class for sum will have been an idea to separate the responsibility, and the class order will maybe have encapsulate the menu, eventually basket for the selection of the dishes and the sum class for the total.

lib/order.rb
```diff
@@ -0,0 +1,20 @@
...
+       @dishes.each do |item|
+           @total += item[:price]
+       end
+       return @total
+    end
+
+    def place_order(phone_string)
```
CelineKaslin: sorry I might have miss something, but what is the phone_string parameter ?

send_sms.rb
```diff
@@ -0,0 +1,15 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+
```
CelineKaslin: are those method wrapped in a class ?

spec/menu_spec.rb
CelineKaslin: Same question as Stonefarmer9, what is the purpose of this test ? What do you expect from the subject ?

README.md
```diff
@@ -13,6 +13,45 @@ Takeaway Challenge
...
+
+Menu includes :Select allowing us to choose a dish
+ 
+Order includes the functionality to :add, :sum and :close.
+
+Customer includes : check_menu, :add, check_order_total, :new_order : place_order
+
+all methods are designed to be single responsiblity. I made Order a part of customer as they couldnt exist without each other. 
+
+is the logic ok??
```
rednblack99: I really like this approach but I can see why it was tricky to test. My only suggestion for a solution is to double every single one of the hashes to predict what they would contain but that would be very tricky to write.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+            end
+        end
+        raise 'dish not found'
+    end
+end
```
rednblack99: Did you run Rubocop? There are a few minor formatting issues such as not having the newlines at the end of your files to ensure it's compliant with the usual Ruby approach.

PR #1187, opened 2018-12-10
Takeaway Challenge - Jethro Whaley

jaywayawyaj:
My attempt at the Takeaway Challenge for week 2.


README.md
```diff
@@ -1,3 +1,31 @@
+TAKEAWAY CHALLENGE
```
rednblack99: 10/10 would README again

README.md
```diff
@@ -1,3 +1,31 @@
...
+Menu can be viewed by #show_order
+Menu items are selected by #choose_order. User is prompted to enter item number
+Order is submitted by #confirm_order
+User receives SMS confirmation that order will be delivered by the time given (one hour from present)
+Running tests
+Run rspec
+
+NOTES
+
+Test coverage is not high currently, nor of high quality. This is very unfinished! TDD went out the window a little on this challenge - if I were to go back to it I'd like to give myself more time, approach it with a stricter process and plan ahead a bit more. Not my finest work!
```
rednblack99: Will keep in mind when reviewing. Really smart to mention this in the README 👍

lib/order.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  attr_reader :menu, :total_order, :bill
+
+  def initialize(menu = Menu.new)
+    @total_order = []
+    @menu = menu
+  end
+
+  def input
+    puts "Enter item number:"
```
rednblack99: This works but is there a way to use an array so the user can input multiple items to their order at once?

lib/order.rb
```diff
@@ -0,0 +1,45 @@
...
+  def add(input)
+    total_order << @menu.loaded_menu[input - 1]
+    puts "#{@menu.loaded_menu[input.to_i - 1][:name]} added to your order"
+    order_complete
+  end
+
+  def order_complete
+    puts "Would you like to add to your order? (y/n)"
+    answer = gets.chomp
+    if answer == "y"
```
rednblack99: You can use ternary operators for simple if/else statements like this. eg. answer == "y" ? input : nil

lib/sms_confirm.rb
```diff
@@ -0,0 +1,18 @@
...
+    account_sid = "ACa807f58093c9262c30d6551be355db48"
+    auth_token = "ad862830a9ba96f9b5e3031c73eb4d94"
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+  end
+
+  def send_confirmation
+    @client.api.account.messages.create(
+      to: "447561848422",
+      from: "441725762059",
+      body: "Order confirmed! It will arrive by #{(Time.new + 3600).strftime("%-I:%M%P")}"
```
rednblack99: Really well done getting the SMS and time to work correctly. Particularly the strftime thingy because the symbols for that largely look like gibberish!

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'menu'
```
rednblack99: No methods over three lines and all have a single responsibility. NICE 🥇

spec/bill_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'bill'
+
+describe Bill do
+
+  let(:bill) { Bill.new }
+  let(:order) { double :order, name: 'pizza', price: 10 }
```
rednblack99: Good use of doubles

spec/menu_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      expect(menu).to respond_to(:show)
+    end
+
+    it 'adds items to list' do
+      menu.show
+      expect(menu.list).not_to be_empty
+    end
+  end
+
+    # THIS TEST NEEDS WORK, CAN'T FATHOM HOW TO DO IT
```
rednblack99: I can't think of a way to do this with doubles but you could just add the specific list you're expecting the code to return. Not ideal but it would work!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'takeaway'
+
+describe Takeaway do
+
+  let(:takeaway) { Takeaway.new }
```
rednblack99: These aren't actually doubles. You may as well write takeaway = Takeaway.new ;)

specsms_confirm_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'text_alert'
+
+describe TextAlert do
+  # No tests for external API
```
rednblack99: You can't test the API but you can test the methods using doubles. For reference here is once method (with your method names instead obvs): 

    `expect(@sms).to receive(:send_confirmation)
    @sms.send_confirmation(@time)`

lib/sms_confirm.rb
scass91: Echoing above comment - Consider taking "%-I:%M%P" out and setting it equal to a variable at the start of your program, for example TIME_FORMAT = "%-I:%M%P"

Then at the end of your file use ".strftime(TIME_FORMAT) - this will remove the 'magic number'-ish style of the current program. Also consider changing the format to "%H:%M" which will do a similar thing to yours, expect will be a 24 hour clock which removes the need for %P (AM/PM)

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :menu, :order, :bill, :sms
+
+  def initialize(menu = Menu, order = Order.new, bill = Bill.new, sms = SMS.new)
+    @menu = menu
+    @order = order
+    @bill = bill
+    @sms = sms
+  end
+
+  def show_menu
```
scass91: Should this be menu.show, then in your initializer should it state menu = Menu.new

spec/menu_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'menu'
+
+describe Menu do
+
+  let(:menu) { Menu.new }
+
+  describe '#load_menu' do
+
+    it 'should load menu from csv file into @menu' do
```
scass91: Does this specific test test for the existence of a CSV file, or is a further test needed to clarify that the menu is being loaded from a CSV file?

PR #1188, opened 2018-12-10
First user story only, partial takeaway challenge 

beacourage:


lib/dishes.rb
```diff
@@ -0,0 +1,12 @@
+class Dishes
+
+  attr_reader :menu
+
+  def initialize
+    @menu = {"Seeds" => 50,
+      "Eyeballs" => 25,
+      "Camel Breast" => 40,
```
tomlovesgithub: ill take two please!

lib/takeaway.rb
```diff
@@ -0,0 +1,5 @@
+require 'dishes'
+
+class Takeaway
```
tomlovesgithub: I cant work out what this class represents, could you explain?
Could you refactor to remove this class completely?

spec/dishes_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'dishes'
+
+describe Dishes do
+
```
tomlovesgithub: if you had 
` let(:dishes) { Dishes.new } `
then ....

spec/dishes_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  it { is_expected.to respond_to(:menu)}
+
+
+  it 'allows user to see a list of dishes with prices' do
+    dish = Dishes.new
```
tomlovesgithub: ....(my bad didnt see this!)...

lib/dishes.rb
```diff
@@ -0,0 +1,15 @@
+class Dishes
+
+  attr_reader :menu
+
+  def initialize
+    @menu = {"Seeds" => 50, "Eyeballs" => 25, "Camel Breast" => 40, "shoes" => 78 }
```
CazaBelle: Interesting dishes? haha

spec/dishes_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    expect(dish.menu).to include({ "shoes" => 78 })
```
tomlovesgithub: ...this could have been something like 
' expect(dishes.menu).to contain_exactly({"Seeds" => 50, "Eyeballs" => 25, "Camel Breast" => 40, "shoes" => 78 })'
Which i guess is only a little more comprehensive, would this test pass?

spec/dishes_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'dishes'
+
+describe Dishes do
+
+
+  it { is_expected.to respond_to(:menu)}
```
tomlovesgithub: This weekend i learned about the [have_attributes matcher](https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers/have-attributes-matcher), check it out! How could you use this to achieve the same/a similar test?

lib/dishes.rb
```diff
@@ -0,0 +1,12 @@
+class Dishes
+
+  attr_reader :menu
+
+  def initialize
+    @menu = {"Seeds" => 50,
+      "Eyeballs" => 25,
+      "Camel Breast" => 40,
+      "shoes" => 78 }
```
rednblack99: This is a fantastic idea for a startup. Only peasants go to stores to buy their shoes.

PR #1195, opened 2019-01-20
Completed takeaway challenge (Nikhil)

nkhil:
## This is a completed solution to the takeaway challenge.

- All tests passing
- 99.42% test coverage
- feature and unit tests implemented

**Goals I'd like feedback on**

- I can model anything
- I can program fluently

### I can model anything

I've used [mermaidjs](https://mermaidjs.github.io/) to include my domain model diagram. I've used a sequence diagram to map my domain model, which helped me plan my code immensely, and helped me avoid any major refactors. Initially, I started off with some rough diagrams (not domain models) to get an idea of the delegation involved, and the sequence diagram helped cement my ideas before writing a single line of code. **Please have a look at the domain model in my README for more information.**

### I can program fluently

This has been an excellent opportunity to put into practice some of the things covered in 2 weeks of the Makers course, namely: 

- Single responsibility principle
- Cohesion
- Encapsulation
- Semantic naming for methods and classes
- Limiting methods to < 5 lines
- Following the user stories and not making assumptions
- Working with an external API (Twilio) which included reading documentation and further testing

**Please have a look at my README for further elaboration of the points above**


lib/calculate.rb
```diff
@@ -0,0 +1,7 @@
+module Calculate
```
ibrahimbutt: I'm not sure if a `Calculate` module was necessary? As an order system does have a responsibility to tell you the total. Think about an invoice – it's the invoices responsibility to show a total based on the billed items. I would have had a `#total` inside `Order`.

Thoughts?

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+  attr_reader :basket
+
+  ERROR = "Item not found or unavailable"
+
+  def initialize(menu = Menu.new)
+    @basket = []
+    @menu = menu.display
+  end
+
+  def take(dish, quantity = 1)
```
ibrahimbutt: The `#take` method had me a little confused on what it does. `Order.take` doesn't` translate well for me.  Since it's being added to `@basket`, in my opinion `#add/add_to_ basket` would make it easier to understand what the method does at a glance.

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+  
+  private
+
+  def basket_with_prices
+    @basket.map { |arr|
+      [arr[0], arr[1], @menu[arr[0]]]
+    }
+  end
+
+  def repeat_order(dish, quantity)
```
ibrahimbutt: What is the purpose of `#repeat_order`? It sounds as if it repeats the order, as in places a duplicate order. But when I read the code, I can see the item, quantity, and that it was added to the basket.

spec/feature_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'takeaway'
+
+RSpec.describe Takeaway do
+  describe "Feature Tests" do
+    before(:each) do
+      @t = Takeaway.new
```
ibrahimbutt: You should use `subject(:takeaway) { described_class.new }`. This way you can have an explicit subject. Also, in general, if you're making variables that are going to be used repeatedly, use `let`, as RSpec knows how to deal with these variables efficiently. In large applications, there can be 100s of tests so it can be expensive to refresh them before each test.

PR #1199, opened 2019-01-20
Takeaway challenge Debora Piu

Debora38:
### **Project: Week 2 Weekend Challenge - Takeaway Challenge**

**To achieve these goals:**

- I can learn anything by myself
- I can refactor anything
- I can debug anything
- I can TDD anything

**Action Plan**

- create a TakeAway with a menu
- show the menu
- implement placing orders
- check total cost
- send SMS confirmation with delivery time

**Evidence Plan**
The program should use Twilio API to send SMS confirmation of an order placed, but RSpec should use mock not to send SMS when running tests.

**Evaluation**
How did it go?
The program is fully functional and Twilio API was implemented and mocked in RSpec.

**Goal: I can TDD anything**
Test coverage in this challenge is lower than 95% because I couldn't test Twilio in order for RSpec not to send SMS. RSpec syntax is getting easier and more familiar.

**Goal: I can debug anything**
As my program was initially more complicated, it got to the point where I couldn't solve my bugs and I had to get rid of not requested functionality to achieve the weekend goal.

**Goal: I can refactor anything**
After implementing a more complicated program than I had to, I applied the YAGNI principle and refactored the entire program getting rid of all what was not requested.

**Goal: I can learn anything by myself**
Learning how to use Twilio API was a challenge and teamwork helped.


lib/takeaway.rb
```diff
@@ -0,0 +1,38 @@
...
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @new_order_list = Array.new
+    @total_cost = 0
+  end
+
+  def show_menu
+    @menu
+  end
+
```
thomasmarano: comment about your 'show_menu' method^:

I put this method in my menu class. I'm not sure if this is necessarily better, but I felt the functionality of 'displaying the menu' was more in line with a menu class with a hardcoded menu, than a takeaway/order class. As it stands, you're menu class doesn't really have much functionality beyond displaying the menu (i.e. there are no other methods/you can't change the menu there) so you might as well have just inserted it as a constant in your takeaway class. By adding the 'display' menu method to it, it seems to have more of a function

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'takeaway'
+
+RSpec.describe Takeaway do
+
+  context "#show_menu" do
+    it "should print the menu" do
+      expect(subject.show_menu).to be_kind_of Menu
+    end
+  end
+
```
thomasmarano: I'm not sure what a better practice would be, but I'm not sure this is necessarily the best test for your 'show_menu' method. I understand that, given the current objectives of your program, and given the fact that @menu initializes with 'Menu.new' (i.e. an instance of the method class), if your program works successfully, this test will pass. But I feel like this isn't testing the show_menu in as much isolation as it could be (i.e. you're testing not only that show_menu returns @menu, but also that @menu is an instance of the Menu class). Additionally, because it is possible to initialize the Takeaway class with menu equal to something other than the Menu.class, if you end up changing the name of the menu class or making some other change there, it will also mistakenly throw off your show_menu class

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  context "#place_order" do
+    it "should return a hash of quantity and Dish instance" do
+      expect(subject.place_order(5, "Cheeseburger")).to eq(33.95)
+    end
+
```
thomasmarano: I also did this in my program but I think there must be a better way to test the Place_Order method. As it stands, for this test to pass, both the place_order method & the update_total_cost method must work. This test therefore is testing 2 methods in 1 test (so if it fails, that could be because update_total_cost is broken or it could be because place_order is broken)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it "should return multiple orders in a hash" do
+      subject.place_order(2, "Cheeseburger")
+      expect(subject.place_order(2, "Chicken Wings")).to eq(25.56)
+    end
+  end
+
```
thomasmarano: The above comment also applies to this test

PR #1200, opened 2019-01-20
Takeaway Challenge - Jess NS Leung

jesslns:
Still working on testing the text function.


jesslns:
Revised code to further delegate tasks to Order and better use of mocking and stubbing in my tests.

lib/takeaway.rb
```diff
@@ -0,0 +1,28 @@
...
+attr_reader :menu, :order, :text
+def initialize(menu=Menu.new, order=Order.new, text=Text.new)
+  @menu = menu.read
+  @order = order
+  @text = text
+end
+
+  def add_order(dish,number)
+    @order.add_order(dish,number)
+    print "#{number}x #{dish.to_s}(s) is added to your basket"
```
alicelieutier: You have `puts` and `prints` in many places in your app. That means that the app would hard to adapt to a different interface. Is there a way you can extract the printing to something else?

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_reader :dish
+
+  DEFAULT_MENU = { margherita: 8, romana: 10, four_cheese: 12, seafood: 14, garlic_bread: 5 }
+
+  def initialize (dish=DEFAULT_MENU)
+    @dish = dish.sort
+  end
+
+  def read
+    @dish.to_h
```
alicelieutier: I'm wondering why you are doing this here.
You may want to read this to understand better how hashed and arrays differ, especially around ordering: http://www.rubyinside.com/how-to/ruby-sort-hash

PR #1204, opened 2019-01-20
Working version of the takeaway challenge without the sms message receiving feature.

JPProgrammingTime:


spec/list_of_dishes_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:fish_dish) { double("fake_fish") }
+  let(:toast_dish) { double("fake_toast") }
+
+  before(:each) do
+    @list_of_dishes = ListOfDishes.new
+  end
+
+  it "should be an instance of ListOfDishes" do
+    expect(@list_of_dishes).to be_instance_of(ListOfDishes)
+  end
```
kaihoffman: I know I"m not the expert on this, but is this an example of testing state? You create a thing and then immediately test that it is an instance of that class. Very clear and descriptive names though!

lib/order_processor.rb
```diff
@@ -0,0 +1,38 @@
...
+class OrderProcessor
+
+  attr_reader :list_of_dishes, :orders, :number_of_orders, :overall_price, :sms_message_sender
+
+  def initialize(list_of_dishes = ListOfDishes.new)
+    @list_of_dishes = list_of_dishes
+    @orders = {}
+    @number_of_orders = 0
+    @overall_price = 0
+    @sms_message_sender = SMSMessage.new
```
kaihoffman: Nice use of dependency injection within the method here.

lib/order_processor.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def add_order(order_number, quantity)
+    if @orders.key?(order_number.to_s.to_sym)
```
kaihoffman: Is it necessary for you to go to string and then to symbol? Would it not be able to convert straight to a symbol?

PR #1205, opened 2019-01-20
Takeaway challenge solution

thielsen:
Complete solution for the takeaway challenge

- I can TDD anything

Took the usual red/green/refactor approach with a bit of overall refactoring at the end. Use of Instance and Object doubles to ensure texts aren't sent during testing while maintaining 100% coverage. Injecting of dependencies meant I could inject these doubles into the Takeaway application.

- I write code that is easy to change

I've been reading 'Practical Object Oriented Design in Ruby' and have tried to apply the principles I've been learning, injecting dependencies so they can be more easily changed, following the single responsibility model and using hashes to pass arguments when larger numbers (for Twilio for instance) are required to remove the dependency on argument order/completeness.

Didn't quite get as much time as I'd have liked this weekend so think there could be some more refactoring to get methods to a smaller size..


lib/takeaway.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def check_total(given_total)
+    true if given_total == calculate_total
+  end
+
+  def complete_order(given_total)
+    raise 'Total incorrect' unless check_total(given_total)
+
+    @sms = @texter.new({
```
ellieturno: could this be extracted to shorten the method?

PR #1212, opened 2019-01-20
Solution

vaithschmitz:
### I HAVE A METHODICAL APPROACH TO SOLVIN PROBLEMS 

> There was a fair amount of planning, scheduling and diagramming before writing a single character of code. I've used several models (inc. UML, dependency graphs with a step component) to ensure I could visualize the product as a whole, while gaining better visibility on the actual program-critical components. 

### I CAN LEARN ANYTHING MYSELF

> Lots of research was conducted into all things APIs, which ended up as a much broader reflection on OSI model (esp. application layer) and different relevant protocols to ensure security and reliable client-server requests.

### I WRITE CODE THAT IS EASY TO CHANGE

> My codebase is very modular without being fragmented. Precautions were taken that pragmatic entities stayed intact while putting a strong emphasis on SOLID principles.

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,73 @@
+require 'twilio-ruby'
+
+class Order
+attr_reader :menu, :order, :total, :total_items, :meal_choice, :meal_count, :total_cost
+
+def initialize
+  @menu = {Pizza: 10, Pasta: 7.50, Cheesesteak: 8.50, Fries: 3.50, Donut: 2.50}
+  @order = []
+  @total_cost = 0
+  @total_items = 0
+end
+
+def take_order
+  p "What would you like to order?"
+  @menu.map{|key,value| p "#{key.to_s}: £#{value}"}
+  meal_choice = gets.chomp.capitalize.to_sym
+  if !@menu.include?(meal_choice)
+    p "Unfortunately we're out of your selected product. Please choose one of the following:"
+    take_order
+  else
+    p "How many #{meal_choice} would you like?"
+    meal_count = gets.chomp.to_i
+    create_order(meal_choice, meal_count)
+  end
+  p "Enter 1 to continue ordering or 9 to checkout!"
+    answer = gets.chomp.to_i
+    until answer == 1 || answer == 9
+      answer = gets.chomp.to_i
+    end
+  answer == 1 ? take_order : show_order     
+end
+
+def create_order(meal_choice, meal_count)
+  dish_total = @menu[meal_choice] * meal_count 
+  order << [meal_choice.to_s, meal_count, dish_total]
+  calc_total(order)
+end
+
+def calc_total(order)
+  order.map{|x,y,z| @total_items += y; @total_cost += z}
+end
+
+def show_order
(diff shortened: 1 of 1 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,73 @@
+require 'twilio-ruby'
+
+class Order
+attr_reader :menu, :order, :total, :total_items, :meal_choice, :meal_count, :total_cost
+
+def initialize
+  @menu = {Pizza: 10, Pasta: 7.50, Cheesesteak: 8.50, Fries: 3.50, Donut: 2.50}
```
david-mears: can you make it such that the menu isn't hard-coded into the program, eg by using a Menu class?

lib/takeaway.rb
```diff
@@ -0,0 +1,73 @@
+require 'twilio-ruby'
+
```
david-mears: Would be good to have tests!
Run rubocop to pick up 33 suggestions for improving the code.

lib/takeaway.rb
```diff
@@ -0,0 +1,73 @@
...
+  p "What would you like to order?"
+  @menu.map{|key,value| p "#{key.to_s}: £#{value}"}
+  meal_choice = gets.chomp.capitalize.to_sym
+  if !@menu.include?(meal_choice)
+    p "Unfortunately we're out of your selected product. Please choose one of the following:"
+    take_order
+  else
+    p "How many #{meal_choice} would you like?"
+    meal_count = gets.chomp.to_i
+    create_order(meal_choice, meal_count)
```
david-mears: I like this solution to user input (two separate questions)

lib/takeaway.rb
```diff
@@ -0,0 +1,73 @@
...
+
+def text_confirm
+  p "Enter SMS if you would like a text to confirm your order, else just press enter"
+  text = gets.chomp.upcase
+  text == "SMS" ? send_sms : exit
+end
+
+def send_sms
+  account_sid = ENV["TWILIO_SID"]
+  auth_token = ENV["TWILIO_TOKEN"]
```
david-mears: +1 for using environment variables

lib/takeaway.rb
david-mears: > For example the menu items should not be hard coded into a restaurant class. Arguably they should not be in the business logic at all, e.g. being added at runtime (i.e. in IRB) or loaded from an external hash or maybe even a file. We are concerned that the menu items will likely change

lib/takeaway.rb
```diff
@@ -0,0 +1,73 @@
...
+attr_reader :menu, :order, :total, :total_items, :meal_choice, :meal_count, :total_cost
+
+def initialize
+  @menu = {Pizza: 10, Pasta: 7.50, Cheesesteak: 8.50, Fries: 3.50, Donut: 2.50}
+  @order = []
+  @total_cost = 0
+  @total_items = 0
+end
+
+def take_order
```
david-mears: (from rubric) > You may find it difficult to remember the indentation rules, but one helpful rule of thumb for indentation is to ensure that you are two space indented inside any do ... end, class ... end or def ... end block.

PR #1215, opened 2019-01-20
wpdunk takeaway submission

wpdunk:
Implemented the user stories, but confused myself a lot in the process, struggled with writing tests and class delegation.

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,9 @@
+class Menu
+
+  attr_reader :dishes
+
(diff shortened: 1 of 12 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+class Menu
+
+  attr_reader :dishes
+
+  def initialize(dishes)
+    @dishes = dishes
```
dangroze: I like that you didn't hard-code the list of dishes!

PR #1217, opened 2019-01-21
Finished weekend challenge

david-mears:
See the README.md for explanation of how to use.

**I write code that is easy to change** - plenty of dependency injection. But I'm not sure if anyone but me can understand my method names in Order class (?)
**I can TDD anything** - I feel I'm getting a bit more sophisticated in rspec, testing unusual things like stdout and stdin, though I'm sure you can spot things I've done suboptimally. High test coverage. Still need to learn how to send different things to a mocked `gets` within a single test.
**I can refactor anything** - I realised for [this commit](https://github.com/makersacademy/takeaway-challenge/commit/2e4abeb10cdc4494d8f5713a29e9a49622d69165) that I needed to make MenuDisplay not dependent on a Menu object that had to somehow be remembered by Order. That was getting really complicated, so I refactored it so that MenuDisplay can be made anew every time you want to print something, without any need to pass it a menu object (it just eats hashes now).

### DIFF
--- a/lib/menu_display.rb
+++ b/lib/menu_display.rb
@@ -8,10 +8,15 @@ class MenuDisplay
   end
 
   def list_menu
+
   end
 
   def price_to_sterling(item)
     '£' + menu[item].to_s
+  end
+
+  def print_item(item)
+    item.to_s.split('_').each{ |item| item.capitalize! }.join(' ')
   end
 
 end
--- a/lib/menu_display.rb
+++ b/lib/menu_display.rb
@@ -2,9 +2,9 @@ class MenuDisplay
 
   def list_menu(hash:)
     list = []
-    hash.each_with_index { |key_value_pair, index|
-      numerator = '#' + (index + 1).to_s + '.'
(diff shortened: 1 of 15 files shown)


lib/menu_display.rb
```diff
@@ -1,7 +1,17 @@
...
+  attr_reader :menu
 
+  def initialize(menu = Menu.new)
+    @menu = menu.menu
+  end
 
+  def list_menu
+  end
+
+  def price_to_sterling(item)
```
vaithschmitz: Could you use string interpolation instead of calling a function per item?

lib/order.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def take_order
+    puts menu_display_class.new.list_menu
+    puts ORDER_MESSAGE
+    get_input
+  end
+
+  def get_input
+    loop do
```
vaithschmitz: Could you shorten this function by using a different type of loop? Maybe try using a while loop.

PR #1222, opened 2019-01-21
Takeaway Challenge - Partial solution

aidengleave:
**I can TDD anything**

I feel that while my understanding of TDD has improved in the last week, I am unsure as to whether I have written enough tests to properly cover the implemented code, even though I have managed to keep the coverage at 100%.

Other than this, I am becoming more used to using doubles and how to call methods in RSpec to test functionality.

I plan to try to continue with this project as I found the idea of using Twilio interesting and also feel that the print functions should have its own class, but ran out of time when trying to implement the class extraction.

**I have a methodical approach to solving problems**

I have been trying to incorporate modelling into my testing and writing process and am increasingly happy with maintaining a tighter cycle when I don't feel I'm understanding something properly... git stash-ing any changes, going back to a clean commit where I feel I understood what was happening and trying a different approach.

However, I am a little frustrated that this is the second weekend challenge I haven't managed to complete. While I didn't take as much of a psychological back-step as I did last week. 
I definitely found it difficult to motivate myself to make a start on Saturday, and still had my fair share of self doubt.

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,2 @@
+class Menu
+end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,16 +1,22 @@
 class Menu
   attr_reader :menu_items
 
-  FOOD = {pizza: 2.99, burger: 3.50, burrito: 4.50, 
-  lobster: 50.00}
+  FOOD = { pizza: 2.99, burger: 3.50, burrito: 4.50, 
+  lobster: 50.00 }
 
   def initialize(menu_items = FOOD)
     @menu_items = menu_items
   end
 
   def print
-    puts "Takeout Menu"
-    @menu_items.each {|food, price| puts "#{food.capitalize} £#{'%.2f' % price}" }
+    puts "Takeout Menu".center(30)
+    formatted_menu
   end
 
+  private
+  def formatted_menu
+    @menu_items.each do 
+      |food, price| puts "#{food.capitalize} £#{'%.2f' % price}".center(30)
+    end 
+  end
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,14 +1,16 @@
 class Menu
   attr_reader :menu_items
 
-  FOOD = {}
+  FOOD = {pizza: 2.99, burger: 3.50, burrito: 4.50, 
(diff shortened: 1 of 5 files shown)


spec/features/feature_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'menu'
+require 'order'
+
+RSpec.describe 'features' do
```
ch359: Potentially should have specific RSpec tests in here (rather than implictly testing to see if Ruby crashes) - i.e. expect statements.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader :menu_items
+
+  FOOD = { pizza: 2.99, burger: 3.50, burrito: 4.50, 
+  lobster: 50.00 }
+
+  def initialize(menu_items = FOOD)
+    @menu_items = menu_items
+  end
+
+  def print
```
ch359: Just checking that you realise the print and formatted_menu methods are both inaccessible to the user (as menu is injected into order class); so no way to view what's on offer to place an order without separately creating a new menu class. 

Consider having a method in Order that delegates responsibility to Menu? 
```suggestion
(in order.rb)
  def print
    menu.print
  end
  
```

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class Order
+  attr_reader :take_order, :menu
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @take_order = []
+  end
+
+  def customer_order(item, amount)
```
ch359:
```suggestion
  def customer_order(item, amount = 1)
```
Setting a default would mean the customer doesn't need a quantity

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+    @take_order << { item: item, amount: amount }
+  end
+
+  def total_cost
+    total = 0
+
+    @take_order.each do |food_item|
+      total += @menu.menu_items[food_item[:item]] * food_item[:amount]
```
ch359: This bombs out when the user enters a string - consider a `to_sym` when taking customer orders (which will probably be a string if it comes via `gets` for example. The customer doesn't know to use a method otherwise.

PR #1224, opened 2019-01-21
Partial Solution

thomasmarano:


lib/order.rb
```diff
@@ -0,0 +1,42 @@
+require 'meatdish'
+require 'menu'
+
+class Order
```
Debora38: creating a class Order makes sense!

lib/order.rb
```diff
@@ -0,0 +1,42 @@
...
+      @basket = {}
+      @balance = 0
+    end
+
+
+    def view_menu
+        @menu
+    end
+
+    def add_food (dish, quantity = 1)
```
Debora38: quantity default to 1. good idea!

lib/order.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+    end
+
+    def add_balance(sum)
+        @balance += sum
+    end
+
+    def check_out(total)
+        raise "Incorrect amount" if total != @balance
+
```
Debora38: Twilio is missing however a string saying 'order confirmed' would make your program functional for now.

send_sms.rb
```diff
@@ -0,0 +1,14 @@
+require 'twilio-ruby'
+
+account_sid = 'ACaa3e4ba30b28658a9d7b1a051b424a1f'
+auth_token = 'f09ba6909500522a26c1fdb6a2d6a823'
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+from = '+441359366051' # Your Twilio number
+to = '+447779889503' # Your mobile phone number
+
+client.messages.create(
```
Debora38: All you're missing is to complete these lines below, give them a method name and pass it on to your program. class (or module) name on top as well

lib/order.rb
```diff
@@ -0,0 +1,42 @@
+require 'meatdish'
+require 'menu'
+
+class Order
+    attr_reader :menu
+    attr_accessor :balance
+
+    def initialize
+      @menu = {
```
Debora38: As you have a Menu class, you should have the hardcoded menu in there

PR #1237, opened 2019-02-17
Main functionality completed and tested

simon-norman:


lib/dish.rb
```diff
@@ -0,0 +1,8 @@
+class Dish
```
Harrison25000: Nice, I like this class.

PR #1241, opened 2019-02-17
Takeaway challenge

Brendao1:


lib/takeaway.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'menu'
+require_relative 'order'
+
+class Takeaway
+
+  attr_reader :menu, :order
+
+  def initialize(menu:, order: nil)
```
Brendao1: If I change this to def initialize(menu, order: nil), 
I can test this in IRB with the following: t = Takeaway.new(menu = "chicken")

PR #1246, opened 2019-02-17
Completed to an extent

Harrison25000:
Runs in IRB & sends the text however, fails in rspec due to the require 'Twilio-ruby' feature failing, didn't know how to fix this so my rspec testing is lacking.

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,24 @@
+require 'takeaway'
+
+class Ordermeal
+  attr_reader :dishes
+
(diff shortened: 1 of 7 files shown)


lib/order.rb
```diff
@@ -0,0 +1,24 @@
+require 'takeaway'
+
+class Ordermeal
+  attr_reader :dishes
+
```
simon-norman: Might be clearer to use named variables rather than single letters

PR #1247, opened 2019-02-18
Awesome challenge

dafuloth:
I love that I am getting a taste of integrating external services by using their API. The free Twilio trial worked but seemed to stop sending me SMS after 7 messages - and then started working after 40 minutes.

I hope whoever I do code review with is better with RSpec syntax than I am! I wasn't able to get it to test that an instance method was able pass it's parent object to the initializer of another class.

I am having a better time with weekend challenges now that I'm pacing myself better - and not sinking so much time into RSpec minutiae.


README.md
```diff
@@ -1,82 +1,180 @@
...
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
------
-
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+# Weekend Challenge 2: Takeaway
+
```
clarepins: My overarching comment is to focus on the SRP for both classes and methods. They are doing a lot. When I looked at my mentor's code, he said we should be aiming to get to a place where each method is usually 1 line and just calls in a bunch of other methods.

lib/submission.rb
```diff
@@ -0,0 +1,32 @@
...
+    @order = order
+
+    error_message = "Add items to order first. Cannot submit empty order."
+    raise error_message if order.total.zero?
+
+    submit_order
+  end
+
+  def submit_order
+    sms_text = "Thank you! Your order was placed at 17:52 and should be with you by 18:52."
```
clarepins: You can use Time.new to send a msg with the actual time.

lib/submission.rb
```diff
@@ -0,0 +1,32 @@
+require 'dotenv/load'
+require 'twilio-ruby'
+
+class Submission
+  def initialize(order = Order.new)
```
clarepins: What's the benefit of having a default to start a new order when you submit the previous order? Sorry if I'm missing something.

lib/menu.rb
```diff
@@ -0,0 +1,33 @@
...
+    { name: "Potato Wedges", price: 2.00 },
+    { name: "Chocolate Cake Slice", price: 1.50 },
+    { name: "Coconut Cake Slice", price: 1.50 }
+  ]
+
+  def list_dishes
+    list_of_dishes = ""
+
+    DISHES.each_with_index do |dish, index|
+      list_of_dishes << "#{index + 1}. #{dish[:name]} - \u00A3%0.2f\n" % dish[:price]
```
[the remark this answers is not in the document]
dafuloth: Agreed. For some reason I was concerned that Ruby might not be able to work with the '£' character. But I've checked and `"\u00A3" == "£" => true`

README.md
dafuloth: Definitely many possibilities for refactoring

lib/submission.rb
dafuloth: dependency injection, or I could be doing it wrong...

lib/submission.rb
dafuloth: Good idea

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+
+class Order
+  attr_reader :total, :selected_dishes
+
+  def initialize
+    @selected_dishes = Array.new
+    @total = 0
+  end
+
+  def add(dish, quantity = 1)
```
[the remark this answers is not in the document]
dafuloth: Definitely many good candidates for refactoring

PR #1249, opened 2019-02-18
Takeaway-challenge

SurekaStoeckigt:


spec/order_spec.rb
```diff
@@ -0,0 +1,142 @@
...
+  dishes_class_double = double :dishes_class, new: dishes_double
+  hash_double = double :hash
+
+  allow(dishes_class_double).to receive(:new).and_return(dishes_double)
+  allow(dishes_double).to receive(:dish_list).and_return(list)
+
+  order = Order.new(dishes_class_double)
+  expect(order.has_dish(food_item)).to eq true
+end
+
```
Schofield88: Would declare dishes_double and dishes_class_double with let at the top of the describe block and keep reusing them. The code would DRY out.

lib/order.rb
```diff
@@ -0,0 +1,98 @@
...
+    @ordered_items_price.each { |counts|  @ordered_food_price[counts] += 1 }
+    p "You have ordered:"
+    @ordered_food_name.each do |dish, counts|
+       p "#{dish}: #{@sorted_by_name[dish]} x #{counts}"
+    end
+  end
+
+end 
+
+
```
Schofield88: This was a long class. Just needs a bit of extraction.

PR #1250, opened 2019-02-18
Takeaway-challenge

mmar8:
Not finished, but did as much as I could.

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,33 @@
+class Restaurant
+    attr_reader :menu, :order
+
+    def initialize
+        @menu = { 'burger': 5,
+                  'pizza': 6,
+                  'lasagna': 4,
+                  'ice_cream': 2
+                }
+        @order = []
+    end 
+
+    def menu_print
+        @menu.each { |key, value| puts "#{key}: £#{value}" }
+    end 
+
+    def add_item(item_from_menu)
+        @order << item_from_menu
+    end 
+
+    def include?(item)
+        @menu.include?(item)
+    end
+
+    def  price(item)
+        @menu.fetch(item).to_f
+     end 
+    
+     def total 
+        @total = []
+        @order.each { |hash| @total << hash}
+
+end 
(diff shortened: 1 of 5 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
...
+        @menu = { 'burger': 5,
+                  'pizza': 6,
+                  'lasagna': 4,
+                  'ice_cream': 2
+                }
+        @order = []
+    end 
+
+    def menu_print
+        @menu.each { |key, value| puts "#{key}: £#{value}" }
```
steve-f-kahn: Is there a way you could output the price in correct format £6.00. You could try '%.2f' % value

lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
...
+    end 
+
+    def add_item(item_from_menu)
+        @order << item_from_menu
```
steve-f-kahn: The user wants to be able to add multiple of the same item at the same time. That is not possible at the moment. How could you do this? Anything they pass will go onto the order. It doesn't matter what it is. This will break your code later when they try to total the order as they are not inputing a hash with an integer assigned as the value. Given more time it is important to test for edge cases.

lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
...
+    end
+
+    def  price(item)
+        @menu.fetch(item).to_f
+     end 
+    
+     def total 
+        @total = []
+        @order.each { |hash| @total << hash}
+
```
steve-f-kahn: def total is missing an end, your code can't be working. The 'total' method is not tested at all. This is a reason why we write test before code because if we run out of time our production code (the code we function test) still works.

lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
+class Restaurant
+    attr_reader :menu, :order
+
+    def initialize
+        @menu = { 'burger': 5,
+                  'pizza': 6,
+                  'lasagna': 4,
+                  'ice_cream': 2
+                }
+        @order = []
```
steve-f-kahn: Are you going to create a new restaurant for each order? I suggest a new class is needed for the customer.

PR #1255, opened 2019-03-16
Takeaway challenge completed without bonus task.

saypop:


README.md
```diff
@@ -12,71 +12,93 @@ Takeaway Challenge
...
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+## Acknowledgments
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this weekend.
+* Shout out to my peeps over at Twilio - nice looking out yo (https://www.twilio.com).
+* Big up to the Mad March squad over at Makers Academy - I couldn't see you but I felt you the whole way along this journey.
+* Huge thanks to meditation - namaste, jai ma, om shanti shanti, mo' fire.
+* Mad respect to da real OG... my brain - you're taking some hard knocks like a champ and you keep getting up for more, fo real tho homie I don't know where I'd be without you. Much love.
```
Hives: 🤣

lib/basket.rb
```diff
@@ -0,0 +1,54 @@
...
+    { :name => item, :price => price, :quantity => quantity }
+  end
+
+  def get_quantity(item)
+    quantity = 1
+    quantity += @basket[get_index(item)][:quantity] unless new_item(item)
+    quantity
+  end
+
+  def get_index(item)
```
justcallmegoyo: Your methods are all concise and you are using delegation habitually.

spec/takeaway_menu_spec.rb
```diff
@@ -2,10 +2,10 @@
 
 describe TakeAwayMenu do
 
-  let(:subject) {described_class.new}
+  let(:subject) { described_class.new }
 
   it 'contains a readable menu' do
-      expect(subject.read["spring roll"]).to eq 0.99
+    expect(subject.read["spring roll"]).to eq 0.99
```
justcallmegoyo: You use the read method here.  Is this widely available in other programming languages?

spec/takeaway_spec.rb
```diff
@@ -2,17 +2,37 @@
 
 describe TakeAway do
 
-  let(:menu_class_double) {double(:menu_class)}
-  let(:menu_double) {double(:menu)}
-  let(:subject) {described_class.new(menu_class_double, menu_double)}
+  let(:menu_class) { double(:menu_class) }
+  let(:menu) { double(:menu) }
+  let(:basket) { double(:basket) }
```
justcallmegoyo: You set up doubles appropriately

spec/takeaway_spec.rb
```diff
@@ -2,17 +2,37 @@
...
+      expect(basket).to receive(:add).with(dish)
+      subject.add_to_basket(dish)
+    end
+
+  end
+
+  describe '#view_basket' do
+
+    it 'displays the basket' do
+      expect(basket).to receive(:view)
```
justcallmegoyo: Doubles deployed effectively.

README.md
```diff
@@ -12,71 +12,93 @@ Takeaway Challenge
...
-As a customer
-So that I can check if I want to order something
-I would like to see a list of dishes with prices
 
-As a customer
-So that I can order the meal I want
-I would like to be able to select some number of several available dishes
+And you'll need Twilio.
+```sh
+$ gem 'twilio-ruby'
```
Hives: add this to your Gemfile so users can install the whole program with `bundle install`

Hives: it'll also make sure that compatible versions of all gems are installed

lib/messager.rb
```diff
@@ -0,0 +1,37 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+
+class Messager
```
justcallmegoyo: Messenger?

lib/messager.rb
```diff
@@ -0,0 +1,37 @@
...
+    @receive_number = customer_number
+
+  end
+
+  def send_sms(message)
+
+    unless message.nil? || @send_number.nil?
+
+      account_sid = 'AC97abd4b506ad5d4ec59ca886227a21ea'
+      auth_token = '6e48f0377408fd3c364daad172641952'
```
justcallmegoyo: Remove this from public view?

README.md
```diff
@@ -12,71 +12,93 @@ Takeaway Challenge
...
+```
```
Hives: this is in your Gemfile, so you should say so! you can install everything with `bundle install`

lib/messager.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    client.messages.create(
+    from: from,
+    to: to,
+    body: message
+    )
+
+  end
+
+end
```
Hives: can't see any tests for this class! you should be able to test it by using a double for the Twilio client

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  describe '#complete_order' do
+
+    before do
+      allow(messager).to receive(:send_sms)
+    end
+
+    it 'sends an sms' do
+      expect(messager).to receive(:send_sms).with("Thank you! Your order was placed and will be delivered before 06:09PM")
+      subject.complete_order(phone_number, messager, "06:09PM")
```
Hives: this passes three arguments, but the method only takes two!

lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
...
+    @basket.add(dish)
+    "#{dish} added to basket"
+  end
+
+  def view_basket
+    @basket.view
+  end
+
+  def complete_order(customer_number, messager = Messager.new(customer_number),
+    time_in_30 = (Time.now + 30 * 60).strftime("%I:%M%p"))
```
Hives: better to indent this line so it aligns with the first parameter on the line above to make it clearer that you've split the parameters over two lines

spec/takeaway_spec.rb
Hives: because you're sending in a double for time_in_30, you're not testing that the default setting works.

lib/basket.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def view
+    "Your basket total is #{basket_total} and it contains:\n#{basket_list}"
+  end
+
+  def add(item, price = @menu.read[item], quantity = get_quantity(item))
+    @basket << hash(item, price, quantity)
+  end
+
+  private
```
Hives: you're getting low test coverage on this class because it's saying that some of your private methods aren't covered. although it's right not to test private method directly you should still be able to test their behaviour indirectly by writing appropriate tests for the public methods where they're used... i think.... for instance respec is saying that your private methos `basket_total`, `basket_list` and `compose_line` are covered, even though you haven't tested them directly. so i think you should be able to get coverage for the other ones. maybe bring this up with katerina...

lib/basket.rb
Hives: yes, lots of nice, short focussed methods 👍

Hives:
unit tests for each class still passed when i commented out the other classes, so looks like you've doubled out dependent behaviour well

PR #1258, opened 2019-03-17
Sam - takeaway challenge

samanthaixer:
I must confess to having a takeaway mid-way through this challenge - power of suggestion!

### DIFF
--- a/lib/dish.rb
+++ b/lib/dish.rb
@@ -1,0 +1,25 @@
+class Dish
+  def initialize(menu_number, description, price, available = true)
+    @menu_number = menu_number
+    @description = description
+    @price = price
+    @available = available
+  end
+
+  def display
+    if available?
+      "#{@menu_number}. #{@description} £#{@price}"
+    else
+      "Dish unavailable"
+    end
+  end
+
+  def make_unavailable
+    @available = false
+  end
+
+  private
+  def available?
+    @available
+  end
+end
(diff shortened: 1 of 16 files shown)


lib/dish.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :description, :price
+  def initialize(menu_number, description, price, available = true)
+    @menu_number = menu_number
+    @description = description
+    @price = price.to_f
+    @available = available
+  end
+
+  def display
+    if available?
```
mattTea: would a guard clause have been possible to make this method a bit shorter?

lib/message.rb
```diff
@@ -0,0 +1,36 @@
...
+    "Thank you! Your order was placed and will be delivered before #{arrival}"
+  end
+
+  private
+
+  def connect(text_content)
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
+    client = Twilio::REST::Client.new(account_sid, auth_token)
+
```
mattTea: great use of env variables!

lib/order.rb
```diff
@@ -0,0 +1,20 @@
...
+  def place(order_list, total_cost)
+    order_cost = order_list.calculate_cost(@menu_items)
+    check_total(order_cost, total_cost.to_f)
+  end
+
+  private
+
+  def check_total(order_cost, total_cost)
+    raise "order cost does not match expected cost" if order_cost != total_cost
+
```
mattTea: is this checking a user entered value?

PR #1259, opened 2019-03-17
James Palmer pull requests

jpalmerr:
3 user stories implemented.


lib/menu_list.rb
```diff
@@ -0,0 +1,8 @@
+class List
+
+  attr_reader :list
+
+  LIST = { "Margherita" => 6, "Pepperoni" => 7,
```
jpalmerr: Using variables makes it difficult to be flexible in the future, say for example want to change in future. Can we put it in the initialize instead?

PR #1260, opened 2019-03-17
takeaway challenge solution - Matt Thompson

mattTea:


README.md
```diff
@@ -1,5 +1,99 @@
-Takeaway Challenge
-==================
+Matt Thompson - Takeaway Challenge
```
samanthaixer: Nice clear README with good description and instructions

lib/menu.rb
```diff
@@ -0,0 +1,28 @@
...
+  def initialize
+    @dishes = []
+    @orders = []
+  end
+  
+  def add_dish(name, price, dish_class = Dish)
+    @dishes << dish_class.new(name, price)
+  end
+
+  def view_dishes
```
samanthaixer: Would you consider adding a String version of view so that it prints a menu?

spec/menu_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "menu"
+
+describe Menu do
+  subject { Menu.new }
+  
+  let(:dish) { double :dish }
+  let(:dish_class) { double :dish_class, new: dish }
+
+  context "when looking for dishes" do
+    it { is_expected.to respond_to(:view_dishes) }
```
samanthaixer: This could be removed as your next test also tests this behaviour

spec/menu_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    it "displays dishes" do
```
samanthaixer: Good use of dependency injection

lib/order.rb
```diff
@@ -0,0 +1,49 @@
+require_relative "send_sms"
+require 'dotenv/load'
+
+class Order
+  def initialize(dish, quantity = 1)
+    @selected_dishes = []
+    quantity.times { @selected_dishes.push(dish) }
+    @confirmed = false
+    view_total
```
samanthaixer: Does this call the method at initialise? If so, is this to set up the @total variable?

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def view_total
+    total = 0
+    @selected_dishes.each do |dish|
+      total += dish.price
+    end
+    @total = total
+  end
+
+  def confirm_order(
```
samanthaixer: This seems to be accessed via the menu and the orders variable. How does this work? Is this a case of applying a method from order on a property owned by Menu?

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    _to = ENV['TWILIO_TO_NUMBER'],
+    message_class = Sms,
+    time = calculate_time
+  )
+    sms = message_class.new
+    @confirmed = true
```
samanthaixer: Do you check the total of the order against a passed-in total? "Place the order by giving the list of dishes, their quantities and a number that should be the exact total. If the sum is not correct the method should raise an error"

spec/order_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      subject.confirm_order(to, message_class)
+      expect(subject).to be_confirmed
+    end
+
+    it "responds to sms send method" do
+      expect(sms).to respond_to(:send)
+    end
+
+    before do
+      allow(sms).to receive(:send)
```
samanthaixer: I don't think you need to allow this as well as expecting it in the test below?

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+    @total = total
+  end
+
+  def confirm_order(
+    _to = ENV['TWILIO_TO_NUMBER'],
+    message_class = Sms,
+    time = calculate_time
+  )
+    sms = message_class.new
```
samanthaixer: I like how you've made the type of message able to change, but your variable name suggests SMS. Could the variable name change to something more generic?

PR #1264, opened 2019-03-17
Takeaway Challenge - Submission - I documented my approach. Lots to improve on as always :)

kimdiep:


lib/takeaway.rb
```diff
@@ -0,0 +1,8 @@
+class Takeaway
+
+
+  def see_dishes
+    "Sweet and Sour Chicken: £4.00"
```
andrewmclelland23: Good and clear use of TDD, making the test pass as simply as possible.

lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
...
+    @food_menu = food_menu
+    @order_management = order_management
+    @message = message
+  end
+
+  def see_dishes
+    food_menu.print
+  end
+
+  def place_the_order(selected_dishes)
```
andrewmclelland23: There was a requirement to pass in a total value here which should return an error if the total didn't match the calculated total. This could be something to add if you revisit the code

lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
...
+    selected_dishes.each do |dish, number|
```
andrewmclelland23: For SRP, you could extract the each loop that adds the orders into a private method and then call this in your place_order method. This would also make the method more readable.

spec/food_menu_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'food_menu'
+
+describe FoodMenu do
+
```
andrewmclelland23: Good use of doubles to encapsulate tests

spec/food_menu_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  it 'contains a list of yummy dishes with their prices' do
+    expect(food_menu.food).to eq(food)
+  end
+
+  context '#print' do
+
+    it 'prints the list of yummy dishes with their prices' do
+      expect(food_menu.print).to eq("pizza: 5.00")
+    end
+
```
andrewmclelland23: Would be good to add tests that tested adding more than one dish and printing more than one dish to ensure they work as expected.

PR #1266, opened 2019-03-17
Takeaway challenge

EvelynAleLeo:


lib/takeaway.rb
```diff
@@ -0,0 +1,23 @@
+class Takeaway
+  
+  attr_reader :dishes, :order, :price
+
+  def initialize(items)
```
hybridbad: items could be delegated to a new Menu class object that holds a hash of dishes + prices

lib/takeaway.rb
```diff
@@ -0,0 +1,23 @@
...
+    @dishes = items
+    @order = []
+    @price = 0
+  end
+
+  def show_list
+    @dishes
+  end
+
+  def add_to_order(plate)
```
hybridbad: good use of key? to check dish is there

lib/takeaway.rb
```diff
@@ -0,0 +1,23 @@
...
+    @order.push(plate) if @dishes.has_key?(plate) && @price += @dishes[plate] 
+  end
+    
+  def total_price
```
hybridbad: you could remove the += dishes plate and insert into this method to separate the responsibility

spec/takeway_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'takeaway'
+
+describe Takeaway do
+
+# As a customer
+# So that I can check if I want to order something
+# I would like to see a list of dishes with prices
+
+  it 'should show list of dishes with prices' do
```
hybridbad: you could assign the dishes hash to a variable and call it where you need it

hybridbad: dishes = { "rice" => 1.25, "banana" => 1.00, "fries" => 2.10 }

subject(:takeaway) { Takeaway.new(dishes) }

then just do your tests without having to create a new takeaway object each time

lib/takeaway.rb
hybridbad: probably not - seems you use the total to verify order is correct by comparing total to the sum of dishes. Different predicate method?

PR #1268, opened 2019-03-17
first draft

alexanderchen6:


lib/menu.rb
```diff
@@ -0,0 +1,39 @@
+require 'order'
+require 'takeaway'
+
+class Menu
+  # Dish class here is like Menu of dishes
+  attr_reader :dishes
+
+  def initialize(dishes = [{ name: 'chinese', price: 8 },
```
micosmin: What do you think about not initializing a class with a data structure? It creates a hard dependency on that data structure

lib/menu.rb
```diff
@@ -0,0 +1,39 @@
...
+  attr_reader :dishes
+
+  def initialize(dishes = [{ name: 'chinese', price: 8 }, 
+  { name: 'fishandchips', price: 7 }, 
+  { name: 'chicken', price: 5 }, { name: 'beef', price: 6 }])
+    @dishes = dishes
+  end 
+
+  def display_list
+    @dishes.each do |x|
```
micosmin: Use descriptive names for local variables, instead of x

lib/menu.rb
```diff
@@ -0,0 +1,39 @@
...
+      puts "#{x[:name]} #{x[:price]}"
+     # how to make this a return statement and not puts
+    end 
+  end 
+
+  def dish?(dish)
```
micosmin: Maybe this method should be made private. It is only used by this class and maybe you don't need to have it publicly accessible

lib/order.rb
```diff
@@ -0,0 +1,29 @@
+require 'menu'
+
+class Order
+  attr_reader :dishes, :menu
```
micosmin: You could wrap these in methods so that you don't expose the state of the object

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(order = Order.new(menu), sms = nil , config = {})
+   
+    @order = order
+    @sms = sms || SMS.new(config)
+    # this is order object
+    # @total = 0
+  end 
+
+  def check_menu(all_dishes = Menu.new)
```
micosmin: Where is the Menu.new coming from? It has not been injected and the menu file has not been required in this file?

lib/menu.rb
```diff
@@ -0,0 +1,39 @@
...
+  # Dish class here is like Menu of dishes
+  attr_reader :dishes
+
+  def initialize(dishes = [{ name: 'chinese', price: 8 }, 
+  { name: 'fishandchips', price: 7 }, 
+  { name: 'chicken', price: 5 }, { name: 'beef', price: 6 }])
+    @dishes = dishes
+  end 
+
+  def display_list
```
micosmin: What do you think about extracting this to a Display class?

spec/menu_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'menu'
+require 'takeaway'
+
+describe Menu do 
+  subject(:menu) { described_class.new }
```
micosmin: Good use of named subjects

spec/menu_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  describe '#display_list' do 
+
+    it 'displays all the dishes and prices' do   
+      expect { subject.display_list }.to output("chinese 8\nfishandchips 7\nchicken 5\nbeef 6\n").to_stdout
```
micosmin: good application of testing behaviour and not state

micosmin: Because you have a named subject (:menu), you can use menu instead of subject

spec/order_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+require 'menu'
+
+describe Order do 
+  subject(:order) { described_class.new(menu) }
+
+  let(:dishes) do 
+    { chicken: 1, chinese: 2 }
+  end 
+
+  let(:menu) { double(:menu) }
```
micosmin: As you are injecting a menu inside an Order class, while this is good, maybe for readability if should be placed next to the subject(:order)

spec/order_spec.rb
```diff
@@ -0,0 +1,41 @@
+
```
micosmin: TESTING looks good!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  describe '#check_menu' do 
+    it 'displays list of dishes and prices' do 
+      
+      menu = menu_class.new
+      # how to use the double dish above, instead of creating a new double here. (it didnt work)
+      p menu
+      expect(menu).to receive(:display_list) 
+      # couldnt put dish_class.new here
+      subject.check_menu(menu)
```
micosmin: Is there any return tested here?

PR #1280, opened 2019-04-14
Takeaway Challenge - callum marshall

callum-marshall:
-program outline, with good test coverage
-needs sms functionality
-hoping to inject dependency if possible

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,12 @@
+class Menu
+
+  def list
+    {
+    burger: 6,
+    pizza: 10,
+    burrito: 7,
+    noodles: 8
+    }
+  end
+
+end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -2,10 +2,10 @@ class Menu
 
   def list
     {
-    burger: 6,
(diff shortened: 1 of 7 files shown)


CalumDarroch:
Your code makes good use of the Single Responsibility Principle. Methods and classes have clear responsibilities, and the methods themselves are short. I think the number of classes is just about right.

Good use of dependency injection for the `initialize` method on the Takeaway class. The use of quantities of items is good, allowing you to display multiples more neatly and total the cost.

You could maybe slim down the number of methods. For example, do `send`, `client_info` and `create(client)` need to be separate? The code currently invokes a lot of methods in sequence from just calling one from the terminal.

Also, as we saw from running your code in irb, the environment variables are printed to the console! Consider refactoring the methods to prevent this, or using the `dotenv` gem with an .env file(put in your .gitignore) in the challenge's root directory.

Test coverage is decent at ~90% but could be higher, and as you say there are a few tests you still need to write.

This is a good partial solution. With a bit of refactoring, upping the test coverage/completing tests and implementing a simple text interface I think this challenge would be rounded off for you.

PR #1282, opened 2019-04-14
Takeaway challenge

jordantroberts:


lib/menu.rb
```diff
@@ -13,4 +23,10 @@ def view_menu
     MAIN_MENU
   end
 
+  def place_order(dish, quantity)
```
tsankhalpara: I like how you have used the arguments in this method so that the user can input a quantity. I was trying to implement this in mine but could not find a way to link the use the quantity parameter to give me a total amount.

lib/menu.rb
```diff
@@ -1,32 +1,20 @@
...
-      "Tofu Katsu Curry" => 8,
-      "Prawn Tempura" => 5,
-      "Vegetable Tempura" => 5,
-      "Yakisoba" => 7,
-      "Miso Soup" => 2
-    }
-
-  def view_menu
-    MAIN_MENU
+  def initialize(dishes)
```
tsankhalpara: What was the reason for removing the full menu from here?

tsankhalpara:
Really great clean code and easy to understand. I like how each method was separated into different class and spec files. I would suggest moving sent_text to a private method so that you do not need to run a test on it.

PR #1284, opened 2019-04-14
Checking in partial solution

Xa1n:


lib/menu.rb
```diff
@@ -2,39 +2,19 @@
...
-  private
-
-  def main_list
-    MAINS.collect { |item, price| "#{item}: £#{price}0" }.join(", ")
-  end
-
-  def sides_list
-    SIDES.collect { |item, price| "#{item}: £#{price}0" }.join(", ")
+  def show
+    @options.each {|item,price| puts "#{item}: £#{price}0"}
```
NapperJLG: Definitely a way to print the list without printing the hash at the end (see Sarpong's code!)

lib/takeaway.rb
```diff
@@ -19,14 +19,15 @@ def menu
   def place_order(dish, quantity = 1)
     fail "Sorry we don't have that, check our menu" if !@menu.available?(dish)
 
-    @basket[dish.to_sym] += quantity
+    @basket[dish] += quantity
   end
 
   def total
-    @basket
+
```
NapperJLG: Maybe consider changing your basket to an array - seems easier to iterate over an array than a hash (or it was for me anyway)

spec/menu_spec.rb
```diff
@@ -2,12 +2,16 @@
...
 RSpec.describe Menu do
 
-  let(:mains) { Menu::MAINS }
-  let(:list) { subject.send(:sides_list) }
-
-  describe '#list' do
+  describe '#menu' do
     it 'prints a full list of available dishes' do
-      expect(subject.show).to include list
+      expect{subject.show}.to output.to_stdout
```
NapperJLG: Maybe stub the menu and expect to output the contents of the stub? output.to_stdout feels like it would always pass.

spec/menu_spec.rb
```diff
@@ -2,12 +2,16 @@
...
+    end
+  end
+
+  describe '#available?' do
+    it 'raises error if chosen item is unavailable' do
```
NapperJLG: Consider rewording as this isn't what the test seems to be testing for - maybe "it 'checks if dish is available' do"

PR #1288, opened 2019-04-14
Week 2 Makers: Takeaway Challenger Compare

kiahjade:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+describe Takeaway do
+  let(:menu) { double(:menu, print: printed_menu) }
+  let(:printed_menu) { {
+    "Chicken bits" => 5,
+    "Burger" => 5,
+    "Pizza" => 8,
+    "Chips" => 3
+  } }
+  let(:order) { Order.new }
```
shektor: Could this be updated to use a London style test, where the Order object is doubled.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  describe '#add_item' do
+    it 'add an item to an order' do
+      expect(subject.add_item("item", 2)).to eq(order.add("item", 2))
+    end
+  end
+
+  # describe '#view_order' do
+  #   it 'includes the current order summary' do
+  #     allow(subject.view_order).to receive @order.order_summary
```
shektor: An Order object double that can respond to the order_summary and basket_total could be used here. While this is not really testing real behaviour, you can still test that it is calling the correct methods on the Order instance.

lib/takeaway.rb
```diff
@@ -0,0 +1,31 @@
...
+require_relative 'order'
+require_relative 'message'
+
+class Takeaway
+  attr_reader :menu, :order
+
+  def initialize
+    @menu = Menu.new
+    @order = Order.new
+    @message = Message.new
```
shektor: Could dependency injection here be used with the Menu, Order and Message objects.

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+require_relative 'order'
+require_relative 'takeaway'
```
shektor: Are these require needed for the program to run?

PR #1292, opened 2019-04-14
Takeaway Challenge

shektor:


lib/send_sms.rb
```diff
@@ -0,0 +1,28 @@
...
+      body: "Thank you! Your order was placed and will be delivered before #{delivery_time}"
+      )
+  end
+
+  private
+
+  def delivery_time
+    Time.now + (60 * 60)
+  end
+end
```
kiahjade: 85.7% coverage, how could this be higher?

README.md
```diff
@@ -1,82 +1,32 @@
...
+fish - £5
+chips - £2
+curry-sauce - £1
+ => {"fish"=>5, "chips"=>2, "curry-sauce"=>1} 
+2.5.0 :004 > order.create("chips*2,curry-sauce*3")
+ => "chips*2,curry-sauce*3" 
+2.5.0 :005 > order.verify(7)
+ => "Order placed and sms message sent to confirm delivery" 
+2.5.0 :006 > exit
+```
```
kiahjade: README makes feature testing very easy to follow

lib/order.rb
```diff
@@ -0,0 +1,40 @@
...
+    total = 0
+    dish_split = selection.split(",")
+    dish_split.each do |item|
+      dish = item.split("*")[0]
+      quantity = item.split("*")[1].to_i
+      total += dishes.price_list[dish] * quantity
+    end
+    total
+  end
+end
```
kiahjade: Easy to follow methods, named well. It's clear what the code is doing. 
It's nice that the user can add multiple items at once. I did not think of that!
Maybe this could just be formatted slightly differently so it's less specific?  
Would be cool if the user could see the calculated sub-total of their order before verifying but it's handy that they have to type the total themselves to ensure it's checked.

lib/send_sms.rb
kiahjade: Everything here is clear and easy to read! I assume this would work if I had the right Twilio access.

spec/send_sms_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'send_sms'
+
+describe SendSMS do
+  describe '#send' do
+    it 'sends a sms' do
+      expect(subject).to receive(:send)
+      subject.send
+    end
+  end
+end
```
kiahjade: RSPEC runs perfectly and this is a well written test so I assume your texting function is working!

spec/order_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  
+  context 'total provided does not equal calculated total' do
+    describe '#verify' do
+      it 'raises error' do
+        subject = described_class.new(dishes_double, sms_double)
+        subject.create("chips*2,curry-sauce*3")
+        user_total = 8
+        msg = "Order Error - Total does not match to price list"
+        expect { subject.verify(user_total) }.to raise_error msg
+      end
```
kiahjade: Like that you've added an extra layer of testing here with an error!

lib/order.rb
```diff
@@ -0,0 +1,40 @@
...
+  def verify(user_total)
+    error_msg = "Order Error - Total does not match to price list"
+    raise error_msg if calculate_total != user_total
+    sms.send
+    "Order placed and sms message sent to confirm delivery"
+  end
+
+  private
+
+  def calculate_total
```
kiahjade: Yay! Private method.

lib/order.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def view_dishes
+    dishes.view
+  end
+
+  def create(selection)
+    @selection = selection
+  end
+
```
kiahjade: Could there have been and additional "review_basket" method in here? Or something similar. So that the user can review their order before verifying?

lib/order.rb
```diff
@@ -0,0 +1,40 @@
...
+    @sms = sms
+  end
+
+  def view_dishes
+    dishes.view
+  end
+
+  def create(selection)
+    @selection = selection
+  end
```
kiahjade: Maybe a hash could have been used here? Or was there a reason for otherwise?

kiahjade: Using a hash would then make it easier to manage separate items/counts and a delete function could even be implemented later?

lib/order.rb
kiahjade: Maybe there should be a separate restaurant/takeaway class which can create a new instance of both order and dishes/menu?

kiahjade (commented):
Overall, very slick code (as far as I know), readable, clear and well tested. 
Possibly a few improvements to functionality from a user and restaurants perspective but I'm impressed. Most boxes from the review helper ticked :)

PR #1295, opened 2019-04-14
Takeaway challenge first try

sejinkay:
Tried to make it work till second user story (select some number of several available dishes).

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,20 +1,14 @@
 class Menu
   attr_reader :food_list
 
-  def initialize
-    @food_list = []
-  end
-
-  def list
-    food =
-      { name: "Chicken Satay", price: 6 },
-      { name: "Fried Spring Rolls", price: 6 },
-      { name: "Green Curry", price: 10 },
-      { name: "Salmon Red Curry", price: 12 },
-      { name: "Stir Fried Aubergine", price: 8 },
-      { name: "Grilled Steak", price: 13 },
-      { name: "Egg Fried Rice", price: 2 },
-      { name: "Organic Coconut Water", price: 3 }
-    @food_list << food
+  def food_list
+    { "Chicken Satay" => 6,
+      "Fried Spring Rolls" => 6,
+      "Green Curry" => 10,
+      "Salmon Red Curry" => 12,
+      "Stir Fried Aubergine" => 8,
+      "Grilled Steak" => 13,
+      "Egg Fried Rice" => 2,
+      "Organic Coconut Water" => 3 }
   end
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,20 @@
+class Menu
+  attr_reader :food_list
+
+  def initialize
+    @food_list = []
+  end
+
+  def list
+    food =
+      { name: "Chicken Satay", price: 6 },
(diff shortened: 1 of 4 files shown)


lib/takeaway.rb
```diff
@@ -1,2 +1,10 @@
+require_relative 'menu.rb'
+
 class Takeaway
+
+  def show_menu
+    menu_person = Menu.new
```
MaxRobertsDear: what does menu_person mean? perhaps a more obvious variable name could be used here?

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
```
MaxRobertsDear: great! I like that you have delegated the task of hosting the menu to a separate class .. follows the principles of object-orientated programming.

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+  attr_reader :food_list
+
+  def initialize
+    @food_list = []
```
MaxRobertsDear: you've made it obvious here that the food_list object is an array - great for the reader!

spec/menu_spec.rb
```diff
@@ -1 +1,12 @@
 require 'menu'
+
+describe Menu do
+  describe '#list' do
+    # This test might be not nessasary as there's similar one already.
+    it 'shows the list of dishes with prices' do
+      food_person = Takeaway.new
```
MaxRobertsDear: well written tests. reads easily

spec/menu_spec.rb
```diff
@@ -1 +1,12 @@
...
+      subject.list
```
MaxRobertsDear: it would be nice if you used an explicit subject name here

spec/takeaway_spec.rb
```diff
@@ -1,4 +1,10 @@
 require 'takeaway'
+require 'menu'
 
 describe Takeaway do
+  it 'shows the list of dishes with prices' do
+    food_person = Menu.new
+    food_person.list
+    expect(subject.show_menu).to eq food_person.food_list
```
MaxRobertsDear: this expectation could benefit from more appropriate variable names such that it reads like a sentence

lib/takeaway.rb
```diff
@@ -1,10 +1,28 @@
 require_relative 'menu.rb'
 
 class Takeaway
+attr_reader :cart
+
+  def initialize
+    @cart = []
+  end
```
MaxRobertsDear: What is the responsibility of the Takeaway class? One might be tempted to create a whole different class that looks after the 'cart' (or the order of the customer), thereby delegating the responsibility of storing the selection

spec/takeaway_spec.rb
```diff
@@ -4,14 +4,15 @@
...
     expect(subject.show_menu).to eq food_person.food_list
   end
 
   describe '#select' do
-    let(:food){ double :food }
+    let(:food) { double :food }
+    let(:quantity) { double :quantity }
     it 'puts the selected dishes to the cart' do
-      expect(subject.select(food)).to eq "#{food} added to your cart!"
+      expect(subject.select(food, quantity)).to eq "#{food} added to your cart!"
```
MaxRobertsDear: nicely written test! would read even better if the subject was explicitly defined

PR #1296, opened 2019-04-15
Partial solution - need to refactor

ayelishaamoah:
Struggled with getting the sms to consistently send - there are no unit tests for this class as I'm not sure how to implement.

Need to add more detail to my README.

### DIFF
--- a/lib/confirmation.rb
+++ b/lib/confirmation.rb
@@ -8,13 +8,13 @@ Dotenv.load('./twilio.env')
 class Confirmation
   def send_confirmation(order)
     @order = order
-    @time = DateTime.now
-    @delivery_time = @time + Rational(1,24)
+    @time = Time.now
+    @delivery_time = @time + Rational(1, 24)
 
     account_sid = ENV['TWILIO_ACCOUNT_SID']
     auth_token = ENV['TWILIO_AUTH_TOKEN']
     client = Twilio::REST::Client.new(account_sid, auth_token)
-    from = ENV['TWILIO_FROM_NUMBER']# Your Twilio number
+    from = ENV['TWILIO_FROM_NUMBER'] # Your Twilio number
     to = ENV['TWILIO_TO_NUMBER'] # Your mobile phone number
 
     client.messages.create(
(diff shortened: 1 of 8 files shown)


lib/confirmation.rb
```diff
@@ -0,0 +1,30 @@
+require 'dotenv/load'
```
CarlyJ88: This sounds like a really interesting way to set up your environmental variables for twilio, I'd be interested in hearing more about it :)

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+class Menu
```
CarlyJ88: I like how you have set up your class in a separate file.

spec/order_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    before do
+      subject.add_item('Spring rolls' => 2.99)
+      subject.add_item('Special Chow Mein' => 6.50)
+      subject.add_item('Egg Fried Rice' => 2.99)
+      subject.add_price(2.99)
+      subject.add_price(6.50)
+      subject.add_price(2.99)
+    end
+
+    describe '#display_order' do
```
CarlyJ88: I like this method of checking the order is displayed correctly.

spec/order_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'order'
+
+describe Order do
+  let(:subject) { described_class.new }
```
CarlyJ88: Just wondering what the described_class.new is referring to. It wasn't clear for me.

spec/menu_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'menu'
```
CarlyJ88: Not sure what this is testing and how, please explain.

lib/takeaway.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'menu'
+require_relative 'order'
+require_relative 'confirmation'
+require 'date'
```
CarlyJ88: Not sure what 'date' is, please explain what file you are requiring as I couldn't find it.

lib/confirmation.rb
```diff
@@ -0,0 +1,30 @@
...
+    from = ENV['TWILIO_FROM_NUMBER'] # Your Twilio number
+    to = ENV['TWILIO_TO_NUMBER'] # Your mobile phone number
+
+    client.messages.create(
+    from: from,
+    to: to,
+    body: <<~SMS
+    Thank you for your order.
+    Order time:#{@time.hour}:#{@time.min}
+    Estimated Delivery Time: #{@delivery_time.hour}:#{@delivery_time.min}
```
CarlyJ88: I like this method. I am curious how you made sure the new time would be one hour in the future.

lib/takeaway.rb
ayelishaamoah: This can be deleted.

spec/order_spec.rb
ayelishaamoah: described class just means the class that you are testing - in this case it is order.

PR #1297, opened 2019-04-15
Week 2 Takeaway Challenge

NapperJLG:
Still some features and tests to be written but the program currently functions


MYREADME.md
```diff
@@ -0,0 +1,42 @@
...
+`x.confirm_order(4.50)`
+
+## To be added
+This program is currently quite bare bones and I would like to add to it in future. Features I would add include:
+- Look into moving the order methods into a seperate class
+- Better tests - currently no tests written for the textmessage class
+- Output strings to let the user know that their items have been added to the order
+- Raise an error if a user tries to order a dish that isn't on the menu
+- Set the textmessage to include the time one hour from when the order was place
+- Use environment variables to hide my Twilio details (will make this repo private until that is done)
```
Xa1n: Just to avoid any risks, you could replace your information with pseudo-text just to show a user that they can add their own details in instead of having yours up there.

MYREADME.md
```diff
@@ -0,0 +1,42 @@
...
+
+`x.order`
+
+`x.order_total`
+
+Confirm the order by passing the correct order total. If the order total is incorrect, an error will be raised. If the total is correct, a text will be sent confirming your order
+
+`x.confirm_order(4.50)`
+
+## To be added
```
Xa1n: Really like this addition, shows continued interest in the project.

lib/takeaway.rb
```diff
@@ -1,38 +1,36 @@
...
   end
-  @order_total
-end
 
-def confirm_order(total)
-  raise("That is not the correct total!") if total != @order_total
-text = TextMessage.new
-text.send_text
-end
+  def confirm_order(total)
```
Xa1n: this makes sense according to challenge description, so less an issue with you and more an issue with how they wrote it. But I think confirm_order would either be a function of the actual takeaway system (i.e. their system processes the order_total), or the user would simply be given the order_total automatically underneath a list of individual orders with their respective amounts and prices. That's more complicated though, so this is a good start.

PR #1300, opened 2019-04-15
Takeaway challenge update

tsankhalpara:
Currently all the tests pass and coverage is 100%.
I have not yet completed my own README. I haven't been able to mock test the SMS method so that it does not send a message each time I run the tests. Also I have not yet been able to split my methods into different classes which have their own lib and spec files.


spec/menu_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe '#calculate_total' do
+    it 'prints copy of order and total' do
+      subject.select("pizza")
+      subject.select("pasta")
+      copy = "pizza 6.0, pasta 8.0, Total: 14.0"
+      expect(subject.calculate_total).to eq copy
+    end
+  end
+
```
jordantroberts: Do you test for the text message method?

PR #1309, opened 2019-05-12
Weekend Takeaway Challenge

lbarber729:


lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+  def add(dish, quantity = 1)
+    raise "Item(s) not available, please order again!" unless
+    @menu.has_key?(dish)
+
+    @order_list[dish] = 0 unless @order_list[dish]
+    @order_list[dish] += quantity
+    @order_list
+  end
+
+  def total
```
LisaMcCormack: when I call total on irb I don' get a total?

LisaMcCormack: Sorry I think this is a problem with my irb

PR #1310, opened 2019-05-12
TakeAway

hughbric:


lib/take_away.rb
```diff
@@ -0,0 +1,70 @@
...
+    puts "Total: £#{convert(@total)}"
+    @total
+  end
+
+  def checkout(sum = 0)
+    raise "Incorrect amount, please enter the total." if sum != total
+
+    send_text
+  end
+
```
olliebrownlow: I found it tricky to follow what was happening in each public method. It would be better to be more descriptive in a way faithful to what each method is trying to achieve, for example, what is "convert"? I know it is converting, but from what to what?

lib/take_away.rb
```diff
@@ -0,0 +1,70 @@
...
+      puts "#{item[:index]}. #{item[:item]}: £#{item[:amount]}"
+    end
+  end
+
+  def order(item_number, quantity = 1)
+    dish = menu[item_number - 1]
+    @basket << { item_number: item_number, quantity: quantity }
+    "You have added #{quantity} x #{dish[:item]} to your basket."
+  end
+
```
olliebrownlow: Maybe the name of the method could be changed? "order" could imply that it is the final order (which you have called "checkout" which is good), so is ambiguous. Maybe "add_to_basket" would be clearer.

lib/send_sms.rb
```diff
@@ -0,0 +1,32 @@
...
+  def send_text
+    account_sid = ENV["TWILIO_ACCOUNT_SID"]
+    auth_token = ENV["TWILIO_AUTH_TOKEN"]
+
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+
+    @client.messages.create(
+      to: ENV["PHONE_NUMBER"],
+      from: "+447723431770",
+      body: text_message
```
olliebrownlow: Perhaps best to hide this phone number too!

lib/menu.csv
```diff
@@ -0,0 +1,17 @@
...
+Panko crusted prawns,6.50
+Miso Dengaku,6.50
+Agadashi tofu,4.00
+Nippon 'mock duck',5.95
+Kale and Quinoa Salad,6.00
+Japanese Buffalo Wings,4.95
+Sesame Brocoli,2.50
+Extra Soy Sauce,0.50
+Extra Ginger,0.50
+Extra Wasabi,0.50
```
olliebrownlow: I love Japanese food!!

spec/take_away_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  context '#order' do
+    it 'selects dish' do
+      allow(subject).to receive(:menu).and_return([{ :index => 1, :item => "Miso Soup", :amount => "2.00" }])
+      item_number = 1
+      expect(subject.order(item_number)).to eq("You have added 1 x Miso Soup to your basket.")
+    end
+
```
olliebrownlow: Do you need to stub here? Can you not just load your @menu, and then call menu on your subject? You are not using another class in this test right?

PR #1311, opened 2019-05-12
Takeaway Challenge

yem82:


lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+class Menu
+  attr_reader :menu
+
+  MENU = { "spaghetti bolognese" => 4.55, "ham and cheese pancake" => 3.75,
```
breycarr: For the purposes of testing, would it be worth limiting this to just one item?

Process.md
```diff
@@ -0,0 +1,519 @@
...
+
+
+Inspecting 9 files<br/>
+.........
+<br/>
+9 files inspected, no offenses detected<br/><br/>
+
+**THREE**
+
+As a customer
```
breycarr: What would you approach be to attempt into this user story?

PR #1313, opened 2019-05-13
Pull request - Takeaway challenge

olliebrownlow:


spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  it 'so customers can place an order, they can view a menu with dishes and prices' do
+    tas = TakeawayService.new
+    menu = Menu.new
+    expect { tas.view_menu }.to output(menu.view).to_stdout
+  end
+
+  # As a customer
+  # So that I can order the meal I want
+  # I would like to be able to select some number of several available dishes
+
```
hughbric: Your tests would be more readable if you grouped them according to a contexts. It would also allow you to DRY it out by putting the 3 repeated lines into a before method that would execute before the beginning of each test:`before(:each) do end`. Also, instead of creating a new instance of `TakeawayService.new`, you can use RSpec's `subject` which does the same thing.

lib/takeaway_service.rb
```diff
@@ -0,0 +1,92 @@
+require_relative 'menu'
+require_relative 'dish'
+require_relative 'send_sms'
+
+class TakeawayService
+  attr_reader :basket
+
```
hughbric: I like the use of dependency injection with the menu_class. I don't think you need to add the text_service_class to the initialize of the TakeawayService class.

lib/takeaway_service.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def total_cost
+    print cost
+  end
+
+  def place_order
+    @text_service_class.new.message
+  end
+
```
hughbric: Some of your private methods look like they do exactly the same as the public ones (other than add the word print). Is it worth adding a new method just for that? It might help reduce your private methods by just doing the work in the public methods. It depends on what you want to go for.

PR #1314, opened 2019-05-13
Ben pull request for weekend challenge

breycarr:
Completed user story 2. Delegation has been utilised, but I will need to work on moving as once of my learning goals this week. 

I have laid out how I would proceed through the other user stories.


lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
+
+  attr_reader :dishes
+
+  def initialize
+    # I could have the menu read from a file in a private method
+    # at initialisation. I have chosen to keep things simple here for
+    # ease of testing
+    @dishes = { burrito: '6.80' }
```
yem82: The customer doesn't have many choices here ;). Perhaps you could add more items to the menu so that the customer can add various dishes to the order.

lib/order.rb
```diff
@@ -0,0 +1,19 @@
+class Order
```
yem82: You could require the file for the menu class here for ease of use in feature testing.

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def view
```
yem82: The user may find it more useful if they are provided with a list that provides the available dishes and prices. This can be achieved by using a block.

PR #1319, opened 2019-05-24
take away challenge 

JaydaKad:
struggling with the mocking

### DIFF
--- a/lib/dish_list.rb
+++ b/lib/dish_list.rb
@@ -1,0 +1,16 @@
+class DishList
+
+  attr_reader :dish_list_prices
+
+  def initialize
+    @dish_list_prices = { spaghetti_bolognese: 30,
+                          chicken_pesto: 44,
+                          fettuccine_beef: 42,
+                          chiili_chicken: 45,
+                          sabor_salad: 32,
+                          tiramisu: 15,
+                          banoffee_pie: 15
+                        }
+  end
+
+end
--- a/lib/dish_list.rb
+++ b/lib/dish_list.rb
@@ -19,4 +19,8 @@ class DishList
     end
   end
 
+  def price(item)
+    @dish_list_prices[item]
+  end
+
 end
--- a/lib/dish_list.rb
+++ b/lib/dish_list.rb
@@ -6,11 +6,17 @@ class DishList
     @dish_list_prices = { spaghetti_bolognese: 30,
                           chicken_pesto: 44,
                           fettuccine_beef: 42,
-                          chiili_chicken: 45,
+                          chilli_chicken: 45,
                           sabor_salad: 32,
                           tiramisu: 15,
                           banoffee_pie: 15
                         }
+  end
+
+  def menu
+    @dish_list_prices.each do |item, price|
+      puts "#{item} - £#{price}"
+    end
   end
 
 end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -13,8 +13,9 @@ class Order
     @dish_list.menu
   end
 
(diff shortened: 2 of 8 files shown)


lib/dish_list.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def menu
+    @dish_list_prices.each do |item, price|
+      puts "#{item} - £#{price}"
+    end
+  end
+
+  def price(item)
+    @dish_list_prices[item]
+  end
```
lindseydew: Good use of creating a separate object for managing the menu here

lib/dish_list.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :dish_list_prices
+
+  def initialize
+    @dish_list_prices = { spaghetti_bolognese: 30,
+                          chicken_pesto: 44,
+                          fettuccine_beef: 42,
+                          chilli_chicken: 45,
+                          sabor_salad: 32,
+                          tiramisu: 15,
+                          banoffee_pie: 15
```
lindseydew: Good use of an instance variable

lib/order.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/dish_list.rb'
+require './lib/text_message.rb'
+
+class Order
+
+  attr_reader :price_tracker, :dish_list
+
+  def initialize(dish_list = DishList.new)
```
lindseydew: Nice use of dependency injection here

lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+    quantity.times { @price_tracker << @dish_list.price(item) }
+  end
+
+  def check_total(estimate = 0)
+    puts @price_tracker.sum
+    @price_tracker.sum == estimate
+  end
+
+  def place_order
+    @sms.send_sms
```
lindseydew: Nice use of delegation here

lib/text_message.rb
```diff
@@ -0,0 +1,18 @@
+require 'twilio-ruby'
+
+class TextMessage
+  attr_reader :account_sid, :auth_token, :client
+
+  def initialize
+    @account_sid = ENV['accsid'] # Your Account SID from www.twilio.com/console
+    @auth_token = ENV['token']   # Your Auth Token from www.twilio.com/console
+    @client = Twilio::REST::Client.new account_sid, auth_token
```
lindseydew: There's an opportunity for the Twilio client to be passed in as a constructor argument here, so that this class could be tested using mocking

spec/order_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  let(:menu) { double(:dish_list, :menu => "Pizza - 6.5") }
+  let(:dish_list_class) { double(:dish_list_class, :new => menu) }
+
+  it 'can show all dishes with prices in a nice format' do
+    order
+    expect(order.view_dishes).to eq("Pizza - 6.5")
+  end
+
+  it 'can select a specific dish and display price' do
+    o = Order.new
```
lindseydew: You're super close with mocking here! 
Wondering if there is a reason you chose to create a new instance of `Order` here? If you used the `order` variable defined on line 6 you could make use of the mocked `dish_list_class` instead?

lib/text_message.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+
+  def send_sms
+    @client.messages.create(
+    body: "Your order has been placed and will be with you by #{(Time.now + 3600).strftime("%H:%M")}",
```
lindseydew: Great job on setting up the Twilio client!

PR #1320, opened 2019-05-24
Completed takeaway challenge

kks110:


lib/dish.rb
```diff
@@ -0,0 +1,10 @@
+class Dish
+
+  attr_reader :name, :price
+
+  def initialize(name, price)
```
lindseydew: Nice idea to model this as a separate object

lib/text_sender.rb
```diff
@@ -0,0 +1,24 @@
...
+  def send_message
+    @client.messages.create(
+                           from: ENV.fetch('TWILIO_SEND_NUMBER'),
+                           body: generate_message_body,
+                           to: ENV.fetch('MOBILE_NUMBER')
+                           )
+  end
+
+  def generate_message_body
+    "Thank you! Your order was placed and will be delivered before #{@time.hour + 1}:#{@time.min}"
```
lindseydew: Lovely idea to include the time in this

lib/text_sender.rb
```diff
@@ -0,0 +1,24 @@
...
+    @time = time_class.new
+    @account_sid = ENV.fetch('TWILIO_ACCOUNT_SID')
+    @auth_token = ENV.fetch('TWILIO_AUTH_TOKEN')
+    @client = client_class.new(@account_sid, @auth_token)
+  end
+
+  def send_message
+    @client.messages.create(
+                           from: ENV.fetch('TWILIO_SEND_NUMBER'),
+                           body: generate_message_body,
```
lindseydew: Does the twilio client give a response after the message has been created? A stretch task could be to use the response from Twilio to determine whether you need to retry.

PR #1321, opened 2019-05-24
Order takeaway and get a text message

AnnaWeakclaw:


lib/customer.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def choice(name, amount = ONLY_ONE)
+    @order.add(name, amount)
+  end
+
+  def checkout(my_idea_of_the_cost)
+    message = "You are charging me the incorrect price"
+    raise message if @order.cost != my_idea_of_the_cost
+    puts "You will pay £#{@order.cost}"
+    Pub.new(@order)
```
lindseydew: The dependency on `Pub.new()` could be factored into the `initialize` method to break the dependency between `Customer` and `Pub`

lib/dish.rb
```diff
@@ -0,0 +1,9 @@
+class Dish
+  attr_reader :name, :price, :item
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+    @item = { :name => @name, :price => @price }
+  end
```
lindseydew: Nice to have factored this out into its own class - was the intention for the menu class to reference it?

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+  def add(dish)
+    @menu.push(dish)
+  end
+
+  def show
+    i = 1
+    @menu.each { |dish|
+      puts "#{i}. #{dish[:name]} for £#{dish[:price]}"
+      i += 1
+    }
```
lindseydew: Nice idea to put the item number in the display

lib/send_sms.rb
```diff
@@ -0,0 +1,24 @@
...
+require "rubygems"
+require "twilio-ruby"
+require "time"
+
+# Your Account Sid and Auth Token from twilio.com/console
+# DANGER! This is insecure. See http://twil.io/secure
+
+class Message
+  def initialize
+  account_sid = ENV["TWILIO_ACCOUNT_SID"]
```
lindseydew: The `TwilioClient` could be a dependency on this class so that this could be tested

spec/menu_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'menu'
+
+describe Menu do
+  let(:menu) { Menu.new }
+  let(:dish) { double(:dish) }
+
```
lindseydew: Nice use of doubles here

spec/feature_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'customer'
+
+describe 'customer can plane an order' do
+
```
lindseydew: Were you blocked on writing this test or did you run out of time?

PR #1323, opened 2019-05-25
This is my attempt at the takeaway challenge.

gps2601:
Steps to demo it can be found in the feature_test.rb : )

### DIFF
--- a/lib/dish.rb
+++ b/lib/dish.rb
@@ -1,0 +1,16 @@
+require 'money'
+
+class Dish
+  Money.locale_backend = :currency
+
+  attr_reader :name, :price
+
+  def initialize(name, price)
+    @name = name
+    @price = Money.new(price, "GBP")
+  end
+
+  def formatted
+    "#{name}: #{price.format}"
+  end
+end
--- a/lib/dish.rb
+++ b/lib/dish.rb
@@ -7,7 +7,7 @@ class Dish
 
   def initialize(name, price)
     @name = name
-    @price = price
+    @price = price.to_i
   end
 
   def formatted
--- a/lib/dish.rb
+++ b/lib/dish.rb
@@ -7,10 +7,10 @@ class Dish
 
   def initialize(name, price)
     @name = name
-    @price = Money.new(price, "GBP")
+    @price = price
   end
 
   def formatted
-    "#{name}: #{price.format}"
+    "#{name}: #{Money.new(price, "GBP").format}"
   end
 end
--- a/lib/food_list.rb
+++ b/lib/food_list.rb
@@ -1,0 +1,25 @@
+require 'csv'
+require './lib/dish'
+
+module FoodList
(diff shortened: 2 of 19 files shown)


lib/dish.rb
```diff
@@ -0,0 +1,16 @@
+require 'money'
+
+class Dish
+  Money.locale_backend = :currency
```
lindseydew: Nice idea to use a money class

lib/food_list.rb
```diff
@@ -0,0 +1,25 @@
...
+    @items = items
+    @item_class = item_class
+  end
+
+  def view_items
+    to_return = ""
+    items.each do |item|
+      to_return += item.formatted + " "
+    end
+    to_return
```
lindseydew: I think this can be achieved without iteration. eg:
```items.join(" ")```

File: lib/order.rb [hunk 3]
```diff
@@ -0,0 +1,14 @@
+require './lib/food_list.rb'
+require 'money'
+
+class Order
+  include FoodList
+
+  def show_total
+    Money.new(items.inject(0) { |sum, item| sum + item.price }, "GBP").format
```
lindseydew: To be a little more strictly `Single Responsibility` this logic:
```items.inject(0) { |sum, item| sum + item.price }``` could be separated into its own function to test the logic independently of presenting this.

File: lib/sms_sender.rb [hunk 4]
```diff
@@ -0,0 +1,26 @@
+require 'twilio-ruby'
+
+class SmsSender
+  def initialize(sms_client = Twilio::REST::Client, time = Time)
```
lindseydew: Great use of dependency injection for the Twilio Client and the time object

spec/dish_spec.rb
```diff
@@ -0,0 +1,17 @@
+require './lib/dish'
+require 'money'
+
+describe Dish do
+  let(:my_dish) { Dish.new("Rice", 100) }
+  it 'can be initialized with a name' do
+    expect(my_dish.name).to eq("Rice")
```
lindseydew: I like that you have chosen to test the data structure and the presentation of the data separately in this class

lib/sms_sender.rb
```diff
@@ -0,0 +1,26 @@
...
+    @account_sid = ENV.fetch('ACCOUNT_SID')
+    @auth_token = ENV.fetch('AUTH_TOKEN')
+    @from = ENV.fetch('TWILIO_NUMBER') # Your Twilio number
+    @to = ENV.fetch('GEOFFS_NUMBER')
+    @client = sms_client.new(@account_sid, @auth_token)
+    @time = time.new
+  end
+
+  def confirm_order
+    @client.messages.create(
```
lindseydew: An extension opportunity here could be to check the success of a response from twilio and potentially retry in the event of a failure

PR #1324, opened 2019-05-26
submission of takeaway challenge

LaurenOM:


spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect(order.items).to be_empty
+  end 
+
+  it 'adds selections to an order' do 
+    updated_items = order.add_to_cart("Margherita")
+    expect(updated_items).to include("Margherita")
+  end 
+  
+  it 'calculates the total of an order' do
+    order.add_to_cart({ item: "Margherita", price: 10, quantity: 2 })
```
lindseydew: I see that you're using a selection from menu to add to an order in your feature test. This could be shown here by mocking the `select` method on Menu and using that in your order class

lib/menu.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+  
+  def select(selection, quant)
+    selected = {}
+    items.each do |item|
+      if selection == item[:item] 
+        selected = { item: item[:item], price: item[:price], quantity: quant }
+      end
+    end
+    selected
```
lindseydew: This could also be achieved with a select and a map. eg:

``` items.select { |item| item[:item] == selection }.map { |item| { item: item[:item], price: item[:price], quantity: quant } ```

File: lib/twilio.rb [hunk 3]
```diff
@@ -0,0 +1,28 @@
+require 'twilio-ruby'
+
+class Sender
+  attr_reader :client
+  def initialize
```
lindseydew: Could use dependency injection on the twilio client here, and write a test to check that the create message is called

PR #1325, opened 2019-05-26
Peter's challenge submission

peterjcole:


lib/dish.rb
```diff
@@ -0,0 +1,20 @@
...
+    @price = BigDecimal(price, 4)
+  end
+
+  def to_s
+    "#{name}: £#{format(price)}"
+  end
+
+  private
+
+  def format(amount)
```
lindseydew: Nice use of a private method here

lib/order_handler.rb
```diff
@@ -0,0 +1,60 @@
...
+  end
+
+  def add_all_to_order(order_string)
+    order_arr = split_order_string(order_string)
+    order_arr.each_cons(2) do |item_arr| 
+      item_arr[1].to_i.times { @current_order.add(get_dish(item_arr[0])) }
+    end
+  end
+
+  def split_order_string(order_string)
```
lindseydew: This method could make the item and the quantity separate variables so that you don't need to split the string on `*`

spec/messenger_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'messenger'
+require 'dotenv/load'
+
+describe 'messenger' do
+  let(:messenger) { Messenger.new(twilio) }
+  let(:twilio) { double(:twilio) }  
+  
+  before(:each) do
+    allow(twilio).to receive_message_chain(:api, :account, :messages, :create)
```
lindseydew: Nice use of chaining here!

spec/order_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    it 'returns the correct total when a large number of items have been added' do
+      10.times { order.add(sushi) }
+      order.add(pizza)
+      expect(order.total).to eq(BigDecimal(161.51, 5))
+    end
+  end
+  
+  context '#delivery_time' do
+    it 'returns the order time plus an hour, formatted as HH:MM' do
+      expect(order.delivery_time).to eq((time + (60 * 60)).strftime('%H:%M'))
```
lindseydew: Could even mock the time here so that we can test the exact example of a string.

PR #1327, opened 2019-05-27
Takeaway Challenge Submission - Megan

megan-e-goode:
Had to merge my repos as work was not on the forked version of this challenge.


README.md
```diff
@@ -1,4 +1,4 @@
-Takeaway Challenge
+# Takeaway Challenge
```
samjones1001: Great README! Covers all of the things I think are useful to see.

lib/customer.rb
```diff
@@ -0,0 +1,23 @@
...
+class Customer
+  attr_reader :order
+
+  def initialize(order = Order)
+    @order = order.new
+  end
+
+  def view_menu
+    @order.view_menu.each_with_index do |dish, index|
+      puts "#{index + 1}. #{dish[:dish]}: £#{'%.2f' % dish[:price]}"
```
samjones1001: Normally I would suggest extracting this kind of presentation logic out into a class of it's own to keep it separate from your business logic, but given that this class is mostly just delegating responsibility down to an `Order` object I think you're probably ok here.

lib/dish.rb
```diff
@@ -0,0 +1,12 @@
+class Dish
+
+  def initialize(name, quantity, price)
+    @name = name
+    @price = price
+    @quantity = quantity
+  end
+
+  def add
+    { dish: @name, price: @price, quantity: @quantity }
```
samjones1001: This feels a little strange to me - you have a class which allows you to create an object with all of the information you need to represent a dish but then have a method to translate that object into a hash containing the same information. I'm not sure what you gain from doing this 🤔

samjones1001: That said, I can't actually see anywhere in the code base that this method is being used, so maybe it should just be deleted 🙂

lib/message.rb
```diff
@@ -0,0 +1,18 @@
+require 'twilio-ruby'
+
+class Message
+  SUCCESS_MSG = "Thank you! Your order was placed and will be delivered before 18:52"
+
+  def initialize
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
```
samjones1001: By creating a new instance of the twilio client inside of your initialize method like this you're setting up a hard dependency between your two objects and making it very difficult to isolate your unit tests. The biggest problem with that in this particular case is that you wont be able to test this class without sending yourself a text message every time you run rspec 🙂 

You can start to get around this by making use of dependency injection:

```ruby
def initialize(client = Twilio::REST::Client.new(account_sid, auth_token))
  @client = client
end
```

lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+    { dish: "Stir-fry Squid with Veg", price: 5.75 },
+    { dish: "Mixed Seafood", price: 4.75 },
+    { dish: "Spring Rolls (6)", price: 3.50 }
+    ]
+
+  def initialize(message = Message)
+    @menu = MENU
+    @basket = []
+    @total = 0
+    @message = message.new
```
samjones1001: Same thing here - you could use dependency injection here too.

lib/dish.rb
megan-e-goode: I did wonder about this when I was writing it as I thought it was a bit useless, however I just included it because I thought the challenge might want to see how we link the use of the classes together, but I agree that I could just remove it :)

PR #1328, opened 2019-05-27
Takeaway Challenge - Helen

HelenHTech:
Create Menu, Order and Takeaway Class. Create Twillo_text file
Didn't finish Takeaway class, received errors, such as  #<Double> received unexpected message :new with (no args), and couldn't fix. Might need some guidance to review where it all went so wrong!

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -5,6 +5,6 @@ class Menu
   end
 
   def list_menu
-    return @menu_dishes
+    @menu_dishes
   end
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -2,11 +2,9 @@ class Menu
   attr_accessor :menu_dishes
   def initialize
     @menu_dishes = { :pizza => 4.00, :sausage => 2.00, :curry => 6.50, :rice => 2.00, :chilli => 6.00, :salad => 3.50 }
-    #  The above @menu_disheshash should come from the Menu class. Need to figure out how to access it
   end
 
   def list_menu
     return @menu_dishes
   end
-
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,12 @@
+class Menu
+  attr_accessor :menu_dishes
+  def initialize
+    @menu_dishes = { :pizza => 4.00, :sausage => 2.00, :curry => 6.50, :rice => 2.00, :chilli => 6.00, :salad => 3.50 }
+    #  The above @menu_disheshash should come from the Menu class. Need to figure out how to access it
+  end
+
+  def list_menu
+    return @menu_dishes
+  end
+
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,26 @@
+class Order
+  attr_reader :ordered_items, :total
+  def initialize
+    @ordered_items = []
+    @total = 0
+    @menu_dishes = { :pizza => 4.00, :sausage => 2.00, :curry => 6.50, :rice => 2.00, :chilli => 6.00, :salad => 3.50 }
+  end
+
+  def select_dishes(item_chosen)
+    @ordered_items.push(item_chosen)
+    @total += @menu_dishes[item_chosen]
+  end
+
+  def price
+    @total
+  end
+
+  def text_order_confirmation
+    if !@ordered_items.empty? && @total > 0
+      return "Thank you! Your order was placed and will be delivered before 18:52"
+    else
+      return "Order incomplete"
+    end
+  end
+
+end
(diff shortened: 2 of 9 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,11 @@
+class Menu
+  attr_accessor :menu_dishes
+  def initialize
+    @menu_dishes = { :pizza => 4.00, :sausage => 2.00, :curry => 6.50,
+      :rice => 2.00, :chilli => 6.00, :salad => 3.50 }
+  end
+
+  def list_menu
+    @menu_dishes
```
samjones1001: A very small thing, but the fact that you have an accessor for your @menu_dishes instance variable means that you don't need the `@` here

lib/order.rb
```diff
@@ -0,0 +1,24 @@
+class Order
+  attr_reader :ordered_items, :total, :menu_dishes
+  def initialize 
+    menu = Menu.new
```
samjones1001: Creating an instance of another object from inside of the initialize method like this sets up a very hard dependency between the tow objects, and makes it very difficult to properly isolate your unit tests.

You can get past this problem by making use of dependency injection:

```ruby
  @menu = menu 
end
```

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+    @menu_dishes = menu.menu_dishes
+    @ordered_items = []
+    @total = 0
+  end
+
+  def display_menu
```
samjones1001: This method feels like a good use case for delegation. As it stands, your Order class is responsible for displaying the menu which feels a bit off. Your menu class has a `list_menu` method which also does the same thing. You could delegate responsibility for this action down to the menu class by making a call to that method here.

```ruby
def display_menu
  menu.list_menu
end
```

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+    @menu_dishes
+  end
+
+  def select_dishes(item_chosen, quantity)
+    @ordered_items.push(item_chosen, quantity)
```
samjones1001: Storing these values in an array like this feels a bit risky - it means that throughout your system, you'll be reliant on knowing that the structure of the array should be `[item, quantity, item, quantity]` - if you forgot this order, or if the order in which you were storing information ever changed this could cause problems. Can you think of a more structured way of storing the info which would remove this dependence on a specific order?

lib/takeaway.rb
```diff
@@ -0,0 +1,36 @@
+class Takeaway
+  attr_reader :menu_class, :order_class
+  def initialize(order_class = Order, menu_class = Menu,
```
samjones1001: Nice use of dependency injection here.

samjones1001: Interesting though that this class contains an instance of the `Menu` and an instance of `Order` which, in turn, also holds onto an instance of `Menu`. I wonder if you could remove this duplication be changing your object model a little?

lib/takeaway.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def display_menu
+    menu_class.list_menu
+  end
+
+  def ordered_items(item_chosen, quantity)
+    @order_class.select_dishes(item_chosen, quantity)
+  end
+
+  def send_text_message
```
samjones1001: This feels like it could probably be a private method.  Unless you expect either a user or another object to make direct use of a method then a good rule of thumb is to always default to private.

lib/twillo_text.rb
```diff
@@ -0,0 +1,24 @@
...
+class SendMessage
+  def initialize
+    @account_sid = ENV['TWILIO_ACCOUNT_SID']
+    @auth_token = ENV['TWILIO_AUTH_TOKEN']
+    @recipient_phone = ENV['Recipient_phone']
+    @twilio_number = ENV['Twilio_number']
+  end
+
+  def send_customer_message
+    @client = Twilio::REST::Client.new @account_sid, @auth_token
```
samjones1001: Again, creating a new instance of the twilio client from within a method creates a hard dependency. This time it's even more problematic as you wont be able to test your code without sending yourself a text message 🙂 

You can use dependency injection again here, although the process of doubling and stubbing required to get everything working correctly is a bit more complex this time.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'takeaway'
+
+describe Takeaway do
```
samjones1001: It looks like your making good use of doubles and stubs here - this means these tests should be fully isolated.

Right now though I don't think your tests for the `Order` class are though.

PR #1329, opened 2019-05-27
Lin's Takeaway Friday challenge

LinTrieu:
Friday challenge submission for Lin Trieu

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,3 +1,4 @@
+# Creates a menu of takeaway items and displays menu
 class Menu
   attr_reader :display , :items
 
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -8,7 +8,7 @@ class Menu
   end
   
   def items
-    { "pasta" => 4.99, "bread" => 0.99, "rice" => 5.99 }
+    { "pasta" => 4.99, "bread" => 0.99, "pizza" => 5.99 }
   end
   
   def price(dish)
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -4,7 +4,7 @@ class Menu
 
   def display
     items.each do |dish,price|
-      puts "#{dish}: #{price}"
+      "#{dish}: #{price}"
     end
   end
   
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -9,7 +9,8 @@ class Menu
   end
   
   def items
-    { "pasta" => 4.99, "bread" => 0.99, "pizza" => 5.99 }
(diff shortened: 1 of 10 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,19 @@
+# Creates a menu of takeaway items and displays menu
+class Menu
+  attr_reader :display , :items
```
samjones1001: Right now, these readers aren't doing anything for you - you're setting them up but don't have any instance variables with matching names, and then defining some methods with the same names so you're overwriting them anyway.

lib/menu.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def display
+    items.each do |dish,price|
+      "#{dish}: #{price}"
+    end
+  end
+  
+  def items
+    { "pasta" => 4.99, "bread" => 0.99, "pizza" => 5.99 } # => muttable
```
samjones1001: This feels to me like it belongs in an instance variable rather than as a return value from a method.

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+    @basket_summary[dish] += quantity
+  end
+
+  def display_total
+    price_calc
+    "Your order total is £#{@total}"
+  end
+
+  # private
+  def price_calc
```
samjones1001: I think you were right in the first place - this method should probably be private. It seems as though a user would interact with the `display_total` method rather than this one, so I don't think it needs to be exposed.

My guess is that you removed the privacy so you could test it? Hopefully what we spoke about in kickoff this morning has helped you understand the correct approach here 🙂

lib/takeaway.rb
```diff
@@ -0,0 +1,19 @@
+# Allows takeaway to checkout current order and send confirmation text message using Twilio API
+require_relative 'send_sms' #twilio-ruby code hidden in .gitignore
```
samjones1001: Great that you're thinking about not exposing your api keys and other info by adding sensitive data to .gitignore.

Rather than ignoring the whole implementation of the twilio api, however, I'd suggest just setting up some variables containing the specific information you want to keep secure in a seperate file, ignoring that file and pulling the variables into you Message class code.

lib/takeaway.rb
```diff
@@ -0,0 +1,19 @@
+# Allows takeaway to checkout current order and send confirmation text message using Twilio API
+require_relative 'send_sms' #twilio-ruby code hidden in .gitignore 
+
+class Takeaway
+
+  def initialize(order = Order.new)
+    @message = Message.new
```
samjones1001: I'd suggest using the same approach as you have for the Order class here and making use of dependency injection.  That would decrease the coupling between your objects and make testing easier.

PR #1330, opened 2019-05-27
Egle takeaway_challenge

ebernotaite:
Somewhat incomplete, needs some more work - see README


lib/basket.rb
```diff
@@ -0,0 +1,56 @@
...
+    h = Hash.new 0
+    @contents.each {|p| h[p] += 1}
+    @contents = h.keys.map{|k| k[:count] = h[k]; k}
+    @contents.each do |item|
+      price = item[:price] * item[:count]
+      puts "#{item[:pizza]} x#{item[:count]}, £#{price}"
+    end
+  end
+
+  def total
```
samjones1001: It's a small improvement but I feel as though you can probably use inject here to avoid needing the accumulator variable and the iteration.

lib/message.rb
```diff
@@ -0,0 +1,17 @@
+require 'twilio-ruby'
+require 'dotenv/load'
+
+class Message
+
+def initialize
+  @client = Twilio::REST::Client.new(ENV['SID'], ENV['TOKEN'])
```
samjones1001: Right now, any time you create an instance of your Message class, you're hardwired to create an instance of the real Twilio client too. This is a creating a hard dependency between those objects and making it so that you'll send yourself a text message every time that you run your unit tests.

A simple fix for this would be to make use of dependency injection

```ruby
def initialize(client = Twilio::REST::Client.new(ENV['SID'], ENV['TOKEN']))
  @client = client
end
```

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+class Menu
+attr_reader :items
+  def initialize
+    @items = [{:pizza => "Margherita", :price =>  9.99}, 
+              {:pizza => "Neapolitan", :price => 12.99},
+              {:pizza => "Greek", :price => 13.99},
+              {:pizza => "Padana", :price => 11.99}]
+  end
+
+  def view
```
samjones1001: A pretty common pattern in OOD is to separate your business logic (anything dealing with data or state) from presentation logic (anything dealing with displaying information to human users).

Right now this method is dealing with presentation logic where as the rest of the class is dealing with business logic - there could be a good argument for extracting out a `Printer` class or something similar

lib/takeaway.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_reader :basket
+  
+  def initialize(menu_class = Menu, basket = Basket.new)
+    @menu_class = menu_class
+    @menu = nil
+    @basket = basket
+  end
+
+  def view_menu
+    @menu = @menu_class.new
```
samjones1001: This line sort of implies that every time I want to look at the menu a new one should be created. That's probably ok here, but it feels to me that it would make more sense for the menu to be created once when the `Takeaway` object is instantiated and then be held in an instance variable.

spec/basket_spec.rb
```diff
@@ -0,0 +1,37 @@
+describe 'Basket' do
+
+  let(:basket) {Basket.new}
+  let(:message_dbl) {double(:message, :send => "Success")}
+  let(:basket_dbl) {Basket.new(message_dbl)}
```
samjones1001: I think the naming here is slightly confusing - at first I assumed that this was going to be a double of the basket class itself

PR #1331, opened 2019-05-27
Takeout app completed without text order confirmation functionality

OlliePorter:


spec/order_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'order'
+
+describe Order do
+
+  subject(:order)       { Order.new(menu) }
+  let(:menu)            { double(:menu, :dishes => { "Tom Yum" => 9.90, "Pad Thai" => 10.00, "Green Curry" => 6.99, "Red Curry" => 6.00, "Aloe Drink" => 2.00, "Chocolate Ice Cream" => 4.00, "California Roll" => 12.00 }) }
```
samjones1001: This looks as though you've got the hang of doubles 👍 

Your tests all look to be using them correctly too.

samjones1001 (commented):
This looks like a very solid effort for this point in the course.

I've left a couple of thoughts below - not that much feedback here as there's not actually that much I would change in what you've implemented so far 🙂 

Let me know if you have any questions!

PR #1333, opened 2019-05-28
Food order menu

Ma9sha:
Hi,

I tried to get the totals working for all items ordered this morning. I have not completed the following:

raise an exception
get the text message

I will try and do it over this week if time permits. I would still like more understanding on how objects are called.

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,19 @@
+class Menu
+  attr_reader :list
+  def initialize
+    @list = [
+      {:name => "panini", :cost => "2"},
+      {:name => "biriyani", :cost => "3"},
+      {:name => "cake", :cost => "4"}
+    ]
+  end
+
+  # def select_dishes(item)
+  #   @choosen_items = []
+  #   @list.each do |dish|
+  #     @choosen_items.push(item) if dish[:name] == item
+  #   end
+  #   return @choosen_items
+  # end
+end
+
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -3,41 +3,31 @@ require './lib/menu.rb'
 class Order
 
 attr_reader :list_of_dishes, :selected_items
+
 def initialize(m = Menu.new)
- 
   @list_of_dishes = m
   @selected_items = []
 end
+
 def view_dishes
   @item = @list_of_dishes.list
 end
 
 def select_dishes(item, quantity)
   @total = 0
+  i = 0
   selection = @list_of_dishes.list.select do |hash|
-    hash[:name] == item
+  hash[:name] == item
   end
-  @selected_items.push(selection[0])
-
+  quantity.times {@selected_items.push(selection[i])}
+  i += 1
   @selected_items.each do |hash|
-    @total += (hash[:cost].to_i * quantity)
-end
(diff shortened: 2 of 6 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,19 @@
...
+  attr_reader :list
+  def initialize
+    @list = [
+      {:name => "panini", :cost => "2"},
+      {:name => "biriyani", :cost => "3"},
+      {:name => "cake", :cost => "4"}
+    ]
+  end
+
+  # def select_dishes(item)
```
samjones1001: I'm guessing this method was removed from this class and then moved up to the Order class?

If that's the case I would recommend deleting the code here rather than just commenting it out - it's an easy way of keeping your code looking neat and tidy.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
+require './lib/menu.rb'
+
+class Order
+
+attr_reader :list_of_dishes, :selected_items
+
+def initialize(m = Menu.new)
```
samjones1001: Good use of dependency injection here 👍

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+  @list_of_dishes = m
+  @selected_items = []
+end
+
+def view_dishes
+  @item = @list_of_dishes.list
```
samjones1001: Is it necessary to do a variable assignment here and create this new @item instance variable?

It seems to me like you could achieve the same thing using
```ruby
def view_dishes
  list_of_dishes.list
end
```

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+end
+
+def select_dishes(item, quantity)
+  @total = 0
+  i = 0
+  selection = @list_of_dishes.list.select do |hash|
+  hash[:name] == item
```
samjones1001: anywhere we use a do end block, it's convention to indent the code which sits inside of it

So:

```ruby
my_array.each do |item|
   puts item
end
```

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+def initialize(m = Menu.new)
+  @list_of_dishes = m
+  @selected_items = []
+end
+
+def view_dishes
+  @item = @list_of_dishes.list
+end
+
+def select_dishes(item, quantity)
```
samjones1001: I'm not clear on what the `i` variable in this method is doing for you. It seems like you could only ever enter one name of a dish at a time, so you shouldn't need a counter to keep the place of where you are in the array of selected dishes?

spec/menu_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'menu'
+
+describe Menu do
```
samjones1001: No tests 😱

lib/order1.rb
```diff
@@ -0,0 +1,41 @@
+require './lib/menu.rb'
+
+class Order1
```
samjones1001: Is this another implementation of the same Order class from above? 🙂 

I actually prefer this version as I think the implementation of the `add_item` method is much cleaner. I also like that it delegates some responsibility down to the `Menu` class.

PR #1340, opened 2019-06-09
Partial Solution (all but Twilio part)

LucyMHall:
See MYREADME.md for more details


spec/cost_calculator_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  before(:each) do
+    @calc = CostCalculator.new
+  end
+
+  describe "#price_for_dish" do
+    it "gets the price for an ordered dish from the menu" do
+      @menu = double("menu", menu_contents: { "Fish" => 2 })
+      @order = double("order", current_order: { "dish" => "Fish", "quantity" => 3 })
+      expect(@calc.price_for_dish(@order.current_order, @menu.menu_contents)).to eq(2)
```
alicelieutier: You don't actually need doubles here as you are passing hashes, and not objects into your function. You could write:
```
expect(@calc.price_for_dish({ "dish" => "Fish", "quantity" => 3 }, { "Fish" => 2 })).to eq(2)
```

**However** this feels like a bit of a design flaw. You probably should pass objects to that function directly, and the function would then call the necessary methods on them. And in that case, you would indeed need doubles.

lib/cost_calculator.rb
```diff
@@ -0,0 +1,23 @@
+class CostCalculator
+
+  def calculate(current_order, menu_contents)
+    cost_of_order = 0
+    current_order.each do |index|
+      cost_of_order += price_for_quantity(index, menu_contents)
+    end
+    cost_of_order
+  end
+
```
alicelieutier: All methods below feel like they should be private. 
As a side note, you should not test private methods.

spec/menu_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  before(:each) do
+    @menu_contents = { "Fish" => 3, "Chips" => 2, "Veggie Burger" => 2, "Mushy Peas" => 0.5, "Curry Sauce" => 0.5, "Ketchup" => 0.2 }
+    @displayed_menu = "--Menu--\nFish: £3.0\nChips: £2.0\nVeggie Burger: £2.0\nMushy Peas: £0.5\nCurry Sauce: £0.5\nKetchup: £0.2\n"
+    @formatted_menu = ["Fish: £3.0", "Chips: £2.0", "Veggie Burger: £2.0", "Mushy Peas: £0.5", "Curry Sauce: £0.5", "Ketchup: £0.2"]
+    @menu = Menu.new(@menu_contents)
+    @header = "--Menu--"
+  end
+
+  describe 'initialize' do
```
alicelieutier: I would not test initialize - it has no behaviour (no output to the user)

spec/menu_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'menu'
+
+RSpec.describe "Menu" do
+
+  before(:each) do
+    @menu_contents = { "Fish" => 3, "Chips" => 2, "Veggie Burger" => 2, "Mushy Peas" => 0.5, "Curry Sauce" => 0.5, "Ketchup" => 0.2 }
+    @displayed_menu = "--Menu--\nFish: £3.0\nChips: £2.0\nVeggie Burger: £2.0\nMushy Peas: £0.5\nCurry Sauce: £0.5\nKetchup: £0.2\n"
```
alicelieutier: Why do you declare these at the top? I'd understand if they were used a number of times, but that is not the case. It feels less readable to have to always go up and down to know what is passed around when reading the tests.

spec/order_spec.rb
```diff
@@ -0,0 +1,77 @@
+require "order"
+
+RSpec.describe "Order" do
+
+  before(:each) do
+    @menu = double("menu", menu_contents: { "Fish" => 3, "Chips" => 2 }, view: "expected_menu")
+    @menu_contents = { "Fish" => 3, "Chips" => 2 }
```
alicelieutier: Is this line necessary?

spec/order_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+    @cost_calculator = double("cost_calculator", calculate: 8)
+    @order = Order.new(@menu, @cost_calculator)
```
alicelieutier: Nice use of doubles + dependency injection here 🎉

PR #1341, opened 2019-06-09
Celia takeaway-challenge partial solution

ghost:


bengscott2 (commented):
I think you had a great start! Looks like you were on the right track. Nice short methods and good interaction between classes. You were definitely on your way with this one. Great work.

PR #1343, opened 2019-06-09
Ruth Baker partial solution

ruthmoog:


app.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/takeaway.rb'
+require './lib/menu.rb'
+require './lib/order.rb'
+
+mochi_shop = Takeaway.new(menu: Menu.new)
```
zepherine2006DEV: I think this could be easily turned into a feature test, by adding some expectations. Sam Morgan's video is very good for this.

lib/takeaway.rb
```diff
@@ -0,0 +1,22 @@
...
+  def order_summary(order)
+    order.reduce(0) do |total, line_item|
+      price = @menu.dishes[line_item[:dish]]
+      total + line_item[:qty] * price
+    end
+  end
+
+  private
+
+  attr_reader :menu
```
zepherine2006DEV: I wasn't quite sure what the benefits were of making the attr_reader for :menu private or if the attribute reader was in fact used anywhere inside the class? All the code in the class seems to refer to the instance variable directly (@menu). I tried commenting out the attr_reader and all the tests still passed, perhaps indicating that the attr_reader isn't required.

PR #1345, opened 2019-06-09
partial solution

chris-groves:


lib/order.rb
```diff
@@ -0,0 +1,35 @@
...
+    @final_order = []
+  end
+
+  def select_dish(dish)
+    @menu_items.each_key do |key|
+    @order_dishes << dish if key == dish
+    end
+  end
+
+  def final_order_compile
```
aniasobo: maybe refactor this method for readability and to have it only do one thing, and delegate some things away to other methods?

lib/order.rb
```diff
@@ -0,0 +1,35 @@
+class Order
+  attr_reader :order_dishes, :menu_items, :final_order
+
+
+  def initialize
```
aniasobo: Hardcoded menu is anti-SRP so would be good to extract it to another class

PR #1349, opened 2019-06-09
Takeaway challenge - incomplete attempt

Danielandro:


README.md
```diff
@@ -48,35 +37,82 @@ So that I am reassured that my order will be delivered on time
...
-* Hints on functionality to implement:
-  * Ensure you have a list of dishes with prices
-  * Place the order by giving the list of dishes, their quantities and a number that should be the exact total. If the sum is not correct the method should raise an error, otherwise the customer is sent a text saying that the order was placed successfully and that it will be delivered 1 hour from now, e.g. "Thank you! Your order was placed and will be delivered before 18:52".
-  * The text sending functionality should be implemented using Twilio API. You'll need to register for it. It’s free.
-  * Use the twilio-ruby gem to access the API
-  * Use the Gemfile to manage your gems
-  * Make sure that your Takeaway is thoroughly tested and that you use mocks and/or stubs, as necessary to not to send texts when your tests are run
-  * However, if your Takeaway is loaded into IRB and the order is placed, the text should actually be sent
-  * Note that you can only send texts in the same country as you have your account. I.e. if you have a UK account you can only send to UK numbers.
+Instructions
```
ajosephides: Would be good to get an example of how to run the program. e.g should I use IRB? If so how do I use it - ie what are examples of how to use a menu, an order?

lib/restaurant.rb
```diff
@@ -0,0 +1,27 @@
+require 'order'
```
ajosephides: Think this line is causing an error. Prevents me from running the program locally in IRB. In this case being explicit with the dependent file location with `require relative` would be good.

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+class Menu  
+  def print_menu
+    format_menu
+  end
+
+  def dishes
```
ajosephides: I don't think this method is being used by any of the other classes right now. If it isn't it should be private or as it is a constant I believe already accessible to other internal classes.
If it is there purely for a test it shouldn't be.

lib/restaurant.rb
```diff
@@ -0,0 +1,27 @@
...
+    @order = order
+  end
+
+  def print_menu
+    menu.print_menu
+  end
+
+  def add_to_order(dish, quantity = 1)
+    raise "Cannot add to order: Item is not on menu" if not_on_menu?(dish) 
+    current_order << @order.new(dish, quantity)
```
ajosephides: Nice use of the dependency injection

lib/restaurant.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  private
+  attr_reader :menu, :order
```
ajosephides: Are these needed in a private section? Anything in private still has accesses to the instance variables.

lib/restaurant.rb
```diff
@@ -0,0 +1,27 @@
...
+  
+  def not_on_menu?(dish)
```
ajosephides: Nice and clean breaking this one out.

spec/menu_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'menu'
+
+describe Menu do
+  subject(:menu) { described_class.new }
+
```
ajosephides: As per comment on `menu.rb` - there is no test for `dishes` so just delete it from code.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'restaurant'
+
+describe Restaurant do 
+  subject(:restaurant) { described_class.new(menu: menu) }
```
ajosephides: This is cool.  I've seen this syntax before but not sure what it means. Would be good to get an explanation from you on it.

PR #1350, opened 2019-06-09
Takeaway Challenge

petraartep:
Twilio API worked perfectly when I hardcoded the phone numbers, but after implementing env variables, can't receive text messages. Due to time constraints I was not able to fix it yet.


lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+  
+  DEFAULT_MENU = {
+      spring_rolls: 5,
+      wonton: 10,
+      congee: 10,
+      chow_mein: 15,
+      hotpot: 20
+    }
+
+  def initialize(dishes = DEFAULT_MENU)
```
LucyMHall: I like the fact your menu is not hardcoded in but there is a default option.

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+    @dishes = dishes
+  end
+
+  def print_menu
+    @dishes.map do |name, price|
+      "#{name.to_s.split("_").map(&:capitalize).join(" ")}: £#{price}"
```
LucyMHall: Maybe split this bit of code up e.g. formatted_name = name.to_s.split("_") then do `#{formatted_name.map....)

lib/order.rb
```diff
@@ -0,0 +1,22 @@
+require_relative 'menu'
+
+class Order
+
+  attr_reader :current_order
+
+  def initialize(menu)
```
LucyMHall: def initialize(menu= Menu.new) ... you could pass a new instance of the menu class, if someone just does order = Order.new the @menu will refer to your default menu, but you can always do order =Order.new({new_menu_item: 5}) to add in a new menu

This way you don't have to make a menu first then pass it through, it combines these two steps

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+    @menu = menu
+    @current_order = Hash.new(0)
```
LucyMHall: @current_order = {}

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+    @total = 0
+  end
+
+  def add(dish, quantity)
+    raise 'Item is not available' if @menu.dishes[dish] == nil
```
LucyMHall: @menu.dishes[dish] -> what this does it goes into your the instance of the menu class you passed into order, it looks at the hash you created e.g.{ spring_rolls: 5} and it asks for the value that the key (dish) exposes. So if the item isn't in the hash then the value will be nil

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def choose(dishes)
+    dishes.each do |name, quantity|
+      @order.add(name, quantity)
+    end
+  end
+
+  def place_order
+    @sms.send_sms
+    "Thanks for your order! You will be charged £#{@order.total}. You will receive a confirmation message shortly."
```
LucyMHall: if you want to see the instance variable total you will need to add it to your order attr_reader

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+    @current_order[dish] = quantity 
+  end
+
+  def total_order(dish, quantity)
```
LucyMHall: This isn't being called so the total wont be updated

LucyMHall: Maybe call it in your add method

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'menu'
```
LucyMHall: You might not need this class, I'm not sure

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  
+  context 'initialize' do
+    it 'has a list of dishes and prices' do
+      expect(menu.dishes).to eq(dishes_on_menu)
+    end
+  end
+
+  describe '#print_menu' do
+    it 'prints formatted menu' do
+      printed_menu = 'Spring Rolls: £5, Wonton: £10, Congee: £10, Chow Mein: £15, Hotpot: £20'
```
LucyMHall: I like the fact that you have extracted this onto an earlier line, it makes for easier to read expectations

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      p expect(menu.print_menu).to eq(printed_menu)
+    end
```
LucyMHall: syntax for testing anything you are printing out rather than just returning: expect { menu.print_menu}.to output(printed_menu).to_stdout

spec/order_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'order'
+
+RSpec.describe Order do
+  let(:order) { described_class.new(menu) }
+  let(:menu) { double :menu, :dishes => dishes_on_menu }
```
LucyMHall: alternate syntax: @menu = double("Menu", dishes: dishes_on_menu)

I like that you have the :dishes return just a simple phrase rather than fake dishes. You are only testing what you need to test to get the job done :)

spec/order_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  let(:dishes_on_menu) do {
+      spring_rolls: 5,
+      wonton: 10,
+      congee: 10,
+      chow_mein: 15,
+      hotpot: 20
+    }
```
LucyMHall: You could combine this: 
@dishes_on_menu = {spring_rolls: 5, wonton: 10, congee: 10, chow_mein: 15, hotpot: 20}
@menu = double("Menu", dishes: {@dishes_on_menu)... I don't think you need to use 'let' syntax

LucyMHall: Also in your code you only ever order congee and wontons so you can actually have these as the only dishes on the menu without compromising your code. Doubles don't need to actually look like a real instance of that class, they just need to be the simplest form possible to work with the class your testing

spec/order_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      expect { order.add(:congee, 1) }.to change { order.current_order.count }.by(1)
+    end
+  end
+
+  describe '#total_order' do
+    it 'shows the total order' do
+      dish = :wonton
+      quantity = 2
+      total = 20
+      message = "Your total order is £%.2f" % [total]
```
LucyMHall: Nice bit of code to make sure that your total is always displayed with 2 decimals points :)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  describe '#choose' do
+    it 'allows user to select dishes' do
+      expect(takeaway.choose(dishes_on_menu)).to eq(dishes_on_menu)
+    end 
+  end
+
+  describe '#place_order' do
+    before(:each) do
+      allow(takeaway).to receive(:send_sms).and_return(true)
```
LucyMHall: this bit of stubbing means that the test below will always pass

PR #1360, opened 2019-07-06
Toby Brock Restaurant

TBrockster:
Finished all user stories, not fully tested for edgecases, and could do with some refactoring/rewriting.

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,4 +1,5 @@
 require 'csv'
+require 'twilio-ruby'
 require_relative 'order'
 require_relative 'load_menu_module'
 
@@ -7,6 +8,11 @@ class Menu
   def initialize(restaurant: "fishnchips.csv", order: Order.new(restaurant: restaurant))
     @order = order
     load_menu(restaurant)
+    account_sid = ENV['TWILIOSID']
+    auth_token = ENV['TWILIOAUTH']
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+    @from = ENV['TWILIONUMBER'] # Your Twilio number
+    @to = ENV['PHONENUMBER'] # Your mobile phone number
   end
 
   def display_menu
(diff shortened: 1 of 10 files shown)


lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+    @order
+  end
+
+  # compares order hash against menu hash to determine the total cost of order
+  def receipt
+    receipt_array = []
+    total = 0
+    @order.each do |order_dish, order_quantity|
+      @menu_hash.each do |menu_dish, menu_cost| 
+        if order_dish == menu_dish
```
hemser1: could you break this into another method to shorten this method?

lib/menu.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def check_order(order: @order)
+    order.receipt
+  end
+
+  def confirm_order(order: @order)
+    puts order.receipt
+    puts 'Would you like to confirm your order? (y/n)'
+    user_input = gets.chomp
+    if user_input == 'y'
```
hemser1: Could you break this into a method so if input Y then you run a method, which will help keep the methods more simplistic?

lib/menu.rb
```diff
@@ -0,0 +1,50 @@
...
+require 'twilio-ruby'
+require_relative 'order'
+require_relative 'load_menu_module'
+
+class Menu
+  include LoadMenu
+  def initialize(restaurant: "fishnchips.csv", order: Order.new(restaurant: restaurant))
+    @order = order
+    load_menu(restaurant)
+    account_sid = ENV['TWILIOSID']
```
hemser1: Would you break the Twilio function into another class or module to separate it out?

spec/order_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'order'
+
+describe Order do
+  subject(:order) { described_class.new(restaurant: "fishnchips.csv") }
+  describe '#add' do
+    it 'adds dishes and quantities to the order hash' do
+      expect(order.add('Fish', 3)).to eq({ 'Fish' => 3 })
+    end
+  end
+end
```
hemser1: Could we create a test for receipt method to test the method is working correctly?

PR #1363, opened 2019-07-07
Will's Chippy

willhowes:


lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def select_dish(dish, quantity)
+    add_dish(dish, quantity)
+    { dish => quantity }
+  end
+
+  def order_total
+    total = 0
+    current_order.each do |item|
```
renatapop: Smart!

renatapop: Smart!

README.md
```diff
@@ -1,82 +1,28 @@
...
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+The ```Order``` class is breaking single responsibility principal since it deals with selecting dishes to add to the order and also calculating the total of the order and checking this against the user's given total. This would be interesting discussion point. Selecting a dish could have been extracted into a separate class if I had time.
 
-```
-As a customer
-So that I can check if I want to order something
-I would like to see a list of dishes with prices
+Also, the ```Text``` class calculates the current time and delivery time as well as ordering the API to send the text. A separate ```DeilveryTime``` class could have been created.
```
renatapop: I think simply calling the class "Delivery" and including Text and Delivery as functions of this class would have been sufficient :)

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+    raise 'Total given does not match' unless user_total == order_total
+    user_total == order_total
+  end
+
+  def confirm_order(mobile)
+    Text.new(mobile).send
+    @current_order = []
+  end
+
+  private
```
renatapop: Good use of private methods

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
+class Menu
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = { 'fish' => 5, 'chips' => 3, 'gravy' => 1, 'savoly' => 2 }
+  end
+
+  def read
+    @dishes
```
renatapop: I believe as it stands the read method will simply return the hash to the user. Could have used extrapolation to print out menu in a reader friendly way

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  # As a customer
+  # So that I am reassured that my order will be delivered on time
+  # I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
+
+  xit 'sends a text confirmation to the user when order placed' do
+    # TESTED THIS WITH VALID MOBILE NUMBER AT IT WORKS
+    expect { text.send }.to change { text.message_sent }.from(false).to(true)
+  end
+end
```
renatapop: Overall very clear specs

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'order'
+
+describe Order do
+  subject(:menu) { described_class.new }
+  subject(:order) { described_class.new }
+  let(:dish) { :dish }
+  let(:price) { 5 }
```
renatapop: Would the test still work if   let(:price) { price } instead of a fixed example?

PR #1365, opened 2019-07-07
Tim's Takeaway

Timdavidcole:
Managed to get it all working I think!  I started off by making it WAY too complicated with a user menu/interface but went back to square one after I realized it was overkill.


lib/order.rb
```diff
@@ -0,0 +1,34 @@
...
+    "#{amount}x #{item}(s) has been added to your basket."
+  end
+
+  def summary
+    @items.map do |key, value|
+      "#{key.capitalize} x #{value}"
+    end.join(', ') + " = £#{total}"
+  end
+
+  def total
```
richieganney: could you keep this less than 4 lines?

lib/takeaway.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'order'
+require 'twilio-ruby'
+
+class TakeAway
```
richieganney: have you considered using private methods?

lib/twilio.rb
```diff
@@ -0,0 +1,12 @@
+require 'twilio-ruby'
+
+account_sid = "AC3dacdc49158a31763601f3cd87e88f48" # Your Account SID from www.twilio.com/console
```
richieganney: you can use environment variables to keep this sensitive information hidden

not entirely sure how to do this but you'd need another file to store them

spec/feature_tests_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+end
+
+# As a customer
+# So that I can order the meal I want
+# I would like to be able to select some number of several available dishes
+
+describe "Takeaway features" do
+  it "a customer can add an order of several available dishes" do
+    takeaway = TakeAway.new
```
richieganney: can you not use subject?

also if you put 'subject(:takeaway) { described_class.new }' then this renames subject for you throughout the entire spec file

spec/feature_tests_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+end
+
+# As a customer
+# So that I can verify that my order is correct
+# I would like to check that the total I have been given matches the sum of the various dishes in my order
+
+describe "Takeaway features" do
+  it "check that total is correct" do
+    takeaway = TakeAway.new
```
richieganney: same as above

PR #1368, opened 2019-07-07
Big *Ri* Mac's Takeaway

riannemcc:


lib/order.rb
```diff
@@ -0,0 +1,42 @@
+require_relative './text_service'
```
richieganney: can just do 'text_service' i believe

lib/order.rb
```diff
@@ -0,0 +1,42 @@
...
+  def submit_order(expected_total)
+    raise "Total does not match" if basket_total != expected_total
+
+    @text_service.send_text(order_confirmation_message)
+    "Confirmation text sent"
+  end
+
+  private
+
+  def order_confirmation_message
```
richieganney: i didn't do a realtime one. can we discuss?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+      "Pizza - £6\n" +
+      "Fried chicken - £4\n" +
+      "Cheeseburger - £5\n" +
+      "Chips - £3\n"
+    ).to_stdout
+  end
+
+  it 'selects items from the menu by index' do
+    expected_item = { item: 'Fried chicken', price: 4 }
+    expect(subject.select_item(1)).to eq expected_item
```
richieganney: you can change the name of subject if you want at the beginning of your describe block.
this isn't wrong but its a handy tool
can't remember off top of my head

PR #1369, opened 2019-07-07
Genny Allcroft - takeaway challenge

gennyallcroft:


lib/dish.rb
```diff
@@ -0,0 +1,16 @@
+class Dish
+
+  attr_reader :dish, :price
```
fvarnals: Yaara suggested that it's better to do these as defined methods in the class, although tbh I much prefer attr_reader too hahaha

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+  attr_reader :menu
+
+  def initialize
+    @list_of_dishes = []
+    @menu
+  end
+
+  def add_dish(dish)
+    @list_of_dishes <<  dish.dish_for_menu
+    @list_of_dishes.flatten!
```
fvarnals: What does flatten do?

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def view_menu
+    @list_of_dishes.each do |dish|
+      puts "Dish: #{dish[:dish]}, Price: #{dish[:price]}"
+    end
+  end
+
+  def is_available?(dish)
+    @list_of_dishes.any? {|dish| dish['dish'] == @dish_to_order}
```
fvarnals: Maybe I'm missing it but I don't see @dish_to_order

lib/total_price.rb
```diff
@@ -0,0 +1,26 @@
+require 'menu'
+
+class TotalPrice
+
+  attr_reader :list_of_prices
+  attr_reader :total_price
+
+  def get_prices_for_order(menu, order)
+    @list_of_prices = [3,5]
```
fvarnals: Why [3,5]?

PR #1370, opened 2019-07-07
Josh Hemsley - Takeaway Challenge

hemser1:


lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+require_relative 'menu'
+require_relative 'pricecalculation'
+require_relative 'send_sms'
+
+class Order
+
+  attr_reader :menu
+  attr_reader :summary
+  attr_reader :quantity
+  attr_reader :current_order
```
TBrockster: This seems like a lot of attr_readers?

lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  include SendMessage
+  def initialize(menu = Menu.new, quantity = 1,
+                  calc = PriceCalculation.new(self))
+    @menu = menu
+    @quantity = quantity
```
TBrockster: Could this just be a constant?

lib/order.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def total_cost
+    "Total: £#{total_price}"
+  end
+
+  def checkout
+    send_sms
+  end
+
+  def order_string
```
TBrockster: I think this should be in the module?

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'order'
+require 'pricecalculation'
+require 'send_sms'
+
+describe 'user stories' do
+  let(:subject) { Order.new }
```
TBrockster: I don't think lines like this should be used in feature tests, as they are supposed to reflect the inputs from the command line.

spec/features/user_stories_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  it 'returns current order in string format shwoing cost of each selection' do
+    subject.select('Beef Burger', 3)
+    subject.select('Onion Rings', 5)
+    expect(subject.order_summary).to eq "Beef Burger X 3 = £20.97, Onion Rings X 5 = £9.95"
+  end
+
+  # As a customer
+  # So that I am reassured that my order will be delivered on time
+  # I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
+  it 'sends text with order info including delevery time and price' do
```
TBrockster: Price is not asked for in the user story.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'order'
```
TBrockster: This should be split into a spec file for each class, and they shouldn't 'require' files other than the class they are testing.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+describe Order do
+  let(:subject) { Order.new }
+  let(:checkout) { double :checkout }
+
+  it 'shows menu including list of dishes and prices' do
+
+    expect(subject.read_menu).to eq subject.read_menu
+  end
+
+  it 'should allow you to select menu items and add them to order' do
```
TBrockster: It would be nice to have these tests formatted into describe '#method' block.

lib/order.rb
hemser1: correct - managed to remove 2

lib/order.rb
hemser1: Correct - I should be! Or could be...

spec/features/user_stories_spec.rb
hemser1: correct.

spec/features/user_stories_spec.rb
hemser1: true...

spec/takeaway_spec.rb
hemser1: correct

spec/takeaway_spec.rb
hemser1: also correct

PR #1371, opened 2019-07-07
Takeaway challenge

richieganney:


lib/text.rb
```diff
@@ -0,0 +1,19 @@
+require 'twilio-ruby'
+
+class Text
+
+  def send_sms
+    account_sid = 'ACb72247efd4758c41dbdbdedf140fc3e7'
```
riannemcc: This information (sid, token and phone numbers) should be hidden by using an environment variable. At the moment, someone could access your code and hack into your account (i think) but an env variable means you can gitignore the file and it wont show on github.

lib/text.rb
```diff
@@ -0,0 +1,19 @@
...
+    auth_token = 'a393d90dc34b98d353b5642ccfdc4893'
+    client = Twilio::REST::Client.new(account_sid, auth_token)
+        
+    from = '+441582214721' # Your Twilio number
+    to = '+447825849151' # Your mobile phone number
+        
+    client.messages.create(
+    from: from,
+    to: to,
+    body: "Thank you! Your order was placed and will be delivered in a week! (we're a bit slower than our competitors)"
```
riannemcc: Could use timecop or 'Time' in Ruby to set the actual time

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    it 'allows a customer to order from the menu' do
+      restaurant.add_to_menu("burger", 9)
+      restaurant.add_to_menu("chips", 3)
+      expect(restaurant.select_meal(2)).to eq [{ "chips" => 3 }]
+    end
+  end
+
+  describe '#request_receipt' do
+    it 'generates a receipt for the customer and verifies that their meal is correct' do
+      restaurant.add_to_menu("burger", 9)
```
riannemcc: You use this code a few times so could refactor and include it at the top of the group instead.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'restaurant'
+
+describe Restaurant do
+  subject(:restaurant) { described_class.new }
```
riannemcc: Whats the benefit in using described_class rather than Restaurant?
Question for you as I dont know the answer!

lib/restaurant.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def request_receipt(final_order = Receipt.new)
+    final_order.receipt << @one_order
+    return "Order complete. Total cost: £#{final_order.receipt[0].map do |i| i.values end.flatten.inject(&:+)}"
+    final_order.send_sms
+  end
+
+  def start_order
+    @one_order = []
```
Timdavidcole: Might be an idea to have an Order class maybe?  Feel like restaurant is starting to get pretty busy.

PR #1373, opened 2019-07-08
takeaway

renatapop:
Takeaway takes order and stores into array. Stuck on how to select the correct index from "menu" hash. Suggestions very welcome :)

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,119 @@
+require "Pry"
+
+
+ class Takeaway
+
+  attr_reader :text, :menu, :order  
+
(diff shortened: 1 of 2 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,119 @@
...
+ class Takeaway
+
+  attr_reader :text, :menu, :order  
+
+  def initialize
+      @menu = Menu.new 
+      @order = Order.new 
+  end 
+ 
+   def greeter
```
willhowes: This is a great idea to give the user all the different options at the start. Wish I'd done this!

PR #1398, opened 2019-08-19
takeaway chandelle

claredowson:
I know I should have wrote test code first but as I have been struggling with the coding side, I really wanted to get that bit working first

### DIFF
--- a/lib/takeaway_dishes.rb
+++ b/lib/takeaway_dishes.rb
@@ -1,0 +1,32 @@
+class Dishes
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = [
+              { item: "korma", price: 3.80 },
+              { item: "egg fried rice", price: 2.00 },
+              { item: "plain naan", price: 1.50 },
+              { item: "chicken tikka", price: 4.00 },
+              { item: "special fried rice", price: 2.80 },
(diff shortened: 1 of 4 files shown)


spec/takeaway_dishes_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  it 'view a dish and get the dishes name and cost' do
+    dishes = Dishes.new
+    expect(dishes.select_dishes("korma")).to eq({ :item => "korma", :price => 3.80 })
+  end
+
+  it 'displays the cost of a dish on the menu' do
+    dishes = Dishes.new
+    expect(dishes.display_price("korma")).to eq(3.80)
```
lindseydew: Good use of unit tests

lib/takeaway_dishes.rb
```diff
@@ -0,0 +1,57 @@
...
+        @food_order << element[:item]
+      end
+    }
+    # total = 0
+    # dishes.each { |element|
+    #   if element[:item] == item
+    #     return element[:price]
+    #   end
+    # }
+    # total += element ["price"]
```
lindseydew: Ah yes I can see where you are going here - very close to having a total

PR #1399, opened 2019-08-19
Code review from coach

tuyetngo1:


spec/menu_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'menu'
+describe Menu do
+
+it 'can create a new menu with list of dishes' do
+menu = Menu.new
+expect(menu).to be_a_kind_of(Menu)
+end
+
+it 'can list all the dishes in the menu with prices' do
```
lindseydew: Good unit testing here.

Please be careful about indentation - this can make a difference for readability.

lib/order.rb
```diff
@@ -0,0 +1,37 @@
+class Order
+attr_reader :dishes, :selected, :prices
+def initialize(menu=Menu.new)
+  @dishes = menu.dishes
+  @selected = []
+  @prices =[]
+end
+
+def select(dish,quantity)
+  @dishes.select do |hash|
```
lindseydew: Great use of the `select` method here.

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+    hash2 = hash.merge({quantity:quantity})
+    @selected << hash2
+    @prices.push(hash[:price]*quantity)
+  return @selected
+  end
+end
+end
+
+def total
+  @total = @prices.inject(:+)
```
lindseydew: Great use of the `inject` method - well done

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  allow(mock_order).to receive(:check_amount).and_return(10)
+  expect{ order.done(mock_order) }.to raise_error("Total is not correct!")
+end
+
+  it 'can give you a confirmation when amount is correct' do
+    order = Order.new
+    order.select("Teriyaki",2)
+    order.select("Pho",1)
+    order.total
+    mock_order = double(:mock_order)
```
lindseydew: Nice one for getting the double syntax correct! 
We actually don't need to rely on mocking here because the method is in the same object.

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  it 'can select number of several dishes' do
+    order = Order.new
+    menu = Menu.new
+  expect(order.select("Sushi",3)).to eq([{:dish=>"Sushi", :price=>7.0, :quantity=>3}])
+  end
+
+  it 'gives the total matches with the sum of various dishes in my order' do
+    order = Order.new
+    menu = Menu.new
```
lindseydew: I'm not sure this menu is needed here? Ideally this menu could be mocked so that we are keeping our objects loosely coupled.

PR #1400, opened 2019-08-19
Eloise Takeaway Challenge

EloisePrasad:


spec/dishes_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    it 'can return the menu with price' do
+      expect(subject.menu).to include("chicken burger, £10\nfish and chips, £8\nsteak, £20\nchips, £5\nbeef burger, £10")
+    end
+  end
+
+  describe '#price' do
+    it 'can return the price of the dish' do
+      expect(subject.price("chicken burger")).to eq(10)
+    end
+  end
```
lindseydew: Nice unit testing!

spec/order_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    #   allow(dishes).to receive(:price)
+    #   order = Order.new(dishes_class)
+    #   order.select(dish)
+    # end
+    it 'can respond to select with an argument' do
+      expect(subject).to respond_to(:select).with(1).argument
+    end
+    it 'can add prices to total' do
+      allow(dishes_class).to receive(:new).and_return(dishes)
+      allow(dishes).to receive(:price).and_return(10)
```
lindseydew: Nice use of dependency injection 💯 

Small point, and I can see why you are doing it this way because it's a similar approach to the `Till` code we were looking at last week, but we probably can just give an `instance` of `Dishes` to `Order` rather than a reference of the class. The difference is whether we want to create a new instance of `Dishes` everytime we make an order, or whether our instance of `Order` can refer to the same `Dishes` instance

spec/order_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      order = Order.new(dishes_class)
+      order.select("chicken burger")
+      expect(order.select("chicken burger")).to eq("chicken burger has been added to your order")
+    end
+  end
+  describe '#order_summary' do
+    it' can print the order summary with prices' do
+      allow(dishes_class).to receive(:new).and_return(dishes)
+      allow(dishes).to receive(:price).and_return(10)
+      order = Order.new
```
lindseydew: Careful, if we want to make use of the double here we need to pass it in as an argument to the `Order` class (otherwise it will use its default arg which is the real instance of `Dishes`)

PR #1445, opened 2019-10-03
Takeaway attempt 

egunnery:
Had the text working correctly but then removed personal info

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,47 @@
+require_relative ('text')
+
+
+class Takeaway
+
+  attr_accessor :order
+
+  def initialize
+    @menu = ['chips' => 2.00, 'snackbox' => 3.50, 'battered sausage' => 1.50,
+      'curry sauce' => 1.00, 'battered cod' => 3.00]
+      @ordered_items = []
+    end
+
+    def menu
+      @menu
+    end
+
+    def order(item, quantity)
+      x = 0
+      while x < quantity do
+        @ordered_items << item
+        x += 1
+      end
+      return @ordered_items
+    end
+
+    def cost_of_order
+      y = []
+      @ordered_items.each do |n|
+        y << @menu[0][n]
+      end
+      y = y.inject(:+)
+      y = ('%.2f' % y)
+      return ("£" + y.to_s)
+    end
+
+    # A LESS WEIRD WAY OF PHRASING USER STORY 3, BUT REALLY TESTS THE SAME THING
+    def pay(card_payment)
+      raise 'payment does not reach total' if card_payment != cost_of_order
+      text
+      return 'payment processed'
+    end
+
+    def text
+      Text.new
+    end
+end
--- a/lib/text.rb
+++ b/lib/text.rb
@@ -1,0 +1,17 @@
+class Text
+
+  def initialize
+
+    require 'twilio-ruby'
+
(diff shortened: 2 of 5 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+      'curry sauce' => 1.00, 'battered cod' => 3.00]
+      @ordered_items = []
+    end
+
+    def menu
+      @menu
+    end
+
+    def order(item, quantity)
+      x = 0
```
dearshrewdwit: Variable name readibility
- when in doubt, name the variable what the value represents.

dearshrewdwit: 2. It could be cleaner to store this information ```name=>quantity```, so  ```{ "pizza" => 3 }```, instead of using a while loop, incrementing a counter, and storing copies of the item

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+      while x < quantity do
+        @ordered_items << item
+        x += 1
+      end
+      return @ordered_items
+    end
+
+    def cost_of_order
+      y = []
```
dearshrewdwit: Explore using ruby's #map method
> https://ruby-doc.org/core-2.5.0/Array.html#method-i-map

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    # A LESS WEIRD WAY OF PHRASING USER STORY 3, BUT REALLY TESTS THE SAME THING
+    def pay(card_payment)
+      raise 'payment does not reach total' if card_payment != cost_of_order
+      text
+      return 'payment processed'
+    end
+
+    def text
+      Text.new
```
dearshrewdwit: Where possible - inject dependencies via an argument.

lib/text.rb
```diff
@@ -0,0 +1,17 @@
+class Text
+
+  def initialize
+
+    require 'twilio-ruby'
+
+    account_sid = '****************'
```
dearshrewdwit: Explore storing data as environment variables on your machine, and referencing them by their key within your app.
> https://www.rubyguides.com/2019/01/ruby-environment-variables/

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  it 'calculates the cost of an order' do
+    subject.order('chips', 2)
+    subject.order('battered cod', 1)
+    expect(subject.cost_of_order).to eq("£7.00")
+  end
+
+  it 'raises an error if cusotmer gives different total' do
+    subject.stub(:cost_of_order).and_return("£9")
```
dearshrewdwit: Almost always avoid stubbing the subject under test. Instead think about your test setup - how could you set up your subject under test in a different way to be able to test this example?

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
+require_relative ('text')
+
+
+class Takeaway
```
dearshrewdwit: If you had to describe the purpose of this object's behaviour - what would it be?

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  attr_accessor :order
+
+  def initialize
+    @menu = ['chips' => 2.00, 'snackbox' => 3.50, 'battered sausage' => 1.50,
```
dearshrewdwit: This information is likely to change - how could you allow for this?

PR #1446, opened 2019-10-11
Committing project files

zt3h:


twilio/send_sms.rb
```diff
@@ -0,0 +1,14 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+
+account_sid = 'AC579e4968fbbdc84cd908d0b213589106'
+auth_token = 'a45da61c48a7b9e44cae7851ca460dfa'
```
zero201926: you shouldn't really post these

lib/order.rb
```diff
@@ -0,0 +1,27 @@
...
+        @order = [] 
+    end
+
+    def add_item(name, amount)
+        @menu.menu_items.each do |dish, price|
+            if name == dish
+                #add the dish
+                @order.push ([name, amount])
+                total = amount * price
+            raise 'This item is not on the menu' if name != dish
```
zero201926: I like this, nice way to add an item.

lib/order.rb
```diff
@@ -0,0 +1,27 @@
+require 'menu'
+
+class Order
+    attr_reader :order
+
+    def initialize
+        @menu = Menu.new
```
zero201926: could have (@menu = menu || Menu.new) that way you could of used doubles for tests

twilio/send_sms.rb
```diff
@@ -0,0 +1,13 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+
+no more
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+from = '+19735595553' # Your Twilio number
+to = '+447597870934' # Your mobile phone number
```
zero201926: if this is your real number you probably shouldn't post it either

twilio/send_sms.rb
jamesbingham: Removed those and phone numbers

twilio/send_sms.rb
jamesbingham: Removed as well thanks for the text :)

spec/menu_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'menu'
+
+describe Menu do
+    it "can print the menu items" do
+        expect(subject.print_menu)
```
jamesjoshuahill: Please add an expectation like `expect { subject.print_menu }.to output("...some menu here...").to_stdout`

spec/order_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'order'
+
+describe Order do
+    it "raises error when item is not displayed in menu" do
+        expect {subject.add_item('Bread', 2).to raise_error('This item is not on the menu')}
```
jamesjoshuahill: The `expect { ... }` curly braces are in the wrong place here. Take a look at the RSpec docs and see if you can fix it: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/raise-error-matcher

Make sure you can make the test fail by changing the expectation.

spec/order_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    end
+
+    it "can add add a item on the menu to the order array" do
+        subject.add_item('Chips', 2.00)
+        expect {subject.order.to include(["Chips", 2.00])}
```
jamesjoshuahill: Again, the `expect { ... }` curly braces are in the wrong place here.

spec/order_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+    end
+
+    it "should print total order" do #need to add test to calculate total order
+        expect(subject.total)
```
jamesjoshuahill: Please add an expectation, for example: `expect(subject.total).to eq(42)`

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+class Menu
+    attr_reader :menu_items
```
jamesjoshuahill: I can't see a test for this method. Please add one.

lib/order.rb
jamesjoshuahill: Please change `initialize`, so that you can inject a menu into an order. For example:
```ruby
menu = Menu.new
order = Order.new(menu)
```
Then you will be able to use test doubles instead of real `Menu` instances in your Order tests.

spec/order_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'order'
+
+describe Order do
```
jamesjoshuahill: After injecting Menu into Order, please use test doubles for the `Menu` class in your Order tests.

Step 19 of Boris Bikes has some examples: https://github.com/makersacademy/course/blob/master/boris_bikes/19_isolating_tests_with_doubles.md

spec/menu_spec.rb
jamesbingham: Ive added the menu items to this output statement.

spec/order_spec.rb
jamesbingham: Hi thanks for this I've moved the braces to after the add item arguments where it should be.

spec/order_spec.rb
jamesbingham: Hi this has been corrected but for some reason when I use curly braces for this expect statement it throws an error but with () it works fine 😕

spec/order_spec.rb
jamesbingham: Done :)

lib/menu.rb
jamesbingham: Hi, How can I add a test for a attr_reader?

lib/order.rb
jamesbingham: Changed the order tests from subject to order double

PR #1447, opened 2019-10-11
stories

zero201926:


lib/order.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_accessor :cost, :dishes, :menu
+  def initialize(menu)
+    @cost = 0
+    @dishes = {}
+    @menu = menu || Menu.new
+  end
+
+  def make_order(choice, quantity)
+    fail "#{choice} is not on the menu!" if !menu.has_food?(choice)
+    dishes[choice] = quantity
```
jamesbingham: Nice way of adding item to the order :)

lib/order.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def total
+    cost = totals.inject(:+)
+    cost
+  end
```
jamesbingham: not sure on what this .inject method does?

spec/send_sms_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:client) { double(:client, messages: messages) }
+  # client = Twilio::REST::Client.new(config[:account_sid], config[:auth_token])
+  let(:messages) { spy(:messages) }
+  let(:config) do
+    {
+      account_sid: "AC4599064055b8d324b89925cd5942e372",
+      auth_token: "86293f309beb51dc8b09da7e7d51d720",
+      from: "+44 1133201183",
+      to: "+44 7808275626",
+      body: "Thank you! Your order will be delivered before %s"
```
jamesbingham: I am able to see these credentials not sure if you were supposed to remove them?

zero201926:
I got stuck at some point, can’t remember which one so I opened the the Rspec from the exemplar and then did TDD from there

PR #1448, opened 2019-10-11
Update README.md

lukegittins01:


lib/customer.rb
```diff
@@ -2,9 +2,11 @@
 class Customer
 
 attr_reader :customer_dishes
+attr_reader :order_msg
```
luke-jones-1: these two could be on the same line if a comma was used. this would avoid repeating attr_reader

lib/customer.rb
```diff
@@ -0,0 +1,8 @@
+require_relative "dishes"
+class Customer
+
+  def check
+    @dishes
+  end
+
+end
```
luke-jones-1: nice use of customer class

spec/dishes_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'customer'
+require 'dishes'
+
+describe Dishes do
+
+  it 'should display the dishes and prices to the customer' do
+    customer = Customer.new
+    dish = Dishes.new
+    expect(customer.check).to eq(@dishes)
```
luke-jones-1: could have used doubles here

lib/dishes.rb
```diff
@@ -3,6 +3,8 @@ class Dishes
   attr_reader :dishes
 
   def initialize
-    @dishes = ["Chicken = £5", "Noodles = £3" , "Steak = £6"]
+    @dishes = ["Chicken = £5",
+               "Noodles = £3",
+               "Steak = £6"]
```
luke-jones-1: I like this

lib/customer.rb
lukegittins01: Thank you 👍

lib/dishes.rb
lukegittins01: Thanks I thought it would look better when you have multiple things in an array to separate them out

lib/customer.rb
```diff
@@ -0,0 +1,45 @@
+require_relative "dishes"
+
+class Customer
```
luke-jones-1: could have split this class into multiple classes as it has multiple responsibilities

spec/customer_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+#----- I HAVE OUTPUT THE TEXT MESSAGE TO THE TERMINAL -----
+  it 'will place an order' do
+    @customer.select(@dish.dishes[1])
+    @customer.select(@dish.dishes[2])
+    @customer.place_order
+    expect(@customer.place_order).to eq(@customer.order_msg)
+  end
+
```
luke-jones-1: tests are very easy to understand, well done

spec/customer_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'customer'
+require 'dishes'
+
+describe Customer do
+  it { is_expected.to respond_to :check }
+
+  before do
+    @dish = double(:dishes)
```
jamesjoshuahill: This test double only appears to be used in the test code. So I don't think you need a double. An array would be enough:
```ruby
describe Customer do
  let(:dishes) { ["Chicken = £5", "Noodles = £3", "Steak = £6"] }

  it 'can select a dish from the menu' do
    subject.select(dishes[1])
    ...
```
Please could you remove this double?

spec/customer_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  it 'can select a dish from the menu' do
+    subject.select(@dish.dishes[1])
+    expect(subject.customer_dishes).to include("Noodles = £3")
+  end
+
+  it 'will check the total of dishes' do
+    subject.select(@dish.dishes[1])
+    subject.select(@dish.dishes[2])
+    expect(subject.amount_of_dishes).to eq(2)
```
jamesjoshuahill: I think the story is ambiguous when it says:

> I would like to check that the total I have been given matches the sum of the various dishes in my order

Please could you try to calculate the total price for the order. For example:

1 x Chicken: £5
2 x Noodles: £6
**Total price: £11**

You will need to change the way you represent the list of dishes so that you can access the dish name and the price separately.

spec/dishes_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'customer'
+require 'dishes'
+
+describe Dishes do
+
+  it 'should display the dishes and prices to the customer' do
+    customer = Customer.new
+    expect(customer.check).to eq(@dishes)
```
jamesjoshuahill: I think `@dishes` has no value here, so it would be `nil`. Please could you test for what `customer.check` should return?

lib/customer.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def select(dish)
+    @customer_dishes << dish
+  end
+
+  def amount_of_dishes
+    @customer_dishes.count
+  end
+
+#----- I HAVE OUTPUT THE TEXT MESSAGE TO THE TERMINAL -----
```
jamesjoshuahill: It's great that you got the text message working 👍

spec/customer_spec.rb
lukegittins01: I have removed this double in the recent push and left it as the array like it was before peer review 👍

spec/customer_spec.rb
lukegittins01: I have done the calculations of the dish items in the recent push and it now calculates the total price not the total amount of items. I must have misread the task but its sorted now 👍

spec/dishes_spec.rb
lukegittins01: Thank you josh, I noticed that now I've fixed it. I used p to get visibility and then I checked the customer.check in irb and referenced the correct @dishes variable instead of referencing a new one in the code 👍

lib/customer.rb
lukegittins01: Thanks Josh it was fun and rewarding to do 👍

jamesjoshuahill:
Good work @lukegittins01, you made good progress on this challenge 🥇

lukegittins01:
> Good work @lukegittins01, you made good progress on this challenge 🥇

Thank you Josh, it made sense to change it to total price and not total number of items in the end. Worked out well. Cheers 👍

PR #1450, opened 2019-10-11
takeaway-challenge

josiepollard:


spec/order_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      expect(order).to respond_to :add
+    end
+  end
+
+  def confirm_order(time = calculate_time)
+    message = "Thank you! Your order was placed and will be delivered before #{time}"
+    return message if @confirmation == "Confirm"
+  end 
+
+end
```
josiepollard: I did not add the SMS function

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+    total = 0
+    @dishes.each { |_item, quantity| total += quantity }
+    total
+  end
+
+  def calculate_time
+    time = Time.now
+    "#{(time + 3600).strftime("%H:%M")}"
+  end
+
```
eagle310: good time calculation

spec/order_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'order'
+require 'menu'
+
+describe Order do
+  let(:menu) { double :menu }
+  subject(:order) { described_class.new(menu) }
+  it "can calculate the total cost of the order" do
+    expect(order).to respond_to(:total)
+  end
+
```
eagle310: maybe you could add some dishes to check if the total is correct?

spec/order_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      expect(test_order.total).to eq 10
+    end
+  end
+
+  describe "#add" do
+    it "adds dishes to the order" do
+      expect(order).to respond_to :add
+    end
+  end
+
```
eagle310: maybe you could add a test to check if the correct dishes have been added to the order?

PR #1451, opened 2019-10-11
heading added

adamr27:


lib/order.rb
```diff
@@ -19,4 +20,28 @@ def add_selection_to_order
     @shopping_cart << @selection
   end
 
+  def work_out_price
+    @shopping_cart.each do |food_item|
+      food_item.each do |item, quantity|
+        @price = menu.menu_items[item]
+        puts "#{quantity} x #{item} at £#{@price} = £#{@price *= quantity}"
+        @prices << @price
```
EarthWormGym: similar variable names could cause confusion.

lib/order.rb
```diff
@@ -19,4 +20,28 @@ def add_selection_to_order
     @shopping_cart << @selection
   end
 
+  def work_out_price
+    @shopping_cart.each do |food_item|
+      food_item.each do |item, quantity|
```
EarthWormGym: could refactor for better readability.

lib/order.rb
EarthWormGym: similar variable names could cause confusion.

spec/order_spec.rb
```diff
@@ -12,4 +12,32 @@
     expect(order.selection).to eq({ :rice => 2 })
   end
 
+  it "should add the selection to the order" do
+    expect(order.shopping_cart).to eq([{ :rice => 2 }])
+  end
+
+  it "should get the price of each item in the order" do
+    expect(order.work_out_price).to eq([{ :rice => 2 }])
```
EarthWormGym: great use of one line syntax.

spec/order_spec.rb
```diff
@@ -12,4 +12,32 @@
...
+  end
+
+  context "#price_calculation" do
+    before do
+      order.price_calculation
+    end
+
+    it "should display to total price" do
+      order.calculate_total_price
+      expect(order.display_total_price).to eq "Your total is £4"
```
EarthWormGym: good concise testing.

lib/order.rb
david-mears: Yes, I'll elaborate further on this: `each`es inside `each`es gets confusing. There is a cleaner way to look up whether an item exists on the menu, and look up an items price: splitting those functionalities out into methods. But where should those methods live? Ideally, on your Menu class: that preserves encapsulation, i.e. only the methods that have a direct need-to-know about the menu list are able to know about it.

lib/order.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+    end
+  end
+
+  def calculate_total_price
+    @total = @prices.sum
+  end
+
+  def price_calculation
+    puts work_out_price
```
david-mears: `puts` the return value of `work_out_price`? I can't see at a glance what `work_out_price` returns (as distinct from what it `puts`).

lib/order.rb
```diff
@@ -0,0 +1,47 @@
...
+    calculate_total_price
+    puts display_total_price
```
david-mears: I'm skeptical that making `display_total_price` into a method is useful - it returns an interpolated string which could have been written in this method anyway. Which would justify changing this method's name.

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "order"
+
+describe Order do
+  subject(:order) { described_class.new }
+  let(:menu) { double(:menu) }
```
david-mears: You've made a double, but you didn't do anything with it. I think you can skip this line!

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  it "should add the selection to the order" do
+    expect(order.shopping_cart).to eq([{ :rice => 2 }])
+  end
+
+  it "should get the price of each item in the order" do
+    expect(order.work_out_price).to eq([{ :rice => 2 }])
+  end
+
+  context "#price_calculation" do
```
david-mears: `context` is used to describe the conditions that you're testing under. E.g. if you want to test whether your food-pipe still works when you're upside-down, you might have the `context "when I'm upside-down"`.
`describe` means 'I'm about to describe the behaviour of the following method or a class'.
But they both actually _do_ the same thing.

lib/order.rb
david-mears: Separating these out also creates the problem that someone might call `display_total_price` without remembering to call `calculate_total_price` beforehand.

david-mears:
This is a good submission - well done for getting 100% test coverage. Have a look at the comment about doubles and see if that makes sense. If you implement this, or have questions, ping me.

PR #1455, opened 2019-10-14
still working through

EarthWormGym:
got stuck on this challenge so not very far through but have achieved 100% coverage and working tests for where i got to.

### DIFF
--- a/docs/customer.rb
+++ b/docs/customer.rb
@@ -1,0 +1,18 @@
+class Customer
+  attr_reader :customer_order_array
+
+  def initialize
+    @customer_order_array = []
+  end
+
+  def customer_order
+    p "What meal would you like: "
+    user_meal_input = gets.chomp
+    p '----------------------'
+    p "How many of those would you like?: "
+    user_amount_input = gets.chomp
+    p '-----------------------'
+    @customer_order_array << { :customer_meal => user_meal_input,
+      :amount_ordered => user_amount_input }
+  end
+end
--- a/docs/restaurant.rb
+++ b/docs/restaurant.rb
@@ -1,0 +1,16 @@
+class Restaurant
+  attr_accessor :list_dishes
+
+  def initialize
+    @list_dishes = { "pesto pasta" => 5,
+      "ravioli" => 10,
+      "pigeon" => 5,
+      "arabic herbs" => 10,
+      "mud pie" => 5 }
+  end
+
+  def menu
+    p "Here is a list of our dishes: #{@list_dishes}"
+  end
+
+end
--- a/spec/customer_spec.rb
+++ b/spec/customer_spec.rb
@@ -1,0 +1,15 @@
+require './docs/customer'
+
+describe Customer do
+
+  it 'orders meals' do
+    # Arrange
+    restaurant = Restaurant.new
+    # Act
+    restaurant.menu
+    subject.customer_order
+    # Assert
+    expect(subject.customer_order_array).to include({:amount_ordered=>"3",
+       :customer_meal=>"pesto pasta"})
+  end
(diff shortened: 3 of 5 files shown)


docs/customer.rb
```diff
@@ -0,0 +1,18 @@
...
+  def customer_order
+    p "What meal would you like: "
+    user_meal_input = gets.chomp
+    p '----------------------'
+    p "How many of those would you like?: "
+    user_amount_input = gets.chomp
+    p '-----------------------'
+    @customer_order_array << { :customer_meal => user_meal_input,
+      :amount_ordered => user_amount_input }
+  end
```
adamr27: This could be refactored to take in two arguments into the customer order method so that user_meal_input is set to one of the arguments and user_amount_input could be set to the other one so that means when you call the method you can declare what you want to order and how much u want to order in one go

adamr27: I like the way you have named each variable in the throughout the code as it makes it easier to understand what each variable means

docs/customer.rb
```diff
@@ -0,0 +1,18 @@
+class Customer
```
david-mears: By convention, the folder containing the code is 'lib' (created by rspec); in this case 'docs' exists in order to contain the 'review.md' file.

docs/restaurant.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def initialize
+    @list_dishes = { "pesto pasta" => 5,
+      "ravioli" => 10,
+      "pigeon" => 5,
+      "arabic herbs" => 10,
+      "mud pie" => 5 }
+  end
+
+  def menu
```
david-mears: Good method names are usually verbs, giving you an idea of the one thing the method does. I'd call this `print_menu` or similar.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './docs/restaurant'
+
+describe Restaurant do
+  it 'lists the dishes with their prices' do
+    # Arrange
+    restaurant = Restaurant.new
+    # Act
+    show_menu = restaurant.menu
+    # Assert
+    expect(show_menu).to eq("Here is a list of our dishes: #{subject.list_dishes}")
```
david-mears: You don't always need to use Arrange Act Assert. 'Act' is something that changes the world; e.g. docking a bike into a station changes where the bike is. But printing isn't like that. You can therefore refactor to `expect(restaurant.menu)` etc. For extra brownie points, there's a way to test what is being printed to 'stdout' (standard output): https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/output-matcher (Hint - you'll need to take newlines into consideration with printing.)

spec/customer_spec.rb
```diff
@@ -0,0 +1,18 @@
+require './docs/customer'
+
+describe Customer do
+
+  it 'orders meals' do
+    # Arrange
+    restaurant = Restaurant.new
```
david-mears: This line makes the test for the `Customer` class depend on the `Restaurant` class working. How can you get around this?

david-mears:
Well done on 100% test coverage! To get up to a green, I’d like to see progress on [this issue](https://github.com/makersacademy/takeaway-challenge/pull/1455#pullrequestreview-302010647) and [this issue](https://github.com/makersacademy/takeaway-challenge/pull/1455#pullrequestreview-302002291). Please ping me when you’ve done that.

spec/customer_spec.rb
EarthWormGym: oh sorry I think I left that there from when I was trying to work out the next user story, I shall remove it now.

PR #1456, opened 2019-10-14
Take-away 

luke-jones-1:


lib/receipt.rb
```diff
@@ -0,0 +1,15 @@
+require 'order'
+
+module Receipt
```
lukegittins01: I like the way you have made a separate class for the total order that makes it all look very clear 👍

lib/customer.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def initialize
+    @order = Order.new
+  end
+
+  def list_of_menu_items
+    Menu.list
+  end
+
+  def print_list
```
lukegittins01: Nice extra feature to print the menu items to the user 👍

lib/receipt.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def generate_receipt(order)
+    order.items.each { |item| "#{item}\n" }
+  end
+
+  def receipt_total(order)
+    total = 0
+    order.each { |item| total += item[:quantity]*item[:price].to_f}
+    "Total = #{sprintf "%.2f", total}"
```
lukegittins01: What does the %.2f mean?

luke-jones-1: it makes sure that the number has 2 decimal places

lib/customer.rb
luke-jones-1: thank you 👍

spec/customer_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'customer'
+
+describe Customer do
```
jamesjoshuahill: These tests use a real `Order` in the `Customer` class. How could you use dependency injection to isolate these unit tests from the `Order` class?

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+  def items
+    @order.map { |item| "#{item[:item]} #{item[:quantity]} * #{item[:price]} #{sprintf "%.2f", (item[:quantity]*item[:price].to_f)}" }
+  end
+
+  private
+
+  def menu_item
+    index = MENU.index {|h| h[:item] == item }
+    MENU[index]
+  end
```
jamesjoshuahill: This method reaches into the `MENU` constant. Could you extract this method into the `Menu` module?

PR #1457, opened 2019-10-14
all

albie-b:


lib/deliverytime.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
+class DeliveryTime
+
+  def initialize(number)
+    @number, @time = number, Time.new
```
Adam1824: I like how you've assigned the two instance variables on one line

lib/takeaway.rb
```diff
@@ -0,0 +1,26 @@
...
+class Order
+
+  attr_reader :items
+
+  def initialize(items = nil)
+    @items = items
+  end
+
+  def set_menu
+    @menu = MainMenu.new
```
david-mears: This is a 'hard-coded' dependency: the Order class will only be bug-free as long as the MainMenu class is also bug-free, so the classes are 'coupled'. This is something we want to avoid in Object-Oriented Design: the more that classes are dependent on each other, the more tangled and knotted the whole system will gradually become, until progress slows to a stop. The way to avoid this? _Dependency injection_. This is handled by the `initialize` method.

Once the dependency is injected (not hard-coded), then it becomes possible to test the classes independently too ('unit testing'), by using a double to stand in the place of MainMenu object. But as it is, line 13 is forced to use a real object, so there's no way to use a double. That means you are stuck when it comes to testing Order separately from testing MainMenu.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    expect(Order.new('pizza, curry').items).to eq(('pizza, curry'))
+  end
+  it 'should confirm total' do
+    customer = Order.new('pizza, curry')
+    customer.set_menu
+    expect(customer.confirm_total).to eq("pizza = 5.0 curry = 2.2 total = £7.2")
+  end
+  it 'texts conformation through order' do
+    puts 'enter number'
+    number = gets.chomp
```
david-mears: The trouble here is that to run the test you need a human there typing in the answers, and the human needs to know what they're supposed to type in. Ideally, all the tests will run automatically, so the test suite can be done in seconds. We need to stub `gets` so that when `gets` is called, it returns something pre-defined in your test, instead of using the normal `gets` method.

david-mears:
Well done for implementing all the user stories!

You have a high level of test coverage, but also three classes which aren't tested. So I guess that the test for Order is running code in the other classes. I can also guess that the creation of those classes wasn't test-driven because there aren't unit tests for them.

Could you add tests for the other classes, stub `gets` in the tests, and isolate the Order tests by using a double for MainMenu? That would make this a solid green or supergreen, for me. Are my comments above enough for you to do that? Ping me when you've made changes, or if you have questions.

spec/takeaway_spec.rb
albie-b: I just had this because i properly shouldn't push my phone number to GitHub

PR #1458, opened 2019-10-14
Pull rq 

2Blagden:


lib/takeaway.rb
```diff
@@ -34,9 +33,7 @@ def list_name
 
 def send_a_text
   @client.messages.create(
-    from: '+13605295607',
-    to: '+447808274842',
-    body: 'Test'
```
Will-Bath: This isn't the most ideal thing to be on your pull request ;)

lib/takeaway.rb
```diff
@@ -0,0 +1,70 @@
...
+
+def process_order(order)
+end
+
+m = Menu.new
+chips = Meal.new("chips", 5)
+kfc = Meal.new("KFC", 3)
+m.list_of_dishes << chips
+m.list_of_dishes << kfc
+make_order
```
Will-Bath: It would be good to have manual tests like this as part of the spec tests instead - but otherwise looks like it would work as intended

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+end
+
+describe "make order" do
+  it 'meals can be added to the order list' do
+    m = Menu.new
+    meal1 = Meal.new("kfc", 2)
+    m.list_of_dishes << meal1
+    expect(m.list_of_dishes.count).to eq(1)
```
Will-Bath: This test could be improved by the condition being to increase by 1 rather than equalling 1 - otherwise it's a good use of arrange act and assert :)

lib/takeaway.rb
```diff
@@ -0,0 +1,70 @@
...
+  def list_price
+    @price
+  end
+
+  def list_name
+    @name
+  end
+end
+
+def send_a_text
```
Will-Bath: Good method names, makes it easy to follow what's going on

PR #1460, opened 2019-10-14
whoops

2Blagden:
sorted key issue out

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,7 +1,8 @@
 require 'twilio-ruby'
 
 # put your own credentials here
-#taken them out for pull request
+account_sid = 'AC71ede767f3f69bf43718d8094b244800'
+auth_token = '88ac7f43909c00b5b8e9a32d5b554c35'
 
 # set up a client to talk to the Twilio REST API
 @client = Twilio::REST::Client.new account_sid, auth_token
@@ -33,7 +34,9 @@ end
 
 def send_a_text
   @client.messages.create(
-
+    from: '+13605295607',
+    to: '+447808274842',
+    body: 'Test'
   )
   puts ("tried to send a text")
 end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,8 +1,7 @@
 require 'twilio-ruby'
 
 # put your own credentials here
-account_sid = 'AC71ede767f3f69bf43718d8094b244800'
-auth_token = '88ac7f43909c00b5b8e9a32d5b554c35'
+#taken them out for pull request
 
 # set up a client to talk to the Twilio REST API
 @client = Twilio::REST::Client.new account_sid, auth_token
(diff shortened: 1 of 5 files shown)


lib/takeaway.rb
```diff
@@ -1,8 +1,7 @@
 require 'twilio-ruby'
 
 # put your own credentials here
-account_sid = 'AC71ede767f3f69bf43718d8094b244800'
-auth_token = '88ac7f43909c00b5b8e9a32d5b554c35'
+#taken them out for pull request
```
david-mears: Deleting them doesn't quite work for security - I can still find them here! Once you have got feedback from a coach on this repo, I'd consider making it a private repo.

lib/takeaway.rb
```diff
@@ -0,0 +1,70 @@
+require 'twilio-ruby'
+
+# put your own credentials here
```
david-mears: A good instruction to add to your README.

lib/takeaway.rb
```diff
@@ -0,0 +1,70 @@
...
+# set up a client to talk to the Twilio REST API
+@client = Twilio::REST::Client.new account_sid, auth_token
+
+class Menu
+  def initialize
+    @list_of_dishes = []
+  end
+  def list_of_dishes
+    @list_of_dishes
+  end
```
david-mears: This is fine; you could also use `attr_reader :list_of_dishes` to achieve the same effect.

david-mears: Also, there's a way to build the program that doesn't involve `Menu` spilling its secrets (= allowing other objects to directly access its list of dishes): [see the exemplar](https://github.com/makersacademy/takeaway-exemplar/blob/master/lib/menu.rb).

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'takeaway'
+
+describe Menu do
+  it 'shows list of dishes' do
+  end
```
david-mears: Did you forget to write this test, because it showed as 'passing'? If you run a
```ruby
it "blah blah" do
end
```
with nothing in it, it will show up as passing. You can tag it as pending by using `xit` instead of `it`.

lib/takeaway.rb
```diff
@@ -0,0 +1,70 @@
...
+  ordername = gets.chomp
+  p("quantity:")
+  orderquan = gets.chomp.to_i
+  order << ordername
+  order << orderquan
+  process_order(order)
+end
+
+def process_order(order)
+end
```
david-mears: All these methods - `send_a_text`, `show_menu`, `make_order` - should live on a class. Think about what classes best suit the responsibilities of these methods.

david-mears:
The two most important things for me are [this issue](https://github.com/makersacademy/takeaway-challenge/pull/1460/commits/c859d607da5a2e124234cf1450c9d2843c09ed48#r335339216) and completing the tests (the `make_order` and `show_menu` methods need tests, and the test for the first user story is empty). For any new test or method, make a commit with the new test first, and then a commit where the test has been passed by the code. This would be good evidence for good progress (green) in OOP. For excellent progress, test-drive a method that returns the total price for the order. Ping me when you've done this.

PR #1461, opened 2019-10-14
started twilio

eagle310:


spec/order_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'order'
+require 'menu'
+
+describe Order do
```
jamesjoshuahill: These tests use a real `Menu` in the `Order` class. How could you isolate these unit tests from the `Menu` class using test doubles?

PR #1464, opened 2019-10-25
Rhys' Takeaway Challenge

rhyse1:


spec/menu_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    it "should init with an empty basket" do
+        expect(subject.basket).to eq(Array.new)
+    end
+
+    it "should display a list of items in the menu" do
+        expect(subject.menu).to include(a_kind_of(Item))
+    end
+
+    it "should allow the user to add various items to the basket" do
+        subject.AddToBasket("Pizza", 1)
```
lindseydew: Oooh this looks like c sharp casing. Convention in ruby is lower case and underscores

spec/menu_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        expect(subject.basket[0].name).to eq("Pizza")
+    end
+    it "should allow the user to add numerous amounts of the same item to the basket" do
+        subject.AddToBasket("Pizza", 2)
+        expect(subject.basket[1].name).to eq("Pizza")
+    end
+
+    it "should return the total of the order to the customer" do
+        subject.AddToBasket("Pizza", 1)
```
lindseydew: Could be worth testing the total method when other items are added in?

src/Menu.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'item'
+#require 'twilio-ruby'
+
+class Menu
+    attr_reader :basket, :menu
+
+    def initialize()
+        @basket = []
+        @menu = [Item.new("Pizza", 3.99), Item.new("Burger", 4.99), Item.new("Fries", 2.99)]
```
lindseydew: We could break this out into two objects - one responsible for displaying the menu and understanding what is available, and another responsible for keeping track of the things that are ordered.

src/Menu.rb
```diff
@@ -0,0 +1,46 @@
...
+        @basket.each do |item|
+            total += item.price
+        end
+        return total
+    end
+
+    def PurchaseOrder
+=begin
+        @client = Twilio::REST::Client.new(
+            ENV['ACCOUNT_SID'],
```
lindseydew: This looks like you are on the right track. Would be hard to test this method without actually sending the message. Is there a way we could design this class so that we can check the message sent back to the user without actually sending a text?

PR #1468, opened 2019-10-25
Initial Pull request

khan2597:


lib/takeaway.rb
```diff
@@ -0,0 +1,25 @@
+class Takeaway
+    attr_reader :order
+    attr_reader :price
+
+    def initialize(dishes = Dishes.new)
```
lindseydew: Nice use of dependency injection! This could be further utilised by using a mock when testing the takeaway class

lib/takeaway.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def order_item(dish, amount = 1)
+        item = @dishes.dishes_arr.select { |list| list[:item] == dish }.first
+        amount.times {
+            @order << item
+        }
+    end
+
+    def order_sum
+        @order.map { |item| item[:price] }.reduce(:+)
```
lindseydew: Nice use of higher order functions :)

PR #1469, opened 2019-10-25
Takeaway Challenge

Slongboro:


spec/dish_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    expect(dish).to be_kind_of(Dish)
+    end
+
+    it "should be able to see a list of dishes" do
+        dish = Dish.new
+        expect(dish.menu).to include("Pizza")
+    end 
+    
+    it "should be able to select a number of dishes from the menu" do
+        dish = Dish.new
```
lindseydew: I can see where you are going and this is really on the right lines. 

In order to be able to call `selection` on dish we need to say:

``` dish.selection("Curry",5) ``` 

To check that this has worked ok we would want to observe the _effect_ of the method, which in this case adds to `order`

So we could say 
``` expect(dish.order).to include("curry") ```

File: spec/dish_spec.rb [hunk 2]
```diff
@@ -0,0 +1,21 @@
+require 'dish'
+
+describe Dish do
+
+    it "should create a new instance of dish" do
+        dish = Dish.new
+    expect(dish).to be_kind_of(Dish)
```
lindseydew: Nice!

spec/dish_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    it "should be able to see a list of dishes" do
+        dish = Dish.new
```
lindseydew: Looks good!

PR #1470, opened 2019-10-25
Progress on the Takeaway Challenge

marcb1988:
Progress was slower than anticipated due to some syntax and general coding errors on my part and rspec/irb errors caused by the gem file installation. The first two user stories were completed at the time the challenge time elapsed.


lib/dish.rb
```diff
@@ -0,0 +1,8 @@
+class Dish
+  attr_reader :name, :price
+
+  def initialize(name, price)
```
lindseydew: Cool 👍

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
...
+      Dish.new("Pizza", 5.20),
+      Dish.new("Kebab", 3.00),
+      Dish.new("Chips", 1.00),
+      Dish.new("Drink", 0.80),
+    ]
+  end
+
+  def view_menu
+    dishes.each do |dish|
+      puts "#{dish.name}" + " £" + ('%.2f' % dish.price).to_s
```
lindseydew: Perhaps the formatting of the dish and price could be abstracted into its own method?

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+  end
+
+  def order(selection)
+    @cust_order = []
+    @dishes.each do |dish|
+      if dish.name == selection
```
lindseydew: Would be good to test this method

PR #1471, opened 2019-10-25
Pull Request

swade19:


spec/dish_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "dish"
+
+describe Dish do
+  let(:dish) { Dish.new }
+
+  it "creates a instance of Dish" do
+    expect(dish).to be_instance_of(Dish)
+  end
+
+  # it "initializes with a hash of dishes " do
```
lindseydew: Nice I can see where you are going with this.

We can be explicit here and say:

``` expect(dish.dishes).to eq({"pizza":5, "coke":2})

File: lib/takeaway.rb [hunk 2]
```diff
@@ -0,0 +1,22 @@
+require_relative "dish"
+
+class Takeaway
+  attr_accessor :food, :order
+
+  def initialize
+    @food =
+      { "pizza": 5,
+        "coke": 2 }
```
lindseydew: If we want our class to reference data in another class we can say:

``` dish = Dish.new
      @food = dish.food
```

PR #1472, opened 2019-10-25
Lena Krasnova' attempt at Takeaway challenge

olkras03:


spec/dish_spec.rb
```diff
@@ -0,0 +1,18 @@
+require "dish"
+
+describe Dish do
+  subject(:dish) { Dish.new }
+
+  it "should be instance of class dish" do
+    expect(dish).to be_instance_of Dish
+  end
```
lindseydew: Nice I can see where you are going here :)

PR #1481, opened 2019-10-27
Alex Clifford - Takeaway Challenge

Clifford2910:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  it 'shows the menu with dishes and prices' do
+    expect(takeaway.print_menu).to include("Margherita: £6")
+  end
+
+  it 'can select some number of several available dishes' do
+    takeaway.add("Pepperoni", 2)
+    takeaway.add("Hawaiian", 1)
+    expect(selected_dishes).to eq(selected_dishes)
```
edcourage: not sure what this is testing? I think it will always pass regardless as your asking if the same thing is the same??? if that makes sense

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
+class Takeaway
```
edcourage: Could be split into more classes

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  MENU_LIST = {
```
edcourage: Think we should avoid hard coding if possible.

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+    sum = 0
+    raise "no dishes to check price on!" if @selected_dishes.length <= 0
+    @selected_dishes.each do |dish|
+      dish.each do |dish, quantity|
+        sum += price(dish) * quantity
+      end
+    end
+    sum
+  end
+
```
edcourage: could be a private method?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+# I would like to be able to select some number of several available dishes
+#
+# As a customer
+# So that I can verify that my order is correct
+# I would like to check that the total I have been given matches the sum of the various dishes in my order
+#
+# As a customer
+# So that I am reassured that my order will be delivered on time
+# I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
+require 'takeaway.rb'
```
edcourage: maybe add nested context/describe so it easier to read.

PR #1491, opened 2019-10-28
Pull request from Andy Jones. Complete.

jonesandy:


lib/basket.rb
```diff
@@ -0,0 +1,17 @@
+class Basket
+  attr_reader :basket
+
+  def initialize
+    @basket = {}
+  end
+
+  def set_basket(item, quantity)
```
ingridbjarman: I like the way you have structured this and it's easy to see what it does and how it will work, but could the method have a more clear name perhaps?

jonesandy: Very good point.  I will remember to sense check method names.

PR #1495, opened 2019-10-28
Topher takeaway challenge

topherb99:


lib/order.rb
```diff
@@ -0,0 +1,39 @@
+require 'twilio-ruby'
```
eliseaston: Would this work better as a gem in your Gemfile? I couldn't get the program to run on my computer because I couldn't find the required 'twilio file

eliseaston: Just spoken to you and realise this is now because it's in your .gitignore file

.gitignore
```diff
@@ -1,2 +1,3 @@
 /**/.DS_Store
```
eliseaston: Just a note in general: README file is always useful for other users for your code. I found this blog post quite useful: https://medium.com/@meakaakka/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3

spec/order_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    it 'Returns current total when order is selected' do
+      expect(subject.select('pizza', 3)).to eq 'Your order will cost £18'
+    end
+
+    it 'returns current total when multiple items are selected' do
+      subject.select('pizza', 3)
+      expect(subject.select('nuggets', 2)).to eq 'Your order will cost £26'
+    end
+  end
+end
```
eliseaston: As we've just discussed, there should be a test to test the send_text method that stubs out the response - more details in the code review rubric file

eliseaston: Ideally you should have an independent test for each of the methods in your code - you might want to add a test for the .total and the .checkout methods

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+class Order
+  DEFAULT_MENU = {
+    curry: 6,
+    chips: 2,
+    pizza: 6,
+    nuggets: 4,
+    salad: 3
+  }
+  attr_reader :menu, :order
+  include Twilio
```
eliseaston: This seems really neat :)

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+    @menu = menu
+    @order = []
+  end
+
+  def checkout
+    meals = []
+    @order.each { |hash| meals << hash[:quantity]
+                 meals << hash[:meal]
+    }
+    body = "Thank you! Your order was placed at #{(Time.now).strftime("%k:%M")} and will be delivered before #{(Time.now + 2700).strftime("%k:%M")}"
```
eliseaston: You could consider splitting the send_sms and body into a separate method

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def select(item, quantity)
+    @order << { meal: item, quantity: quantity, total: (@menu[item.to_sym] * quantity) }
+    "Your order will cost £#{total}"
+  end
+
+  def total
+    total = 0
+    @order.each { |hash| total += hash[:total]}
```
eliseaston: This seems v neat! :)

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+    meals = []
+    @order.each { |hash| meals << hash[:quantity]
+                 meals << hash[:meal]
+    }
+    body = "Thank you! Your order was placed at #{(Time.now).strftime("%k:%M")} and will be delivered before #{(Time.now + 2700).strftime("%k:%M")}"
+    send_text(body)
+  end
+
+  def select(item, quantity)
+    @order << { meal: item, quantity: quantity, total: (@menu[item.to_sym] * quantity) }
```
eliseaston: You could just use this in your .total method (your suggestion!)

PR #1496, opened 2019-10-28
adds text ending class and method

Ben230:


lib/text.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
+
+class Text
+  def initialize
+    account_sid = 'ACa353025d1bc42a06b259ff9332279f8d'
+    auth_token = '30e174a43e758ef20e98c283551427e0'
```
emmavanoss: I think it'd be better to use env variables for these too, as they could let someone get into your account

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+    @prices = []
+    @choice.each { |food| @prices << @menu.menu[food] }
+  end
+
+  def verified?
+    calculate_price == @user_total
+    send_text
+  end
+
+  private
```
emmavanoss: I like that send_text is private. I think some of the other methods could be too

lib/order.rb
```diff
@@ -0,0 +1,33 @@
+require 'menu'
+
+class Order
+  def initialize(*choice, quantities, user_total)
```
emmavanoss: This seems like a lot of arguments to have to pass in, and a lot of work that initialize is doing. Maybe you could split it up between the methods, e.g. add items/quantities one by one through a separate method and then get the user_total after that.

lib/text.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilio-ruby'
```
emmavanoss: I saw this gem was missing from your Gemfile, it'd be good to add it so that anyone running 'bundle' will get it downloaded

lib/order.rb
```diff
@@ -0,0 +1,33 @@
+require 'menu'
```
emmavanoss: I couldn't feature test in irb because this file wouldn't load. You need to use 'require_relative' to make that possible.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+    @menu = Menu.new
+  end
+
+  def calculate_price
+    convert_dishes_to_prices
+    @price = (@prices.zip(@quantities)).map { |x, y| x * y }
+    @price.sum
+  end
+
+  def convert_dishes_to_prices
```
emmavanoss: I like your method and variable names, and this is a good example.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+
+class Order
+  def initialize(*choice, quantities, user_total)
+    @choice = choice
+    @quantities = quantities
+    @user_total = user_total
+    @menu = Menu.new
+  end
+
+  def calculate_price
```
emmavanoss: I'm not sure about your approach to entering dishes and quantities, but this is a nice way to handle them in the method

PR #1498, opened 2019-10-28
Unfinished Takeaway challenge - Ingrid Bjarman

ingridbjarman:


jonesandy:
I'd suggest updating the README as you go through the project. Creating your own README at the start and stripping out any Makers specific information.  You can always refer to the actual Makers repo to look at the challenge while your own README is built out at the stages you are.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
+require 'dish'
+
+class Menu
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = [
```
jonesandy: I love how you have made the option to inject dishes at creation of menu.  Maybe added more work as for the challenge we could just use a preset menu. But code is flexible.

jonesandy:
Code was easy to understand. Names make sense and could follow the flow of how the application takes shape.

PR #1503, opened 2019-11-12
Creating Pull Request

selinaerhabor:


lib/restaurant.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  def menu
+    @menu = [
+      {dish: 'Kebab', price: '£5.00'},
+      {dish: 'Large Fries', price: '£1.50'},
+      {dish: 'Regular Fries', price: '£1.00'},
+      {dish: 'Pounded Yam', price: '£2.00'},
+      {dish: 'Egusi Soup', price: '£4.50'},
+      {dish: 'Jellof Rice', price: '£2.00'},
+    ]
```
rahulshah7: Have a read about the  [open/closed principle](https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). Can you find a way of moving the menu array outside of the source code of the class?

rahulshah7: The menu prices are stored as strings, can you think of a problem this will cause with maths operations on them?

PR #1505, opened 2019-11-12
Rob Grant pull request

djlonboy:


lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  def initialize(menu) #Pass the correct Menu as argument
+    @menu = menu
+    @order_list = [] #Array containing sets of [item, quantity]
+    @status  = "not placed"
+  end
+
+  # Adds the menu item and quantity to the order
+  def add_item(item, quantity)
+    raise "Item not on menu" if on_menu?(item) == false
```
marianneoco: Great to include edge cases! As user has to type name of item, likely to make errors etc

lib/order.rb
```diff
@@ -0,0 +1,70 @@
...
+  attr_reader :order_list, :status
+
+  def initialize(menu) #Pass the correct Menu as argument
+    @menu = menu
+    @order_list = [] #Array containing sets of [item, quantity]
+    @status  = "not placed"
+  end
+
+  # Adds the menu item and quantity to the order
+  def add_item(item, quantity)
```
marianneoco: Useful to be able to specify quantity (in mine, they'd have to add something 3 times)

PR #1507, opened 2019-11-12
Sam King Pull Request - Takeaway challenge

sk52:
Status of solution documented in README.md
3rd user story completed, yet to begin user story 4. 
Edge cases require tests, and use of doubles also required in places.


spec/menu_spec.rb
```diff
@@ -0,0 +1,30 @@
+require './lib/menu.rb'
+
+describe Menu do
+  it { is_expected.to be_instance_of(described_class) }
+
+  it { is_expected.to respond_to(:dishes) }
+
+  describe '#add' do
+    it 'should add item to the menu attribute dishes' do
+      dish = double(:dish)
```
anhnguyenis: Is a double necessary here?

anhnguyenis (commented):
Good code. I can understand what your approach was :)

PR #1509, opened 2019-11-12
AN

anhnguyenis:


spec/menu_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  it 'responds to list' do
+    menu = Menu.new
+    expect(menu).to respond_to(:list)
+  end
+
+  it 'shows menu list of dishes with prices' do
+    menu = Menu.new
+    menu.list
+    expect(:meal => "Fish", :price => 2.00).to include(:meal => "Fish", :price => 2.00)
```
sk52: I'm not sure if this is testing what you think it is, it looks like it might always pass due to the way the test is setup, maybe have a look at this

PR #1511, opened 2019-11-12
Marianne O'Connor - October 2019 Cohort - Pull Request

marianneoco:


lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+  def total_cost
+    @cost_calculator = 0
+    cost_calculator_hash = Hash[ @order.map {|key, value| [key, (value.to_f * Takeaway::MENU[key].to_f)]}]
+    cost_calculator_hash.each do |key, value|
+      @cost_calculator += value
+    end
+    "Total cost is £#{sprintf("%0.02f", @cost_calculator)}"
+  end
+
+  def confirm(message = Message.new)
```
djlonboy: Good work getting the text messaging API to work!

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+    @order = {}
+  end
+
+  def add(item)
+    if @order[item] == nil
+      @order[item] = 1
+    else
+      count = @order[item]
+      @order[item] = count + 1
+    end
```
david-mears: Check out the conditional assignment operator, it might save space here: https://stackoverflow.com/a/14697343

lib/order.rb
david-mears: By reading the Message object, I realised that this method is for sending a confirmation text message. Good dependency injection here.

README.md
```diff
@@ -13,6 +13,19 @@ Takeaway Challenge
...
+1 test failing - attempt to prevent text from sending each time rspec is run.
+Also currently am able to send texts to my own number but have no corresponding rspec tests to confirm this.
+Need to deal with case where user adds something which is not on the menu.
+
+To use:
+* Load takeaway.rb in IRB
+* takeaway = Takeaway.new
+* takeaway.start_order
+* takeaway.new_order.add("Margherita")
+* takeaway.total_cost
```
david-mears: Takeaway doesn't have a `total_cost` method; I presume you mean `takeaway.new_order.total_cost`.

marianneoco: Oops yes, thanks!

lib/order.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/takeaway'
+require './lib/message'
+require 'twilio-ruby'
+
+class Order
+
+  attr_reader :order
```
david-mears: It's a _**bit**_ confusing to have an attribute called `order` on an instance of `Order`. Makes me think of [this](https://www.youtube.com/watch?v=H4v7wddN-Wg).

README.md
```diff
@@ -13,6 +13,19 @@ Takeaway Challenge
...
+
+1 test failing - attempt to prevent text from sending each time rspec is run.
+Also currently am able to send texts to my own number but have no corresponding rspec tests to confirm this.
+Need to deal with case where user adds something which is not on the menu.
+
+To use:
+* Load takeaway.rb in IRB
+* takeaway = Takeaway.new
+* takeaway.start_order
+* takeaway.new_order.add("Margherita")
```
david-mears: Thinking further ahead in the life of the program, I'm not sure of the wisdom of storing the last order on the Takeaway/Restaurant, and then recommending people interact with that. What if two people are ordering at the same time? When the second person starts their order, it will replace the previous unfinished `@new_order`. They will both be adding items to whichever order was most recently created.

david-mears: Just noticed that the above doesn't apply because you are creating a new Takeaway for each user. I made that assumption because of how in real life there will only be one restaurant, which centrally manages what is contained in the menu, and can take orders, etc.

spec/order_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'order'
+
+describe Order do
+
+  let(:order_double) { double :order }
```
david-mears: This is where naming things the same gets confusing. It's unclear if this is doubling an `Order` object or an `@order` instance variable belonging to such an object. I presume it's the latter because we want to be testing real `Order`s, but you never know.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+describe Takeaway do
+
+  it { is_expected.to be_instance_of Takeaway }
+
+  it "should contain a list of dishes with prices" do
+    expect(Takeaway::MENU).to be_instance_of Hash
+  end
+
+  it "should allow the user to place an order" do
```
david-mears: It doesn't test this, it just tests whether a method called 'start_order' exists.

david-mears:
A very good submission :)
Well done for getting 95% test coverage. Killing it.
Bonus points are available for isolating your unit tests with doubles...

lib/message.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class Message
+
+  def initialize
+    account_sid = "AC9e08cf89a069abf78ee15d77cd7a1363"
+    # Your Test Account SID from www.twilio.com/console/settings
+    auth_token = "84f4c2388aca22d5b248fdfe1c396c6e"
+    # Your Test Auth Token from www.twilio.com/console/settings
+
+    @client = Twilio::REST::Client.new account_sid, auth_token
```
david-mears: Good chance to use dependency injection: don't want to bake in a reliance on Twilio.

PR #1513, opened 2019-11-12
First commit

savagevince:


lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+class Menu
```
sujesh-roy: Like the fact that you have created separate class and spec file for menu as suggested in our learning module.

PR #1516, opened 2019-11-12
Implementation of user stories

rahulshah7:


spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'dish.rb'
+
+describe Dish do
+  subject { described_class.new('pizza', 10) }
+  it { is_expected.to be_an_instance_of described_class }
```
selinaerhabor: Created instance of Dish, a useful initial test

PR #1518, opened 2019-11-12
End of day 1

LucyMatejka:


spec/dishes_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'dishes'
+
+describe Dishes do
+  it 'creates an instance of Dishes' do
+    dishes = Dishes.new
+    expect(dishes).to be_an_instance_of(Dishes)
+  end
```
SimoneW2: Good unit test

lib/dishes.rb
```diff
@@ -0,0 +1,14 @@
+class Dishes
+
+  def initialize
+    @list = [ "fish £6",
+            "chicken £7",
+             "curry £5",
+             "pizza £4",
+             "vegetarian £3"]
```
SimoneW2: Good use of an array. Maybe a hash with key and value pairs for later tests when refactoring, however I am not completely sure myself.

PR #1522, opened 2019-11-13
Takeaway challenge - not currently working

JNuckley:


lib/Meals.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def menu
+  #  @dishes.each do |key, value|
+  #  puts "#{key} #{value}"
+  # end
+end
+# only pushing the meal name so far :( and not the price
+  # def order(meal)
+  #   @dishes.find { |item| item[:meal] == meal }
+  #   @ordered << meal
```
amgc3: substitute item where is says << meal

PR #1529, opened 2019-11-24
Takeaway Weekend Challenge

EManifold:


Gemfile
```diff
@@ -4,5 +4,6 @@ gem 'capybara'
 gem 'rake'
 gem 'rspec'
 gem 'rubocop', '0.71.0'
+gem 'twilio-ruby', '~> 5.29.1'
```
AndreaDiotallevi: The 'Rubocop' gem suggests to list the gems in alphabetical order.

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+class Menu
+  attr_reader :dishes
+  def initialize
+    @dishes = [
+      {number: 1, dish: 'Chicken Biryani', price: 8.99},
+      {number: 2, dish: 'Vegetable Dansak', price: 7.99},
+      {number: 3, dish: 'Rogan Josh', price: 9.49},
+      {number: 4, dish: 'Chicken Korma', price: 9.99}
+    ]
```
AndreaDiotallevi: You are hardcoding the dishes. It might be useful to inject the dishes as a dependency so that in the future you can implement a different menu.

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+class Menu
+  attr_reader :dishes
+  def initialize
+    @dishes = [
+      {number: 1, dish: 'Chicken Biryani', price: 8.99},
```
AndreaDiotallevi: To make the code cleaner you could you a Dish class, with @name and @price as instance parameters.

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
...
+      {number: 2, dish: 'Vegetable Dansak', price: 7.99},
+      {number: 3, dish: 'Rogan Josh', price: 9.49},
+      {number: 4, dish: 'Chicken Korma', price: 9.99}
+    ]
+  end
+
+  def list_dishes
+    @dishes.each do |dish|
+      p "Number #{dish[:number]}".center(10) + "||".center(5) + "#{dish[:dish]}".center(15) + "£#{dish[:price]}".center(15)
```
AndreaDiotallevi: It is probably better to 'return' the string, rather than "p"ing it.

lib/my_order.rb
```diff
@@ -0,0 +1,50 @@
+require_relative 'menu'
```
AndreaDiotallevi: I am not sure if you need to "require" menu.

lib/my_order.rb
```diff
@@ -0,0 +1,50 @@
...
+      @actual_total += dish[:price]
+    end
+    p "Actual total: #{@actual_total}"
+    p "Calculated total: #{@calculated_total}"
+    raise "Incorrect total" if @actual_total != @calculated_total
+    "Total is correct"
+  end
+
+  def place_order
+    Text.new.send
```
AndreaDiotallevi: It is probably better to inject Text as a dependency.

lib/my_order.rb
```diff
@@ -0,0 +1,50 @@
+require_relative 'menu'
+
+class MyOrder
+
+  attr_reader :my_order, :menu
+
+  def initialize(menu = Menu.new)
+    @my_order = []
```
AndreaDiotallevi: You could improve the variable naming to @order_list, to be more specific.

lib/text.rb
```diff
@@ -0,0 +1,19 @@
+require 'twilio-ruby'
+
+class Text
+  def send
+    # put your own credentials here
+    account_sid = 'AC335e55ee7a7e11ba978b9d86a0c4a24f'
+    auth_token = 'f8c2febc1532775de9cd9b91f72ef6ca'
```
AndreaDiotallevi: You should hide this sensitive information inside environment variables.

lib/text.rb
```diff
@@ -0,0 +1,19 @@
...
+    # set up a client to talk to the Twilio REST API
+      @client = Twilio::REST::Client.new account_sid, auth_token
+
+      date_time = Time.now + 3600
+      time = date_time.strftime("%H:%M")
+      @client.messages.create(
+        from: '+18506378703',
+        to: '+447812758744',
```
AndreaDiotallevi: You should hide this sensitive information inside environment variables.

spec/menu_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'menu'
+
+describe Menu do
+
+  let(:dish_example) { {number: 1, dish: 'Chicken Biryani', price: 8.99} }
+
+  describe '#list_dishes' do
+    it "should show the available dishes in the menu" do
+      expect(subject.list_dishes).to include dish_example
```
AndreaDiotallevi: I think you are testing state rather than behaviour. Maybe better to test that the return value is a string listing the dishes with prices, like ```"Chicken Biryani (£8.99), etc."

spec/my_order_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'my_order'
+
+describe MyOrder do
+
+  before(:each) do
+    allow(subject).to receive(:actual_total) { :calculated_total }
```
AndreaDiotallevi: I think you should stub doubles and not the actual test subject, otherwise the class code would not run and tested.

Gemfile
```diff
@@ -4,5 +4,6 @@ gem 'capybara'
 gem 'rake'
```
AndreaDiotallevi: Remember to add your readme file.

PR #1531, opened 2019-11-24
My own solution

AndreaDiotallevi:


.gitignore
```diff
@@ -1,2 +1,5 @@
 /**/.DS_Store
 /coverage
+
+#Ignore .env file
```
EManifold: Test comment

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader :dishes
+
+  def initialize
+    @dishes = []
+  end
+
+  def add(dish)
+    fail "This dish is already in the menu" if contains? dish
+
+    @dishes.push(dish)
```
EManifold: Maybe use shovel - shorter/easier to read?

lib/restaurant.rb
```diff
@@ -0,0 +1,24 @@
...
+    @takeaway = takeaway
+    @text_provider = text_provider
+  end
+
+  def checkout(price)
+    fail "The takeaway basket is empty" if @takeaway.empty?
+    fail "Payment amount not correct" unless @takeaway.correct_amount?(price)
+
+    @text_provider.send_text(delivery_time)
+    @takeaway.complete
```
EManifold: Nice addition of takeaway.complete

lib/restaurant.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  private
```
EManifold: How come delivery time is a private method?

lib/text_provider.rb
```diff
@@ -0,0 +1,16 @@
+require "dotenv/load"
```
EManifold: What's dotenv/load?

EManifold: Env variable?

spec/menu_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "menu"
+
+describe Menu do
+  let(:menu)   { Menu.new }
```
EManifold: How come you've used this and not subject?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  describe "#correct_amount?" do
+    it "should return 'true' if the given price matches the total cost of the takeaway" do
+      takeaway.order(dish_1, 5)
+      takeaway.order(dish_2, 2)
+      expect(takeaway.correct_amount?(74)).to eq true
+    end
+
+    it "should return 'false' if the given price matches the total cost of the takeaway" do
```
EManifold: If the given price doesn't match, no?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+      takeaway.order(dish_1, 5)
+      takeaway.order(dish_2, 2)
+      expect(takeaway.correct_amount?(20)).to eq false
+    end
+  end
+
+  describe "#empty?" do
```
EManifold: Some method names with question mark could be more descriptive - we don't know that empty is referring to basket.

spec/text_provider_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+  let(:client) { double :client }
+
+  describe "#send_text" do
+    it "should be able to send a text message to confirm successful order and 1 hour delivery time" do
+      message = "Thank you! Your order was placed and will be delivered before 18:55"
+      twilio_message_body = { from: ENV['TWILIO_PHONE_NUMBER'], to: ENV['MY_PHONE_NUMBER'], body: message }
+      allow(client).to receive_message_chain(:messages, :create).with(twilio_message_body)
+      expect(Twilio::REST::Client).to receive(:new).with(ENV['TWILIO_ACCOUNT_SID'], ENV['TWILIO_AUTH_TOKEN']).and_return(client)
+      text_provider.send_text("18:55")
+    end
```
EManifold: This is good! I couldn't work out how to test for the text output

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+    "#{dish.name} added to the menu"
+  end
+
+  def list
```
alicelieutier: This method could be named to reflect that you are formatting a string, not just returning a list.

alicelieutier: Also, since you are mostly formatting the dish, I'm wondering if it would be better to delegate formatting to the dish class.

lib/restaurant.rb
```diff
@@ -0,0 +1,24 @@
+class Restaurant
+  DELIVERY_TIME = 3600 # one hour
+
+  attr_reader :takeaway, :text_provider
+
+  def initialize(takeaway = Takeaway.new, text_provider = TextProvider.new)
+    @takeaway = takeaway
```
alicelieutier: Basket or order seems like a better name for takeaway.

lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
+class TakeAway
+  attr_reader :menu, :basket
+
+  def initialize(menu = Menu.new)
+    @basket = Hash.new(0)
+    @menu = menu
+  end
+
+  def read_menu
```
alicelieutier: I don't know why this method exist. It does not seem to be doing much.

lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
...
+    @menu.list
+  end
+
+  def order(dish, quantity = 1)
+    fail "This dish is not in the menu" unless @menu.contains?(dish)
+
+    @basket[dish] += quantity
+    "#{quantity}x #{dish.name}(s) added to your basket"
```
alicelieutier: Why does this method return a string?

lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  def basket_summary
+    @basket.map do |dish, quantity|
+      "#{dish.name} x#{quantity} = £#{dish.price * quantity}"
+    end.join(", ")
+  end
+
+  def total
```
alicelieutier: Naming - This should be "Total for display" or something similar.

spec/menu_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  describe "#list" do
+    it "should return an empty string if there are no dishes in the menu" do
+      expect(menu.list).to eq ""
+    end
+
+    it "should return the list of dishes" do
+      menu.add(dish_1)
+      menu.add(dish_2)
+      expect(menu.list).to eq "Carbonara (£10), Puttanesca (£12)"
```
alicelieutier: This is a good test for add, too. 
You don't need to make add return a string just to test it - you can test it using other methods like here.

Basically, your tests should be for behaviours, not for methods.

PR #1533, opened 2019-11-24
Project (incomplete)

JKBero:


lib/take_away.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  def basket_summary
+    @basket
+    # Not passing the tests in the desired format...
+    # ...skipped step due to time constraints
+  end
+
+  def total
+    total_price
```
Katerina-codes: Is this variable needed here?

PR #1536, opened 2019-11-24
Pull request

jj49411:


lib/menu.rb
```diff
@@ -0,0 +1,18 @@
...
+  def initialize
+    @menu = { "Sushi" => 5, 
+    "Curry" => 6, 
+    "Egg roll" => 2, 
+    "Soup" => 1 }
+  end
+
+  def display
+    @menu.each do |dish, price|
+      p "#{dish}: £#{price}"
```
ffgi-es: I think the `p` here is what is causing one of the tests to fail. It might be better to use `puts` here for normal output.

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+    @menu.display
+  end
+
+  def select_dish(dish, quantity)
+    @basket[dish] += quantity
+  end
+
+  def total
+    @sum = @basket.reduce(0) do |sum, (dish, quantity)| 
+    sum + (@menu.menu[dish] * quantity)
```
ffgi-es: It might be better to have a method on the Menu class that allows you to get the dish rather than exposing the menu attribute and allowing people to change the menu whenever they like.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  
+  describe '#select_dish' do
+    it 'should add a dish to the basket' do
+      subject.select_dish(:pie, 2)
+      expect(subject.basket).to include({ pie: 2 })
+    end
+  end
+
+  describe '#total' do
+    #how can I assign { "sausage" => 4 } to the menu double?
```
ffgi-es: you could do `allow(menu).to receive(:menu).and_return({ "sausage" => 4 })` since the test is complaining that your total method is calling `#menu` on the menu double.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      subject.select_dish(:sausage, 6)
+      allow(textsender).to receive(:send_text)
+    end
+    
+    it 'should raise an error if the amount is incorrect' do
+      expect{ subject.place_order(3) }.to raise_error("Incorrect total")
+    end
+
+    it 'should send a confirmation text' do
+      expect(textsender).to receive(:send_text)
```
ffgi-es: You need to actually call the method `#place_order` if you want to test this expectation

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def view_basket
+    @basket.each do |dish, quantity|
+      p "#{dish}x#{quantity}"
+    end
+    p "Total:£#{@sum}"
+  end
+
+  def place_order(amount)
+    fail "Incorrect total" if amount != @sum
```
ffgi-es: What happens if you haven't call `#total` yet before you try and place an order?

PR #1537, opened 2019-11-24
uncompleted, but will be

DawidSzpener:


ajbacon:
Verbal comments on code given directly.

Would be useful (for future reference) to get some instructions in the README

PR #1540, opened 2019-11-25
Takeaway Challenge Attempt

Jamie95187:


lib/menu.csv
```diff
@@ -0,0 +1,8 @@
+dishes,prices
```
IngramCapa: Creating a CSV file for the menu rather than hardcoding the menu in a hash is a good example of creating clear areas of 'responsibility'

lib/menu.rb
```diff
@@ -0,0 +1,28 @@
+require 'csv'
```
IngramCapa: I've learned lots reading this, now I'll know how to load a CSV file in the future :)

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'order'
+
+describe Order do
+
+  describe '#add_item' do
+    context '1 item in the order' do
+      it 'adds an item to be printed later' do
+        item_dbl = double("Burger", :Burger => "£4")
+        order = Order.new
```
IngramCapa: You could add any doubles you plan on using for a specific method between the describe and the context so you don't have to repeat the code in other tests (e.g. check the total_per_dish adds up when you add a dish a second time

lib/menu.rb
IngramCapa: For the purpose of this exercise, a hash would've sufficed, but thumbs up for thinking of real life scenarios

spec/spec/order_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'order'
+
+describe Order do
+
+  describe '#add_item' do
+    context '1 item in the order' do
```
IngramCapa: you could add a test to see how the order behaves when adding a second dish

PR #1547, opened 2019-11-25
takeaway challenge submission

ajbacon:


DawidSzpener:
This is great, although it looked complicated at first

PR #1550, opened 2019-12-20
Masha's takeaway challenge

marleyhill:


spec/order_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'order'
+
+describe Order do
+  let(:menu) { double :menu, list: "juice - £1.00" }
+  let(:menu_class) { double :menu_class, new: menu }
```
lindseydew: Great use of mocking here! 

Curious as to why you are passing in the reference to the class? I think in this case passing in the object itself should be ok since the Order object only needs to refer to one Menu object

lib/order.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+    t
+  end
+
+  def print_receipt
+    # @choice.each do |i|
+    #   i.each {|k, v| puts "#{k} - £%0.2f" % v }
+    # end
+    self.work_out_order
+    puts "Total to pay: £%0.2f" % self.total_price
```
lindseydew: Great use of Single Responsibility here!

lib/order.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def work_out_order
+    number = {}
+    count = 1
+    @choice.each do |h|
+      h.each_key do |k|
+        if !number.has_key?(k)
```
lindseydew: Could this method be simplified? Perhaps the quantity could always be stored (even if it's one) to negate the need of the conditional here?

spec/order_spec.rb
marleyhill: Thank you for your comments, Lindsey! I did actually do the dependency injection challenge during the week before the Takeaway challenge, and was relying on it when creating doubles (hence both class and object doubles as I was just following what was suggested there), but clearly I have misunderstood it :) (am working on resolving the misunderstanding...)

PR #1551, opened 2019-12-20
Laura takeaway challenge

laek:


lib/order.rb
```diff
@@ -0,0 +1,17 @@
...
+  def initialize
+    @ordered_dishes = { }
+  end
+
+  def total
+
+  end
+
+  def add_item(dish, quantity = 1)
+    added_item = {dish => @menu[dish]} #trying to access the hash in another class but not much luck with it.
```
david-mears: RSpec gave the error:

```
NoMethodError:
       undefined method `[]' for nil:NilClass
```

So, Ruby can't use the method written with square brackets on a NilClass object. That is a hint that `@menu` is `nil` in this case. Why would a variable evaluate to nil?

<details><summary>Click here once you've answered the question</summary>
Right, if you haven't initialized a variable, it returns `nil`.

So menu has to be accessible within this class Order. Probably you'll pass in a Menu object as an argument when you create the Order object.
</details>

spec/menu_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+describe Menu do
+
+  it "initialises a new instance of the Menu class" do
+    expect(subject).to be_a Menu
+  end
+
+  it "contains a list of dishes with prices" do
+    expect(subject.menu).to be_a Hash
+    expect(subject.menu["Fried rice"]).to be_a Numeric
+    expect(subject.menu).to have_key("Fried rice")
```
david-mears: This test ties you to a specific menu. You could improve your code by not hard-coding the menu: https://github.com/nboyse/takeaway-challenge/blob/master/docs/review.md#open-closed-principle

PR #1552, opened 2019-12-20
Iryna's Takeaway Challenge 

iryna-deans:
Hi Eoin,

Please find below my takeaway challenge. I have completed all four user stories. I started doing the first two and realised I created a wrong class (Dish) so I went back and renamed it to Menu and did it as a Hash so it resembled a Menu.

For User Story 4, I watched a youtube video which explained how to send a message in Ruby using Twilio. I then tried it out and it worked. However I didn't use an ENV variable(for my account sid & auth token) so I had to delete my original repository and lost all my commits and reupload all of the files into a new repository(hence why there is only one commit). Please refer to my README file as I took detailed notes of my commits and my failing tests there as well :) 

Wishing you a Merry Christmas and wonderful New Year, 
Iryna

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -18,7 +18,7 @@ class Takeaway
 
   def select(item,quantity=1)
     @result = @menu.find { |x| x[item] }
-    @basket.concat([@result]* quantity)
+    @basket.push([@result] * quantity)
     end
 
   def total
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,51 @@
+require_relative 'Menu'
+require("bundler")
+Bundler.require()
+
+
+class Takeaway
+  attr_reader :basket, :result,:menu
+
+  def initialize
+    @basket = []
+    @menu = nil
+    @total = 0
+  end
+
+  def see(menu)
+    @menu = menu.list
+  end
+
+  def select(item,quantity=1)
+    @result = @menu.find { |x| x[item] }
+    @basket.concat([@result]* quantity)
+    end
+
+  def total
+    @basket.each { |x|
+      x.each { |key, value|
+      @total = @total + value }}
+      @total
+  end
+
+  def is_amount_correct(total)
+    time = Time.now + 1*60*60
+    if @total == total
+      send_text( "Thank you! Your order was placed and will be delivered before #{time}")
+    elsif
+      fail "Error"
+    end
+  end
+
+  def send_text(message)
+    account_sid = 'ACXXXXXXXXXX'
+    auth_token = 'XXXXXXX'
+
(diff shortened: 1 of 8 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    @menu = nil
+    @total = 0
+  end
+
+  def see(menu)
+    @menu = menu.list
+  end
+
+  def select(item,quantity=1)
+    @result = @menu.find { |x| x[item] }
```
lindseydew: Nice, this looks like it would work. I think you could simplify the method here to not need concat, by only pushing the item into the array.

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+  def total
+    @basket.each { |x|
+      x.each { |key, value|
+      @total = @total + value }}
+      @total
+  end
+
+  def is_amount_correct(total)
```
lindseydew: Nice! Could you think about a way of testing this method to check it gives the right response with the time? Perhaps Time.new could be passed in as a dependency to the object?

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    time = Time.now + 1*60*60
+    if @total == total
+      send_text( "Thank you! Your order was placed and will be delivered before #{time}")
+    elsif
+      fail "Error"
+    end
+  end
+
+  def send_text(message)
```
lindseydew: Have you any thoughts on how you might test this?

lindseydew (commented):
Great job on this! Looks like you have a good understanding of how objects can interact. There are some opportunities in your code to make use of dependency injections so we can safely unit test the functionality. This activity might be helpful for you to look at to practise the concept :) https://github.com/makersacademy/skills-workshops/blob/master/practicals/object_oriented_design/dependency_injection.md

lib/takeaway.rb
iryna-jurkiv: Hi Lindsey, Thank you for your suggestion. I have changed my code accordingly to use push (instead of concat) and it works :)

lib/takeaway.rb
iryna-jurkiv: Hi Lindsey, 

Thank you for your suggestion. I did some research and was thinking I could test this method with something like this: 

allow(Time).to receive(:now).and_return(@time_now)

I am going to try this with my Chitter challenge as I used Time.now there in a method as well :)

lib/takeaway.rb
iryna-jurkiv: Hi Lindsey, 

I did some research and I think I could use a stub/mock outlined in this stackoverflow post https://stackoverflow.com/questions/35095857/rspec-trying-to-stub-a-method-that-returns-its-own-argument.

PR #1554, opened 2019-12-20
Callum's takeaway challenge

CalMcD:


lib/takeaway.rb
```diff
@@ -0,0 +1,39 @@
...
+  def display_menu
+    @items.collect do |k, v|
+      "#{k}, £#{v}0"
+    end
+  end
+end
+
+class Order
+
+  attr_reader :basket, :main_order
```
david-mears: I'm not sure I understand what `@main_order` is for. I can see that you use it to create a hash to put in `@basket`. But I can't see why (a) it needs to be an instance variable and (b) it needs to be exposed to things outside the class with an attribute reader.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'takeaway'
+
+describe Menu do
+  it 'can initialize a constant of menu' do
+    menu = Menu.new
+  end
+
+  it 'holds a list of available items' do
+    menu = Menu.new
+    expect(menu.items).to include :chicken => 5.0
```
david-mears: This test ties you to a specific, hard-coded menu. https://github.com/nboyse/takeaway-challenge/blob/master/docs/review.md#open-closed-principle

PR #1555, opened 2019-12-20
TakeAway-Challenge

Davidbaez87:


lib/menu.rb
```diff
@@ -0,0 +1,19 @@
...
+require 'order.rb'
+class Menu
+  attr_reader :dishes
+  DISHES_LIST = {
+    "Chicken Rice" => 4.50,
+    "Fried Chicken" => 5.50,
+    "Fried Rice" => 2.50,
+    "Side Salad" => 1.50,
+    "Noodle Soup" => 3.50
+  }
```
david-mears: You could improve your code by not hard-coding the menu: https://github.com/nboyse/takeaway-challenge/blob/master/docs/review.md#open-closed-principle

lib/order.rb
```diff
@@ -0,0 +1,20 @@
+class Order
+  attr_reader :basket, :item, :quantity
```
david-mears: I don't see the need to expose the value of `item` and `quantity` to things outside the class with these attribute readers. These attribute readers at the moment will return `nil` if called, because there are no instance variables called `@item` or `@quantity`. You're correct not to make them into instance variables, as you've only used them as local variables.

lib/takeaway.rb
```diff
@@ -0,0 +1,19 @@
+require 'menu.rb'
+require 'order.rb'
```
david-mears: You probably don't need all these `require`s here, because any program/interface/test suite using the classes should load them itself. Then you can control precisely which ones it loads.

lib/takeaway.rb
```diff
@@ -0,0 +1,19 @@
...
+class Takeaway
+  attr_reader :menu, :place_order
```
david-mears: You have an attribute reader here for an attribute that doesn't exist (`@place_order`). I'm guessing this appeared in order to pass the last test in takeaway_spec.rb?

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'menu.rb'
+describe Menu do
+  subject(:menu) { described_class.new }
+  context "#initialize" do
```
david-mears: `context` introduces a description of a situation, e.g. "when the user is not logged in". `describe` announces that the next tests are testing a certain method or class.

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    it 'create a list if dishes' do
+      menu = Menu.new
+      expect(menu.dishes).to eq described_class::DISHES_LIST
+    end
+  end
+
+  context "#print_menu" do
+    it 'prints the menu' do
+      menu = Menu.new
+      show_menu = menu.dishes.each{ |dish, price| "#{dish}: £#{price}, " }
```
david-mears: The test would be more robust if you wrote out the string you meant. What if there were a bug in this line, or in `.dishes`?

spec/order_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'order.rb'
+require 'takeaway.rb'
+describe Order do
+let(:item) {'Chicken Rice'}
+let(:quantity) { 1 }
+let(:menu) { double(:menu, dishes: { 'Chicken Rice' => 4.50 }) }
+it 'creates the instance of order'do
+order = Order.new
+end
```
david-mears: This is very hard to read without indentation.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+takeaway = Takeaway.new
+  allow(menu).to receive(:print) { menu.dishes }
+  expect(takeaway.print_menu).to eq menu.print
+end
+end
+it 'place an order' do
+takeaway = Takeaway.new
+expect(takeaway).to respond_to :place_order
+end
+end
```
david-mears: This is very hard to read without / with inconsistent indentation.

PR #1556, opened 2019-12-20
Takeaway challenge

Rhodine:
partial solution

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,7 +1,7 @@
 class Takeaway
   def initialize(menu:menu)
     @menu = menu
-    @order = []
+    @order = order
   end
 
   def dishes_with_prices
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,3 +1,6 @@
 class Takeaway
+  def dishes_with_prices
+
+  end
 
 end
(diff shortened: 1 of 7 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,18 @@
+class Takeaway
+  def initialize(menu:menu, order:nil)
```
david-mears: Nice use of required keyword arguments. In `menu:menu`, you're saying that the default value of the thing on the left is the thing on the right. But the thing on the right hasn't been initialized, so this is effectively the same as `menu:nil`.

lib/takeaway.rb
```diff
@@ -0,0 +1,18 @@
...
+    @menu = menu
+    @order = order
+  end
+
+  def dishes_with_prices
```
david-mears: I'd probably make it explicit in the method name that this is a method for displaying/printing.

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+      chips: 3.00
+    }
+  end
+
+  it "creates an instance of Menu" do
+    expect(menu).to be_a Menu
+  end
+
+  it "has a list of dishes with prices" do
+    expect(menu.dishes).to eq(dishes)
```
david-mears: Great test - you dynamically generated the menu instead of hard coding it.

PR #1557, opened 2019-12-20
User_stories 1 and 2

carmen-alina:
Merry Christmas & Happy New Year!

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,18 @@
+require_relative 'dishes_list'
+
+class Menu
+  attr_reader :menu
+
+  def initialize
+    @menu = {}
+  end
+
+  def add(dish)
+    @menu[dish.name] = dish.price
+  end
+
+  def print_menu
+    @menu
+  end
+
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,3 @@
+class Order
+
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -9,6 +9,7 @@ class Order
   end
 
   def select(dish, quantity)
+    @order_list[quantity] = dish
   end
 
 end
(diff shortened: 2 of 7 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'dishes_list'
+
+class Menu
+  attr_reader :menu
+
+  def initialize
+    @menu = {}
+  end
+
+  def add(dish)
```
lindseydew: Nice, this menu class looks great!

spec/dishes_list_spec.rb
```diff
@@ -0,0 +1,19 @@
+require './lib/dishes_list.rb'
+
+describe DishesList do
+
+  subject {described_class.new("Salad", 4)}
+
+  it 'instantiates the class with 2 arguments' do
+    expect(subject).to be_an_instance_of(DishesList)
```
lindseydew: Lovely!

lib/order.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class Order
+  attr_reader :order_list, :total
+
+  def initialize
+    @order_list ={}
+    @total = 0
+  end
+
+  def select(dish, quantity)
```
lindseydew: This looks great!

spec/order_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  it 'initializes an empty order list' do
+    expect(subject.order_list).to be_empty
+  end
+
+  it { is_expected.to respond_to(:select).with(2).argument }
+
+  describe ".select" do
+
+    let(:dish){ double :dish}
```
lindseydew: Looks like you are making good attempts at a double here! How could we make sure the double is able to call the thing that the method calls?

PR #1558, opened 2019-12-20
Takeaway Challenge - Story 2 somewhat complete

annelouD:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:order) { double :order }
+  let(:printed_menu) { "Special Fried Rice: £6.50" }
+  let(:dishes) { { rice: 2, curry: 1 } }
+
+  it 'creates instance of class' do
+    expect(takeaway).to be_a(Takeaway)
+  end
+  
+  it 'allows user to see a list of dishes with prices' do
+    expect(takeaway.print_menu).to eq(printed_menu)
```
david-mears: This tests that `takeaway.print_menu` successfully calls `menu.print`. It does call it, and the test passes because you've defined `menu.print` on line 5. But the real menu object has no `print` method (it's called `prints`). So the test is giving a false positive.

To avoid this, RSpec has [verifying doubles](https://github.com/rspec/rspec-mocks/tree/master/features/verifying_doubles):

> Verifying doubles are a stricter alternative to normal doubles that provide guarantees about what is being verified. When using verifying doubles, RSpec will check that the methods being stubbed are actually present on the underlying object if it is available. Prefer using verifying doubles over normal doubles.

> No checking will happen if the underlying object or class is not defined, but when run with it present (either as a full spec run or by explicitly preloading collaborators) a failure will be triggered if an invalid method is being stubbed or a method is called with an invalid number of arguments.

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
+class Menu
+ # open/close principle - open to extension closed to modification
+  attr_reader :dishlist
+  def initialize(dishlist)
```
david-mears: This is great - injecting the list of dishes dynamically instead of hard-coding them.

PR #1577, opened 2020-01-06
Christy Takeaway Challenge

christymasterson:
Takeaway challenge from before Christmas break. 
Started on user story 2, not yet got the 'select dishes' method to work. 
Struggled with this challenge - in process of following walkthrough on YouTube.

### DIFF
--- a/lib/order_manager.rb
+++ b/lib/order_manager.rb
@@ -1,0 +1,20 @@
+require_relative 'menu'
+
+class OrderManager
+
+  def initialize
+    @menu = Menu.new
+    @order = []
+  end
+
+  def show_menu
+    @menu.show_dishes
+  end
+
+  def select(dishes)
+    dishes.each do |dish, quantity|
+    order.add(dish, quantity)
+    end
+  end
+
+end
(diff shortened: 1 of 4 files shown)


lib/order_manager.rb
```diff
@@ -0,0 +1,20 @@
...
+    @menu = Menu.new
+    @order = []
+  end
+
+  def show_menu
+    @menu.show_dishes
+  end
+
+  def select(dishes)
+    dishes.each do |dish, quantity|
```
david-mears: RSpec gave me this error here:

```
NoMethodError:
       undefined method `each' for "pad thai":String
```

What types of object can have the method `each` called on them?

lib/order_manager.rb
```diff
@@ -0,0 +1,20 @@
+require_relative 'menu'
+
+class OrderManager
+
+  def initialize
+    @menu = Menu.new
+    @order = []
+  end
+
+  def show_menu
```
david-mears: I think this method is untested!

PR #1578, opened 2020-01-06
Complete - no receiving order by text

basselalsayed:


lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+require_relative 'item'
+
+class Menu
+  attr_reader :items
+
+  def initialize
+    @items = {}
+  end
+
+  def add_item(item, name, price)
```
Uvalente: Could specify a default Item class

lib/send_sms.rb
```diff
@@ -0,0 +1,22 @@
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load('sensitive.env')
```
Uvalente: Should refer to the ENV variables after loading them

PR #1579, opened 2020-01-06
Takeaway Challenge 20.12.19

nboyse:
Hi,

This is the code I wrote for the Takeaway Challenge we completed before Christmas break.

Natasha.

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,75 @@
+class Takeaway
+
+  def initialize(menu = Menu.new, checkout = Checkout.new)
+    @menu = menu
+    @checkout = checkout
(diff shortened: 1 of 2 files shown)


david-mears:
Well tested (100% coverage!) and well divided into classes.
By convention, separate classes live in separate files named after the class, e.g. `class Checkout` in checkout.rb
I find it confusing to have attributes of an object which have the same name as the class. You have menu.menu and checkout.checkout, and people (such as me) might get confused as to what each part represents.
You could also improve this by not hard-coding the menu: https://github.com/nboyse/takeaway-challenge/blob/master/docs/review.md#open-closed-principle
Generally, excellent.

PR #1582, opened 2020-01-09
Object

lindseydew:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+      dishes = {
+        "Chicken Korma" => 4.50,
+        "Veggie Burrito" => 3.00,
+        "Fries" => 3.00,
+        "Beef Burger" => 7.00,
+        "Steak" => 7.00,
+        "Kinder Bueno Waffle" => 5.00,
+        "Vanilla Ice Cream" => 5.00
+      }
+      expect(Takeaway.new.menu(dishes)).to include("Chicken Korma" => 4.50)
```
lindseydew: Nice, it's great that you have got this test to work! I think you actually don't need to call the `menu` method, as you are defining dishes as a variable in Takeaway. 

so this could be simplified to:

```
takeaway = Takeaway.new
expect(takeaway.dishes).to include("Chicken Korma" => 4.50)
```

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+   end
+ end
+   describe '#dishes_selection' do
+     it { is_expected.to respond_to(:dishes_selection).with(1).argument }
+
+     it'putes dishes selection' do
+      expect(Takeaway.new.dishes_selection("Chicken Korma")).to eq("Item: Chicken Korma costs £4.5")
```
lindseydew: It's lovely that you have got this formatting well!

lib/order.rb
```diff
@@ -0,0 +1,12 @@
+require_relative 'takeaway'
+
+def initialize
+  @balance
+end
+# verify that my order is correct
```
lindseydew: Wondering whether you were intending to make this into it's own class?

PR #1586, opened 2020-02-02
Takeaway challenge completed

jasylwong:
All user requirements met. Customised text sent upon correct payment by the user.

### DIFF
--- a/lib/basket.rb
+++ b/lib/basket.rb
@@ -1,11 +1,10 @@
 class Basket
 
-  attr_reader :dishes # , :quantities, :estimated_total
+  attr_reader :dishes
   
   def initialize(menu)
     @dishes = []
     @menu = menu
-    # @quantities = []
   end
 
   def add(dish, quantity = 1)
@@ -29,7 +28,7 @@ class Basket
     summary = ""
     @dishes.uniq.each do |dish|
       summary += "#{dish.capitalize} x #{@dishes.count(dish)}: "\
-        "£#{@menu.dishes[dish] * @dishes.count(dish)}\n"
+        "£#{@menu.dishes[dish.to_sym] * @dishes.count(dish)}\n"
     end
     summary += "Total: £#{total}\n"
   end
--- a/lib/basket.rb
+++ b/lib/basket.rb
@@ -1,0 +1,37 @@
+class Basket
+
+  attr_reader :dishes # , :quantities, :estimated_total
+  
+  def initialize(menu)
+    @dishes = []
+    @menu = menu
+    # @quantities = []
+  end
+
+  def add(dish, quantity = 1)
+    @menu.search(dish)
+    quantity.times { @dishes << dish }
+  end  
+  
+  def remove(dish, quantity = 1)
+    quantity.times do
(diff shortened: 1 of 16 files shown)


spec/order_spec.rb
```diff
@@ -0,0 +1,58 @@
+require 'order'
+
+describe Order do
+  let(:menu) { Menu.new({ pizza: 4, burger: 3, hot_dog: 2 }) }
```
Kefuri: Can implement a double here to isolate tests?  `let(:menu) { double("menu double") }`
And then allow menu to receive x method and return y

spec/basket_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'basket'
+
+describe Basket do
+  # let(:menu) { double :menu }#, new: ({ pizza: 4, burger: 3, hot_dog: 2 })) }
+  # :menu.new({ pizza: 4, burger: 3, hot_dog: 2 })
+  # allow(:menu).to receive(:new).with({ pizza: 4, burger: 3, hot_dog: 2 })
+  let(:menu) { Menu.new({ pizza: 4, burger: 3, hot_dog: 2 }) }
```
Kefuri: isolation needed here too!

lib/sms.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def initialize(account_sid, auth_token, from_num, to_num)
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+    @from_num = from_num
+    @to_num = to_num
+  end
+
+  def send
+    @client.messages.create(from: @from_num, to: @to_num,
+    body: "Thank you for ordering with Jasiveroo!"\
```
Kefuri: I'd like a Jasiveroo web app soon

lib/basket.rb
```diff
@@ -0,0 +1,36 @@
...
+    @dishes = []
+    @menu = menu
+  end
+
+  def add(dish, quantity = 1)
+    @menu.search(dish)
+    quantity.times { @dishes << dish }
+  end  
+  
+  def remove(dish, quantity = 1)
```
Kefuri: if you'er defaulting the quantity in the method that calls this method, I don't think you need to default it again!

- Furthermore, is there a reason to have a call for this inside the order class? or can it be called directly in IRB? order.remove calls basket.remove but could you not do basket.remove straight?

lib/order.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'basket'
+require_relative 'menu'
+require_relative 'sms'
+
+class Order
+  attr_reader :basket, :menu, :sms
+  
+  def initialize(menu, account_sid, auth_token, from_num, to_num)
+    @basket = Basket.new(menu)
+    @menu = menu
```
Kefuri: reassigning menu here might be unnecessary - its available in menu and basket

lib/basket.rb
```diff
@@ -0,0 +1,36 @@
+class Basket
```
Kefuri: Potential restructure of basket? Move the functionality into order and let basket be a way of storing the dishes. That way you can just keep a write_to method in here and add dishes as they come.

PR #1587, opened 2020-02-02
Weekend Challenge - TakeAway

hansa-lallu:
Created a takeaway menu 
Completed the first three user stories. 
Did not complete the send message function

### DIFF
--- a/lib/notifier.rb
+++ b/lib/notifier.rb
@@ -1,4 +1,6 @@
+require 'twilio-ruby'
+
 class Notifier
-  def send_text
+  def send_text(messgae)
   end
 end
--- a/lib/notifier.rb
+++ b/lib/notifier.rb
@@ -1,0 +1,3 @@
+class Notifier
+
+end
--- a/lib/notifier.rb
+++ b/lib/notifier.rb
@@ -1,3 +1,4 @@
 class Notifier
-
+  def send_text
+  end
 end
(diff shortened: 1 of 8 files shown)


lib/order.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'menu'
+require_relative 'notifier'
+
+class Order
+
+  attr_reader :basket, :menu
+
+  def initialize(menu = Menu.new)
```
davmcgregor: Good idea using a dependency injection

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    @basket = Hash.new(0)
+    @menu = menu
+  end
+
+  def add_item(item, quantity = 1)
```
davmcgregor: Smart move setting default quantity

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    check_item_exits(item)
+    @basket[item] = @basket[item] + quantity
+    add_basket_confirmation(item, quantity)
+  end
+
+  def total_order
```
davmcgregor: Overall very impressive code. Would be good to see a README showing how it all works together

lib/notifier.rb
```diff
@@ -0,0 +1,6 @@
+require 'twilio-ruby'
```
davmcgregor: Next steps would be to connect this to the Twilio API: https://www.twilio.com/docs/usage/api

PR #1588, opened 2020-02-02
incomplete ☹

Mrtly:
I can’t make both the feature and the unit tests pass at the same time. 🤔
I’ve left it where the feature test works, as I trust it more.
Most likely the code works (up to half of the reqs), but I don’t know how to write proper spec tests.
Did not attempt the text part.

### DIFF
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,0 +1,20 @@
+require "menu"
+
+describe Menu do
+  subject(:menu) { described_class.new(dishes) }
+
+  let(:dishes) do
+    {
+      chicken: 3.50,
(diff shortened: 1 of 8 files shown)


spec/menu_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+  let(:dishes) do
+    {
+      salad: 3.50,
+      falafel: 4.50,
+    }
+  end
+  it "has a list of dishes and prices" do
+    expect(menu.dishes).to eq(dishes)
+  end
+
```
Peter2-71828: can you delete this first test? Is it redundant now you have the second test?

PR #1593, opened 2020-02-02
Incomplete Challenge

Ovy95:


lib/order.rb
```diff
@@ -0,0 +1,21 @@
+class Order
+  attr_reader :dishes
+
+  def initialize(menu)
+    @dishes = {}
+    @menu = menu
+  end
+
+  def add(dish, quantity)
+    fail NoItemError, "#{dish.capitalize} is not on the menu!" unless menu.has_dish?(dish)
```
saidbyced: Great to check if item exists before adding to order! I totally didn't think of this.

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  subject(:menu) { described_class.new(dishes) }
+    #injects the Hash making it easier for future updates to the menu
+  let(:dishes) do 
+    {
+      pizza: 8.99,
+      wedges: 2.50,
+      soda: 1.00
+      
+    }
```
saidbyced: Great double - enough info to test while using not using all your time to create it.

README.md
```diff
@@ -82,3 +82,6 @@ Notes on Test Coverage
 ------------------
 
 You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+
+Not my code, followed the walkthough from youtube,to understand what had to be done
+As my code went to pot.
```
saidbyced: Adding a direct reference to resources would ensure your helper would get some credit

lib/takeaway.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def place_order(dishes)
+    dishes.each do |dish, quantity|
+      order.add(dish, quantity)
+    end
+  end
+
+  private
+
+  attr_reader :menu, :order
```
saidbyced: Great use of `private` to ensure minimal data availability outside of class

PR #1596, opened 2020-02-03
Takeaway DL

Dlibmanw:


spec/menu_spec.rb
```diff
@@ -0,0 +1,7 @@
+require './lib/menu.rb'
+describe Menu do
+  let(:menu) { double :menu }
```
jasylwong: 1. Redundant double; doesn't get used?
2. Also, no point in creating a double of the described class in its own test file.

lib/menu.rb
```diff
@@ -0,0 +1,11 @@
+class Menu
+  The_menu = {
+    "Pad_thai" => 8,
+    "Red_curry" => 9,
+    "Green_curry" => 10
+    }
+
+  def show_menu
+    The_menu
+  end
```
jasylwong: If  the only thing show_menu does is return The_menu, The_menu should be accessible by referring to the constant without having to write/call a method.

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    it 'presents the amount to be paid for each dish' do
+      expect{ subject.add_items("Pad_thai", 3, 24) }.to change { subject.my_bill }.to [24]
+    end
+  end
+  describe '#total_amount' do
+    it 'shows the total amount for all dishes in the order' do
+      order = Order.new
+      order.add_items("Pad_thai", 2, 16)
```
jasylwong: You can use 'subject' instead of Order.new. Saves you having to define a new variable for each test.

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
+require './lib/order.rb'
+describe Order do
+  describe '#add_items' do
+    it 'allows to enter the dish name, quantity and total amount' do
+      expect{ subject.add_items("Pad_thai", 2, 16) }.to change { subject.my_order }.to ["Pad_thai * 2"]
```
jasylwong: Good use of the 'change' matcher! :)

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
```
jasylwong: Include another test to check that adding other subsequent dishes, after the pad thais above were added, does not delete the pad thais (ie it doesn't reset the my_order each time).

lib/order.rb
```diff
@@ -0,0 +1,22 @@
+class Order
+  attr_reader :my_order, :my_bill, :orde_details, :self_calculated_total, :total_amount
+  def initialize
+    @my_order = []
+    @my_bill = []
+    @self_calculated_total = 0
+  end
+
+  def add_items(dish, quantity, total)
```
jasylwong: Good use of including the quantity as a second argument.

Total price though wouldn't be something the user would normally be asked to input when making an order for an individual dish. I think the instructions are telling us to ask the user to input the total at the very end ie if he pays the right amount, that doubles as a check that what they've input is the same as what the actual total cost is.

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+    @my_order.push("#{dish} * #{quantity}")
+    @my_bill.push(Menu::The_menu[dish] * quantity)
+    @self_calculated_total += total
+  end
+
+  def total_amount
+    @total_amount = @my_bill.inject(0, :+)
+  end
```
jasylwong: Nice!! Great use of inject to keep your code succinct.

PR #1597, opened 2020-02-03
Takeaway_challenge

blu3skies:


lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+
+class Menu
+  attr_reader :menu_list
+
+  def initialize
```
zsoobratty: Might be worth having the menu as a constant? Unsure if it is more efficient in any way but it could be an alternative to make the initialize cleaner?

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+    @total = 0
+  end 
+
+  def view_menu
+    menu.view
+  end 
+
+  def add_item(dish, quantity)
+    order_form[dish] = quantity
+    quantity.times do
```
zsoobratty: Nice way of adding things to the @items instance array using .times!

lib/order.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/menu.rb'
+class Order 
+
+    attr_reader :order_form, :menu, :balance
```
zsoobratty: The :balance doesn't seem to be linked to anything - maybe you meant total!

PR #1599, opened 2020-02-03
takeaway-challenge

Pazoia:
Created menu and displaying full list of dishes


lib/menu.rb
```diff
@@ -0,0 +1,26 @@
+class Menu
+
```
BarriF13: Codes are really clear, and they are working. I really like the diagram that you made. I would like to see more comments on the code. I would like to see more refactoring later on.

PR #1600, opened 2020-02-03
Code review pull request - Takeaway weekend 2 challenge

saidbyced:


spec/dish_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'dish'
+
```
Ovy95: I like your tests they're very readable

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+      Dish.new("Cookie Dough", 5.0),
+      Dish.new("Neapolitan", 4.7),
+      Dish.new("Rocky Road", 9.0),
+      Dish.new("Vanilla Bean", 6.0),
+      Dish.new("Coffee", 3.2),
+      Dish.new("Chocolate Brownie", 5.0)
+    ]
+  end
+  
+  def list
```
Ovy95: Nice use of the .each method to print out the items with their related values

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'dish'
+
+class Menu
+  
+  def initialize
+    @menu = [
+      Dish.new("Cookies & Cream", 5.0),
+      Dish.new("Chocolate", 4.3),
```
Ovy95: What about maybe having the hash inside spec class and calling from it inside the tests. This will make it easier to update for future prices and additions to the menu.

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+      Dish.new("Rocky Road", 9.0),
+      Dish.new("Vanilla Bean", 6.0),
+      Dish.new("Coffee", 3.2),
+      Dish.new("Chocolate Brownie", 5.0)
+    ]
+  end
+  
+  def list
+    count = 1
+    @menu.each do |item|
```
Ovy95: Maybe change item to Dish, as it will be a bit easier to read, then use a .index method rather than using the count method

saidbyced: Good shout! Minimises code. Wish I'd thought of that

PR #1603, opened 2020-02-03
Week 2 challenge

EWright212:


lib/feature.rb
```diff
@@ -0,0 +1,16 @@
+require './lib/place_order'
+require './lib/menu'
+my_menu = Menu.new
+
```
DavidStewartLDN: Really nicely defined methods to make clear their function. Really great! Helps make your code cleaner by not using comments.

lib/menu.rb
```diff
@@ -0,0 +1,19 @@
+class Menu
+
+  attr_reader :nested_menu
+
+  def initialize
+    @nested_menu = [
+      ["chocolate", 0.50],
+      ["rasperries", 0.25],
+      ["ice_cream", 3.72]
+    ]
```
DavidStewartLDN: Well named again, and thank you for using multi line to define menu to increase clarity.

lib/menu.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def print_menu
+    print "Menu is as follows:\n"
+    @nested_menu.each_with_index do |item, index|
+      print "#{index+1}. #{item[0]}, £#{item[1]}\n"
+    end
```
DavidStewartLDN: Your code is very readable and I thank you for that. I have some lessons to learn from you!

lib/place_order.rb
```diff
@@ -0,0 +1,69 @@
...
+
+  def take_order
+      @order = @order_start
+  end
+
+  def ask_for_total
+    p "Please write your expected order total below e.g 12.36"
+  end
+
+  def take_user_total
```
DavidStewartLDN: Maybe total could be moved out into a separate class for calculating. Separating responsibility and all that jazz (hands)

lib/place_order.rb
```diff
@@ -0,0 +1,69 @@
...
+    @users_total =  gets.chomp.to_f
+  end
+
+  def calculate_total
```
DavidStewartLDN: Sam as above, could remove to make it so place order class only places the order. I am sure with more time you might have chosen to move some functionality out

lib/place_order.rb
```diff
@@ -0,0 +1,69 @@
...
+    # take first item, assume quantity 1
+    p @order
+    menu_item = Menu.new.nested_menu
+    if @order.length == 2
+      first_index = @order[0].to_i - 1
+      second_index = @order[1].to_f
+      @total = (menu_item[first_index][1] * second_index)
+    else
+      first_index = @order[0][0].to_i - 1
```
DavidStewartLDN: Maybe you could use an iterative process here though it looks like this could be the simplest possible code for a menu of three items so maybe if the menu was larger this would make sense

spec/menu_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'menu'
+
+describe Menu do
+
```
DavidStewartLDN: Maybe could mock the menu so that you can be clear that the menu works with other menu items

spec/place_order_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'place_order'
+
+describe PlaceOrder do
+
+  describe "#takes_dishes" do
+    it { is_expected.to respond_to(:ask_for_order) }
+    it 'asks for an order' do
+      expect(subject.ask_for_order).to eq "Please place your order in the form of order number, quantity, and start on a new line for each item E.g. 1, 7[return key]2, 4,[return key][return key]"
+    end
```
DavidStewartLDN: Maybe at return after end to give the eye some white space 🥶

spec/place_order_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      #TODO - test with stub total calc
+    it 'calculates simple total' do
+      subject.order = ["1", "1"]
+      subject.calculate_total
+      expect(subject.total).to eq 0.5
+    end
+    it 'calculates simple total' do
+      subject.order = ["2", "1"]
+      subject.calculate_total
+      expect(subject.total).to eq 0.25
```
DavidStewartLDN: You have tested all options together, which maximises confidence but may take extra time. Maybe two tests would fill you with the confidence you need 👩🏼‍🌾

spec/place_order_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      subject.calculate_total
+      expect(subject.total).to eq 1.0
+    end
+    it 'calculates total with two items' do
+      subject.order = ["3, 1","2, 4"]
+      subject.calculate_total
+      expect(subject.total).to eq 7.44
+    end
+  end
+
```
DavidStewartLDN: Get some more of those stubs and mocks going 💆🏼

PR #1604, opened 2020-02-03
Take away attempt

zsoobratty:


lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+class Menu
+    attr_reader :menu
+    MENU = { "pizza" => 10, "pasta" => 8, "chips" => 3, "burger" => 7, "milkshake" => 5 }
```
blu3skies: I like the use of a constant for the Menu

lib/basket.rb
```diff
@@ -0,0 +1,27 @@
...
+
+class Basket
+    attr_reader :basket, :menu, :total
+    def initialize(menu = Menu.new)
+        @menu = menu
+        @basket = {}
+    end
+
+    def add(item, quantity)
+        @menu.menu.select { |dish, price| @basket[item] = quantity if dish == item }
```
blu3skies: great use of the select method. I needed something like this.

lib/basket.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    def show_basket
+        basket = []
```
blu3skies: suggest using different name for array to avoid confusion with hash

lib/basket.rb
```diff
@@ -0,0 +1,27 @@
...
+        @basket.each { |item, quantity| basket << "#{item} x #{quantity}" }
+        basket.join(", ")
+    end
+
+    def checkout
+        checkout = []
```
blu3skies: similar array variable in the show_basket method. Could use same instant variable to get total.

spec/basket_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        expect(basket.basket).to eq({})
+    end
+
+    it 'starts with a balance of 0' do
+        expect(basket.total).to eq(nil)
+    end
+
+    describe '#add' do
+        it 'can add an item to the basket' do
+            expect { basket.add("pizza", 1) }.to change { basket.basket }.from({}).to({"pizza" => 1})
```
blu3skies: I like the use of 'change from.. to ..' I haven't really seen this much.

spec/menu_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'menu'
+
+describe Menu do
+    let(:menu) { Menu.new }
+    
+    it 'shows the list of items in the menu' do
+        expect(menu.menu).to eq(described_class::MENU)
```
blu3skies: menu.menu is easy to understand but not the best coding practice i think.

PR #1605, opened 2020-02-03
Takeaway challenge

davmcgregor:


lib/order.rb
```diff
@@ -0,0 +1,79 @@
+require_relative "takeaway"
+require_relative "message"
+
+class Order
+  attr_reader :basket, :menu, :total
+
+  def initialize
```
hansa-lallu: possible opportunity to do a dependency injection here

lib/order.rb
```diff
@@ -0,0 +1,79 @@
...
+          break
+        else 
+          puts "How many would you like to order? "
+          quantity = gets.chomp.to_i
+          add_item(selection, quantity)
+        end  
+    end
+  end
+
+  def add_item(item, quantity)
```
hansa-lallu: good work, this took me so long to figure out. You could set your quantity to a default of 1 so if user doesn't enter how many you have a default value. 

 def add_item(item, quantity = 1)

lib/order.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+    end
+  end
+
+  def order_summary
+    @basket.each do |dish, quantity|
+      puts "#{quantity} x #{dish}" 
+    end
+    puts "Coming to a total of: £#{@total}"
+  end
```
hansa-lallu: Nice work on this method, very easy to read

lib/order.rb
```diff
@@ -0,0 +1,79 @@
...
+    if confirmation == "Y"
+      send_confirmation
+    end
+  end
+
+  def send_confirmation
+    time = Time.new.+(1800).strftime("%I:%M %p")
+    message = Message.new
+    message.send("Thank you! Your order was placed and will be delivered before #{time}")
+  end
```
hansa-lallu: Love this, great job.

lib/order.rb
```diff
@@ -0,0 +1,79 @@
...
+      end
+    end
+    
+    if dish_sum == @total
+      return "Your order total matches the sum of dishes"
+    else
+      return "Your order total does not match the sum of dishes"
+    end
+  end
+end
```
hansa-lallu: You could split this method into two. 
Possibly so you have a confirmation method in a private section and call it within your verify method

PR #1606, opened 2020-02-03
Takeaway challenge

tobydawson1:
Incomplete

### DIFF
--- a/lib/restaurant.rb
+++ b/lib/restaurant.rb
@@ -1,0 +1,15 @@
+class Restaurant 
+
+    attr_reader :menu
+
+    def initialize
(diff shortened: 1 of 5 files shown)


lib/restaurant.rb
```diff
@@ -0,0 +1,15 @@
+class Restaurant 
+
+    attr_reader :menu
+
+    def initialize
+        @menu = [{10 => "Pad Thai"}, {11 => "Massaman Curry"}, {5 => "Crispy Squid"}, {3 => 'Jasmine Rice'} ]
```
jlblumberg: You could consider changing the menu into one big hash rather than an array of hashes - think it will make it easier to work with in future steps. And maybe also switching the order of key and value, so that if someone enters a dish, it returns the price, rather than the other way around.

PR #1607, opened 2020-03-01
Takeaway challenge

dariathompson:


poliver24:
**Appreciations**

`def add(dish, number = DEFAULT_NUMBER)
    fail "This item is not on the menu" unless exists?(dish)`

- I like that you have a default value for the quantity.

- Nicely laid out code
- Easy to follow
- Tests are readable.
- Your logic is sound. 

**Suggestions**



` def send_message
    account_sid = "xxxxxxx" 
    auth_token = "xxxxxxx"`

- look into Environment variables for more efficient ways to protect security

`it "stores multiple dishes that have been order" do
      subject.add("Vodka", 1)
      subject.add("Chicken Kiev", 1)
      expect(subject.order_list).to eq([{ "Vodka" => 5 }, { "Chicken Kiev" => 22 }])
    end`

- look into using doubles so that your tests do not depend on multiple methods to run. 

- Look into how you could test your SMS application.

`body: "Your order is on the way and will be delivered at #{Time.now + 3600}",`

- Make time pretty :(

naeglinghaff:
Love the feature tests on the README - really helpful to understand the responsibility of different classes. 

The default number is a great idea to make your code simpler.

All your methods are short and sophisticated, easy to read.

I like your tests for order - you've separated them out really well and covered a lot of bases. Great work!

Suggestions:

- Environment variables - dotenv is a really cool tool that can help protect your confidential information.
- Consider using mocking, stubs and doubles to make sure your other object dependencies don't interfere with your tests. Eg. double menu when you are testing order
- Consider making some of your order methods private so that they can't be accessed by other objects such as the total, checking the order exists and sending the message.

PR #1608, opened 2020-03-01
Take-away challenge with send_message functionality

stellenberger:


Swarbzz:
You code is easy to read and I understand your thought processes as I go through it. Your comments are relevant as are your tests. Perhaps look at refactoring you code and tests a tad. Adding a method in your tests will allow you to call on it instead of repeating multiply lines. 
P.S. looking forward to holy paws.

PR #1610, opened 2020-03-01
Takeaway challenge

sgneha:


lilawalker:
Appreciation: your code is very well laid out and easy to read. Well done!
Suggestion: key and value names in hash can be named to dish and quantity, and possibly extract to a private method the looping over the hashes as it is the same and can be used in multiple places in the code.

PR #1611, opened 2020-03-01
Weekend Project!

jarasmar:
I started focusing on having a good delegating system within classes but I realize my biggest gap was mocking. Although all my code was written following TDD (tests first) I was really struggling to made them pass. I left all my code at a basic working status to cover all stories and focus to understand mocking in other projects. At the end I came back and manage to refactor all the tests in restaurant class to be fully independent!

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -9,25 +9,30 @@ class Order
     @prices_to_sum = []
   end
 
-  def add_dish
-    while true do
-      puts "Select a dish (type end to finish)"
-      dish = gets.chomp.to_s
-      break if dish == "end"
+  def add_dish(dish)
+    fail "Sorry, that dish is not available" if existent_dish?(dish) == false
 
-      if existent_dish?(dish) == false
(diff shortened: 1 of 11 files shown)


Kittaru87:
Hello! Happy Monday!

I really liked your README and the inclusion of the domain model. I think I'll be stealing that idea. Also, good number of tests and I can see that you're getting to grips with doubles and mocks. Completely understand your frustration with the environmental variables.

You can probably get rid of a few of the tests and the beginning of your spec files as they are tested later on (ones like testing that it initializes with instance variables).

lib/order.rb
```diff
@@ -0,0 +1,65 @@
...
+  def finish_order
+    sum_price
+    print_order
+  end
+
+# private (HOW TO CALL THEM FROM TESTS?)
+
+  def existent_dish?(dish)
+    Menu.new.menu_list.has_key?(dish) ? true : false
+  end
```
Kittaru87: This can be refactored

PR #1613, opened 2020-03-01
Take_away Weekend Challenge. A partial solution.

rafahg:
I have implemented a partial solution to the challenge. Everything seems to work properly.
Only the lack of time have prevented me to finish this challenge, I have struggled making the tests and i have spend many hours researching how to obtain one particular result and test it in rspec.

The only method missing the final one, #place_an_order. It is going to be quite an unsuccessful business if people cannot place an order so, i will implement the solution as soon as possible.

As long as, it would had been easy to copy the code for any other pull request with the final solution, i have decided to remain honest to myself and look for the solution when i have a little more time, and only offer a partial one, but mine, ONLY MINE!!!!. (super villain laugh here)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,7 @@
 source 'https://rubygems.org'
 
+ruby '2.6.5'
+
 group :test do
   gem 'rspec'
   gem 'rubocop', '0.72.0'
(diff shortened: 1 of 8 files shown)


john-baxter:
I like how it is simple and easy to follow. Especially the _spec files; this makes it a lot easier to follow how you built the code up.
An improvement I would suggest is that you improve the hash with the ordered food - it seems inefficient to have multiple times of one thing.

PR #1614, opened 2020-03-01
takeaway-challenge

Kittaru87:


jarasmar:
Hello Sophia :)
I really liked that you are using a lot of doubles in your tests and that your methods have single responsability. I think we got to more or less the same working structure!
As a suggestion I'd say that spec files should only be requiring the class they are testing, and that way you will be sure that your tests run independently. Also adding a domain model or a structure in your readme will be very useful to see at once what your program is meant to do before reading the actual code.

PR #1615, opened 2020-03-01
Artemis Papanikolaou - Takeaway (partially complete)

artemis-p:


Benjamaker:
I really like your menu and the fact you have kept things simple with your order class. Your code is easy to read.

PR #1618, opened 2020-03-02
TakeAway Challenge

ghost:


stellenberger:
Your code is very efficient and has no redundant code, as far as I can see! The tests look great, and your use of doubles in the test to avoid hard coding is a clever technique that I should have used!! Maybe something to change would be letting the user know their order in a pretty format, that is put'es out onto the screen.

PR #1619, opened 2020-03-02
Progress so far

probablymenext:
SMS not working and could probably improve tests, but this is as far as I got.


spec/menu_spec.rb
```diff
@@ -0,0 +1,68 @@
+require 'menu'
+
+describe Menu do
+
+    menu = Menu.new
+    item1 = Item.new(1, "Eggs and Turnips", 10)
```
WilliamJGrace: When testing a class, you should try to avoid relying on another class. Doubles would work here but it looks like you'd have to make a few..., might be a better way?

spec/menu_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    item2 = Item.new(2, "Cheese, Carrot and Marzipan", 12)
+    item3 = Item.new(3, "Pear, Pork and Parmisan", 20)
+    item4 = Item.new(4, "Cabbage on Toast", 90)
+    menu.instance_variable_set(:@menu_items, [item1, item2, item3, item4])
```
WilliamJGrace: 'nice use of rspec 'instance_variable_set

spec/menu_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+
+    it 'initialises without error' do
+        expect{subject}.not_to raise_error
+    end
+
+    describe 'build_menu' do
```
WilliamJGrace: After describe you could have let(:subject) { subject.build_menu)
Then you don't have to repeat yourself

spec/menu_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    describe 'add_to_basket' do
+        it 'responds to 2 arguments' do
+            expect(subject).to respond_to(:add_to_basket).with(2).arguments
+        end
+        it 'can add item on menu to basket' do
+            subject.add_to_basket(1, 2)
+            expect(subject.instance_variable_get(:@basket)).to include {{item: instance_variable_get(:@menu_items)[0], quantity: 2}}
+            p subject.instance_variable_get(:@basket)
```
WilliamJGrace: You probably don't need the P anymore?

PR #1621, opened 2020-03-02
Takeaway challenge

john-baxter:


rafahg:
- I think the logic is very clear and i particularly like the use of hashes. 
- The way you use code methods to organise the printing is inspiring to me, i am gonna use it my self, for sure.
- The only thing i could suggest you to improve(which i am not entitled at all to do, due my own state of ignorance) is to keep things a little simpler.

PR #1623, opened 2020-03-02
Pull request

WilliamJGrace:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+  # before(:each) do
+  it "creates an order" do
+    takeaway.create_order(order_double)
+    expect(takeaway.current_order).to eq order_double
+  end
+  it "user can add to order" do
+    allow(order_double).to receive(:order) { [] }
+    takeaway.create_order(order_double)
+    takeaway.add_to_order("pizza", 2)
```
probablymenext: One more line to check the value of @current_order would be useful here.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'takeaway'
+
+describe Takeaway do
+  let(:takeaway) { Takeaway.new }
+  let(:menu) { double :menu, see_menu: printed_menu }
+  let(:printed_menu) { "menu" }
```
probablymenext: I like the simplicity of using a small string here instead of a whole dummy menu.

PR #1625, opened 2020-03-02
Takeaway challenge

lukedonov:


Apliz:
What went well IMO
- The extra implementation was great, being able to check for edge cases as standard
- use of recursion is a neat idea; it nicely breaks the code apart and makes it much more readable than it would have been otherwise. (nice private class too)
--
- Testing of edge cases was thorough

things that could improve IMO
- repetition in the Order class, same 'fail' code in total() and checkout(), creating a private class or Lambda that holds the block for you will increase readability. The 'fail' statement in total() might not be necessary for function. 
--
moving towards dependency injection in the Rspec tests would separate spec from lib (mocks, let, double etc)
#adds and item twice {it block} may be redundant because of surrounding tests.

PR #1626, opened 2020-03-02
Code Review : Takeaway : Luke

Apliz:
…t wonky for now;

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,17 @@
+class Menu 
+
+  attr_reader :menu
+
+  def initialize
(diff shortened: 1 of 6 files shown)


lukedonov:
I like how you made a front end class which referenced your other classes. It made the code look straightforward and clean, and easy to understand the functionality of your program. 
I like how concise and easy to read your tests are in their context groups. I like your use of doubles so as not to hardcode in your order spec.

In your order tests you could utilise mocks and stubs to allow you to run rspec without having to input anything each time you run rspec.

PR #1627, opened 2020-03-02
Takeaway challenge code review

OrionReed:


lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+
+
+  attr_reader :menu
+
+  def initialize
+    @menu = [
+      { :name => "Pizza", :price => 15 },
+      { :name => "Burger", :price => 12 },
+      { :name => "Salad", :price => 10 }
+    ]
```
OrionReed: Nice way of doing the menu

lib/order.rb
```diff
@@ -0,0 +1,18 @@
...
+    def initialize
+      @total = 0
+      @dishes = Menu.new
+      @order = []
+    end
+
+    def dish(dish, quantity)
+      @order << dish
+
+    end
```
OrionReed: dish could use quantity by using `quantity.times do ...`

spec/menu_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "menu"
+
+describe Menu do
+
+  describe "#print_menu" do
+    it "gets the list of dishes' names and prices" do
+      expect(subject.print_menu).to eq [{ :name => "Pizza", :price => 15 }, { :name => "Burger", :price => 12 }, { :name => "Salad", :price => 10 }]
```
OrionReed: menu list passed to `eq` could be setup as a double to be easier to read

PR #1633, opened 2020-03-28
Challenge completed apart from the bonus step

PhilipVigus:
All done, and I think I'm going to leave it there.


spec/list_of_dishes_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'list_of_dishes'
+
+describe ListOfDishes do
+  let(:dishes) { [
+    {name: "Stew", price: "£2.99"},
```
Angelica137: I like that you used named parameters

lib/dish.rb
```diff
@@ -0,0 +1,8 @@
+class Dish
```
Angelica137: I really like you have a class for Dish

spec/order_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  context 'functionality' do
+    it 'shows an order summary' do
+      order = Order.new(list_of_dishes, selections, 5)
+      expect(order.summary).to eq({ stew: { quantity: 2, price: 2 }, rice: { quantity: 1, price: 1 } })
+    end
+
+    it 'sends a text with the expected delivery time when an order has been verified' do
+      text_handler = double(:text_handler, send_confirmation_text: 'Text sent')
+      order = Order.new(list_of_dishes, selections, 5)
+      expect(order.send_text(text_handler)).to eq('Text sent')
```
jamesjoshuahill: I notice that the unit test for `send_confirmation_text()` does not expect it to return anything.

Maybe `send_confirmation_text()` does not need to return anything when it succeeds? We could just expect the method to be called?

spec/text_handler_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'text_handler'
+
+describe TextHandler do
+  it 'sends a confirmation text to a specified phone number' do
+    text_client = double(:text_client)
+    allow(text_client).to receive_message_chain(:new, :messages, :create)
```
jamesjoshuahill: Could this unit test be stronger by expecting `create()` to be called with specific arguments?

README.md
```diff
@@ -48,37 +38,203 @@ So that I am reassured that my order will be delivered on time
...
+![class diagram after implementing user story 1](story1_domain_model.png)
+
+My initial approach to this story was a ListOfDishes class containing a simple array with the names and prices of each dish, but I quickly decided to refactor this into the two classes you see here. I decided to use a hash as it was easier and more intuitive to access individual dishes by name rather than their index.
+
+I was (and still am) torn on the name ListOfDishes. The obvious classname here is Menu, and I was itching to use it all through the challenge. However, I decided to stick to ListOfDishes as that is the language used in the user story.
+
+Here is an example of using the feature in irb:
+
+```
+2.6.5 :001 > require './lib/list_of_dishes'
```
jamesjoshuahill: I like that you have included manual acceptance tests for each story 👍

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+      full_dish = list_of_dishes.get_dish(dish)
+      @summary[dish] = { quantity: quantity, price: full_dish.price }
+    end
+  end
+
+  def verify_total(total)
+    total_from_summary = 0
+    @summary.each do |_, quantity_and_price|
+      total_from_summary += quantity_and_price[:quantity] * quantity_and_price[:price]
+    end
```
jamesjoshuahill: I wonder if this could be refactored using `inject`/`reduce`? 🤔

lib/text_handler.rb
```diff
@@ -0,0 +1,35 @@
...
+  def send_confirmation_text(order_number, text_client = Twilio::REST::Client)
+    client = text_client.new(@account_sid, @auth_token)
+    begin
+      client.messages.create(
+        from: @takeaway_number,
+        to: order_number,
+        body: create_message_body
+      )
+    rescue RuntimeError => e
+      raise RuntimeError, "unable to send text message. Error from text client - #{e.message}"
```
jamesjoshuahill: I like that you caught the error and decorated it with useful context. In this case we're catching and throwing it again, which adds a runtime overhead for these errors. Might the stack trace be enough to understand this error without catching it?

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    text_handler.send_confirmation_text(@phone_number)
+  end
+
+  private
+
+  def populate_summary(list_of_dishes, selections)
+    @summary = {}
+
+    selections.each do |dish, quantity|
+      raise ArgumentError, 'ordered dish not on menu' if list_of_dishes.get_dish(dish).nil?
```
jamesjoshuahill: I wonder if we could avoid a `nil` check here. Could `Order` ask `ListOfDishes` if it has a dish?

lib/text_handler.rb
```diff
@@ -0,0 +1,35 @@
+require 'dotenv'
+Dotenv.load('./twilio_credentials.env')
+require 'twilio-ruby'
+
+class TextHandler
+  def initialize
+    read_env_variables
```
jamesjoshuahill: Since there are only three I would prefer to see the instance variables in `initialize`, where I expect to find them.

PR #1637, opened 2020-03-30
Angelica

Angelica137:


lib/list.rb
```diff
@@ -1,5 +1,16 @@
 class List
+  attr_reader :menu
+  def initialize
+    @menu = { 'apple' => 3 }
+  end
   def see
-    menu = { 'apple' => 3 }
+    @menu
```
Angelica137: I forget what this is doing here, but I do not think it should be

lib/list.rb
```diff
@@ -0,0 +1,28 @@
...
+  
+  def initialize
+    @menu = {
+      'apple' => 3,
+      'ham' => 4
+     }
+     @order =[]
+  end
+  
+  def method_missing(m, *args, &blk)
```
alexakearns: A little confused about these argument names and what they are actually representing.

spec/order_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    menu = { 'apple' => 3 }
+    choice = list.apple
+    expect(subject.see_order).to eq(choice)
+  end
+
+  xit 'shows the total of the order' do
+    list = List.new
+    menu = { 'apple' => 3 }
+    choice = list.apple
+    expect(subject.total).to eq(choice.values.sum)
```
alexakearns: I like this! 😄

lib/order.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  def initialize
+    @order = []
+  end
+
+  def order_maker
+  end
+  
+  def total
+    order_total = @order.values.sum
```
alexakearns: I like how simple but effective this is.

lib/list.rb
```diff
@@ -0,0 +1,28 @@
...
+    if args.empty? && blk.nil? && @menu.has_key?(m.to_s)
+      choice = Hash.new
+      choice[m] = @menu[m.to_s]
+      @order << choice
+      puts @order
+#    elsif @menu.has_key?(m.to_s)
+#      [args[0], @menu]
+    else
```
alexakearns: Making sure it doesn't create a new item if not on the menu, would be nice to have a test for this (unless i have missed it)

PR #1638, opened 2020-03-30
Takeway-challenge progress

FayeCarter:
Recap of week 2 learnings and application to new project

Included in the change:

I have completed the first 3 user stories of the challenge

What is incomplete?

total method is currently very basic.
I aim to make it return a bill of items with a total cost at the end

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -12,4 +12,10 @@ class Order
     @order << [item, quantity]
   end
 
+  def total
+    total_items = 0
+    @order.each { |item| total_items += item[1] }
+    "Total items: #{total_items}"
+  end
+
 end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,15 @@
+class Order
+
+  def initialize 
+    @order = []
+  end
+
+  def show_order 
+    @order
+  end
+
+  def add_item(item, quantity = 1)
+    @order << [item, quantity]
+  end
+
+end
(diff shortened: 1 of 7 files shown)


lib/order.rb
```diff
@@ -0,0 +1,15 @@
+class Order
+
+  def initialize 
+    @order = []
+  end
+
+  def show_order 
+    @order
+  end
+
```
grahamfalconer: Does this create duplicates of the item name string?

lib/order.rb
```diff
@@ -12,4 +12,10 @@ def add_item(item, quantity = 1)
     @order << [item, quantity]
   end
 
+  def total
+    total_items = 0
+    @order.each { |item| total_items += item[1] }
+    "Total items: #{total_items}"
```
grahamfalconer: Nice Use of string interpolation!

spec/order_spec.rb
```diff
@@ -14,11 +14,22 @@
...
+  describe '#total' do
+    it 'returns total number of items' do
+      subject.add_item("margherita", 5)
+      expect(subject.total).to eq "Total items: 5"
+    end
 
+    it 'returns total number of items of multiple pizzas' do
+      subject.add_item("margherita", 5)
+      subject.add_item("vegetarian", 3)
+      expect(subject.total).to eq "Total items: 8"
```
grahamfalconer: Good use of mock (I think)

grahamfalconer: Stub*

PR #1639, opened 2020-03-30
Takeaway Challenge - Alexa

alexakearns:
Takeaway challenge:
Completed user story 1
User story 2 tests incomplete due to challenges using doubles in tests.


lib/menu.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'menu_item'
```
PhilipVigus: Beautifully laid out class, well indented with good naming.

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+class Menu
+
+  attr_reader :list
+
+  def initialize
+    @list = Hash.new
+  end
+
+  def add_new_item(name, price)
+    @list[name] = MenuItem.new(name, price)
```
PhilipVigus: It might be worth thinking about what happens if you try and add an item for a name that already exists.

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def print_menu
+    @list.map { |_key, value| 
+      "#{value.name} - £#{value.price}" 
+    }
+  end
+
+  def return_item(key)
+    @list.values_at(key)
```
PhilipVigus: This can be replaced with @list[key]

spec/menu_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'menu'
```
PhilipVigus: Well laid out again. I like the use of contexts to split the tests up.

spec/menu_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  context 'prints menu' do
+
+    it 'prints the name attribute in the value of the hash' do
+      subject.add_new_item("pepperoni", 9)
+      expect(subject.print_menu).to eq ["pepperoni - £9"]
+    end
+  end
+
+  # context 'return selected item' do
```
PhilipVigus: It's worth considering whether this code is still needed. Commented out code is usually better deleted if it can be.

PhilipVigus:
Generally really good. I found your code easy to read and follow, it was well laid out and the choice of names for methods, classes and variables really helped with comprehension.

I would be careful with leaving commented out code. With tests you're better of disabling them instead, but changing the 'it' into 'xit'. This means the test won't run, and you won't have to comment it out.

PR #1643, opened 2020-03-30
Take-away challenge - not fully complete

tommyrharper:


lib/order.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def confirm
+    time = Time.new + DELIVERY_TIME
+    time = time.strftime("%R")
+    @notification.send(time)
+    return "Thank you! Your order was placed and will be delivered before #{time}."
+  end
+
+  private
```
Steven-Klavins: Good use of encapsulation :)

README.md
```diff
@@ -14,71 +16,252 @@ Takeaway Challenge
...
+  - Verbalise my thought process when solving a problem.
+
+## User Story 1
 
 ```
 As a customer
+```
+From the users perspective in IRB:
```
tatsiana-makers: It's really nice that you showed how a user would interact with your program via IRB!

README.md
```diff
@@ -14,71 +16,252 @@ Takeaway Challenge
...
+```
 As a customer
 So that I can order the meal I want
 I would like to be able to select some number of several available dishes
+```
+From the users perspective in IRB:
+```
+order = Order.new
+=> #<Order:0x00007f863783fda0 @menu={:fish=>5, :chips=>3, :salad=>2}> 
+order.update("1 fish, 3 salad, 2 chips")
```
tatsiana-makers: This is a really interesting way of adding things to the menu. As an engineer, before even reading any code, I expect that parsing this order requires a bit of precise and complicated code. I wonder if we could have a customer add items to the order in a way that requires less complex parsing. Some ideas I have are something like:
- `order.update(["fish", "salad", "salad", "salad"])`
- `order.update({"fish": 1, "salad": 3})`

What do you think of these alternatives?

tatsiana-makers: The method name `update` is a bit confusing. Without reading the code, I'm not intuitively sure what would happen if I did
`order.update("1 fish, 3 salad, 2 chips")`
followed by
`order.update("1 fish, 1 salad, 1 chips")`.
Would I end up with 2 fish, 4 salad, and 3 chips or 1 fish, 1 salad, 1 chips. I wonder if calling the method name `add` would make this clearer here?

README.md
```diff
@@ -14,71 +16,252 @@ Takeaway Challenge
...
-  * However, if your Takeaway is loaded into IRB and the order is placed, the text should actually be sent
-  * Note that you can only send texts in the same country as you have your account. I.e. if you have a UK account you can only send to UK numbers.
+## My own Story
 
-* Advanced! (have a go if you're feeling adventurous):
-  * Implement the ability to place orders via text message.
+```
+As a customer
+So that I can change my order if it is wrong
+I would to be able alter my order before I confirm
```
tatsiana-makers: This is really cool!

So this answers my question above - I wonder if there's a clearer name than `update`? to demonstrate this behaviour - I can't think of a good one, but something like `replace`?

 I also wonder if a user might want to be able to just add one order of chips to an existing order without re-adding all of the other items again?

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
+class Menu
+
+  attr_reader :list
+
+  def initialize
```
tatsiana-makers: What if we wanted to provide our own menu - for example we want this code to serve both chinese and pizza restaurants? How could we do this?

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    order = Order.new(menu_double)
+    expect { order.update("1 cat") }.to raise_error 'cat is not on the menu'
+  end
+
+  it 'throws an error if the order number is negative' do
+    order = Order.new(menu_double)
+    expect { order.update("-3 fish") }.to raise_error 'cannot order a negative number'
+  end
+
+  it 'allows the user to change order' do
```
tatsiana-makers: This also implicitly tests that happens if the order number is 0, what do you think about adding this to the description of the test case?

README.md
```diff
@@ -14,71 +16,252 @@ Takeaway Challenge
...
+=> 10
+```
 
-* A free account on Twilio will only allow you to send texts to "verified" numbers. Use your mobile phone number, don't worry about the customer's mobile phone.
+## Domain model
 
-> :warning: **WARNING:** think twice before you push your **mobile number** or **Twilio API Key** to a public space like GitHub :eyes:
->
-> :key: Now is a great time to think about security and how you can keep your private information secret. You might want to explore environment variables.
+Objects | Messages
```
tatsiana-makers: Awesome addition to the README

README.md
```diff
@@ -14,71 +16,252 @@ Takeaway Challenge
...
+- You will need your own [Twilio account](https://www.twilio.com) to set up this up.
+- You will have to create the folowing file:
+```
+lib/send_sms.rb
+```
+- Add these contents to the file
+```
+require 'twilio-ruby'
+
+class Notification
```
tatsiana-makers: It might be a nice exercise to try and write tests for this file - how could we use mocking here to make sure our tests don't actually send text messages every time they're run?

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
...
+    @list = {
```
tatsiana-makers: Each of the items in this list have similar attributes: a `name` and a `cost`. I wonder if we could extract these similarities out into a common class?

PR #1649, opened 2020-04-26
Takeaway Challenge - Marius Brad

mbrad26:


lib/restaurant.rb
```diff
@@ -0,0 +1,40 @@
+class Restaurant
```
will-head, line 1: Like that you have a Restaurant class to hold everything - makes it much easier to understand!

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
+class Menu
+
+  attr_reader :menu_list
+
+  def initialize
+    @menu_list = { 'Fish & Chips' => 4, 'Burger' => 3, 'Shawarma' => 5 }
```
will-head, line 6: Maybe start with an empty list, so the class can be used for any restaurant, not just a Fish and Chip restaurant?

PR #1652, opened 2020-04-26
Takeaway Challenge

Joanneyoung01:


README.md
```diff
@@ -1,3 +1,33 @@
...
+
+As a customer
+So that I can verify that my order is correct
+I would like to check that the total I have been given matches the sum of the various dishes in my order
+
+Thanks
+Jo
+
+
+
```
cmb84scd, line 30: Good, clear readme :)

lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
...
+  def print_menu
+    menu.print
+  end
+
+  def print_basket
+    basket.print
+  end
+
+  def add_to_basket(dish, quantity)
+    quantity.times{ @order_arr << dish}
```
cmb84scd, line 23: I believe there is a missing space in the curly brackets here (after dish) but otherwise it's fine. Good method names too :)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  let(:basket) { double(:basket, print: basket_example) }
+  let(:basket_example) { "Rice 100, Soup 20, Chefs Special 10" }
+  it "shows the basket" do
+    expect(takeaway.print_basket).to eq ("Rice 100, Soup 20, Chefs Special 10")
+  end
+
+  let(:dish) { double :dish }
+  let(:quantity) { double :quantity }
+
+  it "responds to 2 arguements" do
```
cmb84scd, line 28: Not sure about this test as I believe it's testing state rather than behaviour. Would suggest changing if possible.

PR #1655, opened 2020-04-26
Takeaway Challenge - Jed Stanton 

JStant95:


spec/menu_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'menu'
+
+describe Menu do
+  describe '#view_menu' do
+    it 'Allows user to see menu items' do
+      expect{ subject.view_menu }.to output(String).to_stdout
```
dpwdec, line 6: Probably would have been more robust to test for the expected literal output.

```
menu_output = 'Egg Fried Rice - £1.50' \
"Sweet and Sour Chicken - £6.50' \
# etc.
expect{ subject.view_menu }.to output(menu_output).to_stdout
```

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+    puts current_total
+  end
+
+  private
+
+  def calc_total
+    total = 0
+    @order.each do |item|
+      Menu::MENU.each do |menu_item|
```
dpwdec, line 31: Good solution to use `Menu` as a namespace for storage and directly reference it!

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'order'
+require 'menu'
+require 'bill'
+
+describe Order do
+  describe "#add_to_order" do
+    it "Should add items to a user's order" do
+      expect { subject.add_to_order("Egg Fried Rice", 2) }.to change { subject.order.length }.by(1)
```
dpwdec, line 8: Maybe consider testing output directly again, as you're not instantiating an object just creating an order hash, it would useful to also confirm that you `"Egg Fried Rice", 2` really was added and is a hash as this test could still pass by just adding an empty object to `subject.order`.

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    it "should only add an item if the item is present on the menu" do
+      expect { subject.add_to_order("Chips", 1) }.to raise_error("This item is not on the menu")
+    end
```
dpwdec, line 13: Good error checking.

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  describe "Order info" do
```
dpwdec, line 16: Not really clear what this `describe` block is for.

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    let(:menu) { Menu.new }
+
+    before(:each) do
+      menu.order
+      menu.my_order.add_to_order("Egg Fried Rice", 2)
+      menu.my_order.add_to_order("Sweet and Sour Chicken", 2)
+    end
```
dpwdec, line 23: good use of `before` to keep your code DRY! 🥵

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    describe "#current_total" do
+      it "calculates total price of current order" do
+        expect(menu.my_order.current_total).to eq(16.0)
+      end
+    end
+
+    describe "#view order" do
+      it "Prints out order" do
+        expect{ menu.my_order.view_order }.to output(String).to_stdout
+      end
```
dpwdec, line 34: Again, would probably recommended testing direct output as at the moment this test would pass by just submitting an empty string to this. All the set up with `before` becomes redundant for this test if you don't test that its printing the correct thing.

spec/bill_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+describe Bill do
+  let(:menu) { Menu.new }
+
+  describe "#pay" do
+    it "Should let the user know what they've paid" do
+      menu.order
+      menu.my_order.add_to_order("Egg Fried Rice", 2)
+      bill = Bill.new(menu.my_order)
+      expect{ bill.pay }.to output("You've paid 3.0\n").to_stdout
+    end
```
dpwdec, line 14: This could probably benefit from being re-written with a dependency injection of the `Menu` class.

PR #1656, opened 2020-04-27
Submit Takeaway Challenge

will-head:
Competes all User Stories, except actually sending text message (puts output instead).

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -2,6 +2,7 @@ class Menu
   attr_reader :dishes
   
   ERROR_DISH_EXISTS = "Error: that dish number already exists"
+  ERROR_DISH_NOT_EXISTS = "Error: that dish is not on the menu"
   
   def initialize 
     @dishes = Array.new 
@@ -15,7 +16,7 @@ class Menu
   end
   
   def get(number)
-    raise "Error: that dish is not on the menu"
+    raise ERROR_DISH_NOT_EXISTS
   end
   
   private
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -26,17 +26,17 @@ class Menu
     raise ERROR_DISH_NOT_EXISTS
   end
   
-  def check?(number)
-    return false if @dishes.empty?
-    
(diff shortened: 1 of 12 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,39 @@
...
+    raise ERROR_DISH_EXISTS if include?(dish)
+    
+    @dishes.push(dish)
+    @dishes.last
+  end
+  
+  def get(number)
+    raise ERROR_DISH_NOT_EXISTS if @dishes.empty?
+    
+    @dishes.each do |on_menu|
```
mbrad26, line 21: Maybe you could extract another method and make it private.

lib/menu.rb
```diff
@@ -0,0 +1,39 @@
+class Menu
+  attr_reader :dishes
+  
+  ERROR_DISH_EXISTS = "Error: that dish number already exists"
+  ERROR_DISH_NOT_EXISTS = "Error: that dish is not on the menu"
+  
+  def initialize
```
mbrad26, line 7: Try and use Dependency Injection.
`def initialize(dishes = Array.new)
    @dishes = dishes
end
`

lib/order.rb
```diff
@@ -0,0 +1,31 @@
+class Order
+  attr_reader :order, :time
+  
+  ERROR_NOT_INTEGER = "Error: only numbers can be supplied"
+
```
mbrad26, line 5: Here you can use dependency injection as well.

spec/dish_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'dish'
+
+context Dish do
```
mbrad26, line 3: You could use describe instead.

spec/order_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'order'
+require 'menu'
+require 'dish'
+
+context Order do
```
mbrad26, line 5: describe as well

PR #1660, opened 2020-04-27
Takeaway Challenge

cjm106:
Work done so far on the WK2 weekend challenge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,34 +1,40 @@
 Takeaway Challenge
 ==================
-```
-                            _________
-              r==           |       |
-           _  //            |  M.A. |   ))))
-          |_)//(''''':      |       |
-            //  \_____:_____.-------D     )))))
-           //   | ===  |   /        \
-       .:'//.   \ \=|   \ /  .:'':./    )))))
-      :' // ':   \ \ ''..'--:'-.. ':
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
 
- ```
+    \\\\\\\|///////
+     \\\\\\|//////
+      \\\\\|/////
+       \\\\|////
+       ./#####\.
+      -|#######|-
+      -|#######|-                         Jock'O'Byte Restaurant
+       '\#####/`            .    
+        '  |  `     .      /   London's Favourite Virtual Scottish Restaurant
+           |-  .   |\_ _//              
+          -|    |\_|   //___.            ** NOW DOING TAKEAWAY **
+           |    | '  / __--'
+           |- . |  /  <___.
+           |  |` /  __--'
+          -|  |/___<.
+           | /
+           |-
+           |
+          -|
+
+
 
 Instructions
 -------
 
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+```
+    $ git clone git@github.com:cjm106/takeaway-challenge.git
+    $ cd takeaway-challenge
+    $ rspec   # to run tests for User Stories 1-3
 
-Task
------
+```
 
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+User Stories
+-------
 
 ```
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,34 +1,40 @@
...
-          |_)//(''''':      |       |
-            //  \_____:_____.-------D     )))))
-           //   | ===  |   /        \
-       .:'//.   \ \=|   \ /  .:'':./    )))))
-      :' // ':   \ \ ''..'--:'-.. ':
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
 
- ```
+    \\\\\\\|///////
```
collier-jo, line 4: Maybe consider having: Domain model // IRB. An IRB would be helpful to see how this interacts i.e. in mine its like a new user is made and then does x y z so you can (in theory) follow the flow.

README.md
```diff
@@ -48,37 +54,12 @@ So that I am reassured that my order will be delivered on time
...
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this weekend.
-
-Notes on Test Coverage
-------------------
-
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+To Do
```
collier-jo, line 57: Maybe as well as having how your code works or how you went about making them work

lib/interface.rb
```diff
@@ -0,0 +1,43 @@
+require 'menu'
+
+class Interface
+
+  attr_reader :food
+
+  def initialize
+    @food = Menu.new
```
collier-jo, line 8: Could have this as a dependancy injection so it is not so hard coded?

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
+class Menu
+
+  attr_reader :items
+
+  def initialize
+    @items = [{ name: "Haggis Supper", price: "10.99" },
```
collier-jo, line 6: As this is in initialize if a user updates the menu this will be overridden next time the menu is created. Maybe consider having an empty array of hashes.

lib/place_order.rb
```diff
@@ -0,0 +1,31 @@
+require 'menu'
+
+class PlaceOrder
+
+  attr_reader :basket
+  def initialize
+    @basket = []
+    @menu = Menu.new
```
collier-jo, line 8: Hardcoded.. Maybe could be injected?

spec/interface_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(subject.food).to receive(:menu)
+    subject.food_menu
+  end
+
+  it "Is able to display the options_menu." do
+    options = "------------------------------------------\n1. Place your order\n2. View your order\n9. Exit\n"
+    expect { subject.options_menu }.to output(options).to_stdout
+  end
+
+  it "Is able to accept user input '1' and call on :place_order." do
```
collier-jo, line 15: I would double check how this is being called. As you interface does not call an PalaceOrder class and place_order is a method in that class this should in theory fail without some kind of double... We can maybe go through together and find an appropriate  double as I am slightly unsure too.

lib/interface.rb
```diff
@@ -0,0 +1,43 @@
...
+  def options_menu
+    puts "------------------------------------------"
+    puts "1. Place your order"
+    puts "2. View your order"
+    puts "9. Exit"
+  end
+
+  def input(option)
+    case option
+    when "1"
```
collier-jo, line 24: See spec comment around this

spec/menu_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'menu'
+
+describe Menu do
+
+  it "Is able to store a list of dishes with name and price." do
+    subject.add_dish("Special", "8.50")
```
collier-jo, line 6: This is where you could have a let double i.e.

let(:food_item) { double:menu }  or { "Special" }
let(:price) { double:menu } or { 8.50 } 

The bits in the {} are place holders so could be literally anything 
The let is referring to the parameter name

README.md
cjm106: Yes I need to start adding these to my readme.

PR #1664, opened 2020-04-27
Pull request - takeaway 

collier-jo:


lib/order.rb
```diff
@@ -0,0 +1,33 @@
+require "./lib/menu"
+
+class Order
+  attr_reader :current_items
+  def initialize(menu = Menu.new.full_menu)
```
cjm106, line 5: I like the way you're calling on the full_menu method here.

spec/order_spec.rb
```diff
@@ -0,0 +1,41 @@
+require './lib/order'
+require "./lib/menu"
+require "./lib/user"
+
+describe Order do
+
+  let(:item) { double :order }
```
cjm106, line 7: Great to see some doubles in your tests - I really should've spent time implementing these as well.

spec/order_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+  describe "#basket_summary" do
+    it"responds to baskey_summary" do 
+      expect(subject).to respond_to(:basket_summary)
+    end 
+
+    it "adds the hash arrays up and assigns value " do 
+      subject.add_item("tofu curry", 5)
+      subject.add_item("tofu curry", 6)
+      
+      expect(subject.basket_summary).to eq(["tofu curry, x11, £61.6"])
```
cjm106, line 38: Yum, tofu curry, and a good way to test that your basket works!

lib/user.rb
```diff
@@ -0,0 +1,27 @@
+require "./lib/menu"
```
cjm106, line 1: Nice clean code and some SRP to boot - Its really clear to me what each part is doing!

spec/user_spec.rb
```diff
@@ -0,0 +1,45 @@
+require './lib/user'
+require "./lib/menu"
+require "./lib/order"
+
+describe User do
+  let(:unicorn) { Order.new }
```
cjm106, line 6: Unicorns need to eat to!

PR #1665, opened 2020-04-27
first commit - created menu csv file, set up takeaway spec and lib files

Tracht:
unable to load menu file, issues with creating a test for that

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,25 @@
+# Takeaway class
+
+class Takeaway
+
+  def initializer(menu='TonysPizzaMenu.csv')
+    @menu = menu
+  end
+
+  def show_menu
+    file = File.open(@menu, "r")
+    file.readlines.each do |line|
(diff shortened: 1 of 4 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,25 @@
+# Takeaway class
+
+class Takeaway
+
+  def initializer(menu='TonysPizzaMenu.csv')
```
ZsofiaS, line 5: Good use of external file

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'takeaway'
+
+describe Takeaway do
+
+  context 'when initialized' do
+    it 'accepts 1 argument: default menu' do
+      menu = double()
+      expect(subject).to respond_to(:menu)
+    end
+    describe '#show_menu' do
```
ZsofiaS, line 10: As a next step, you can write another test for show_menu, to not only test that the method exists, but on what it does.

PR #1667, opened 2020-05-25
Finished weekend task 2

ChrisCooney05:
Everything is complete and working as I would like it to, the only thing I have not managed to do was put together an RSpec test for the Twilio API, I couldn't work out how to do it without sending and SMS.


README.md
```diff
@@ -14,71 +14,47 @@ Takeaway Challenge
...
 
-Notes on Test Coverage
-------------------
+```
+chinese.confirm("+44your_number_here")
+```
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+By doing this you will get a confirmation, by text to let you know that your food will be with you
+within an hour from point of order.
```
tatsiana-makers, line 60: Really nice to see an example of how someone would use your program!

lib/message.rb
```diff
@@ -0,0 +1,24 @@
+require 'twilio-ruby'
+require 'time'
+
+class Message
+
+  attr_reader :time
+
+  def initialize
+    @time = Time.now + 1 * 60 * 60
```
tatsiana-makers, line 9: Without a comment, it's difficult to understand why you're adding 3600 to `Time.now`. What do you think about adding a comment here?

Is this something to do with british summer time? Does `Time` have an inbuilt way of giving you the correct time accounting for british summer time?

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+
+require_relative 'menu'
+require_relative 'message'
+
+class Order
+
+  attr_reader :current_order
+
+  def initialize
+    @menu = Menu.new
```
tatsiana-makers, line 13: Your order class is tightly coupled with your Menu class. Could we use dependency injection here?

https://github.com/makersacademy/skills-workshops/blob/f5b4801840fe07d26ff70341652dc81dcda12289/practicals/object_oriented_design/dependency_injection.md might be a helpful resource

lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+  def total
+    @current_order.each do |item|
+      puts "#{item[:item]}: #{item[:price]}"
+      @order_total += item[:price]
+    end
+    puts "current order total is £%0.2f" % [@order_total]
+  end
+
+  def confirm(number)
+    @message = Message.new
```
tatsiana-makers, line 35: Could we also use dependency injection for the Message class? 🙂

spec/menu_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+describe Menu do
+  it 'Should have a menu class' do
+    expect(Menu).to respond_to(:new)
+  end
+
+  describe '#read' do
+    it 'Should print out a list of items' do
+      expect { subject.view }.to output.to_stdout
+    end
+  end
```
tatsiana-makers, line 12: Can you think of  a test for the `select` method?

spec/message_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    expect(Message).to respond_to(:new)
+  end
+
+  it 'Should initialize with the current time + 1 hour' do
+    expect(subject.time).to eq((Time.now + 1 * 60 * 60).strftime("%k:%M"))
+  end
+
+  describe '#text_confrim' do
+    it 'Should have a #text_message method' do
+      expect(subject).to respond_to(:text_message)
```
tatsiana-makers, line 15: Are there any more details we can check about the `text_message` method? What about what we expect it to receive?

lib/message.rb
ChrisCooney05: Whilst looking up how to add 1 hour to a new instance of time, everything pointed to the syntax I have used. On testing, the text message does send confirming that the food will be with the customer 1 hour from the current time.

lib/order.rb
ChrisCooney05: Thanks for pointing this out, totaly forgot about DI

lib/message.rb
tatsiana-makers: Ah, that makes sense! Then a comment to explain that you want the text to be sent an hour after the order would help the reader understand the purpose behind this line of code 🙂

PR #1668, opened 2020-05-25
Tristan Takeaway Challenge

tristanlangford:


spec/restaurant_spec.rb
```diff
@@ -0,0 +1,84 @@
+require 'restaurant'
+
+describe Restaurant do
```
tatsiana-makers, line 3: Really nice tests!

PR #1669, opened 2020-05-25
As far as I got with the challenge

Sumner1185:
I really struggled with the SMS class and could not get any tests to work. The first two classes (menu and order) went OK but then problems began when dealing with Twilio.


lib/order.rb
```diff
@@ -0,0 +1,25 @@
...
+    @dishes = {}
+  end
+
+  def add(dish, quantity)
+    raise "Apologies, #{dish} is not available" unless @menu.dish_available?(dish)
+  
+    @dishes[dish] = quantity
+  end
+
+  def bill
```
tatsiana-makers, line 16: Could we make this a private method to achieve more encapsulation? Are there other methods that could be private?

lib/sms.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def body
+    "Thank you, order received! It will be delivered before #{delivery_time}"
+  end
+
+  def delivery_time
+    (Time.now + MAX_DELIVERY_TIME).strftime(TIME_FORMAT)
+  end
+
+  def variables
```
tatsiana-makers, line 34: Nice use of environment variables! When is this `variables` method called?

lib/takeaway.rb
```diff
@@ -0,0 +1,13 @@
+class Takeaway
+
+  def initialize(menu, order, sms)
```
tatsiana-makers, line 3: If you do decide that we need this class, the `order`  and `sms` inputs here are not currently being used. How might using this variable help us use dependency injection?

https://github.com/makersacademy/skills-workshops/blob/f5b4801840fe07d26ff70341652dc81dcda12289/practicals/object_oriented_design/dependency_injection.md might be useful 🙂

spec/order_spec.rb
```diff
@@ -0,0 +1,34 @@
+require_relative '../lib/order'
+require_relative '../lib/menu'
```
tatsiana-makers, line 2: Do we need to require the `Menu` class here? Could we isolate the unit tests to not use the real menu class?

spec/sms_spec.rb
```diff
@@ -0,0 +1,31 @@
+require_relative '../lib/sms'
+require 'time'
+
+describe SMS do
```
tatsiana-makers, line 4: I'd recommend checking out https://github.com/makersacademy/takeaway-challenge/blob/master/docs/review.md#stubbing-the-twilio-api (which I also linked in your cohort channel earlier today) as inspiration of how you can test the SMS class without actually sending the messages

lib/takeaway.rb
```diff
@@ -0,0 +1,13 @@
+class Takeaway
```
tatsiana-makers, line 1: Do we need the Takeaway class? It looks like its only responsibility is to delegate to the menu class to display the menu.

README.md
```diff
@@ -47,38 +37,31 @@ As a customer
...
 ```
+**My Approach**
+-----
+
+**Domain Model Diagrams**
+
+
+**Responsibilities**
+
+![Responsibilities](https://github.com/Sumner1185/takeaway-challenge/blob/master/ResponsibilityTable.png)
```
tatsiana-makers, line 48: Really nice format for the class diagram for the responsibilities of each class! Is the Takeaway class responsible for allowing the order to be placed or sending the text message? Usually responsibilities can be derived by looking at the public methods within a class

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,15 @@
+require_relative '../lib/takeaway'
+
+describe Takeaway do
+  subject(:takeaway) { described_class.new(menu: menu, order: order, sms: sms)}
```
tatsiana-makers, line 4: It looks like this test fails with a useful error message as to what is going wrong. Can you use this message to fix the error?

PR #1672, opened 2020-05-25
Week 2 weekend challenge: Takeaway Challenge

ad13380:


README.md
```diff
@@ -48,37 +33,11 @@ So that I am reassured that my order will be delivered on time
...
-
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this weekend.
-
-Notes on Test Coverage
-------------------
+## Considerations
```
tatsiana-makers, line 36: Really nice description of your considerations, giving a reviewer a nice look into your thought process (:

lib/basket.rb
```diff
@@ -0,0 +1,67 @@
...
+  def sub_total(dish, quantity)
+    (@menu_data[dish].to_f * quantity).round(2)
+  end
+
+  def capitalise_item(item)
+    item.split.map { |word| word.capitalize }.join(" ")
+  end
+
+  def store_menu_data
+    @menu_data = load_data('../menu.csv')
```
tatsiana-makers, line 56: My understanding is that the reason the code is structured this way is to avoid many instances of the same menu existing. What if the `initialize` method for this class accepted an instance of Menu as input? The reason this might be a better approach is that imagine someone creates a menu, makes an order, and then a dev accidentally edits this csv. This means that the view of the menu items will be inconsistent between the Order and Menu. I also expect that this will reduce some repeated code between your classes, as each currently needs to load and process the csv data

lib/menu.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+    return
+  end
+
+  def capitalise_item(dish)
+    dish.split.map { |word| word.capitalize }.join(" ")
+  end
+
+  def store_menu_data
+    @menu_data = load_data('../menu.csv')
```
tatsiana-makers, line 34: What if we wanted a second restaurant with another menu? How could we achieve this? Might an initialise method with appropriate inputs help here?

lib/order.rb
```diff
@@ -0,0 +1,75 @@
+require 'csv'
+
+class Order
+  def initialize
```
tatsiana-makers, line 4: Could this also accept an instance of Menu as input?

lib/takeaway.rb
```diff
@@ -0,0 +1,76 @@
+require_relative 'menu'
+require_relative 'order'
+require_relative 'basket'
+require_relative 'checkout'
+
+class TakeAway
+  def initialize(menu = Menu.new, order = Order.new, checkout = Checkout.new)
```
tatsiana-makers, line 7: Awesome use of dependency injection!

spec/checkout_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+describe Checkout do
+  let(:user_prompt) { "Please enter your phone number to complete the order:\n" + "+44 " }
+  let(:confirmation_prompt) { "Your order is on it's way! Please check your phone for confirmation.\n" }
+
+  describe '#place_order' do
+    it { is_expected.to respond_to(:place_order) }
+
+    it 'prompts the user for their phone number' do
+      allow(subject).to receive(:gets).and_return("1234567890")
+      allow(subject).to receive(:send_sms)
```
tatsiana-makers, line 12: It looks like you stubbed the sending of the Twilio API, nice! For more information, I recommend you check out https://github.com/makersacademy/takeaway-challenge/blob/master/docs/review.md#stubbing-the-twilio-api

spec/checkout_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+
+describe Checkout do
+  let(:user_prompt) { "Please enter your phone number to complete the order:\n" + "+44 " }
+  let(:confirmation_prompt) { "Your order is on it's way! Please check your phone for confirmation.\n" }
+
+  describe '#place_order' do
+    it { is_expected.to respond_to(:place_order) }
+
+    it 'prompts the user for their phone number' do
+      allow(subject).to receive(:gets).and_return("1234567890")
```
tatsiana-makers, line 11: This stubbing for the user input looks great!

ad13380: Thanks for the input! Would you say generally this was the correct way to test an app that has so much user input? Or is there a better framework for it that would allow better test coverage?

tatsiana-makers: This is the way that I've seen it done in the past. I'm not aware of any frameworks off the top of my head

PR #1673, opened 2020-05-25
Rich Takeaway

RichEwin:
Unable to complete last user story. 

I wanted to add a variety of additional features throughout the app but had trouble writing the unit tests. I wanted to add the sms functionality and confirm the order had been added to basket with a output statement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,34 +1,4 @@
-Takeaway Challenge
-==================
-```
-                            _________
-              r==           |       |
-           _  //            |  M.A. |   ))))
-          |_)//(''''':      |       |
-            //  \_____:_____.-------D     )))))
-           //   | ===  |   /        \
-       .:'//.   \ \=|   \ /  .:'':./    )))))
-      :' // ':   \ \ ''..'--:'-.. ':
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
-
- ```
-
-Instructions
--------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
------
-
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+## User Stories
 
 ```
 As a customer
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -48,37 +18,40 @@ So that I am reassured that my order will be delivered on time
...
+- Class ----> menu
+- Method (behaviour) ----> menu
 
-Notes on Test Coverage
-------------------
+## How to run
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+```
+irb
```
Katie-McDonagh, line 45: using an irb example in the readme makes it really easy to understand how your program works, as does the domain modelling

spec/menu_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'menu'
+
+describe Menu do
+
+  it 'creates a new instance of menu' do
+    expect(subject).to be_instance_of Menu
+  end
+
```
Katie-McDonagh, line 8: you could think about using mocks/doubles here to test the behaviour of the .menu method instead of using the actual menu to test, which I think falls more into testing behaviour rather than state.

spec/message_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'message'
+
```
Katie-McDonagh, line 2: you could think about changing the nature of the test to something like it "delivers a message with estimated delivery time" using expect(subject).to eq "thanks, order to be delivered before 17:00"

spec/message_spec.rb
```diff
@@ -0,0 +1,7 @@
...
+describe Message do
+  it "creates an instance of message" do
+    expect(subject).to be_instance_of Message
+  end
+end
```
Katie-McDonagh, line 7: you could think about changing the nature of the test to something like it "delivers a message with estimated delivery time" using expect(subject).to eq "thanks, order to be delivered before 17:00"

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    'Mango Martini Cocktail' => 7
+
+    }
+  }
+
+  it 'creates a new instance of takeaway' do
+    expect(subject).to be_instance_of Takeaway
+  end
+
+  describe '#view_menu' do
```
Katie-McDonagh, line 27: I like the clearly defined describe blocks for different methods, makes it easy to read

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    'Hawaiian Margarita Cocktail' => 7,
+    'Mango Martini Cocktail' => 7
+
+    }
+  }
+
+  it 'creates a new instance of takeaway' do
+    expect(subject).to be_instance_of Takeaway
+  end
+
```
Katie-McDonagh, line 26: using the double you have created :printed menu, the next test could be  

it "returns the menu of dishes and prices" do
    expect(takeaway.printed_menu).to eq(view_menu)
  end

lib/order.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'menu'
+require_relative 'takeaway'
+
+class Order
+  attr_reader :basket, :menu
+
+  def initialize
```
tatsiana-makers, line 7: Could we use dependency injection here to decouple the classes? https://github.com/makersacademy/skills-workshops/blob/f5b4801840fe07d26ff70341652dc81dcda12289/practicals/object_oriented_design/dependency_injection.md

lib/order.rb
```diff
@@ -0,0 +1,29 @@
...
+    @basket = {}
+    @basket_total = 0
+    @menu = Menu.new
+  end
+
+  def add_to_order(dish, quantity)
+    @basket[dish] = quantity
```
tatsiana-makers, line 14: Can we add a dish that isn't on the menu?

lib/order.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def order_summary
+    @basket.each do |item, quantity|
+      puts "#{item} x #{quantity}: £#{Menu.new.menu[item] * quantity}"
+    end
+  end
+
+  def basket_total
+    @basket.inject(0) do |total, (item, quantity)|
+      @item_total = total + @menu.menu[item] * quantity
```
tatsiana-makers, line 25: What's the difference between `@basket_total` and `@item_total`?

lib/order.rb
```diff
@@ -0,0 +1,29 @@
...
+    @menu = Menu.new
+  end
+
+  def add_to_order(dish, quantity)
+    @basket[dish] = quantity
+  end
+
+  def order_summary
+    @basket.each do |item, quantity|
+      puts "#{item} x #{quantity}: £#{Menu.new.menu[item] * quantity}"
```
tatsiana-makers, line 19: rather than relying on the `Menu` class directly here, could we use `@menu`? Combined with dependency injection this would allow us to isolate the two classes

lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'menu'
+require_relative 'order'
+
+class Takeaway
```
tatsiana-makers, line 4: Can a takeaway have only ever have one order?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'takeaway'
+require 'menu'
+require 'order'
+
+describe Takeaway do
+
+  let(:order) { Order.new }
```
tatsiana-makers, line 7: We should use doubles rather than actual instances of other classes when unit testing a class- this allows us to isolate the implementations as I was mentioning in the kick-off this morning.

PR #1675, opened 2020-05-25
Nikita's Takeaway Challenge

NikitaDouglas:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'takeaway'
+
+describe Takeaway do
+  subject(:takeaway) { Takeaway.new }
```
ooduola: # Appreciate how you've defined takeaway instead of subject. It's easier to follow

ooduola: # Appreciate how you've defined takeaway instead of subject. It's easier to follow

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  describe '#see_menu' do
+    it 'returns a formatted list of menu items to the customer' do
+      expect { takeaway.see_menu }.to output("Cod & Chips - 5.40\nWhite Sausage - 2.30\n").to_stdout
```
ooduola: You could look to define the menu using let rather than hardcoding.

PR #1676, opened 2020-05-25
Takeaway challenge - Bene Arinci

BeneArinci:


lib/menu.rb
```diff
@@ -0,0 +1,21 @@
...
+  def initialize
+    @dishes = {
+      "marinara" => 5,
+      "margherita" => 6,
+      "napoli" => 6,
+      "calzone" => 8
+    }
+  end
+
+  def show_menu
```
ad13380, line 14: You can try putting this sort of thing in a "presentation class" which will take the menu data and format it to look nice (which I didn't do either, but I heard it was useful)

lib/message.rb
```diff
@@ -0,0 +1,19 @@
+class Message
+  require 'twilio-ruby'
```
ad13380, line 2: Try putting this in the gem file so that you don't have to put it here

lib/order.rb
```diff
@@ -0,0 +1,25 @@
+require './lib/takeaway.rb'
```
ad13380, line 1: Try using 'require_relative' for linking to other classes. It probably doesn't make a huge difference here, but I've had some very annoying bugs that took hours to solve - only to realise that 'require_relative' is supposed to be used for linking to other classes, and 'require' should be used for gems

lib/order.rb
```diff
@@ -0,0 +1,25 @@
...
+class Order
+  attr_reader :items, :menu
+  def initialize(menu)
+    @items = {}
+    @menu = menu
+  end
+
+  def add_dish(dish, quantity = 1)
+    @items[dish] = 0 unless items[dish]
```
ad13380, line 10: You can define the hash on line 5 as @items = Hash.new(0), which means you don't need to put this line in

lib/order.rb
```diff
@@ -0,0 +1,25 @@
...
+    @items[dish] += quantity
+  end
+
+  def single_item_price(dish)
+    @menu.dishes[dish]
+  end
+
+  def order_price
+    total = 0
```
ad13380, line 19: Try looking at the .reduce method here, it would mean you don't need to have line 19 or 23

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def visualise_menu
+    @menu.show_menu
+  end
+
+  def add_to_order(dish, quantity = 1)
+    raise "item not available" unless @menu.in_menu?(dish)
+
+    @order.add_dish(dish, quantity)
+    "#{dish} X #{quantity} added to your order"
```
ad13380, line 21: Good idea putting a confirmation message in

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'menu'
+
+describe Menu do
+  let(:menu) { Menu.new }
+  let(:list) { ["marinara - £5", "margherita - £6", "napoli - £6", "calzone - £8"] }
```
ad13380, line 5: Good use of the let() method to help keep your tests DRY

spec/message_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'message'
+describe Message do
+  subject(:message) { described_class.new }
+  it 'creates an instance variable for @my_number' do
+    expect(subject.my_number).to eq('+447510060718')
```
ad13380, line 5: maybe this should be expect(message.my_number), seeing as you defined it that way in line 3

spec/order_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'order'
+
+describe Order do
```
ad13380, line 3: Great use of mocking throughout

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'takeaway'
+require 'menu'
```
ad13380, line 2: since you are isolating tests with doubles, you can probably take this out

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+describe Takeaway do
+  let(:menu) { double :menu }
+  let(:order) { double :order }
+  let(:message) { double :message }
+  subject(:takeaway) { described_class.new(menu, order, message) }
+  let(:list) { ["marinara - £5", "margherita - £6", "napoli - £6", "calzone - £8"] }
```
ad13380, line 9: Good idea for keeping tests DRY

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  it 'uses the class menu, show_menu method, to make the user visualise dishes in a menu format' do
+    allow(menu).to receive(:show_menu).and_return(list)
+    expect(subject.visualise_menu).to eq(["marinara - £5", "margherita - £6", "napoli - £6", "calzone - £8"])
```
ad13380, line 13: You can set the eq() to be eq(list) to save you from having to type it out

lib/message.rb
ad13380: Forget that - I was wrong

PR #1677, opened 2020-05-25
Tim's Takeaway Challenge Attempt

TimCPB:


TimCPB:
Not quite sure what exactly what functionality the user stories require but they are all satisfied in feature tests, to some degree.

I didn't know how to test twilio features in rspec and ran out of time, so there are no unit tests for the fourth user story.

The spec tests don't really use any mocks or doubles, so I know they're not very well written.

After dead ends on Saturday/Sunday I had to go back to the drawing board and ask for some guidance from my mentor at the planning/modelling stage.

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
...
+    puts "\nMenu\n\n"
+    @list_of_dishes.each { |dish, price|
+      puts "#{dish}........£#{price}"
+    }
+    puts "\n"
+  end
+
+  def on_menu?(dish)
+    @list_of_dishes.has_key?(dish.to_sym)
+
```
joshuaabrookuk, line 20: Extra space?

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+class Menu
+
+  attr_reader :list_of_dishes
+  DEFAULT_MENU = { "chicken tikka": 9, "beef madras": 10, "vegetable vindaloo": 8 }
```
joshuaabrookuk, line 4: Nice clean constant

lib/order.rb
```diff
@@ -0,0 +1,24 @@
+require_relative 'menu'
```
joshuaabrookuk, line 1: It seams like you understand high coupling and low cohesion.

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class Order
+
+  attr_reader :selection, :menu
+  DEFAULT_MENU = { "chicken tikka": 9, "beef madras": 10, "vegetable vindaloo": 8 }
```
joshuaabrookuk, line 6: Is there a way to use DEFAULT_MENU from your menu.rb file.

PR #1681, opened 2020-05-26
Nick Butlers attempt at the Takeaway Challenge

nwmbutler:


lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
...
+require_relative 'order'
+
+class TakeAway
+
+  attr_reader :menu
+
+  def initialize
+    @menu = Menu.new
+    @order = Order.new
+  end
```
RaeRachael, line 11: def initialize (menu = Menu.new)

lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def seemenu
+    @menu.menu
+  end
```
RaeRachael, line 15: see_menu?

PR #1685, opened 2020-05-26
Week2 - Takeaway Challenge: Code Review

sophiewo:
I haven't full TDD'ed the Twilio implementation yet but it should work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,9 +18,10 @@ I would like to see a list of dishes with prices
 
 * Menu Class:
   * menu_list: 
-    1.  variable with restaurant name
+    1. variable with restaurant name
+    2. change menu_list to hashes with number values and prices
   * show_list: 
-    1.  method to show variable.
+    1.  method to show hash interpolation.
 
 ------
 ```
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ I would like to see a list of dishes with prices
   * menu_list: 
     1. variable with restaurant name
     2. change menu_list to hashes with number values and prices
-  * show_list: 
-    1.  method to show hash interpolation if entire menu list.
+  * #show_list: 
+    1. method to show hash interpolation if entire menu list.
     2. prints restaurant name and menu.
 
 ------
@@ -30,7 +30,13 @@ As a customer
 So that I can order the meal I want
 I would like to be able to select some number of several available dishes
 ```
-```
+##### To satisfy second user story TDD step implimentation:
+
+* Order Class:
+  * #select_items : ask for customer order
+
+------
+<!-- ```
 As a customer
 So that I can verify that my order is correct
 I would like to check that the total I have been given matches the sum of the various dishes in my order
@@ -39,5 +45,5 @@ I would like to check that the total I have been given matches the sum of the va
 As a customer
 So that I am reassured that my order will be delivered on time
 I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
-```
+``` -->
 
--- a/README.md
+++ b/README.md
@@ -1,84 +1,36 @@
-Takeaway Challenge
-==================
-```
-                            _________
-              r==           |       |
-           _  //            |  M.A. |   ))))
-          |_)//(''''':      |       |
-            //  \_____:_____.-------D     )))))
-           //   | ===  |   /        \
-       .:'//.   \ \=|   \ /  .:'':./    )))))
-      :' // ':   \ \ ''..'--:'-.. ':
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
+# Takeaway Challenge
 
- ```
+This programme was created using TDD of the following user stories
 
-Instructions
--------
-
-* Challenge time: rest of the day and weekend, until Monday 9am
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
------
-
+### To install
 * Fork this repo
 * Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
 
-```
+### User Stories
+````
 As a customer
 So that I can check if I want to order something
 I would like to see a list of dishes with prices
+````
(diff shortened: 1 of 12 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
+
+  TAKE_AWAY_NAME = "Costa del Oval"
+
+  attr_reader :menu_list, :take_away_name
+
+  def initialize
+    @take_away_name = TAKE_AWAY_NAME
+    @menu_list = [{ item_number: 1, item: "Veggie Burger and Sweet Potato Fries", price: 3 },
```
Emanuele-20, line 9: Great idea put Menu in initialize. It'll useful for each new instance of Menu

spec/order_process_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+context ' #show_menu' do
+  let(:expected_output) { ["Costa del Oval Menu",
+                             "1. Veggie Burger and Sweet Potato Fries: £3",
+                             "2. Veggie Lasagna: £5",   
+                             "3. Falafel and Hallumi Wrap: £4"].join("\n") + "\n" 
+    }
+
+  it 'shows entire menu list' do
+    expect { subject.show_menu }.to output(expected_output).to_stdout
```
Emanuele-20, line 17: Good syntax

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    it 'responds to #show_list' do
+      expect(subject).to respond_to(:show_list)
+    end
+
+    it 'shows entire menu list' do
+      expect { subject.show_list }.to output(expected_output).to_stdout
+    end
+  end 
+end
```
Emanuele-20, line 29: Clear Rspec

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      order.select_items(1)
+      order.customer_order_summary
+      order.subtotal
+      subtotal_test = order.subtotal
+      expect(order.subtotal).to eq(subtotal_test)
+    end
+  end
+end
+
+
```
Emanuele-20, line 56: Refactoring scenario

lib/order.rb
```diff
@@ -0,0 +1,42 @@
...
+  def customer_order_summary
+    @customer_basket.each do |order|
+      return "#{order[:item]}, price £#{order[:price]}"
+    end
+  end
+  
+  def subtotal
+    @customer_basket.each do |value| 
+      basket_subtotal << (value[:price].to_i) 
+      return basket_subtotal.sum
```
tatsiana-makers, line 39: What happens if we `return` inside of a loop- will we iterate over the entire customer basket?

lib/order.rb
```diff
@@ -0,0 +1,42 @@
...
+  # def show_menu
+  #   Menu.new.show_list
+  # end
+
+  def select_items(customer_choice) 
+    raise "Invalid number, please choose again" if customer_choice > @menu.length
+
+    @menu.each do |choice|
+      if customer_choice == choice[:item_number]
+        customer_basket << choice
```
tatsiana-makers, line 25: Should this be `@customer_basket`?

lib/order_process.rb
```diff
@@ -0,0 +1,47 @@
...
+require_relative 'send_sms'
+
+class OrderProcess
+
+  def initialize
+    @restaurant_name = Menu.new.take_away_name
+    @order = Order.new
+  end
+
+  def show_menu
```
tatsiana-makers, line 13: Does the `OrderProcess` class need this method? Consider what actions belong to which class and only have the actions as methods inside of its owning class

lib/order_process.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'order'
+require_relative 'menu'
+require 'twilio-ruby'
+require_relative 'send_sms'
+
+class OrderProcess
+
+  def initialize
```
tatsiana-makers, line 8: We can use dependency injection here as well 🙂

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+    it { is_expected.to respond_to(:subtotal) }
+
+    xit 'returns sum of all items in customer basket' do
+      customer_basket = [2, 4]
+      order.select_items(2)
+      order.select_items(1)
+      order.customer_order_summary
+      order.subtotal
+      subtotal_test = order.subtotal
+      expect(order.subtotal).to eq(subtotal_test)
```
tatsiana-makers, line 51: Try and think of the assertion of a unit test as a mathematical equation. On the left hand side you have the _actual_ value, where you call the method and on the right hand side you have the _expected_ value. So in this case the expected value is `8`.

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  describe '#subtotal' do
+      
+    it { is_expected.to respond_to(:subtotal) }
+
+    xit 'returns sum of all items in customer basket' do
+      customer_basket = [2, 4]
+      order.select_items(2)
+      order.select_items(1)
+      order.customer_order_summary
```
tatsiana-makers, line 48: Do we need to call `customer_order_summary` or `subtotal` outside of the assertion on line 51?

spec/send_sms_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'send_sms'
+
+describe Message do
+  it 'should instantiate a new instance' do
+    expect(Message.new).to be_instance_of Message
```
tatsiana-makers, line 5: As shared in the slack channel this morning, https://github.com/makersacademy/takeaway-challenge/blob/master/docs/review.md#stubbing-the-twilio-api is a resource you may find helpful 🙂

README.md
```diff
@@ -1,84 +1,62 @@
...
 
+----
+## User Stories
+----
 ```
 As a customer
+```
+##### To satisfy first user story TDD step implimentation:
```
tatsiana-makers, line 18: Nice description of your approach and thought process to these user stories

lib/order_process.rb
```diff
@@ -0,0 +1,47 @@
...
+    customer_order_summary
+  end
+
+  def confirmation_text
+    Message.new.confirmation_text
+  end
+
+  private
+
+  def order_asker
```
tatsiana-makers, line 44: Nice use of private methods for encapsulation!

PR #1688, opened 2020-06-21
Takeaway Challenge - Dhara

Dhara-95:
- Completed user stories 1 and 2.
- Started user story 3 but not complete - all tests not passed. 
- In future I would improve the relationship between my classes and not have the menu hard-coded into the @dishes instance variable. I would also have liked to adhere to SRP as currently there are methods that could be moved into separate classes.

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,15 +1,23 @@
 class Menu
   attr_reader :dishes
-  MENU = [
-    { "Magherita" => "£5.00" },
-    { "Pepperoni" => "£8.00" }
-  ]
+  MENU = {
+    "Margherita" => 5.00,
+    "Pepperoni" => 8.00
+  }
 
   def initialize
     @dishes = MENU
   end
 
   def view_menu
-    @dishes
+    @dishes.map do |dish, price|
+      "#{dish}: £#{('%.2f' % price)}"
+    end
+  end
+
+  def dish_available?(chosen_dish)
+    unless @dishes.include?(chosen_dish) == false
+      true
+    end
   end
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,11 +1,12 @@
 class Menu
   attr_reader :dishes
+  MENU = [
+    { "Magherita" => "£5.00" },
+    { "Pepperoni" => "£8.00" }
+  ]
 
   def initialize
-    @dishes = [
-      {"Magherita" => "£5.00"},
-      {"Pepperoni" => "£8.00"}
-    ]
+    @dishes = MENU
   end
 
   def view_menu
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,12 +1,11 @@
 class Menu
   attr_reader :dishes
-  MENU = {
-    "Margherita" => 5.00,
-    "Pepperoni" => 8.00
-  }
 
   def initialize
-    @dishes = MENU
+    @dishes = {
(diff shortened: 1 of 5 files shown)


spec/menu_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'menu'
+
+describe Menu do
+  subject(:menu) { Menu.new }
```
stupot1, line 4: This is not needed as subject is always an instance of Menu when inside the describe block on line 3

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
+class Menu
+  attr_reader :dishes
+
+  def initialize
+    @dishes = {
+      "Margherita" => 5.00,
+      "Pepperoni" => 8.00
+    }
```
stupot1, line 8: Would a single hash be the best way to store this data? For allowing this class to be expanded later an array of hashes could be better.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def view_menu
+    @dishes.map do |dish, price|
+      "#{dish}: £#{('%.2f' % price)}"
```
stupot1, line 13: This could be refactored onto one line of code.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+  end
+
+  def dish_available?(chosen_dish)
+    unless @dishes.include?(chosen_dish) == false
+      true
+    end
```
stupot1, line 20: An IF statement could be better here for this functionality.

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+
+class Order
+  attr_reader :basket, :dishes
+
+  def initialize
+    @basket = []
+    @dishes = {
+      "Margherita" => 5.00,
+      "Pepperoni" => 8.00
+    }
```
stupot1, line 11: In order initialize, you are creating the menu again.  All you need to do here is create @dishes as an instance of menu and you could access the menu through the attribute reader.

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def add_to_basket(dish, qty = 1)
+    qty.times { @basket << dish }
```
stupot1, line 15: This is a good use of the times method to add the dish to the basket.

lib/order.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  # incomplete method - test does not pass
+  def total_price
+    @basket.map { |x| (x * dishes.values).sum }
```
stupot1, line 20: You could have used the `.sum` method for this.

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'order'
+require 'menu'
+
+describe Order do
+  subject(:order) { Order.new }
```
stupot1, line 5: Not needed line as subject will always be an instance of Order

lib/order.rb
Dhara-95: Thanks, have amended this to '@dishes = Menu.new' so that it calls the menu from the Menu class file.

spec/menu_spec.rb
Dhara-95: Amended now.

lib/menu.rb
Dhara-95: Amended now.

PR #1690, opened 2020-06-21
Chinese takeaway version 1

mattybwoy:


spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
+require './lib/order'
+require './lib/takeaway'
+
+describe Order do
+  subject(:order) {Order.new}
+  let(:dish1) { double :'Chow mein', dish: 'chow mein', price: '3' }
```
JoshuaNg2332, line 6: Great use of double but can also be used in a let statment to do pricing for multiple lines. Helps keep code DRY.

lib/order.rb
```diff
@@ -0,0 +1,19 @@
...
+  def initialize
+    @takeaway = Takeaway.new
+    @orderlist = []
+  end
+
+  def add(dish, price)
+    @orderlist << {dish: price}
+  end
+
+  def total
```
JoshuaNg2332, line 15: Super clean here! Keeping the total sum of dishes to 1 line is awesome!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  describe "takeaway creation" do
+    it "checks for instance of takeaway" do
+      expect(subject).to be_an_instance_of(Takeaway)
+    end
+  end
+  describe "display_menu" do
+    it "displays the menu" do
+      #menulist = takeaway.menu
+      menu = takeaway.menu
+      expect(subject.display_menu).to eq menu
```
JoshuaNg2332, line 17: This is awesome! Just needs to link to the double for it to be able to display the menu.

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    describe "add items" do
+        takeaway = Takeaway.new
+        it { is_expected.to respond_to(:add).with(2).argument }
+        #expect(subject.add).to eq"chicken_chow_mein"
+    end
+    describe "total of order" do
+      it "calculates the sum of the order" do
+        subject.add(dish1, 3)
+        #how to use doubles for adding arguments
+        subject.add(dish2, 2)
```
JoshuaNg2332, line 23: Once you have created a hash with the prices of the meals, you can map through the hash and pull through the prices and sum that here. This is just an alternative way to the way you've already done this.

PR #1695, opened 2020-06-21
Alessa Takeaway challenge pull request

alessa-lou:
Got to halfway through user story 2

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,28 @@
+class Menu
+    
+  attr_reader :full_menu
+    
+  def initialize
+    @full_menu = { 
+        "1. cheeseburger" => "£5",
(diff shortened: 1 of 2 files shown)


katieljones:
I like how Alessa has started this challenge. It's very clear in the code what it is doing. It considers edge cases and what to do if a user picks a number that isn't available. I like the UI and the response she's included after each item is selected. The only thing I can think considering is in the tests, where each numbered dish is chosen, could there be one test that allows the customer to choose a dish as opposed to writing for all 6? Not sure myself how to write this but could definitely save some time writing them all if there is a way. 
`expect(subject.choose_dishes(3)).to eq("You have chosen {\"3. veggie burger\"=>\"£6\"}, thank you for ordering")`

PR #1696, opened 2020-06-21
KJ

katieljones:


alessa-lou:
Katie's tests are set out really well so it is easy to understand what is being tested from her test 'it' statements and their contents. Her method names are also really clear and well named.

PR #1700, opened 2020-06-22
Takeaway Challenge - Josh

JoshuaNg2332:
- Completed user stories 1 and 2.
- Looked into user story 3 but didn't get a chance to start it.
- I would like to improve my time taking to start a project and mapping it. This took the most time but once I got started, I was able to make a decent amount of progress.
- Larger chunks did take me longer to solve such as understanding the injection side of things.


mattybwoy:
Very solid effort on a tough weekend challenge. Good use of doubles, SRP and abstraction. Only improvement I can think of would be not to write out the menu again in the menu_spec file to pass the test.

PR #1702, opened 2020-07-19
Ralph Mallett - takeaway challenge

ralphm10:
- All user stories met 
- Able to send sms confirmation but no unit tests for this
- defends against some edge cases
- need to add doubles/stubs into current tests


lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def place_order(sms = Sms.new)
+
+    fail "Basket empty" if @order_items.empty?
+    order_message
+    sms.send_sms
+  end
+
+  def order_message(time = Time.new)
```
DillonBarker, line 50: It seems a little confusing that this method is within the order class as it is all to do with the sms class and doesnt use any of the functionality from the order class.

lib/time.rb
```diff
@@ -0,0 +1,7 @@
+class Time
```
DillonBarker, line 1: I like how you have separated this out, it makes it more readable in the code where you are using it!

spec/order_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'order'
+
+describe Order do
```
DillonBarker, line 3: To be even clearer, you could separate out the tests with more describe blocks.
`describe '#add_to_order' do` 
Would just make it easier to read and maybe if you came back to it, you could find a troubleshoot all the tests for one method easier.

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+  attr_reader :order_items, :order_total, :menu, :subtotal, :running_total
+
+  def initialize(menu = Takeaway.new)
+    @menu = menu
+    @order_items = []
+    @order_total = []
+    @running_total = 0
+  end
+
+  def add_to_order(item, quantity)
```
DillonBarker, line 16: Really cool method, is there any way of breaking it down further ?

README.md
```diff
@@ -47,37 +39,52 @@ So that I am reassured that my order will be delivered on time
...
-  * Implement the ability to place orders via text message.
-
-* A free account on Twilio will only allow you to send texts to "verified" numbers. Use your mobile phone number, don't worry about the customer's mobile phone.
+Progress
+---------
 
-> :warning: **WARNING:** think twice before you push your **mobile number** or **Twilio API Key** to a public space like GitHub :eyes:
->
-> :key: Now is a great time to think about security and how you can keep your private information secret. You might want to explore environment variables.
+My code satisfies all of the above user stories and does defend against edge cases such as displaying a basket or order total when no items have been added, attempting to place an order with no items, and attempting to add an item not on the menu.
```
02ship, line 45: Very clear on the edge cases you've tested, makes it easy to spot if any have been missed, great work!

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def display_order
+    fail "Basket empty" if @order_items.empty?
+
+    "Current order: #{order_items.uniq.map { |x| "#{order_items.count(x)}x #{x}" }.join(', ')}"
+  end
+
+  def display_total
+    fail "Incorrect total" if order_total.sum != @running_total
+    fail "Basket empty" if @order_items.empty?
```
02ship, line 38: Lines 31 and 38 are identical, any way to separate out this functionality and adhere to DRY?

PR #1710, opened 2020-07-20
Dillon Barker - Takeaway Challenge - July 2020

DillonBarker:
I made good progress through the user stories but was stumped when rspec and dotenv were not playing ball together. When I tried to create a confirmation_spec.rb and run a test it would say: 
`An error occurred while loading ./spec/confirmation_spec.rb. - Did you mean?
                    rspec ./spec/order_spec.rb

Failure/Error: require 'dotenv'

LoadError:
  cannot load such file -- dotenv`


lib/menu.rb
```diff
@@ -0,0 +1,24 @@
+class Menu
+
+  attr_reader :meals
+
+  def initialize
+    @meals = {
```
harrylb14, line 6: I think this may be easier as a Hash of Arrays, with Strings as key values

lib/confirmation.rb
```diff
@@ -0,0 +1,29 @@
+require 'twilio-ruby'
+require 'dotenv'
+
+
```
harrylb14, line 4: are these lines at all necessary mate, bloody hell

lib/menu.rb
DillonBarker: Yes I think so too, I will make the changes later on.

PR #1714, opened 2020-07-20
takeaway challenge completed

melanieanderson1995:


spec/order_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "order"
+
```
phump81, line 2: As you are using 'order = Order.new' in each test, you could perhaps declare it at the beginning of the test as a 'let' statement.

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+      @choices
+    else
+      "Not a menu option"
+    end
+  end
+
+  def verify
+    sum = 0
+    @choices.each do |choice|
+      sum += choice[1]
```
phump81, line 27: I like the use of indexing to get the price of an item.

lib/menu.rb
```diff
@@ -0,0 +1,13 @@
+require_relative "order"
+
```
phump81, line 2: Good idea to have menu as a separate class. Maybe it could have had an option to inititalize with different menus.

spec/order_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      expect(order.verify).to eq "Your current total is 17."
+    end
+  end
+
+  it "returns thank you for ordering text" do
+    order = Order.new
+    order.choose("Lasagna")
+    expect(order.submit).to eq "Thanks for ordering! Your order is on its way!"
+  end
+end
```
phump81, line 25: All variable and method names in both tests and classes are well named, and your code is tidy, and easy to understand.

PR #1717, opened 2020-08-16
Completed takeaway challenge

imogen-k:


fraserbrookhouse:
Where are my tests? I need my tests! TESTS! Seriously though, it is a bit lacking in test coverage. I'm sure if you practice doubling/stubbing some more you will easily be able to improve this.

fraserbrookhouse:
I see you haven't added your own README, this would be a good thing to get in the habit of doing to properly document the functionality of your code.

fraserbrookhouse:
`if price_confirm != @total_rounded
      puts "Wrong amount, please try again."
      order_price
    else
      return send_message
    end`

Instead of a conditional as above you could have written a guard clause like this:

`return send_message unless price_confirm != @total_rounded`

It is a bit neater, however both will work.

fraserbrookhouse:
I like that you have made this a bit more fully featured by taking in user input, as if the program is actually dealing with a real customer.

PR #1718, opened 2020-08-16
Submitting Takeaway Challenge

JR-G:


lib/takeaway_kitchen.rb
```diff
@@ -0,0 +1,19 @@
+class TakeawayKitchen
+
```
Charlotte-91, line 2: I like how the menu is in a separate class. It makes everything clean and tidy when looking at the ordersystem file.

lib/ordersystem.rb
```diff
@@ -0,0 +1,60 @@
+require_relative 'takeaway_kitchen'
+require_relative 'text'
+
```
Charlotte-91, line 3: I love how clean your code is. Every method is named well and it's easy to understand what each of them do.

lib/ordersystem.rb
```diff
@@ -0,0 +1,60 @@
...
+    menu
+  end
+
+  def add_to_order(dish, quantity = 1)
+    fail "This isn't on the menu" if in_menu? dish
+
+    order << { dish: dish, quantity: quantity, price: calculate_price(dish, quantity) }
+  end
+
+  def remove_from_order(dish, _quantity = 1)
```
Charlotte-91, line 22: Love this method! :D

README.md
```diff
@@ -1,4 +1,4 @@
-Takeaway Challenge
+Makers End of Unit Challenge - Takeaway Challenge
 ==================
 ```
```
Charlotte-91, line 3: Readme is nice and detailed. Love how you included the irb tests. This has definitely inspired me, hoping I can write such in-depth readme files in the future :)

lib/takeaway_kitchen.rb
```diff
@@ -0,0 +1,19 @@
...
+      { dish: "Pad Thai", price: 7 },
+      { dish: "Panang Curry", price: 7 },
+      { dish: "Cashew Stir-fry", price: 7 },
+      { dish: "Steamed Rice", price: 2 },
+      { dish: "Egg Fried Rice", price: 2 },
+    ]
+  end
+
+  def menu
+    @menu.dup
```
alicelieutier, line 17: Why do you duplicate menu here? This feels quite inefficient. Why not just have an `attr_reader :menu` instead?

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
+require 'ordersystem'
+
+describe OrderSystem do
+  let(:dish) { double :dish }
+  let(:option_one) { { dish: dish, price: 3 } }
+  let(:option_two) { { dish: dish, price: 5 } }
+  let(:takeaway) { TakeawayKitchen.new.menu }
```
alicelieutier, line 7: This is exactly where you should be using a double. This test is for the OrderSystem class, it should not need to use any other class. So here, you probably instead want something like:
```rb
  let(:short_menu) { [
      { dish: "Chicken Satay", price: 3 }, 
      { dish: "Spring Rolls", price: 2 }
  ]}
  let(:takeaway) { double :takeaway, menu: short_menu }
```

Or something similar.

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  
+  it 'returns the menu from TakeawayKitchen' do
+    expect(subject.menu).to eq takeaway
+  end
+
+  it 'initializes with an empty order' do
+    expect(subject.order).to be_empty
+  end
```
alicelieutier, line 15: Don't test the initialize method / initial state of an object.

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+  describe '#view_menu' do
+    it 'returns a hash' do
+      expect(subject.view_menu).to include({ dish: be_an(String), price: be_an(Integer) })
```
alicelieutier, line 19: The purpose of `view_menu` is to print, not to return things. Printing is what you should test for (using the output matcher from rspec)

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  describe '#add_to_order' do
+    it { is_expected.to respond_to(:add_to_order).with(2).argument }
```
alicelieutier, line 24: These tests are not needed once you have tests for the behaviour of the method.

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+
+    it 'increases order count' do
+      subject.menu << option_one
+      subject.add_to_order dish
+      expect(subject.order.count).to eq 1
```
alicelieutier, line 29: This test feels very stateful. You want to test behaviour instead. How would your user test that this method worked? Maybe they looked at the return, or maybe they had to call view menu, or something else to make sure their order had been updated. You want to test this the way the user of the class would test it.

alicelieutier: See the "testing state vs behaviour" week 2 practical.

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  describe '#remove_from_order' do
+    it { is_expected.to respond_to(:remove_from_order).with(2).argument }
+
+    it 'decreases order count' do
+      subject.order << option_one
+      subject.remove_from_order dish
+      expect(subject.order).to be_empty
```
alicelieutier, line 43: A perfect opportunity to use the `empty?` method of your class instead. Same comment as above.

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+
+    it 'raises an error if already empty' do
+      expect { subject.remove_from_order dish }.to raise_error "The order is currently empty"
+    end
+
+    it 'raises an error if it doesnt exist' do
+      subject.order << option_one
```
alicelieutier, line 51: It would be better to use `add_to_order` instead of directly modifying the state.

spec/ordersystem_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  describe '#view_order' do
+    it { is_expected.to respond_to :view_order }
+
+    it 'shows the correct summary' do
+      subject.order << { dish: dish, price: 3, quantity: 1 }
+      subject.order << { dish: dish, price: 5, quantity: 1 }
+      expect { subject.view_order }.to output("#{dish} x 1 = £3\n#{dish} x 1 = £5\n").to_stdout
```
alicelieutier, line 62: You should be testing with fixed inputs and fixed output.
Something like:
```rb
 subject.order << { dish: "Curry", price: 3, quantity: 1 }
 subject.order << { dish: "Pizza", price: 5, quantity: 1 }
 expect { subject.view_order }.to output("Curry x 1 = £3\nPizza x 1 = £5\n").to_stdout
```

Also, you should test in different cases (different quantities, zero, one or two items in order, etc..)

alicelieutier: Each method should have a few tests.

spec/takeaway_kitchen_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'takeaway_kitchen'
+
+describe TakeawayKitchen do
+
+  it 'has a menu' do
+    expect(subject.menu).not_to be_empty
+  end
+end
```
alicelieutier, line 8: This class is pure state no behaviour. I would not have tests for it at all.

spec/text_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  it 'sends a text' do
+    fake_message = {
+      from: '123',
+      to: '456',
+      body: 'testing my message'
+    }
+
+    expect(subject).to receive(:send_text).with(fake_message)
+    subject.send_text(fake_message)
```
alicelieutier, line 14: You are not testing anything of the code here.
As a rule, you should use a to_receive matcher on the subject like this:
```
expect(subject).to receive
```
or 
```
allow(subject).to receive
```
a to_receive matcher is used for interaction of the class under test **with other objects**.

spec/text_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  it 'has a delivery time of an hour' do
+    allow(Time).to receive(:now).and_return(Time.parse("19:00"))
```
alicelieutier, line 18: This one is better -> Stubbing a class/object that is not the subject. 
In this case, you can achieve similar results using a gem called `timecop`

spec/text_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+    expect(subject.message_body).to eq "Thank you! Your order was placed and will be delivered before 20:00"
+  end
+
+  it 'contains a "from" and "to" number' do
+    expect(subject.customer_message).to include({ from: be_an(String), to: be_an(String), body: be_an(String) })
```
alicelieutier, line 24: Alway test against specific values.

PR #1719, opened 2020-08-16
Submitting takeaway challenege

Charlotte-91:
- Did not get to implement the text messaging feature.
- 95.92% test coverage

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,43 @@
+class Order
+
+  def initialize 
+    @menu = @menu = {
+      "Sushi set" => 12,
+      "Prawn katsu curry" => 9,
+      "Kitsune udon" => 7,
+      "Onigiri" => 3,
+      "Matcha ice-cream" => 4
+    }
+    @total = 0
+    @order = []
+  end
+
+  def print_menu
+    @menu.each { |key, value| puts " #{key}: £#{value}" }
+  end
+
+  def user_order(food)
+    # puts "Please enter the names of the items you would like to order.\nPlease type 'done' when you've finished."
+      if @menu.include?(food) == true
+         @order.push(food)
+         print @order
+      elsif food == 'done'
+        return order_total
+      elsif @menu.include?(food) == false
+        puts "Item not on menu"
+        return
+      end
+  end
+
+  def current_order
+    print @order
+  end
+
+  def order_total
+    @order.each do |item|
+      price = @menu.fetch(item)
+      @total += price
+    end
+    return @total
+  end
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -2,9 +2,9 @@ require_relative 'text'
 
 class Order
 
-  attr_accessor :menu ,:text
+  attr_accessor :menu, :text
 
-  def initialize (text = Text_message.new)
+  def initialize(text = Text_message.new)
     @menu = @menu = {
       "Sushi set" => 12,
       "Prawn katsu curry" => 9,
@@ -22,17 +22,16 @@ class Order
   end
 
   def user_order(food)
-    # puts "Please enter the names of the items you would like to order.\nPlease type 'done' when you've finished."
-      if @menu.include?(food) == true
-         @order.push(food)
-         print @order
-      elsif food == 'done'
-        puts  "Order total £#{order_total}"
-        @text.send_text
-      elsif @menu.include?(food) == false
-        puts "Item not on menu"
-        return
-      end
+    if @menu.include?(food) == true
+      @order.push(food)
+      print @order
+    elsif food == 'done'
+      puts "Order total £#{order_total}"
+      @text.send_text
+    elsif @menu.include?(food) == false
+      puts "Item not on menu"
+      return
+    end
   end
 
   def order_total
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -29,10 +29,6 @@ class Order
       end
   end
 
-  def current_order
-    print @order
-  end
-
   def order_total
     @order.each do |item|
       price = @menu.fetch(item)
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,6 +1,10 @@
+require_relative 'text'
+
 class Order
 
-  def initialize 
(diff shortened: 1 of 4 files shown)


lib/order.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'text'
+
+class Order
+
+  attr_accessor :menu, :text
+
+  def initialize(text = Text_message.new)
```
JR-G, line 7: Good use of Dependency Injection

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    elsif @menu.include?(food) == false
+      puts "Item not on menu"
+      return
+    end
+  end
+
+  def order_total
+    @order.each do |item|
+      price = @menu.fetch(item)
+      @total += price
```
JR-G, line 40: Could the 'reduce' method be used here?

spec/order_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'order'
+
+describe Order do 
+
+  it 'prints the menu' do
+    expect(subject.print_menu).to include("Sushi set" => 12)  
+  end
+
+  it 'tells user that an item is not on the menu' do
+    expect { subject.user_order("cake") }.to output("Item not on menu\n").to_stdout
```
JR-G, line 10: Nice test 🙂

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+      "Matcha ice-cream" => 4
+    }
+    @total = 0
+    @order = []
+    @text = text
+  end
+
+  def print_menu
+    @menu.each { |key, value| puts " #{key}: £#{value}" }
+  end
```
JR-G, line 22: I like this method!

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+  def user_order(food)
+    if @menu.include?(food) == true
+      @order.push(food)
+      print @order
+    elsif food == 'done'
+      puts "Order total £#{order_total}"
+      @text.send_text
+    elsif @menu.include?(food) == false
+      puts "Item not on menu"
+      return
```
JR-G, line 33: For any future refactoring, maybe it's worth seeing if any of this can be moved into separate methods?

lib/text.rb
```diff
@@ -0,0 +1,12 @@
+class Text_message
```
JR-G, line 1: I think this is a great alternative to the Twilio API and provides the functionality on producing a message with the delivery time

spec/text_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'text'
+
+describe Text_message do
+  it 'gets the current time and adds an hour' do
+    time_now = Time.now
+    allow(Time).to receive(:now).and_return(time_now += 1 * 60 * 60)
+    expect(subject.delivery_time).to eq (time_now += 1 * 60 * 60)
```
JR-G, line 7: I like how this test is structured!

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def initialize(text = Text_message.new)
+    @menu = @menu = {
+      "Sushi set" => 12,
+      "Prawn katsu curry" => 9,
+      "Kitsune udon" => 7,
+      "Onigiri" => 3,
+      "Matcha ice-cream" => 4
+    }
+    @total = 0
```
tatsiana-makers, line 15: Does this need to be an instance variable?

Instance variable should only be used when you need to store the state of an object as a whole, and that state has the same shelf life as the object.
If you just need to store information for a short amount of time, use a local variable. If you need to pass information between a method and another method called inside it, use an argument. Using an instance variable that is meant to be “short lived”, just to be used in a single method and none after is a source of bugs.

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+    @order = []
+    @text = text
+  end
+
+  def print_menu
+    @menu.each { |key, value| puts " #{key}: £#{value}" }
+  end
+
+  def user_order(food)
+    if @menu.include?(food) == true
```
tatsiana-makers, line 25: When you have a conditional which looks like:
` if something == true` can be refactored to
`if something`

and 
`if something == false` can be refactored to
`if !something`

lib/order.rb
```diff
@@ -0,0 +1,44 @@
...
+      @order.push(food)
+      print @order
+    elsif food == 'done'
```
tatsiana-makers, line 28: It looks like this method is responsible mainly for adding an item of food to the order 

This branch of the if statement looks like a separate piece of logic responsible for printing the order total and sending the text. We can use this separation of concerns as a great reason to put this logic into a different method.

The way I try to think about this (when it comes to both methods and classes):
- identify the main purpose of the method/class
- check that each part of the code is only implementing that main purpose, and if it isn't move that logic into another method/class

lib/order.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'text'
+
+class Order
+
+  attr_accessor :menu, :text
+
+  def initialize(text = Text_message.new)
+    @menu = @menu = {
```
tatsiana-makers, line 8: Do we need `@menu =` repeated here?

spec/order_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'order'
+
+describe Order do
```
tatsiana-makers, line 3: We should unit test every public method defined in a class. Are there some methods which are missing tests?

spec/order_spec.rb
tatsiana-makers, line 10: Nice test for this branch of the `user_order` method! To achieve even better test coverage, we'll need to test all possible branches of this method 🙂

spec/text_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  end
+
+  it 'sends a text to the user with the expected time' do
+    expect(subject.send_text).to include("Your order was placed")
```
tatsiana-makers, line 11: If the text message only said "Your order was placed", would this fulfil the requirements of the method? Is there a stronger assertion we can make here?

PR #1720, opened 2020-08-16
Tam's Takeaway Challenge

just-tam:
Not quite complete still need to add some functionality and tests.


spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    expect(subject).to be_an_instance_of(Order)
+  end
+
+  it "Expects order to respond to add_to_order" do
+    expect(subject).to respond_to(:add_to_order)
+  end
+
+  it "Expects order to respond to total_of_order" do
+    expect(subject).to respond_to(:order_total)
+  end
```
ZishJawaid, line 14: I really like your use of these unit tests to show how the program has been built up. 

As a refactoring suggestion, I think it is possible to delete them after you have written further tests, as my understanding is that they are implicitly tested by the subsequent tests. i.e. line 9 should be implicitly tested by line 20

lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+    @order = []
+    @total = 0
+    @text = Text.new
+  end
+
+  def add_to_order(item, quantity=1, price)
+    @order << {:item => item, :quantity => quantity, :price => price}
+  end
+
+  def order_total
```
ZishJawaid, line 18: I really like how you've implemented this method. It's really clear to follow.

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'order'
+
+class Menu
```
ZishJawaid, line 3: I was thinking that it might be possible to create a new instance of the menu class when you create a new order (in other words, we could inject Menu.new in to the order class). Just like you've done for @text = Text.new in the order class. 

This should mean that when we create a new order, it will have direct access to the menu. This will allow us to display the menu as a hash in the menu class, where the menu hash only contains dishes as the key and prices as the values (no mention of quantities). And then when we come to placing the order, we can view the menu items and prices, and then just add our chosen dishes and then select the quantity, which will be added to the order hash.  This should make the program run a bit more like it would in real life.

PR #1722, opened 2020-08-17
takeaway-challenge

Djura22:


Josh-Steadman:
Good use of the SRP

Josh-Steadman:
Nice variety of rspec tests used

Josh-Steadman:
not enough use of mocks and stubs

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+require_relative 'takeaway'
+
+class Order
+
+  attr_accessor :current_order
+  attr_reader :menu, :receipt, :basket
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @select = {}
```
tatsiana-makers, line 11: Does this need to be an instance variable?

Instance variable should only be used when you need to store the state of an object as a whole, and that state has the same shelf life as the object.
If you just need to store information for a short amount of time, use a local variable. If you need to pass information between a method and another method called inside it, use an argument. Using an instance variable that is meant to be “short lived”, just to be used in a single method and none after is a source of bugs.

Are there other instance variables in this method that don't need to be instance variables?

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'menu'
+
+describe Menu do
+
+  let(:menu) { double :menu }
```
tatsiana-makers, line 5: It doesn't look like we use this variable- do we need it?

Are there other places in the tests where we can make a similar change?

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  it 'responds to method of available dishes' do
+    expect(subject).to respond_to(:avail_dishes)
+  end
+
+  describe 'dishes' do
+
+    it 'includes salmon in dish list' do
+      expect(subject.avail_dishes).to have_key(:salmon)
+    end
+
```
tatsiana-makers, line 21: We should unit test all public methods in a class - can you think of a test for the `customer_view` method?

spec/order_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  it 'responds to print_basket method' do
+    expect(subject).to respond_to(:print_basket)
+  end
+
+  describe 'print_basket' do
+
+    it 'prints total with basket' do
+      expect(subject.print_basket).to include("Total to pay:")
```
tatsiana-makers, line 54: What if `print_basket` only returns "Total to pay:" - would this be sufficient? 


Is there a stronger assertion we can make? (this question applies to several other assertions in your unit tests)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'takeaway'
+
+describe Takeaway do
+
+  before do
+    @list = Menu.new
```
tatsiana-makers, line 6: Here we're using the actual `Menu` class which means the tests aren't isolated. Dependency injection could help us isolate the classes here: https://github.com/makersacademy/skills-workshops/blob/f5b4801840fe07d26ff70341652dc81dcda12289/practicals/object_oriented_design/dependency_injection.md

PR #1724, opened 2020-08-17
Takeaway-challenge

Josh-Steadman:
unable to implement the send text functionality


Djura22:
Great use of stubs in test code.

Djura22:
Good clean and simple methods, shows clearly the functionality of the codebase.

Djura22:
Perhaps more of the methods could be contained within the order class which are related to it's functionality.

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+
+class Menu
+    attr_reader :menu
```
alicelieutier, line 3: No need for this attribute reader given that the view method below is doing exactly the same thing.

lib/order.rb
```diff
@@ -0,0 +1,17 @@
...
+class Order
+
+   def initialize
+    @total_price = 0
+    
+   end
+
+   def sum(food_list)
+    food_list.each { |hash| hash.each {|food, price| @total_price += price } }
+    @total_price
```
alicelieutier, line 11: `total_price` should not be an instance variable here. A local variable is enough :)The user calls the sum method to get that price.

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+require_relative 'order'
+
+class Takeaway
+    attr_reader :menu, :order, :food_list
+    
+    def initialize(menu= Menu.new)
+        @menu = menu
+        @total_price = 0
+        @food_list = []
+
```
alicelieutier, line 11: It's good practice to initialize the instance variable you use in the class, for readability. So here, you could add a line saying `@order = nil`

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+    attr_reader :menu, :order, :food_list
+    
+    def initialize(menu= Menu.new)
+        @menu = menu
+        @total_price = 0
+        @food_list = []
+        
+    end
+
+    def view
```
alicelieutier, line 14: The name of your methods are not super transparent. Maybe `view_menu` would be better here? Or `get_menu`? (view usually implies you do some formatting for the user that you are not doing here).

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+        @food_list
+    end
+
+    def basket
+      @amount = @order.sum(@food_list) 
+      @amount
+      
+     end
+
+    def new_order(order= Order.new)
```
alicelieutier, line 30: Is `new_order` supposed to be used directly by the user, or should it be private?

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+        new_order
+        number.times { food_list.push(food => @menu.menu[food])}
+        @food_list
+    end
+
+    def basket
+      @amount = @order.sum(@food_list) 
+      @amount
+      
+     end
```
alicelieutier, line 28: Don't create instance variables when not needed. Instance variables are only needed when state is shared between methods of an object. Here, this could be re-written as:
```rb
def basket
      @order.sum(@food_list) 
end
```

Also - basket is again not a very descriptive name. What is this method actually doing? Maybe `order_total` or `total_order_amount` would be better?

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+    def check_sum
+       fail 'incorrect sum' if @amount != add
+       true
+    end
+
+     private
+     def add
+        @food_list.each { |hash| hash.each {|food, price| @total_price += price } }
+        @total_price
+     end
```
alicelieutier, line 43: You should not need these two methods. If the only way to get the amount is to call the `basket` method above, that should be enough.

spec/menu_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'menu'
+
+describe Menu do
+  it 'creates an instance of Menu' do
+    expect(subject).to be_an_instance_of(Menu)   
+  end
+
+  it { is_expected.to respond_to(:view) }
```
alicelieutier, line 8: No need for these tests - you want to only test behaviour (the test below is good)

spec/order_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'order'
+
+describe Order do
+
+it {is_expected.to respond_to(:sum).with(1).argument}
```
alicelieutier, line 5: No need for this test once you have the test below to test the behaviour of sum

spec/order_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  describe '#sum' do
+    it 'returns a sum of cost of items recieved' do
+      expect(subject.sum([{:Burger => 5.00}, {:Burger => 5.00}, {:Fish => 6.00}])).to eq 16
+    end
```
alicelieutier, line 10: It would be nice to have several tests for this. Otherwise, with just this one test, your method sum could return a hardcoded 16 and still work. Test with an empty food list, and at list two different non-empty food lists.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  
+it { is_expected.to respond_to(:check_sum) }
+
+it { is_expected.to respond_to(:basket) }
+
+  describe '#initialize' do
+    it 'creates a menu' do
+    expect(subject.menu).to eq menu
+    end
+  end
```
alicelieutier, line 19: No need for these.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+ describe '#new_order' do
+   
+   it 'creates a new order' do
+    allow(subject).to receive(:new_order).and_return first_order
```
alicelieutier, line 24: Never stub methods on the subject. You only want to mock external dependencies (other classes, the system, etc).

alicelieutier: Here, your `new_order` method is set-up for dependency injection already, so you can easily do something like 
```
subject.new_order(first_order)
```
to achieve the same thing.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    allow(menu).to receive(:menu).and_return({"Burger" => 5.00, "Pizza" => 4.00, "Fish" => 6.00, "Chicken" => 5.00})
+    allow(menu).to receive(:view).and_return("Burger" => 5.00, "Pizza" => 4.00, "Fish" => 6.00, "Chicken" => 5.00)
```
alicelieutier, line 26: The fact that you have these two lines makes it clear to me that view and menu do not both need to exist

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+    
+    subject.pick('Chicken', 1)
+    expect(subject.new_order).to eq first_order
```
alicelieutier, line 29: Your expectation here is only testing that your mock is working. It's not ever running your code.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+   end
+
+ 
+  end
+
+  describe '#check_sum' do
+    it 'raises an error if sum is incorrect' do
+        allow(subject).to receive(:add).and_return 11
```
alicelieutier, line 37: No.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
+require 'takeaway'
+
+describe Takeaway do
+let(:menu) { double :menu }
+let(:first_order) {double :first_order }
+subject {described_class.new(menu)}
```
alicelieutier, line 6: This is good.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+  describe '#basket' do
+    it 'returns total cost of order' do 
+        allow(menu).to receive(:menu).and_return({"Burger" => 5.00, "Pizza" => 4.00, "Fish" => 6.00, "Chicken" => 5.00}) 
+        subject.pick('Chicken', 2)
+        subject.pick('Fish', 1)
+        subject.pick('Pizza', 1)
+        expect(subject.basket).to eq 20
+    end
+  end
```
alicelieutier, line 70: These three tests are great, and use mocking exactly how it should be used! We want more of this 🎉

alicelieutier (commented):
Reading your code, I feel like you have understood mocking, but you have also picked up a few wrong habits. You have used it perfectly in some cases, but are also abusing it.

In terms of code, it would be great to spend some more time cleaning it/refactoring for readbility, and simplicity. There are a number of non-essential pieces of code + tests.

PR #1726, opened 2020-08-17
xx

kristycer:


lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
+
+class Takeaway 
+attr_reader :menu, :payment
+
+    def initialize
+        @menu = { 'Salad'=>5, 'Pizza'=>8, 'Pasta'=>6 }
```
tatsiana-makers, line 6: Similar to how we extracted out a Station class from the journeys in last week's Oystercard pairing challenge, can you see a potential class we might extract out of here?

lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
+
+class Takeaway
```
tatsiana-makers, line 2: Currently this class is fulfilling more than one responsibility:
- storing and displaying the menu
- keeping track of an order

Can you see a way we can refactor so that we have multiple classes which each follow SRP?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+
+# As a customer
+# So that I can verify that my order is correct
+# I would like to check that the total I have been given matches the sum of the various dishes in my order
+
+# As a customer
+# So that I am reassured that my order will be delivered on time
+# I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
+
+describe Takeaway do
```
tatsiana-makers, line 19: Nice tests!

PR #1727, opened 2020-09-10
Take-away challenge VJ

vikjusko:


samcolson4:
Macro comments:
- All code has a clear purpose, classes are well defined. 
- Run rubocop to clean up the code a bit. There is some inconsistency in minor things like indentation, spacing.
- Good test coverage.

Menu:
- Good ‘print’ method to make it clear on view.
- Good use of floats!

Order
- Instead of using another hash to collect the dishes, would it have been more flexible to have to different class variables to collect the bill and the food you want?
- Great use of inject(:+) for the total. A codewars classic.

Restaurant
- Good dependency injection in initialiaze - seems flexible. Perhaps add default values to create your objects there.
- Confirm order could be simplified if the dishes were split into two class variables.

PR #1729, opened 2020-09-11
Takeaway challenge - Johnny Yip

JYip93:


lib/order.rb
```diff
@@ -0,0 +1,57 @@
...
+
+
+private
+
+    def check #iterate through dishes_ordered and get subtotal, then sum all the subtotal and compare with total
+        all_subtotal =0
+            @dishes_ordered.each do |item|
+            all_subtotal +=  item.sub_total
+        end
+    all_subtotal == @total ? true : false
```
cfujiname, line 48: NIce succint method

spec/order_item_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'order_item'
+
+
+
+describe OrderItem do
+    let(:dish_id)   {1}
+    let(:quantity)   {3}
+    let(:sub_total)   {9}
+    let(:order_item_double) {double(:dish_id => 1, :quantity=>2, :sub_total=>9 )}
```
cfujiname, line 9: Maybe using rubocop to make code look standardised

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'menu'
+
+describe Menu do
```
cfujiname, line 3: Create a subject here to pass to the tests

spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    let(:menu){Menu.new}
+
+
```
cfujiname, line 6: before do Menu.new so you don't repeat the context

lib/order.rb
```diff
@@ -0,0 +1,57 @@
...
+            sub_total = menu.get_price(dish_id) * quantity
+            item = @order_item.new(dish_id, quantity, sub_total)
+            @total += sub_total
+            @dishes_ordered.push(item)
+        else
+            raise "Dish doesn't exist"
+        end
+    end
+
+    def place_order
```
cfujiname, line 31: Good logic, I missed this step :P

lib/order.rb
```diff
@@ -0,0 +1,57 @@
...
+        @text_message = text_message
+        @order_item = order_item
+    end
+
+    def display_menu
+        menu.display
+    end
+
+    def add(dish_id, quantity)
+        if menu.price_exist(dish_id)
```
cfujiname, line 21: Easy to understand the logic, maybe in the future when we are more pros, we could refactor and try to make the method be less lengthy

PR #1732, opened 2020-09-11
added sequence diagram

TheDancingClown:


prycey77:
Nice! Clear easy to read code. Didn't check rubocop. More test doubles and mocks would be good. README file could do with updating.

PR #1734, opened 2020-09-12
Takeaway challenge

olliesmith3:


olliesmith3:
The code includes features for all user stories as well as a remove_item feature.
Things I would like to improve: further testing on the text class, testing on the order class for the place order method by using doubles of text_class and text, be able to send the message by creating a new instance of the text class.

spec/menu_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'menu'
+
+describe Menu do
```
vikjusko, line 3: You could define all of the menu options as constant in your menu class in order to avoid listing everything in your #print test.

lib/order.rb
```diff
@@ -0,0 +1,93 @@
...
+    end
+  end
+
+  def increase_total(item, quantity)
+    @menu.menu_list.each do |dish|
+      @total += (dish[:price] * quantity) if item == dish[:name]
+    end
+  end
+
+  def decrease_total(item, quantity)
```
vikjusko, line 73: Maybe the decrease_total method should state "Your basket is empty" or another statement rather than take the total amount to negative?

PR #1737, opened 2020-09-14
I am particularly unhappy with this.

obean:


lib/order.rb
```diff
@@ -0,0 +1,21 @@
...
+class Order
+  attr_reader :order_price, :valid_item
+  def initialize
+    @menu = Menu.new
+   # @basket = Basket.new
+    @basket = Basket.new
+    @order_price = 0
+  end
+
+  def select(*items)
```
Catzkorn, line 11: Think about edge cases here. E.g, if the user input a valid dish, but used Capitals. Possibly use .downcase on the parameter to account for this?

lib/order.rb
```diff
@@ -0,0 +1,21 @@
...
+    fail "Please select at least one dish" if items.empty?
+    items.each do |item|
+    @menu.food_items.each { |dish_hash| @basket.current_order << dish_hash if dish_hash[:food] == item }
```
Catzkorn, line 14: Getting a private method error when trying to run select in IRB. Is this due to not doing 'require 'basket'' at the top of the file?

lib/basket.rb
```diff
@@ -0,0 +1,38 @@
+require 'send_sms'
```
Catzkorn, line 1: required this file, but it doesn't exsist. This throws the error 'LoadError (cannot load such file -- send_sms)' when trying to require basket in rb.

lib/order.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/menu'
+class Order
+  attr_reader :order_price, :valid_item
+  def initialize
```
Catzkorn, line 4: Could use dependency injection here as to not hardcode the order class to menu and basket.

spec/basket_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'basket'
+require 'timecop'
+describe Basket do
+  it "contains the current order array" do
+    expect(subject.instance_variable_get(:@current_order)).to be_an_instance_of(Array)
+  end
+  
+  describe "#price" do
```
Catzkorn, line 8: Good use of #methods throughout tests, but I think this should be #price_total.

PR #1738, opened 2020-09-14
Takeaway challenge- partial solution

CRTW26:
Completed user stories 1,2,4
3 and bonus not completed.


lib/confirm.rb
```diff
@@ -0,0 +1,20 @@
+require 'rubygems'
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load 
+
+class Text
+  
+  def sms
+
+    account_sid = ENV["TWILIO_ACCOUNT_SID"]
```
veronicavlee89, line 10: Great use of environment config

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'customer'
+
+RSpec.describe Order do 
+
+  let(:order) { Order.new }
+  let(:takeaway) { double('takeaway', :view => {}, :menu => { chicken: 4.99, chips: 1.50 }) }
+  
+  it 'customer can view menu' do 
+    expect(order.view).to be_instance_of Hash
```
veronicavlee89, line 9: This test could expect that the order.view is eq to the result of takeaway.view (with some value returned), to make sure it's not overriding and returning just any hash

spec/order_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end 
+
+  it 'customer can add item to basket' do 
+    expect { order.add_to_basket }.to change { order.basket.count }.by(1)
+  end 
+  
+  it 'customer can select dish and add to basket' do 
+    order.select_dish('chicken', 2, 4.99)
+    expected = [{ 'chicken' => 2 }]
```
veronicavlee89, line 18: I like how you've defined 'expected', tests are clear and easy to follow

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+    @selection = {}
+    @total = 0
+  end 
+
+  def view
+    @takeaway.view
+  end 
+
+  def select_dish(dish, quantity, price)
+    @selection[dish] = quantity
```
veronicavlee89, line 22: 'selection' could potentially be a hash defined inside the method rather than an instance variable, and passed through to add_to_basket(selection). Then you won't need to nil it after each use

PR #1739, opened 2020-09-14
Takeaway Challenge 

cfujiname:
Challenge built with help of extra material


lib/SMS.rb
```diff
@@ -0,0 +1,30 @@
+require 'twilio-ruby'
+class SMS
+
+  TIME_FORMAT = "%H:%M"
+
+  def initialize(config, client: nil)
+    @client = client || Twilio::REST::Client.new(config[:account_sid], config[:auth_token])
+    @config = config
```
JYip93, line 8: config parameters could be refactored with ENV variables for it to be called easier and safer to store since the credentials on your machine.

JYip93, line 8: config parameters could be refactored with ENV variables for it to be called easier and safer to store since the credentials on your machine.

lib/takeaway.rb
```diff
@@ -6,4 +7,36 @@ def select_dishes(selected_dishes)
...
+    @menu = menu
+    @ordering = ordering || Ordering.new(menu) #initialize order from order class, order being nil
+    @sms = sms || SMS.new(config)
+  end
+
+  def print_menu # US1
+    menu.print # calling the menu from the Menu class, menu will handle printing the menu
+  end
+
+  def place_order(list_of_dishes)
```
JYip93, line 25: Nice and compact method for place_order!

lib/takeaway.rb
```diff
@@ -6,4 +7,36 @@ def select_dishes(selected_dishes)
...
+    add_dish(list_of_dishes)
+    sms.deliver_sms
```
JYip93, line 27: Maybe change the sms.deliver_sms and ordering.sum around? Since it will send a text before actually adding the sum of the order

lib/menu.rb
```diff
@@ -15,5 +16,29 @@ def list_of_dishes
...
   def print_menu
     list_of_dishes
+=======
+class Menu
+
+  # create attr_reader to pass method
+  attr_reader :list_of_dishes
+
+  def initialize(list_of_dishes)
+    @list_of_dishes = list_of_dishes
```
JYip93, line 26: Can add the menu inside as an instance variable of Menu instead of inputting it each time

lib/takeaway.rb
JYip93, line 27: To fufill US3, you can add another method that will check the total and subtotal equal each other before sending the text. If they're not equal you can throw an error. You will need to hardcode/double a wrong subtotal/total in your tests for this to fail though

PR #1744, opened 2020-10-11
Takeaway Challenge - Dan

dan-holmes:


lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def initialize(dishes = default_dishes)
+    @dishes = dishes
+  end
+
+  def default_dishes
+    dishes = []
+    dishes << Dish.new("Curry", 8)
+    dishes << Dish.new("Rice", 3)
+  end
```
kacesera, line 12: Could you load default dishes from a 'default' menu potentially? What would happen here if I wanted to add my own dishes into @dishes upon initialization?

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  def print
+    puts "Menu:"
+    @dishes.each { |dish| print_dish(dish) }
+  end
```
kacesera, line 17: Can this be added to the printer class? I'm not entirely sure that it should, but given that all other 'print' methods are, it seems like this should be as well.

lib/texter.rb
```diff
@@ -0,0 +1,19 @@
+require "twilio-ruby"
+
+class Texter
+  def initialize
+    account_sid = ENV["SID"]
+    auth_token = ENV["AUTH_TOKEN"]
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+
+    @from = ENV["TWILIO_NUMBER"]
```
kacesera, line 9: AH! I finally understand this. Spent all morning trying to figure out the implementation, and this makes much more sense!

spec/texter_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "texter"
+
+describe Texter do
+  it { expect(subject).to respond_to(:send_text).with(1).argument }
+  it { expect(subject).to respond_to(:send_text).with(2).argument }
+  # I couldn't work out how to test this class further. Are we meant to mock twilio?
```
kacesera, line 6: I mocked it to test that #create was called on #message, which would then send the text.

README.md
```diff
@@ -1,83 +1,36 @@
...
+## How to run
 
-As a customer
-So that I can order the meal I want
-I would like to be able to select some number of several available dishes
+First set up a Twilio account here: https://www.twilio.com/.
 
-As a customer
-So that I can verify that my order is correct
-I would like to check that the total I have been given matches the sum of the various dishes in my order
```
kacesera, line 43: The user currently doesn't get to verify that their order is correct. I think that would be a great feature to add.

PR #1745, opened 2020-10-11
Takeaway-challenge submission

dm-devtech:


lib/BillCalculator.rb
```diff
@@ -0,0 +1,54 @@
...
+          @total += v * @meal_options.values[1]
+        elsif k == @meal_options.keys[2]
+          @total += v * @meal_options.values[2]
+        elsif k == @meal_options.keys[3]
+          @total += v * @meal_options.values[3]
+        elsif k == @meal_options.keys[4]
+          @total += v * @meal_options.values[4]
+        elsif k == @meal_options.keys[5]
+          @total += v * @meal_options.values[5]
+        end
```
tatsiana-makers, line 25: With hashes, you can look up values by the key, so for example:
```
inventory = {:pencil => 5, :notebook => 6, :envelope => 3}

p inventory[:pencil] // prints out 5
```

Can you see how this hash lookup can be used to simplify this code? Are there other places the code can be refactored in this way to improve readability?

lib/Menu.rb
```diff
@@ -0,0 +1,45 @@
+class Menu
+  attr_reader :meal_options, :dishes, :customer_total, :store_order
+
+  MEAL_OPTIONS = {:Hamburger=>5, :Pizza=>6, :Pasta=>4,:Fish_and_chips=>5,:Salad=>3,:Lamb_Shank=>7}
+
+    def initialize(printer = Print.new, bill = BillCalculator.new(MEAL_OPTIONS))
+      @dishes = dishes
```
tatsiana-makers, line 7: What is the `dishes` on the right hand side of this line of code?

To get some visibility on what is happening, try creating a new Menu instance in `irb` and seeing what the value of `dishes` is

lib/Menu.rb
```diff
@@ -0,0 +1,45 @@
...
+      @bill.calculate_total(@dishes)
+
+    end
+
+    def print_all
+      @printer.print_dishes(MEAL_OPTIONS)
+    end
+
+    def my_total(total)
+      @customer_total = total
```
tatsiana-makers, line 27: Does the `@customer_total` need to be an instance variable in this class or can it be a local variable that is passed to the `total_error` method as an input parameter?

Generally variables should have the smallest scope possible (local scope < instance variable scope < global variable scope). The reason for this is because it's easier to reason about smaller scope and therefore eliminates potential bugs. Local scope is small, and by looking at a single method you can see the entire scope of the variable. An instance variable could be accessed by all the methods of the class and potentially by another class, so there is more to reason about for instance variables.

tatsiana-makers: Are there other variables that don't need to be instance variables in your code?

lib/Menu.rb
```diff
@@ -0,0 +1,45 @@
...
+      puts "Amount given is correct.  Food ordered"
+    end
+
+    def store_order
+      @bill.store_order(@dishes)
+    end
+
+    def print_my_dishes
+      store_order
+      puts @bill.order
```
tatsiana-makers, line 42: This method is doing some of the printing, even though there is a `Printer` class. By moving all the responsibility of printing into that dedicated class, the code might follow SRP more closely. What do you think?

lib/README.md
```diff
@@ -0,0 +1,8 @@
+#### Running the program in IRB
```
tatsiana-makers, line 1: It'd be great to include a sample usage of your code in the readme. This would help an end-user see how they should use the classes and methods you have written.

lib/BillCalculator.rb
```diff
@@ -0,0 +1,54 @@
+class BillCalculator
+  attr_accessor :total, :order
+  attr_writer :meal_options
```
tatsiana-makers, line 3: A part of encapsulation is restricting direct access to an object's components. In order to increase encapsulation, we should limit the attribute readers and writers to only those we definitely need. 

Do we need this `attr_writer`? Are there other accessors we can get rid of without affecting the running of the program?

lib/Menu.rb
```diff
@@ -0,0 +1,45 @@
...
+      raise 'Selection must be entered in the format ":dish=>quantity"' if dishes.class != Hash
+      raise 'Quantity must be an integer e.g. 4' if !dishes.values.any?(Integer)
+      raise 'Dish selected must be a symbol e.g. :Hamburger' if !dishes.keys.any?(Symbol)
+      @dishes = dishes
+      @bill.calculate_total(@dishes)
+
+    end
+
+    def print_all
+      @printer.print_dishes(MEAL_OPTIONS)
```
tatsiana-makers, line 23: Nice use of forwarding here!

tatsiana-makers (commented):
This is a nice go at the Takeaway challenge! Your three class structure makes sense to me. I've left a few comments where I think SRP and encapsulation could be followed a bit more closely.

Let me know if you have any questions about the comments I've left and I'd be happy to chat further about them 🙂

PR #1752, opened 2020-10-11
Takeaway Challenge 

CamLogie:


lib/menu.rb
```diff
@@ -0,0 +1,19 @@
+class Menu
+
+  def initialize
+    @menu = {
```
jotkafomat, line 4: It would be better for future to use to have option to initialise menu with different dishes so, I think its better to use constant here and use it as default value.

lib/menu.rb
```diff
@@ -0,0 +1,19 @@
+class Menu
```
jotkafomat, line 1: Nice class separation!

lib/basket.rb
```diff
@@ -0,0 +1,35 @@
...
+    def basket
+      @basket
+    end
+
+    def order_total
+      @order_total
+    end
+
+    def add_to_basket(item, quantity)
+      if valid_item?(item)
```
jotkafomat, line 20: I like the idea that you're checking if item is valid.
For clarity I would put raise statement first before if statement

lib/basket.rb
```diff
@@ -0,0 +1,35 @@
...
+        @basket << {item => quantity}
+        @order_total += @menu[item] * quantity
+      else raise "Not a valid item, please select again"  
+      end
+    end
+
+    private
```
jotkafomat, line 27: good use of private

lib/order_takeaway.rb
```diff
@@ -0,0 +1,20 @@
+require_relative 'basket'
+require_relative 'print'
+require_relative 'text'
+require_relative 'menu'
+
+class OrderTakeaway
```
jotkafomat: Is it more like class Text - which can send a confirmation?

lib/print.rb
```diff
@@ -0,0 +1,29 @@
+require_relative 'basket'
+
+class Print
```
jotkafomat, line 3: Again, very good class separation!

spec/basket_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    end 
+  end
+
+  describe 'order_total' do 
+    it 'begins with a value of zero' do
+    expect(subject.order_total).to eq 0
+    end
+  end 
+
+  describe 'add_to_basket(item)' do
```
jotkafomat, line 17: Instead of hardcoding your menu item you could create mock let (:menu) {double :menu} and then expect menu.menu to receive ([{"Double double" => 1}])
(I'll check the syntax when I finish review)

spec/basket_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    it 'adds an item and quantity to the basket which is a hash' do
+    subject.add_to_basket("Double Double", 1)
+    expect(subject.basket).to include({"Double Double" => 1})
+    end
+
+    it 'adds an items price to the order total' do
```
jotkafomat, line 24: same as above

spec/menu_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'menu'
+
+describe Menu do
```
jotkafomat, line 3: It tests state rather then behaviour but I can't think of how you could change that one.

spec/print_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    let(:menu) { {"Double Double" => 4,
+        "Cheeseburger" => 3,
+        "Hamburger" => 2,
+        "French Fries" => 1,
+        "Milkshake" => 1,
+        "Small Soda" => 1 }
+    }
+    let(:basket) {Basket.new}
+
+    it '#print_menu(menu) should print a hash to the console' do
```
jotkafomat, line 16: Good use of mocking!

spec/print_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        expect { subject.print_menu(menu) }.to output("Double Double: $4\nCheeseburger: $3\nHamburger: $2\nFrench Fries: $1\nMilkshake: $1\nSmall Soda: $1\n").to_stdout
+    end
+
+  describe 'Printing related to baskets' do
+
+    before do
```
jotkafomat, line 22: I'm not sure but I think you should use allow/expect syntax over here.

PR #1754, opened 2020-10-12
KC - Takeaway Challenge

kacesera:
Had some trouble with the API, particularly protecting the sensitive information. Wound up locally saving the information, and adding information into the README to ensure that the file gets created locally by the user.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Takeaway Challenge
 
  ```
 # How to Use:
-First, load in IRB using ```irb -r './lib/takeaway.rb'``
+First, load in IRB using ```irb -r './lib/takeaway.rb'```
 
 If you would like to use a new menu, add your menu to the 'Menus' directory with the following format:
 
--- a/README.md
+++ b/README.md
@@ -13,21 +13,26 @@ Takeaway Challenge
        ':..:'                ':..:'
 
  ```
+# How to Use:
+First, load in IRB using ```irb -r './lib/takeaway.rb'``
 
-Instructions
--------
+If you would like to use a new menu, add your menu to the 'Menus' directory with the following format:
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+```food,price```
 
-Task
------
+Then, you can start your order using: ```my_order = Takeaway.new('my_file_name.csv')```
+Now, you can begin your order! (If you don't want to add a menu, just type ```my_order = Takeaway.new```)
 
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+To see the menu: ```my_order.see_dishes```
+To order: ```my_order.place_order```
+To verify your order: ```my_order.verify_order```
+
+Once you've verified your order, you will be asked to enter your phone number. From here, you will receive a text confirming your order and expected arrival time.
+
+## Note:
+Please enter in your own Twilio API account SID, authentication token, and phone number to use.
+
+# User Stories:
 
 ```
 As a customer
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -13,21 +13,28 @@ Takeaway Challenge
...
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+```food,price```
 
-Task
------
+Then, you can start your order using: ```my_order = Takeaway.new('my_file_name.csv')```
```
dan-holmes, line 23: This didn't work for me, as you're loading in the csv here instead of a class

lib/customer.rb
```diff
@@ -0,0 +1,39 @@
+require '../config'
```
dan-holmes, line 1: require_relative?

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'menu'
+
+describe Menu do
+  let(:new_menu) { Menu.new }
+  
+  describe '#initialize' do
+    it 'loads the menu' do
+      expect(new_menu.menu).to eq([{ item: "Quiche", price: 3 }, { item: "Mac & Cheese", price: 6 }, { item: "Mashed Potatoes", price: 2 }, { item: "Tea", :price => 1 }, { item: "Tiramisu", price: 5 }])
```
dan-holmes, line 8: You could have made a Dish class, but this works too

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'takeaway'
+
+describe Takeaway do
+  let(:menu_double) { double :menu, print_menu: "Quiche - £3", find_price: 3 }
+  let(:customer_double) { double :customer, input_number: true, send_confirmation: true }
+  let(:menu_class_double) { double :menu_class, new: menu_double }
```
dan-holmes, line 6: Excellent mocking

README.md
kacesera: You can load it with your own menu that you would need to add into the Menu directory. Otherwise, it pulls from the default menu.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        @wrong_order = Takeaway.new(menu_class_double, cust_class_double)
+        allow(@wrong_order).to receive(:gets).and_return("N")
+      end
+      it "resets the order" do
+        expect(@wrong_order.verify_order).to eq "Your order has been reset"
+        expect(@wrong_order.order).to be_empty
+      end
+    end
+
+    describe '#text_confirmation' do
```
dan-holmes, line 89: Check that customer has received :send_confirmation

lib/takeaway.rb
```diff
@@ -0,0 +1,95 @@
...
+    @order = []
+    @total = 0
+  end
+
+  def see_dishes
+    @menu.print_menu
+  end
+
+  def place_order
+    item = what_would_you_like?
```
dan-holmes, line 19: By convention any method with a ? should only return true or false... I think!

lib/takeaway.rb
```diff
@@ -0,0 +1,95 @@
...
+    @order = []
+    return "Your order has been reset"
+  end
+
+  private
+
+  def order_wrong?
+    puts "Is everything correct? Y or N"
+    answer = gets.chomp.upcase
+    return false if answer == "Y"
```
dan-holmes, line 58: I think you could have done this in one line:
!(answer == "Y")
or even
!(gets.chomp.upcase == "Y")

kacesera: Yes! Thank you for this! I knew I could do it in one line, but I was heavily struggling with how implement it

lib/takeaway.rb
```diff
@@ -0,0 +1,95 @@
...
+    gets.chomp.to_i
+  end
+
+  def anything_else?
+    puts "Anything else? Y or N"
+    true_or_false?(gets.chomp.upcase)
+  end
+
+  def true_or_false?(choice)
+    return true if choice == "Y"
```
dan-holmes, line 87: Same here with the one line, then you may not need a separate method at all. Although perhaps it's clearer this way

lib/customer.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class Customer
+
+  def initialize(phone = nil)
+    @phone = phone
+  end
+
+  def input_number
+    puts "What number should we text a confirmation to?"
+    @phone = "+44#{gets.chomp}"
```
dan-holmes, line 9: I really like this! Small pedantic point, if someone enters their number as 07955... then you'll need to remove the 0 as well as add the +44

spec/customer_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  describe '#send_text' do
+    let(:message_double) { double :message }
+    let(:client_double) { double :client, messages: message_double }
+    let(:client_class_double) { double :client_class, new: client_double }
+
+    it 'sends a text' do 
+      expect(message_double).to receive(:create)
+      new_customer.send_text(client_class_double)
```
dan-holmes, line 29: It would have been good if you could test that the message sent is correct, including the time one hour from now

PR #1760, opened 2020-10-12
Take away

mace-akkari:


lib/menu.rb
```diff
@@ -0,0 +1,13 @@
+class Menu
+  attr_reader :list
+    
+  def initialize
+    @list = {
```
benlynch1931, line 5: Great use of hash to store menu, easy to access and more readable when coding

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def reformat_order(dish)
+    dishes = dish.split(" ")
+    dish = dishes[1].to_sym
+    number = dishes[0].to_i
+    check_order(dish, number)
+    save_order(dish, number)
+  end 
+
+  def check_order(dish, number)
```
benlynch1931, line 39: I'd look into implementing the check_order() into save_order, OR look at calling it within save_order. This will still be following SRP, but will make the code flow better and save you typing `order.check_order` as well as `order.save_order`

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def save_order(dish, number)
+    @items[dish] = number
+  end
+
+  def calculate_cost
+    @cost = 0
+    @items.each { |meal, number|
+      item_cost = @menu[meal].split("£")[1].to_i
```
benlynch1931, line 50: You could look at. not includign '£' in the hash, but leave it as an integer and then add the '£' when printing.

lib/order.rb
```diff
@@ -0,0 +1,55 @@
+require_relative 'menu'
+
+class Order
+  attr_reader :menu, :cost, :items
+  DELIVERY_TIME = 1800
+
+  def initialize(menu = Menu.new)
+    @menu = menu.list
+    @items = {}
+    cost = 0
```
benlynch1931, line 10: I'd make this cost be `@cost` so you don't have to create it on line 48. But this may be done like so for a reason?

lib/order.rb
```diff
@@ -0,0 +1,55 @@
+require_relative 'menu'
```
benlynch1931, line 1: Great use of SRP, each method is doing as little as possible, and delegating where needed!

spec/order_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'order'
+
+describe Order do
+  let(:menu_double) { double :Menu, list: { :shawarma => "£5", :chips => "£3", :salad => "£2", :pepsi => "£1" } }
+  
+  it 'returns the menu on #menu' do
```
benlynch1931, line 6: This test is run in your `menu_spec.rb`. You could look into using your double to return `true` as opposed to the menu. This way you're checking that the Order instance calls this double, without inferring a link between the two classes. (if this didn't make sense, let me know)

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+  attr_reader :menu, :cost, :items
+  DELIVERY_TIME = 1800
+
+  def initialize(menu = Menu.new)
+    @menu = menu.list
+    @items = {}
+    cost = 0
+  end
+
+  def update(items)
```
benlynch1931, line 13: Perhaps change the name of this method to `add_to_order` etc. At the moment it seems slightly vague as unsure if it is ACTUALLY updating something or adding to something?

spec/order_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    order = Order.new(menu_double)
+    expect(order.menu).to include(
+      :shawarma => "£5",
+      :chips => "£3",
+      :salad => "£2",
+      :pepsi => "£1"
+    )
+  end
+
+  it 'allows the user to #add to their order' do
```
benlynch1931, line 16: Just adding to comment on order.rb:13. You're test context says '#add' but is testing `.update`. Just for clarity I'd look at making them the same as each other. (Just a small thought)

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+    items.each { |dish|
+    reformat_order(dish)
+    }
+  end
+
+  def reformat_order(dish)
+    dishes = dish.split(" ")
+    dish = dishes[1].to_sym
+    number = dishes[0].to_i
+    check_order(dish, number)
```
benlynch1931, line 35: Great to see the guard condition as the first method call. Ensures there are no errors going forward!

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+    @items = {}
+    cost = 0
+  end
+
+  def update(items)
+    manage_order(items)
+    calculate_cost
+  end
+
+  def confirm
```
benlynch1931, line 18: If you get time/ had time, I'd recommend at looking into using the Twilio API. It is a cool feature to add. 
This could easily be something you implement later on if you fancied. But for the moment I can clearly see how the code `confirms` the order and can see where you'd implement it!

PR #1761, opened 2020-10-12
Takeaway Challenge_Tim

timcastillogill:
This challenge is incomplete. I think the best thing I could do with it, would be to scrap and start again, but time does not allow that so this is the version I have.


lib/takeaway.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def print_dishes
+    @dishes
+  end
+
+  def display_menu
+    @dishes.each { |dish, price|
+      "#{dish}: #{price}"
+    }
+  end
```
RastallM, line 28: These two methods seem to be doing the same thing. I would maybe get rid of one of them - or perhaps update if the intention was for them to do different things

PR #1767, opened 2020-10-26
Friday Challenge -Krzysztof

krzy5zt0f-1:


lib/menu.rb
```diff
@@ -0,0 +1,16 @@
+class ListofDishes
+  attr_reader :menu
+  def initialize
+    @menu = [{ dish: "Pierogi (dumplings)", price: 7.95 },
```
danbedford56, line 4: This way works well, EBI: Dish was it's own class

lib/text.rb
```diff
@@ -0,0 +1,18 @@
+require 'twilio-ruby'
+class Text
+  def initialize(phone_number)
+    @phone_number = "+44#{phone_number}"
+  end
+
+  def send_text
+    account_sid = ENV['TWILIO_ACCOUNT_SID']; auth_token = ENV['token']
+    client = Twilio::REST::Client.new(account_sid, auth_token)
```
danbedford56, line 9: ENV vars should be all caps (I think).

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
...
+             { dish: "Krokiety (croquettes)", price: 8.45 },
+             { dish: "Mushroom Soup", price: 2.95 },
+             { dish: "Grilled Trout with mashed potatos", price: 11.95 },
+             { dish: "Gulasz stew", price: 13.95 }]
+  end
+
+  def see_list
+    @menu.each { |dish| puts "#{@menu.index(dish) + 1}. #{dish[:dish]}, price: £#{dish[:price]}" }
+  end
+end
```
danbedford56, line 16: This class could have a method to add a new dish to the menu

spec/text_spec.rb
```diff
@@ -0,0 +1,19 @@
+require_relative '../lib/text'
+
+describe Text do
+  subject { described_class.new(12345678) }
+  it { is_expected.to respond_to(:send_text) }
+
+  describe ".send_text" do
+    it "sends a message to given number" do
+      text = "Thank you! Your order was placed and will be delivered before #{(Time.now + 60 * 60).strftime("%k:%M")}"
```
danbedford56, line 9: This expect is satisfied from within your test I think if you are using expect to receive it should be when this is called in another method.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'menu'
```
danbedford56, line 1: WWW: This class is written very well, EBI: You had a 'order loop' method which called these methods

PR #1768, opened 2020-10-26
takeaway challenge 

Eunicekon:


lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
+class Takeaway
+  attr_reader :menu, :price, :number, :total_sum
+
+  def initialize
+    @menu = ["Pasta: £3.50", "lasagna: £4.50", "water: £1.50"]
+    # @price = price
```
jckmrrs, line 6: I like how you're thinking ahead to what you may want in the future. It shows that you have thought about the user stories and you're prepared to adapt your code in the future.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'takeaway'
+# let(:list_of_dishes) {double("Pasta: £3.50")}
```
jckmrrs, line 2: Is this double used?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'takeaway'
+# let(:list_of_dishes) {double("Pasta: £3.50")} 
+
+RSpec.describe Takeaway do
+
+  describe "#menu" do
```
jckmrrs, line 6: Your tests are very easy to understand what they should be doing!

PR #1769, opened 2020-11-07
Weekend challenge(2) complete!

MykeNuLeng:


lib/order.rb
```diff
@@ -0,0 +1,39 @@
...
+    @menu = menu
+    @my_order = []
+    @cheque = 0
+    @message = message
+  end
+
+  def whole_order(all_the_food)
+    all_the_food.each_pair { |key, value| value.times { order(key) } }
+    @cheque if check_cheque?
+    @message.send("Thank you! Your order was placed and will be delivered before #{Time.now + 3600}")
```
francescoFH: Nice way of using time class to send a message in real-time

spec/order_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'order'
+
+describe Order do
+  context 'giving it a menu' do
```
francescoFH, line 4: I guess you should use "#describe" to describe a method and context to separate the functions of the same method when you are using it in different situations. I can probably explain it better when we make the call.

spec/order_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+    it 'when initialized it gets a menu' do
+      expect(subject.menu).not_to be nil
+    end
+
+    context 'allowing the menu double to store items' do
+      before do
+        hash = {"Goan pork vindaloo" => 14.00, "Shahi murg" => 11.00}
+        allow(menu).to receive(:menu_items).and_return(hash)
+        allow(message).to receive(:send).and_return(true)
+      end
```
francescoFH: Really like the before do to don't repeat yourself

PR #1770, opened 2020-11-08
Weekend challenge 2

alexleesonmill:


lib/menu.rb
```diff
@@ -0,0 +1,18 @@
...
+    'California Temaki' => 5,
+    'Chicken Katsu Curry' => 11,
+  }.freeze
+
+  attr_accessor :dishes
+
+  def initialize(dishes = MENU)
+    @dishes = dishes
+  end
+
```
sedwards93, line 14: Very nice, clear and succinct code with distinct classes. Easy to read 👍 !

spec/text_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+describe TextConfirm do
+  it 'initialises with @from with an environmental variable' do
+    expect(subject.from).to eq ENV['SECRET_TWIL_NUM']
+  end
+
+  it 'initialises with @to with an environmental variable' do
+    expect(subject.to).to eq ENV['SECRET_MOBILE']
+  end
+
```
sedwards93, line 12: Nice - I couldn't think how to test this part of my code. Seeing your's makes perfect sense!

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+    amount.times { @order << dish }
+    amount.times { @total += @menu[dish] }
+  end
+
+  def total_correct?
+    balance = 0
+    @order.map do |dish|
+      balance += @menu[dish]
+    end
+    @total == balance
```
sedwards93, line 25: Might be nice to format this so it reads more like a price. at the moment it just shows the cost with no currency or decimal places.

PR #1772, opened 2020-11-08
First pull request

joshuaretallick:
Looking for feedback on splitting up classes


lib/takeaway.rb
```diff
@@ -0,0 +1,40 @@
...
+    dishes[dish]
+  end
+
+  def item_totals
+    order.map do |dish, quantity|
+      dishes[dish] * quantity
+    end
+  end
+
+  def has_dish?(dish)
```
TobyRet, line 36: This should be a private method as it's only called by the takeaway instance

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,66 @@
+require 'takeaway_1'
```
TobyRet, line 1: This seems strange to me - you are requiring  `takeaway_1` but your TakeAway class is in the file `takeaway`

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  it "tells if a dish is on the menu" do
+    expect(takeaway.has_dish?(:hotdog)).to be true
+  end
+
+  it "tells if a dish is not on the menu" do
+    expect(takeaway.has_dish?(:hamburger)).to be false
+  end
+
+  it "doesn't allow items to be added that are not on the menu" do
```
TobyRet, line 54: I would imagine this spec would pass however its not the norm to stub part of a class under test. See an explanation at https://softwareengineering.stackexchange.com/questions/107361/is-it-okay-to-fake-part-of-the-class-under-test

Rather consider making your menu a dependency of TakeAway. You could then test variants of your menu without stubbing.

e.g 

```
menu_hotdog_and_fries = { 
  hotdog: 5.00,
  fries: 2.00
}

menu_hotdog = {
  hotdog: 5
}

takeaway_one = TakeAway.new(menu_hotdog_and_fries)
takeaway_two = TakeAway.new(menu_hotdog)

your 'menu' could be a simple hash or if you really need a class in its own right.

File: spec/takeaway_spec.rb [hunk 4]
```diff
@@ -0,0 +1,66 @@
...
+  it "selects several dishes from the menu" do
+    takeaway.add(:hotdog, 2)
+    takeaway.add(:fries, 1)
+    expect(takeaway.order).to eq(order)
+  end
+
+  it "calculates the total for the order" do
+    takeaway.add(:hotdog, 2)
+    takeaway.add(:fries, 1)
+    total = 13.50
```
TobyRet, line 42: Just a minor naming issue - may make the code more readable if `total` is renamed to `expected_total`

PR #1776, opened 2020-11-08
Class Split

joshuaretallick:
Ok everything is split into classes now but I am having issues bringing my @dishes from the Menu class across to the others hence I have hardcoded the variable in the other classes.
Or maybe Im not understanding how to feature test? I require takeaway.rb in IRB but then how do I say print the menu which is in the Menu class?


lib/order.rb
```diff
@@ -0,0 +1,35 @@
+require_relative 'menu'
+
+class Order
+
+  attr_reader :dishes, :order
+
+  def initialize
+    @dishes = { hotdog: 5.00, fries: 3.50 }
```
TobyRet, line 8: Looks like you should be passing in an instance of Menu here.

lib/order.rb
```diff
@@ -0,0 +1,35 @@
...
+    fail "#{dish.capitalize} is not on the menu!" unless has_dish?(dish)
+    order[dish] = quantity
+  end
+
+  def total
+    item_totals.inject(:+)
+  end
+
+  def price(dish)
+    @dishes[dish]
```
TobyRet, line 22: Once you have passed in an instance of Menu as a dependency you can then call 
```
@menu.dishes
```

lib/order.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def item_totals
```
TobyRet, line 25: `item_total` seems more closely aligned with the `Order` class's  responsibilities. i.e This method should be moved to the Order class. Perhaps also rename to `order_total`?

PR #1781, opened 2020-11-09
submission 

rafael-oelmann:


lib/ramen_bar.rb
```diff
@@ -0,0 +1,34 @@
+class RamenBar
+
+  attr_accessor :dishes, :order, :total_order
+
+  MENU = {
```
AdamusBG, line 5: Good simple usage of the hash here, definitely overcomplicated mine with an array of hashes!

lib/ramen_bar.rb
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  def initialize(dishes = MENU)
+    @dishes = dishes
+    @order  = []
+    @total_order = 0
+  end
+
+  def menu
+    @dishes.map { |dish, price| "#{dish} - £#{price}" }.join(', ')
```
AdamusBG, line 20: Nice chaining of the methods to produce a menu! Suggest maybe joining them with '\n' instead to have every item on a different line.

lib/ramen_bar.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def order_meal(dish, quantity = 1)
+    quantity.times { @order << dish }
+  end
+
+  def total
+    @order.each do |dish|
+    @total_order += @dishes[dish]
```
AdamusBG, line 29: Super easy totalling because of your data structure, nice!

spec/ramen_bar_spec.rb
```diff
@@ -0,0 +1,48 @@
+require 'ramen_bar'
+
+describe RamenBar do
```
AdamusBG, line 3: Good code coverage with your tests.

PR #1784, opened 2020-11-09
Takeaway Challenge - Pull Request

CodeZeus-dev:


lib/basket.rb
```diff
@@ -0,0 +1,35 @@
...
+  attr_reader :basket, :basket_val
+
+  def initialize
+    @basket = []
+    @basket_val = 0
+  end
+
+  def add_to_basket(item, quantity, item_price)
+    update_basket(item, quantity, item_price)
+    "#{quantity}x #{item}(s) added to your basket."
```
ehwus, line 12: This is a really useful addition to the method, gives the user confirmation and clarity in a well formatted way.

lib/basket.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def summary
+    @basket.each do |item|
+      $stdout.puts "#{item.keys.join()} x#{item[item.keys.join()][0]} = £#{item[item.keys.join()][1]}"
```
ehwus, line 17: Have you sacrificed legibility by using one line for this?

lib/basket.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def total_value
+    @basket_val
+  end
+
+  private
+
+  def update_basket(item, quantity, item_price)
+    @basket << {item => [quantity, quantity * item_price]}
```
ehwus, line 28: I think an item object that contains a price would map more naturally to quantity in a Hash

lib/basket.rb
```diff
@@ -0,0 +1,35 @@
...
+    basket_value(quantity * item_price)
+  end
+
+  def basket_value(price)
+    @basket_val += price
```
ehwus, line 33: It would be great for the user to be kept updated on the value of the basket as they add to their order, this function already does that neatly and elegantly and could be used for this.

lib/menu.rb
```diff
@@ -0,0 +1,21 @@
+class Menu
+
+  attr_reader :food_list
+
+  def initialize
+    @food_list = {'Rainbow dumpling' => 28, 'Half crispy aromatic duck' => 39}
```
ehwus, line 6: Aside from just the hardcoding, it's not immediately apparent to me what the integer refers to - is it quantity or price? As the list is a hash I have no specifications to refer back to

lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
+require_relative 'restaurant'
+require_relative 'basket'
+require_relative 'messaging'
+
+class Takeaway
+
+  attr_reader :restaurant, :basket
+
+  def initialize(restaurant = Restaurant.new, basket = Basket.new, message = Messaging.new)
```
ehwus, line 9: Good use of dependency injection

lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
...
+    @basket = basket
+    @message = message
+  end
+
+  def see_menu
+    @restaurant.show_menu
+  end
+
+  def order(item, quantity = 1)
+    # infinite inventory :: quantity not an argument in check_availability
```
ehwus, line 20: This is the only place that it's explicit that the items are mapped to quantity

spec/basket_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+    it 'returns a message to the user' do
+      dish = 'fish'
+      quantity = 2
+      price = 7
+      expect(subject.add_to_basket(dish, quantity, price)).to eq("#{quantity}x #{dish}(s) added to your basket.")
+    end
+
+    it 'updates the basket' do
+      dish = 'fish'
```
ehwus, line 28: Could you use a double here?

spec/messaging_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'messaging'
+
+describe Messaging do
+  let(:client) { double :client }
+
+  it 'sends an SMS to destination phone' do
```
ehwus, line 6: Great use of mocking to avoid an API call

ehwus (commented):
I think that you've hit the brief really well - all of the user stories are perfectly catered for, and all of my feedback is largely about the more minor implementations so I think you've done a really good job.

I'm desperate for a README here, it took a little while to work out that I had to use the takeaway rather than the restaurant to get your full experience!

The way that the classes are pieced together makes a lot of sense to me, and I think my major suggestions are to "prettify" the user experience rather than returning hashes, to maybe even revisit the use of the hash for the menu, and to refactor some of your code for readability and clarity.

PR #1785, opened 2020-11-09
Takeaway challenge 

kabligh:
I struggled to write tests for adding in the Twilio API functionality so this is threadbare.

My method for displaying the basket to the user is clumsy – they are listed in their {hash brackets} – but I struggled to fix this.


kabligh:
Attempting to remove bundle folder

PR #1786, opened 2020-11-09
Pull Request

joshuaretallick:
I still have some functionality problems.  I haven't done the text user story and also I am struggling to share my menu across the other classes.  I have a fair idea on why, I just didnt have enough time to fix it


lib/menu.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  attr_reader :dishes
+
+  def initialize(dishes)
+    @dishes = dishes
+  end
+
+  def print
+    dishes.map do |title, price|
+      "%s £%.2f" % [title.to_s.capitalize, price]
```
Michaelschats: What does the '%s.2f' mean ?

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
...
+    end.join(", ")
```
Michaelschats, line 17: saw this as well, how does end.join(", ") work ?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'takeaway'
+
+describe Takeaway do
+  subject(:takeaway) { described_class.new(menu: menu) }
```
Michaelschats: why is the menu: menu written as it is, what does it do ?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  let(:menu) { double(:menu, print: printed_menu) }
```
Michaelschats: saw this else where, does this make a double of the menu class and method print ?

lib/takeaway.rb
```diff
@@ -0,0 +1,15 @@
+class Takeaway
+
+  def initialize(menu:)
```
Michaelschats: Does this need to require the Menu class in order to print it ?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  it "Instantiates a new class" do
+    expect(takeaway).to be_a_kind_of Takeaway
+  end
+
+  it "shows the menu when called" do
+    expect(takeaway.print_menu).to eq(printed_menu)
+  end
+
+end
```
Michaelschats, line 34: Really clean code and easy to understand, only questions/comments I have made are for my own understanding rather than possible improvements/refactoring.

lib/order.rb
```diff
@@ -19,17 +20,17 @@ def total
...
 
   def price(dish)
-    @dishes[dish]
+    @menu[dish]
   end
 
   def item_totals
     order.map do |dish, quantity|
-      @dishes[dish] * quantity
+      @menu[dish] * quantity
```
Michaelschats, line 28: How does this output the quantity ?

lib/order.rb
```diff
@@ -19,17 +20,17 @@ def total
...
     end
   end
 
   def has_dish?(dish)
-    !@dishes[dish].nil?
+    !@menu[dish].nil?
```
Michaelschats, line 33: Is this testing whether the dish is included or not ?

PR #1787, opened 2020-11-09
stringiest takeaway-challenge

stringiest:


lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
+require_relative 'menu'
+require_relative 'keys'
+require 'twilio-ruby'
+
+class Takeaway
+  attr_reader :menu, :order, :order_total
+
```
malachispencer, line 7: Good that you injected the Menu class into the Takeaway class instead of making it a hard coded dependency, makes isolating your tests with a double easier.

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+  def total
+    @order.each {
+      |dish, quantity| @order_total += menu.dishes[dish] * quantity
+    }
+    return @order_total
+  end
+
+  def send_message
+    keys = Keys.new
+
```
malachispencer, line 40: Could have wrapped the Twilio Client in its own class and injected the dependency into your Takeaway, just like you did with the Menu class.

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+    client.messages.create(
+      from: keys.host,
+      to: keys.recipient,
+      body: "Thank you! Your order will be delivered before #{time}"
+    )
+  end
+
+  def checkout
+    send_message
```
malachispencer, line 53: No need for the explicit return statement, Ruby returns the last line in the method by default.

lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+class Menu
+  attr_reader :dishes
+
```
malachispencer, line 3: You could store the dishes in a class constant, if you wanted to add many more dishes to the menu, your initialize would become quite messy.

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def total
+    @order.each {
+      |dish, quantity| @order_total += menu.dishes[dish] * quantity
+    }
+    return @order_total
+  end
+
+  def send_message
```
malachispencer, line 38: With the use of the keys class that you've created, if I cloned your repo to my computer, the program wouldn't work because I wouldn't have your keys class. If you use environment variables on the other hand, I could simply set my own environment variables and the program would work for me.

README.md
```diff
@@ -1,83 +1,112 @@
...
+* tests shouldn't send texts, so use mocking or stubs here
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+```
+require './lib/takeaway'
+takeaway = Takeaway.new
+takeaway.place_order("pizza", 3)
+takeaway.place_order("burger")
+takeaway.checkout # should generate text message confirming order, and return confirmation message.
+```
```
tatsiana-makers, line 112: Really nice README! It was great to read your plan for each of the user stories, as well as an example usage of the app!

lib/takeaway.rb
tatsiana-makers, line 38: Could this method be made private to increase encapsulation? Are there other methods which could be made private?

spec/menu_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'menu'
+
+describe Menu do
+  it "creates a new instance of menu" do
+    expect(subject).to be_an_instance_of Menu
+  end
+
+  it "initializes with a full menu" do
+    expect(subject.dishes.length).to eq 5
```
tatsiana-makers, line 9: What if `subject.dishes` was [1,2,3,4,5], would this test pass? Should it? There's probably a stronger assertion we can make here

spec/menu_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'menu'
+
+describe Menu do
```
tatsiana-makers, line 3: One thing you aren't testing yet, is that you can initialise an instance with a custom menu

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'takeaway'
+
+describe Takeaway do
+  it "returns a menu when requested" do
+    expect(subject).to respond_to(:read_menu)
+  end
+
+  it "returns the full menu when requested" do
+    expect(subject.menu.dishes.length).to eq 5
```
tatsiana-makers, line 9: We could call and test the `read_menu` method here instead of mimicking the logic inside of that method:
```
expect(subject.read_menu).to eq(foo)
```

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  describe "#order" do
+    it "responds to order" do
+      expect(subject).to respond_to(:place_order).with(2).arguments
+    end
+
+    it "adds an order to the basket" do
+      subject.place_order("pizza")
+      expect(subject.order.length).to eq 1
+    end
```
tatsiana-makers, line 20: Can we also assert that `subject.place_order` returns the correct output? i.e. `"1 x pizza(s) added to your order"`

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+require_relative 'keys'
+require 'twilio-ruby'
+
+class Takeaway
+  attr_reader :menu, :order, :order_total
+
+  def initialize(menu = Menu.new, order = {}, order_total = 0)
+    @menu = menu
+    @order = order
+    @order_total = order_total
```
tatsiana-makers, line 11: Does `order_total` need to be an instance variable? In this case, we're only using this variable in a single method, which suggests it can just be a local variable that exists only inside of the `total` method. In general, if a variable is used inside of only one method, it can be a local variable rather than an instance variable.

Having as little stored state as possible reduces the risk of another developer, for example, accidentally writing a method which will override `@order_total` with a nonsense value.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  subject(:takeaway) { described_class.new }
+
+  before do
+    allow(takeaway).to receive(:send_message)
+  end
+
+  it 'sends a payment confirmation text message' do
+    expect(takeaway).to receive(:send_message)
+    takeaway.checkout
```
tatsiana-makers, line 47: Can we also assert on what `checkout` should return?

tatsiana-makers (commented):
This is a nice go at this weekend's challenge! I've left a few comments and suggestions of other things you could test. Happy to discuss any points made here further, let me know 🙂

spec/menu_spec.rb
stringiest: Test changed to ensure that subject.dishes includes ("pizza", "burger", "steak", "ice cream", "chow mein")

spec/menu_spec.rb
stringiest: Test added to check this.

spec/takeaway_spec.rb
stringiest: test updated to:
expect(suject.read_menu).to include("pizza", "burger", "chow mein", "steak", "ice. cream")

PR #1789, opened 2020-11-09
Michael Schats attempt at Weekend Challenge 

Michaelschats:


spec/Menu_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'Menu'
+
+describe Menu do
+
+  subject(:menu) { Class.new { extend Menu } }
```
joshuaretallick, line 5: I have never seen this extend thing before, guessing it just creates a new Class yeah?
Another way you can write this is:
subject(:menu) { described_class.new }

spec/Takeaway_spec.rb
```diff
@@ -4,16 +4,10 @@
 describe Takeaway do
   subject(:takeaway) { described_class.new }
```
joshuaretallick, line 5: Ah sorry seen you did that here

spec/Takeaway_spec.rb
```diff
@@ -4,16 +4,10 @@
...
 
-  describe '#read_menu' do
-    it "Shows the menu with the prices of each dish" do
-      expect(subject.read_menu).to eq(Menu::MENU)
```
joshuaretallick: because you declared subject(:takeaway) above you can write expect(takeaway.read_menu)
It doesnt change anything but I think makes the code more readable, you might get confused on subject when you see it alot, but you know exactly what it is testing if you put takeaway

lib/Takeaway.rb
```diff
@@ -0,0 +1,10 @@
+class Takeaway
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = { "spring roll" => 0.99, "char sui bun" => 3.99,
+             "pork dumpling" => 2.99, "peking duck" => 7.99,
+             "fu-king fried rice" => 5.99}
```
joshuaretallick: Love me some fu-king fried rice!

joshuaretallick: I also went down the @dishes / menu thing, then like you created a MENU later,

lib/Menu.rb
```diff
@@ -0,0 +1,5 @@
+module Menu
```
joshuaretallick, line 1: I really need to figure modules out!

joshuaretallick (commented):
Looks like you have the basis for something that will be really smart when you finish it.  Its really clear and easy to read

PR #1790, opened 2020-11-09
submission

adam2021:


spec/order_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+describe Order do
+
+  it "should have an order_name when created" do
+    expect(subject.instance_variable_get(:@order_name)).to eq('patel')
+  end
+
+  it "should contain an empty array called @order_items" do
+    expect(subject.instance_variable_get(:@order_items)).to eq([])
+  end
+
```
horthbynorthwest, line 13: both of these tests are doing the same thing?

lib/takeaway.rb
```diff
@@ -0,0 +1,100 @@
...
+
+class Takeaway
+
+  attr_accessor :active_order, :order_items, :menu_item, :kitchen, :cooking
+
+    def initialize
+      @kitchen = Kitchen.new
+        menu
+    end
+
```
horthbynorthwest, line 11: could you inject a new instance of kitchen rather than hard coding it in? i.e. def initialize(kitchen)
@kitchen = kitchen and you just pass in Kitchen.new as the argument?

lib/takeaway.rb
```diff
@@ -0,0 +1,100 @@
...
+    def create_order(name)
+      @active_order = Order.new(name)
+    end
+
+    def test_method(dish_id)
+
+      chosen_dish = @menu_item.select {|dish| dish[:id] == dish_id}
+
+      dish = chosen_dish.pop
+
```
horthbynorthwest, line 41: is this line necessary? could you just put @active_order.order_items << chosen_dish.pop on line 44 instead?

lib/takeaway.rb
```diff
@@ -0,0 +1,100 @@
...
+          puts ""
+          puts "To checkout, complete your order (.complete)"
+          puts "To start over, clear your order (.empty_cart)"
+
+      end
+
+    def empty_cart
+      @active_order.order_items.clear
+    end
+
```
horthbynorthwest, line 69: I really like this ^^ definitely something I was missing in mine!

spec/kitchen_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    takeaway.test_method(1)
+    takeaway.submit_order
+    expect(takeaway.kitchen.cooking.length).to eq(1)
+  end
+
+  it "sends a message when an order is received" do
+    expect { subject.send_message.process("Message to customer") }.to output.to_stdout

+  end
+
```
horthbynorthwest, line 20: what does .process on line 18 do?

PR #1791, opened 2020-11-09
Takeaway challenge – hopefully without bundle folder

kabligh:


lib/order.rb
```diff
@@ -0,0 +1,47 @@
...
+  def add(dish, quantity = 1)
+    raise "Dish not on menu" unless on_menu?(dish)
+
+    quantity.times { @basket << { dish: dish, price: @menu[dish] } }
+    @total += (@menu[dish] * quantity)
+  end
+
+  def checkout
+    if empty?
+      print "Your basket is empty"
```
samanthagottlieb, line 24: Could you have raised an error here instead?

lib/order.rb
```diff
@@ -0,0 +1,47 @@
...
+require_relative 'text'
+
+class Order
+
+  attr_reader :basket, :total, :menu
+
+  def initialize(total = 0)
+    @basket = []
+    @total = total
+    @menu = Menu::DISHES
```
samanthagottlieb, line 11: Nice use of constant variable

lib/order.rb
```diff
@@ -0,0 +1,47 @@
...
+    @text = Text.new
+  end
+
+  def add(dish, quantity = 1)
+    raise "Dish not on menu" unless on_menu?(dish)
+
+    quantity.times { @basket << { dish: dish, price: @menu[dish] } }
+    @total += (@menu[dish] * quantity)
```
samanthagottlieb, line 19: Good use of the hash to calculate the total. Easy to understand!

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
+require 'order'
+
+describe Order do
+
+  context "#initialize" do
```
samanthagottlieb, line 5: I think context is used when testing behaviour state, and describe when testing a specific feature

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    it "has a bill of 0" do
+      expect(subject.total).to eq 0
+    end
+  end
+
+  context "adding items" do
+
+    it "raises an error if person tries to order item not on menu" do
+      expect { subject.add("foie gras") }.to raise_error("Dish not on menu")
```
samanthagottlieb, line 18: You could use stubbing here to test when on_menu? returns false

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    end
+
+    it "adds items to the user's basket" do
+      subject.add("Burger", 1)
```
samanthagottlieb, line 22: You could use doubles here for dish and quantity

samanthagottlieb (commented):
code review

PR #1792, opened 2020-11-09
.

savannaelbey:


margmartelo (commented):
Love the way your code is so organised and tidy! It's easily readable and has a great dependency injection!

lib/order.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'dish'
+require 'twilio-ruby'
+
+class Order
+
+  attr_reader :order_list, :total
+
+  def initialize
+    @order_list = []
+    @total = 0
```
tatsiana-makers, line 10: I was able to remove `@total` and the attr_reader and have the tests pass - perhaps you forgot to remove the attribute reader? Why might this make the tests fail?

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def add(dish, quantity)
+    add_order_to_list(dish, quantity)
```
tatsiana-makers, line 14: `add_order_to_list` is only called within this method, so we can decrease the number of methods (and therefore the cognitive overload for someone who may read your code) by pulling the logic out of `add_order_to_list` and just putting it in `add`.

[Here](https://softwareengineering.stackexchange.com/a/386674) is a stackoverflow answer that elaborates on this a bit.

Are there other places you are doing this in the code?

spec/menu_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'menu'
+
+describe Menu do
+
+  let(:dish1) { double('dish', name: "Arrabiata", price: 9.99) }
```
tatsiana-makers, line 5: Awesome use of doubles!

tatsiana-makers (commented):
Thanks for submitting your code for review - this is a solid go at the weekend challenge! I've left a few comments. Happy to chat further about any of them 🙂

PR #1794, opened 2020-11-10
Functionality All Working

joshuaretallick:
I have completed the first 3 user stories and I am pretty happy with the code. 
I need to add some mocking but just want to check the basic structure I have now and get some feedback


lib/order.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'menu'
+
+class Order
+
+  attr_reader :dishes, :order, :menu
+
+  def initialize
+    @menu = Menu::MENU
```
tatsiana-makers, line 8: Here we're implicitly relying on a constant in the Menu class. Instead, we could make this dependency explicit and use [dependency injection](https://github.com/makersacademy/skills-workshops/blob/f5b4801840fe07d26ff70341652dc81dcda12289/practicals/object_oriented_design/dependency_injection.md) which will allow us to decouple the classes and isolate the unit tests by using mocking

lib/takeaway.rb
```diff
@@ -0,0 +1,30 @@
...
+  #  @order = order
+  end
+
+  # def order
+  #   @order
+  # end
+
+  def new_order
+    @order = Order.new
+    @menu.print_menu
```
tatsiana-makers, line 19: Nice use of [forwarding](https://github.com/makersacademy/skills-workshops/blob/master/practicals/object_oriented_design/oo_relationships.md#forwarding)!

spec/order_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  it "tells if a dish is on the menu" do
+    expect(main_order.has_dish?(:hotdog)).to be true
+  end
+
+  it "tells if a dish is not on the menu" do
+    expect(main_order.has_dish?(:hamburger)).to be false
+  end
+
+  it "doesn't allow items to be added that are not on the menu" do
+    allow(main_order).to receive(:has_dish?).with(:hotdog).and_return(true)
```
tatsiana-makers, line 33: When we do mocking (doubles and stubs) in a unit test for a class, we want to mock _other_ classes and the methods contained in those _other_ classes. We don't want to mock the class we are actually unit testing, because then we aren't testing the real functionality of the class we are aiming to unit test.

lib/takeaway.rb
```diff
@@ -0,0 +1,30 @@
+require_relative './menu.rb'
+require_relative './order.rb'
+
+class TakeAway
+
+  attr_reader :order, :menu
+
+  def initialize
```
tatsiana-makers, line 8: This is another place where dependency injection can be used to decouple the classes.

`Takeaway` relies on the `Menu` and `Order` classes. We want to use dependency injection so that we can isolate the unit tests.
Without using dependency injection, our unit tests of the `Takeaway` class rely on the _actual_ implementation of the `Menu` and `Order` classes. This means if I break the `Menu` class, the `Takeaway` unit tests will also start failing. 

If instead, I've successfully isolated the unit tests by using doubles and stubs, if I break the `Menu` class, my unit tests for `Takeaway` should continue to pass, because in my unit tests, I am relying on doubles and stubs rather than the actual `Menu` class code .

tatsiana-makers (commented):
I've left a few comments to capture the conversation we had earlier today - hope they're helpful! Let me know if you have any further questions 🙂

PR #1802, opened 2020-12-06
Takeaway weekend challenge

AmanTank187:
Completed the three user stories. I'm aware i could split the objects into classes.

### DIFF
--- a/spec/takeaway_spec.rb
+++ b/spec/takeaway_spec.rb
@@ -1,0 +1,57 @@
+require 'takeaway'
+
+describe Takeaway do
+  it 'Is an instance of the takeaway class' do
+    expect(subject).to be_an_instance_of Takeaway
+  end
+
+  describe '#show_menu' do
+    it 'It should respond to the method show_menu' do
+      expect(subject).to respond_to(:show_menu)
+    end
+
+    it 'It should show the menu when called' do
+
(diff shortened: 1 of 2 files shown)


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'takeaway'
+
+describe Takeaway do
+  it 'Is an instance of the takeaway class' do
+    expect(subject).to be_an_instance_of Takeaway
+  end
```
lisabardelli, line 6: you can consider to use a describe statement for your initialize method as you did for the other methods

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  describe '#check_total' do
+    it 'It should respond to the method check total' do
+      expect(subject).to respond_to(:check_total)
+    end
+
+    it 'should puts a message if total is right' do
+      subject.add_item("burger",2)
+      subject.check_total
+    expect(subject.given_total).to eq subject.total
```
lisabardelli, line 50: it seems your it statement does not describe your expectation as you expect get a float

lisabardelli (commented):
You have used a consistent style both in code and tests and applied the Single Responsibility Principle since your functions are performing a single responsibility.
As you said, you may consider to break one class in two as this might make your code more readable.

spec/takeaway_spec.rb
AmanTank187: At first i wanted to get back true then changed my mind, thanks for pointing this out to me so i can change my test

PR #1804, opened 2020-12-06
Takeaway Challenge - Lisa Bardelli

lisabardelli:


lib/dish.rb
```diff
@@ -0,0 +1,7 @@
+class Dish
+  attr_reader :price, :name
+  def initialize(name, price)
```
AmanTank187, line 3: Good use of multiple classes/files

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu 
+  attr_reader :dishes_list
+  def initialize
+    @dishes_list = []
+      
+  end 
+
+  def add_dish(dish)
+    @dishes_list << dish
```
AmanTank187, line 9: Clear methods

lib/order.rb
```diff
@@ -0,0 +1,13 @@
+class Order
+  attr_reader :order_list
+  def initialize
+    @order_list = []
+  end 
+
+  def order_dish(dish_name, quantity = 1)
+    for _ in 0...quantity
```
AmanTank187, line 8: Could use the .times method here, this would reduce the number of lines and make it clear about what you're trying to accomplish with these lines of code.

spec/menu_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+    let(:dish) { double("Dish") }
+    it { is_expected.to respond_to(:add_dish).with(1).argument }
+   
+    it "adds a dish to the menu" do
+      expect(menu.add_dish('dish')).not_to be_empty
+    end
+  end
+
+  describe "#print_menu" do
+    let(:dish) { double("Dish") }
```
AmanTank187, line 20: Good use of the double

lib/main.rb
```diff
@@ -0,0 +1,28 @@
...
+order1 = Order.new
+order1.order_dish("pizza")
+order1.order_dish("pasta", 2)
+order1.order_list
+
+total_price = restaurant.take_order(order1)
+puts total_price
+expected_price = 10
+if total_price == expected_price
+  restaurant.confirm_order(ENV['CUSTOMER_PHONE_NUMBER'])
```
AmanTank187, line 25: Could these lines be placed into a method? This will reduce the amount of lines and make the code clearer although i know what it's doing someone who doesn't know about the challenge may not

lib/message_sender.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class MessageSender
+
+  def initialize(from = ENV['TWILIO_FROM_PHONE_NUMBER'])
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
+    @twilio_client = Twilio::REST::Client.new(account_sid, auth_token)
+    @from = from
+  end
+
```
AmanTank187, line 12: Good implementation of the twilio-ruby gem

lib/restaurant.rb
```diff
@@ -0,0 +1,48 @@
...
+  def confirm_order(phone_num)
+    @message_sender.send_message(phone_num)
+  end
+
+private
+  def on_the_menu?(dish_name)
+    @menu.dishes_list.each do |dish|
+      if dish.name == dish_name
+        return true
+      end
```
AmanTank187, line 35: Good way of using private methods

spec/order_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'order'
+describe Order do
+    
+  describe "#initialize" do
+    it "returns an instance of the Order class" do
+      expect(subject).to be_an_instance_of Order
+
+    end
+  end
+
```
AmanTank187, line 10: Clear tests i can understand what they're doing

lib/dish.rb
AmanTank187: Although the code is easy to follow and very clear, You could add a README to explain your approach to the challenge and how to use your program.

spec/menu_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    before(:each) do
+      allow(dish).to receive(:name).and_return("pizza")
+      allow(dish).to receive(:price).and_return("£20")
+    end
+
+    it "prints the menu" do
+      menu.add_dish(dish) 
+      expect { menu.print_menu }.to output("Name: #{dish.name} - Price: £#{dish.price}\n").to_stdout
```
AmanTank187, line 29: Great test to check if the text is being outputted to the user

AmanTank187 (commented):
Overall the code and tests for this application are very good ! Code and tests are both clear and i can see what they're intended for

AmanTank187 (commented):
Adding additional comments.

PR #1806, opened 2020-12-06
takeaway-challenge

IrinaG29:


lib/order.rb
```diff
@@ -0,0 +1,53 @@
+require 'twilio-ruby'
+
+class Order
+  attr_reader :order, :menu
+
+  def initialize(menu)
+    @order = []
+    @menu = menu
```
chrismabdo, line 8: One thing that can reduce the need for class Order to be dependant on class Menu is to add (menu = Menu.new) within the initialize argument.

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+  def send_message
+    twilio_client.messages.create(
+      from: '',
+      to: '',
+      body: message
+    )
+  end
+
+  def message
+    "Thank you! Your order was placed and will be delivered before 18:52"
```
chrismabdo, line 51: Can perhaps add a Time.new object and instead of using 18:52, use Time.new.hour + 1.

spec/menu_spec.rb
```diff
@@ -0,0 +1,10 @@
+describe Menu do
+  subject { described_class.new }
+
+  describe "#dish_list" do
+    it "shows list of dishes" do
+      expect(subject.dish_list.size).to eq(2)
```
chrismabdo, line 6: Could perhaps add a double here in case more items were added to the menu

chrismabdo (commented):
Overall I think this is great work! 
- Methods and classes are appropriately named, making the code easy to read.
- Methods follow the SRP very well.
- Good use of private methods in the Order class.
- Testing is thorough and I feel as though they cover the requirements set out in the User Stories.

- A few minor suggestions have been added in the individual comments throughout. 

Well done :)

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+    raise "Incorrect price" unless  price == total_price
+    true
+  end
+
+  def complete
+    send_message
+  end
+
+  private
+
```
chrismabdo, line 29: Could consider creating a whole Text object and putting the methods in there

spec/order_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      subject.add_dish("Beef wellington")
+      expect(subject.check_price(19.98)).to be_truthy
+    end
+
+    it "raises an error if given price does not match total price" do
+      expect { subject.check_price(20.00) }.to raise_error("Incorrect price")
+    end
+  end
+
+  describe '#complete' do
```
chrismabdo, line 34: Great use of doubles here

PR #1807, opened 2020-12-06
Takeaway Challenge

kikidawson:


kikidawson:
Next time, I will have a longer planning session. I think I moved on from planning and I still hadn't a clear idea of what classes I was going to use. I tried to do two classes but could quite make sense of it in my head. Looking at my code now, I can see a clearer image of what the different classes should be.

I didn't get onto the texting part but don't think this was an issue as I was working pretty hard to try and understand the classes.

I was struggling to name things but this could have been made easier with a proper planning session.

kikidawson:
USER STORIES

As a customer

As a customer

As a customer

lib/order.rb
```diff
@@ -0,0 +1,66 @@
+class Order 
+  attr_reader :basket, :selected_dish, :menu, :total_price
+
+  def initialize
+    @menu = [
```
PiperS52, line 5: Some tasty options here!...although could maybe have been placed into a Menu class

PiperS52, line 5: I like the structure of your hash, with the added availability info

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def find_dish_details(dish)
+    # [0] had to select the 0th index because the hash is in an array
+    @selected_dish = @menu.select { |item| 
+      item[:name] == dish
+    }[0]
+  end
+
+  def make_unavailable(dish)
```
PiperS52, line 29: I like this - neat method

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+    }
+  end
+
+  def find_dish_details(dish)
+    # [0] had to select the 0th index because the hash is in an array
+    @selected_dish = @menu.select { |item| 
+      item[:name] == dish
+    }[0]
+  end
+
```
PiperS52, line 28: Some of these methods correspond to attributes you might associate with a Menu rather than the actual order, and could maybe be moved over to a Menu class - ie. find_dish_details, make_unavailable

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+    find_dish_details(dish)
+
+    fail "#{dish} is not available" unless @selected_dish[:available]
+    
+    update_selected_dish(quantity)
+
+    @basket << @selected_dish
+  end
+
+  def update_selected_dish(quantity)
```
PiperS52, line 44: I like this ability to update the order. Maybe I've overlooked something, but couldn't see how to update (quantity of) a specific item in the basket

lib/order.rb
```diff
@@ -0,0 +1,66 @@
...
+      @total_price += item[:price]
+    }
+  end
+
+  def total
+    sum_of_prices
+    @total_price
+  end
+
+  def print_summary
```
PiperS52, line 60: Really nice method. I think the rounding on the value for 'total due' may just need tweaking.

spec/order_spec.rb
```diff
@@ -0,0 +1,61 @@
+require 'order'
+
+describe Order do
```
PiperS52, line 3: Good use of the let method

spec/order_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    end
+  end
+
+  describe '#print_summary' do
+    it 'prints the order so far' do
+      order.add_to_basket(chosen_dish, chosen_quantity)
+      expect(order.print_summary).to eq order.basket
+    end
+  end
+end
```
PiperS52, line 61: All rspec tests pass - excellent. It may just require a few more tests to confirm method functionality.

PiperS52:
Great work. The main functionality is there, it may, as you already mentioned, just be worth exploring how to separate it, assigning responsibility into different classes.

PR #1808, opened 2020-12-06
Takeaway Challenge - Ollie Norman

ollienorman:


lib/takeaway.rb
```diff
@@ -0,0 +1,43 @@
...
+  attr_reader :current_order
+
+  def initialize(menu = Menu, twilio = TwilioSender)
+    @menu = menu.new
+    @twilio = twilio.new
+    @current_order = []
+  end
+
+  def add_to_order(dish)
+    @current_order << dish_finder(dish)
```
Will-Helliwell, line 15: All of your methods are really well abstracted: almost all around 3 lines max and descriptive names so that it is easy to follow the logic.

README.md
```diff
@@ -1,83 +1 @@
...
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
-
-Notes on Test Coverage
-------------------
-
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+Fun challenge! Great to work on thinking about how classes interact and how to incorporate gems. Cool feeling to get sent a text thanks to some code I wrote - makes it feel 'real'. If I were able to spend more time on it I'd particularly like to think about how to refactor the way my Dish class is used and serves information to other classes - I think it's a bit messy. I'd also like to spend some time looking into how I'd have written rspec tests for the Twilio code, and how to mock the stuff it's accessing from the gem. I had some thoughts on refactoring the class that it pulls in as a dependency injected class instead and going from there, but didn't get round to getting it implemented.
```
Will-Helliwell, line 1: Only thing I would suggest adding is a quick 'how to use' section so that I as the 'customer' can clearly see how it is intended to be interacted with.

spec/menu_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'menu'
+
+describe Menu do
+  let(:dish) { double :dish, details: { :name => "Pie", :price => 5 } }
```
Will-Helliwell, line 4: Very good use of mocking and doubling - as far as I can tell without playing with the code your tests are isolated from class to class

lib/takeaway.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def add_dish(name, price)
+    @menu.add_dish(name, price)
+  end
+
+  def print_menu
+    @menu.print_menu
+  end
+
+  private
```
Will-Helliwell, line 33: Good use of private methods

Will-Helliwell (commented):
This is all really good - great structure, well-refactored, clear to follow. 

The only bit I'm not 100% clear on (and I know I'm splitting hairs here) is exactly how it is used. Is it ready for the customer to interact with or does the restaurant have to upload their menu first (if so, it might be an idea to create a separate menu file that can be loaded through requiring)? If it is ready for the customer, what commands should they be typing to use it?

PR #1809, opened 2020-12-06
My attempt at the takeaway challenge

natashamcintyre:
I had several issues with rspec (drove me round the bend), but I'd really like to know if you think my structure makes sense and if my code is readable / easily changeable. I'm certain there are bits that aren't!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,45 @@
-My thoughts on the takeaway challenge
+## Takeaway Challenge ##
+==================
+
+## How to Use
+
+```shell
+irb -r './lib/order.rb'
+```
+Create a new order, then choose any of the following methods:
+view_menu --> displays the menu
+select(dish, quantity) --> eg ("dishname", 3)
+place_order --> this submits your order
+
+eg.
+```irb
+my_order = Order.new
+Welcome! Please use view_menu to see the menu.
+To add a dish to your order, use select(dish, quantity)
+To place your order, use place_order.
+2.6.5 :002 > my_order.view_menu
+Pizza: £6.50
+Spaghetti: £5.75
+Salad: £5.25 => nil
+2.6.5 :003 > my_order.select("pizza",3)
+=> "3 pizza(s) added to your order."
+```
+
+
+## My thoughts on the takeaway challenge
 
 I THINK I'M MAKING A BIT OF A MEAL OUT OF THIS CHALLENGE (pun fully intended)
 
-I have really struggled with this challenge. Mostly because of trying to do RSpec, but also I don't understand what check_total is comparing. I've spent hours on it and not even attempted the phone stuff yet.
+I have really struggled with this challenge and found it very frustrating. I spent hours on it before even attempting the phone stuff!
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -1,12 +1,45 @@
...
+
+```shell
+irb -r './lib/order.rb'
+```
+Create a new order, then choose any of the following methods:
+view_menu --> displays the menu
+select(dish, quantity) --> eg ("dishname", 3)
+place_order --> this submits your order
+
+eg.
```
SarahJoniPower, line 14: This is great! Makes it all super clear and easy to understand how to interact with the programme

README.md
```diff
@@ -1,12 +1,45 @@
...
+2.6.5 :002 > my_order.view_menu
+Pizza: £6.50
+Spaghetti: £5.75
+Salad: £5.25 => nil
+2.6.5 :003 > my_order.select("pizza",3)
+=> "3 pizza(s) added to your order."
+```
+
+
+## My thoughts on the takeaway challenge
```
SarahJoniPower, line 29: This is also really useful, really nice to get a sense of how you feel about everything and an insight into your thinking

lib/order.rb
```diff
@@ -2,23 +2,42 @@
 require_relative './invoice.rb'
 
 class Order
-  attr_reader :order, :total, :order_summary
+  attr_reader :total, :order_summary, :order_output
 
   def initialize(menu = Menu.new, invoice = Invoice.new)
```
SarahJoniPower, line 7: This is great! No hard coding of menu or invoice

README.md
```diff
@@ -1,12 +1,45 @@
...
+I have really struggled with this challenge and found it very frustrating. I spent hours on it before even attempting the phone stuff!
+
+I would like to focus on my understanding of the SOLID principles (the structure of the classes and methods, high cohesion - low coupling etc), but I feel like RSpec syntax, mocks, stubs etc are getting in the way. I'm spending ages trying to find out how to make an RSpec test work when I know the code is (mostly) working because I've tested it multiple times in irb. For example, in my Dish class I format the price of the dish. When I declare price as a double in RSpec, it then errors in this line because it can't format the double because it's not a float. I can't figure out how to tell RSpec to allow this to happen. I have therefore commented out some of the tests, which has reduced my test coverage.
+
+I also don't understand the 3rd user story, as I don't understand what 2 things need comparing. As far as I can tell, you have to work out the sum of the order and compare that to... the sum of the order...??? Please tell me what I'm missing!
```
SarahJoniPower, line 37: I also didn't understand what they wanted here

lib/order.rb
```diff
@@ -2,23 +2,42 @@
 require_relative './invoice.rb'
 
 class Order
```
SarahJoniPower, line 4: One thing I might say is I find this class name here a little confusing because you also have the instance variable 'order', you would end up called order.order - so maybe some re-naming here might make it slightly clearer to a new user

lib/order.rb
```diff
@@ -37,27 +56,17 @@ def calculate_total(order)
   end
 
   def prints_order_summary(order)
```
SarahJoniPower, line 58: I also had this as one method originally but then thought maybe broken into 'summary of items' and 'summary of price' would help with the 'one method does one thing' idea. Then you could write another method which calls one then the other if that makes sense

spec/order_spec.rb
```diff
@@ -2,12 +2,16 @@
 
 describe Order do
   let(:dish) { double("dish") }
-  let(:menu) { double "menu" }
-  subject { described_class.new(menu) }
+  let(:menu) { double("menu") }
+  let(:order) { double("order") }
```
SarahJoniPower, line 6: Great use of doubles! Something which still confuses me so much but this looks great!

SarahJoniPower (commented):
Hi Natasha! 
I feel unable to comment on all of this as I didn't get as far, but I think what I do understand looks great and your communication around what you've done and why is brilliant. My only comment would be you might want to re-name your classes to make them more readable. 
Hope the comments are helpful,
Sarah

PR #1810, opened 2020-12-06
first commit

Ben-glitch-cloud:
This was my first attempt of a weekly challenge.

### DIFF
--- a/docs/menu.rb
+++ b/docs/menu.rb
@@ -1,0 +1,35 @@
+class Menu  
+
+    attr_reader :order
+
+    def initialize
+       @list = { 'Latte' => 2.90, 'Cappuccino' => 2.80, 'Cake' => 3.50, 'Sandwich' => 4.00 }    
+
+       @order = []
+    end  
+
+    def list_menu   
+        @list.each do |key, value|
+            puts "Item #{key} Cost #{value}"
+        end
+    end  
+
+    def select_item(item, num) 
+        if num >= 1 || num == nil
+            @order << { item => num }   
+            puts "#{num}x #{item} have been added to your basket."   
+        else 
+            puts "Plaese input a number of items."
+        end 
+    end  
+
+    def check_total_order  
+        @order
+    end 
+end  
+
+menu = Menu.new 
+
+menu.list_menu  
+
+menu.select_item('Latte', 2)
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,0 +1,15 @@
+require './docs/menu.rb' 
+
+describe 'Menu' do
+    it 'should return a hash with 4 item inside' do  
+        menu = Menu.new
+        expect(menu.list).to eq({ 'Latte' => 2.90, 'Cappuccino' => 2.80, 'Cake' => 3.50, 'Sandwich' => 4.00}) 
+    end 
+
+    describe 'list_menu' do
+        it 'should return alist of strings' do
+            menu = Menu.new 
+            expect(menu.list_menu.class).to eq(String)
+        end
+    end
+end


Ben-glitch-cloud:
No

docs/menu.rb
```diff
@@ -0,0 +1,35 @@
+class Menu  
+
+    attr_reader :order
+
+    def initialize
+       @list = { 'Latte' => 2.90, 'Cappuccino' => 2.80, 'Cake' => 3.50, 'Sandwich' => 4.00 }
```
lukewickens1989, line 6: Might be worth setting this to a constant variable as it is probably rarely that the menu will change.

docs/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+    attr_reader :order
+
+    def initialize
+       @list = { 'Latte' => 2.90, 'Cappuccino' => 2.80, 'Cake' => 3.50, 'Sandwich' => 4.00 }    
+
+       @order = []
+    end  
+
+    def list_menu   
+        @list.each do |key, value|
```
lukewickens1989, line 12: Logical code to display the menu. SRP well used.

docs/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+            puts "Item #{key} Cost #{value}"
+        end
+    end  
+
+    def select_item(item, num) 
+        if num >= 1 || num == nil
+            @order << { item => num }   
+            puts "#{num}x #{item} have been added to your basket."   
+        else 
+            puts "Plaese input a number of items."
```
lukewickens1989, line 22: What you could do here is to set num to equal 0 in the argument. This would reduce the code used as you wouldnt need to account for the number being nill and there wouldn't be a need for the if statement.

docs/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+        end 
+    end  
+
+    def check_total_order  
+        @order
+    end 
+end  
+
+menu = Menu.new
```
lukewickens1989, line 31: This could be instantiated in the initialization of the file (something like menu = Menu.new then attribute it to a global variable.)

docs/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+    def select_item(item, num) 
+        if num >= 1 || num == nil
+            @order << { item => num }   
+            puts "#{num}x #{item} have been added to your basket."   
+        else 
+            puts "Plaese input a number of items."
+        end 
+    end  
+
+    def check_total_order
```
lukewickens1989, line 26: Nice, simple and straight forward method that is well named

spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
+require './docs/menu.rb' 
+
+describe 'Menu' do
+    it 'should return a hash with 4 item inside' do  
+        menu = Menu.new
```
lukewickens1989, line 5: You could use let(:menu) { described_class.new } at the top of your describe block which would mean that you wouldnt need to call it multiple times.

spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
+require './docs/menu.rb'
```
lukewickens1989, line 1: Here you can just put require "menu"

spec/menu_spec.rb
lukewickens1989: It might be worth changing this test so that it is flexible with any menu. If another item is added or a price is changed, this test would fail.

spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe 'Menu' do
+    it 'should return a hash with 4 item inside' do  
+        menu = Menu.new
+        expect(menu.list).to eq({ 'Latte' => 2.90, 'Cappuccino' => 2.80, 'Cake' => 3.50, 'Sandwich' => 4.00}) 
+    end 
+
+    describe 'list_menu' do
+        it 'should return alist of strings' do
+            menu = Menu.new 
+            expect(menu.list_menu.class).to eq(String)
```
lukewickens1989, line 12: Here you could include one of the strings you expect to receive. This would just confirm that you are definately getting back what you expect. expect(menu.list_menu).to include("Item Latte Cost 2.90")

docs/menu.rb
lukewickens1989: Key and Value aren't very descriptive of what they represent. It may be easier for other people to read if you call them item and price or something more indicative of what they represent.

lukewickens1989 (commented):
Nice work! The code you have written is well organised and looks like it functions well. The tests look like they cover some of the code functionality however I think this could be improved/enhanced if more time was available. Overall a good first commit.

PR #1812, opened 2020-12-07
Takeaway challenge

PiperS52:
Partial solution - still to create TextMessage class

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,17 @@
+class Menu
+  attr_reader :dishes
+
+  DISHES = { 'hawaian pizza' => 6,
+             'beef noodles' => 5,
+             'chicken tikka massala' => 6,
+             'pepperoni pizza' => 5 }
+
+  def initialize(dishes = DISHES)
+    @dishes = dishes
+  end
+
+  def display
+    @dishes.map { |k, v| "#{k}: £#{v}" }
+  end
+
+end
--- a/spec/order_spec.rb
+++ b/spec/order_spec.rb
@@ -1,0 +1,56 @@
+require 'order'
+
+describe Order do
+  let(:order) { Order.new }
+  let(:mock_menu) { double({ 'chicken tikka massala' => 6, 'hawaian pizza' => 6 }) }
+  # allow(:mock_menu).to respond_to(:map)
+
+  it 'creates an instance of the Order class' do
+    expect(order).to be_instance_of(Order)
+  end
+
+  it 'begins with an empty trolley' do
+    expect(order.trolley).to be_empty
+  end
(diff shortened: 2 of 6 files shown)


spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'order'
+
+describe Order do
+  let(:order) { Order.new }
+  let(:mock_menu) { double({ 'chicken tikka massala' => 6, 'hawaian pizza' => 6 }) }
```
kikidawson, line 5: Good use of a double.

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  describe '.add_item method' do
+    it 'takes two arguments' do
+      expect(order).to respond_to(:add_item).with(2).arguments
+    end
+
+    it 'adds an item to the trolley' do
+      expect { order.add_item('chicken tikka massala', 2) }.to change{ order.trolley.length }.by(1)
+    end
+  end
+
```
kikidawson, line 25: V handy use of comments

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        order.trolley.each do |key, val|
+          if k == key
+            "#{k}, £#{v*val}"
+          end
+        end
+      end
+      expect(summary).to eq('chicken tikka massala, £12')
+    end
+  end
+
```
kikidawson, line 41: I think, if you added @total to the initialized method in Order and added an attribute reader for total. Then you could remove lines 46 - 53 and have the line below instead. Although, thinking about it, this may be what you were avoiding so you could use mock_menu instead. Please ignore if not relevant. 

`expect(order.total).to eq('Total: £12')`

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+             'beef noodles' => 5,
+             'chicken tikka massala' => 6,
+             'pepperoni pizza' => 5 }
+
+  def initialize(dishes = DISHES)
+    @dishes = dishes
+  end
+
+  def display
+    @dishes.map { |k, v| "#{k}: £#{v}" }
```
kikidawson, line 14: Would it work to use different names for the key and value here. Just to make it clear what they refer to. Alternatively, you could add what they refer to in a comment above.

kikidawson (commented):
Easy to follow and clear to read. Would be even clearer if the key/values in hashes were labeled differently. I enjoyed reading your code. 

I did add a small note in your order_spec but please ignore it if it isn't relevant.

I particularly liked your use of the restaurant class. It makes so much sense now I'm looking at it - it's helped me understand classes a bit better!

PR #1813, opened 2020-12-07
Takeaway Challenge Unfinished

sujee09:
Completed up to user story 2.

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,2 @@
+class Takeaway
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,16 +1,24 @@
 class Takeaway
-  attr_reader :menu
+  attr_reader :menu, :order, :total
 
   def initialize
     @menu = {
-      Fish: 5.00,
-      Scampi: 5.50,
-      Chips: 1.50,
-      Gravy: 2.00 
+      fish: 5.00,
+      scampi: 5.50,
+      chips: 1.50,
+      gravy: 2.00 
     }
+    @order = []
+    @total = 0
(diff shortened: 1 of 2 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
...
+      scampi: 5.50,
+      chips: 1.50,
+      gravy: 2.00 
+    }
+    @order = []
+    @total = 0
+  end
+
+  def show_menu
+    @menu.each { |meal, price| puts "-- #{meal} : £#{price} --" }
```
Aracho1, line 16: Great use of enumerator to print out the menu! Clever

lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def add_item(meal, quantity = 1)
+    quantity.times { @order << meal }
+    quantity.times { @total += @menu[meal.to_sym] }
```
Aracho1, line 21: Clever way of adding to the total! I wish I had thought of that.

lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
+class Takeaway
+  attr_reader :menu, :order, :total
+
+  def initialize
+    @menu = {
```
Aracho1, line 5: In line with single responsibility principle, you could consider making the menu a separate class

Aracho1 (commented):
The methods and variables are clear and easy / logical to follow. Creating a total variable and adding to it is very clever and simple! I wish I had thought of that. Perhaps you could make the menu a separate class since it has a different responsibility? And the 3rd user story requires the total to match the customer's total - which is missing.

PR #1814, opened 2020-12-07
Sarah Power - pull request 

SarahJoniPower:
I really didn't get much of this done unfortunately. These three user stories should be there:

As a customer

As a customer

As a customer
I would like to check that the total I have been given matches the sum of the various dishes in my order.

Then I had started to think menu should be a separate class, and started to move that, but ran out of time, and didn't even get to using Twilio API! :(

### DIFF
--- a/lib/restaurant.rb
+++ b/lib/restaurant.rb
@@ -1,21 +1,23 @@
+require './lib/menu.rb'
+
 class Restaurant
   attr_accessor :menu, :basket, :total_amount
     
-  def initialize
-    @menu = { 'Margarita': 7, 'Bufala': 9, 'Neapolitan': 10, 'Rocket salad': 5 }
+  def initialize(menu = Menu.new)
+    @menu = menu
     @basket = {}
     @total_amount = 0
   end
     
   def list_menu
-    @menu.each do |dish, price|
+    @menu.dishes.each do |dish, price|
       puts "#{dish} : #{price}"
     end 
   end
 
   def order(dish, quantity)
     @basket[dish] = quantity
-    @total_amount += @menu[dish.to_sym] * quantity
+    @total_amount += @menu.dishes[dish.to_sym] * quantity
   end
 
   def total_basket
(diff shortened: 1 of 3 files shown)


lib/restaurant.rb
```diff
@@ -0,0 +1,43 @@
+require './lib/menu.rb'
+
+class Restaurant
+  attr_accessor :menu, :basket, :total_amount
```
natashamcintyre, line 4: I like the names of your variables, they are easy to understand

lib/restaurant.rb
```diff
@@ -0,0 +1,43 @@
...
+    
+  def list_menu
+    @menu.dishes.each do |dish, price|
+      puts "#{dish} : #{price}"
+    end 
+  end
+
+  def order(dish, quantity)
+    @basket[dish] = quantity
+    @total_amount += @menu.dishes[dish.to_sym] * quantity
```
natashamcintyre, line 20: Oh my goodness this is a much easier structure than mine 👏

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  it 'responds to the method .total_basket' do 
+    expect(subject).to respond_to(:total_basket)
+  end
+
+  it 'responds to method .total_price' do
+    expect(subject).to respond_to(:total_price)
+  end
+
+#   describe '#list_menu' do
+#     it 'should output a menu list from the Restaurant' do
```
natashamcintyre, line 25: I don't know if this is an appropriate or even useful test, but there's something where you can check that it sends something to output. expect { subject.list_menu }.to output('optional something').to stdout... or something like that

natashamcintyre (commented):
Your code is really easy to read, very straight forward and I feel could be easily changed 🙌 

The only bit I think would make it harder to change is where you've hard coded dishes into the tests. I think you get around this using doubles, but I am really not confident on how to implement those so I might be wrong.

Thank you for letting me look at this, I'd like to try to apply more simple methods that you have used for the menu and your well named variables to my own code if that's okay 😃

PR #1815, opened 2020-12-07
Takeaway challenge for code review

AJSMonty:
User stories covered 

- print menu
- select dish (although struggled to include a module('Dishes') and initialize method in the same class('Takeaway')
- add dish to order (same issue as above with not having an instance variable('order') to be able to add more than one 'select_dish' to
- send text with confirmation (had a hack, couldn't figure Twilio out, extra reading + hacking needed)

### DIFF
--- a/lib/dishes.rb
+++ b/lib/dishes.rb
@@ -1,0 +1,22 @@
+module Dishes
+
+  def initialize
+    @menu = [
+        {:dish => 'pasta', :price => 8},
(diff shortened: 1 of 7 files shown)


lib/dishes.rb
```diff
@@ -0,0 +1,22 @@
+module Dishes
+
+  def initialize
```
calavell, line 3: Interesting use of module

PR #1820, opened 2021-01-17
Takeaway challenge

mscwilson:


spec/order_manager_spec.rb
```diff
@@ -1,15 +1,16 @@
 require "order_manager"
-require "order"
 
 describe OrderManager do
```
bullhornfixie, line 3: Like the structure you've adopted of subclasses and a masterclass to manage everything.

bullhornfixie (commented):
- Regular and informative commits give an instant snapshot of the project building blocks. 
- Great use of CSV library for menu 
- Code is consistent and neat 
- Advanced syntax such as regex 
- Program design has been thought out 
- Possibly, the volume of classes is high and could simplify

PR #1821, opened 2021-01-17
Takeaway Challenge

HolsDuckett:


spec/menu_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'menu'
+
+describe Menu do
+  subject (:menu) { described_class.new(dishes) }
```
hacaravan, line 4: Does this work this way round? Like putting dishes in there before you create it in the next line? Genuine question cause I don't know

spec/order_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+require 'menu'
+
+describe Order do
+  subject(:order) { described_class.new(menu) }
+  let(:menu) { double(:menu, print: printed_menu) }
+  let(:printed_menu) { "Pasta - £7, Pizza - £9, Canlone - £9" }
+  let(:order_list) { {Pizza: 3} }
+
+  it "should initialze with an empty order list array" do
+    expect(order.order_list).to be_instance_of(Hash)
```
hacaravan, line 11: I think it's usually better to do one expectation per description - so separate this expect and the next one into separate it blocks

spec/order_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'order'
+require 'menu'
```
hacaravan, line 2: Do you need this require menu if in the order.rb file you're requiring menu?

lib/order.rb
```diff
@@ -0,0 +1,21 @@
...
+
+attr_reader :order_list, :menu
+
+  def initialize(menu)
+    @menu = menu
+    @order_list = {}
+  end
+
+  def add(dish, quantity)
+    @order_list[dish] = quantity
```
hacaravan, line 13: What happens if you want to add something more than once? You want pasta but then matey over here wants pasta too and adds another one, it looks like this will only add one lot to the order, rather than two

PR #1822, opened 2021-01-17
Charlie's takeaway challenge

charliefischer:


lib/menu.rb
```diff
@@ -0,0 +1,16 @@
+class Menu
+  attr_accessor :dishes
+  def initialize
+    @dishes = { "onion bhaji" => 2.95, "vegetable samosa" => 2.95, "aloo pakora" => 2.95,
+                "chicken tikka" => 3.95, "mixed kebab" => 4.95 }
+  end
+
+  def read_menu
+    @dishes
```
ddottyler, line 9: Everything good.. Could it be slightly more aesthetically pleasing?

lib/takeaway_menu.rb
```diff
@@ -0,0 +1,58 @@
+require_relative 'menu'
+require_relative 'text'
+
+class Takeaway
+  attr_accessor :total
```
ddottyler, line 5: Do we want to be able to alter this from the command line? is that what accessor does? 
Or do we just want to be able to alter it from within the programme?

lib/takeaway_menu.rb
```diff
@@ -0,0 +1,58 @@
...
+  attr_reader :dishes
+
+  def initialize
+    @dishes = Menu.new.dishes
+    @basket = {}
+    @total = 0
+    @send_text = Text.new
+  end
+
+  def order(dish, quantity = 1)
```
ddottyler, line 15: clever way of getting the quantity in.. I missed this entirely and it caused carnage

lib/text.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
+
+    @client = Twilio::REST::Client.new account_sid, auth_token
+
+    @from = ENV['SECRET_TWILIO_NUMBER']
+    @to = ENV['MY_SECRET_NUMBER']
+  end
+
```
ddottyler, line 14: Good idea to hide account_sid/auth_token/and numbers - wish I had

ddottyler (commented):
I think overall it's really good and there's a lot of functionality in there with some clear rspec to back it up. 

I might be wrong, but it looks like there's not much isolation in the testing so perhaps this could be an area to focus on moving forward (mine weren't isolated either!)

Lastly, the code is readable but there's a lot going on. I wonder if it could be streamlined?

PR #1823, opened 2021-01-17
Weekend challenge - Karsten Finlay

KarstenFinlay:


.gitignore
```diff
@@ -3,3 +3,6 @@
 
 # Local cache of Rubocop remote config
 .rubocop-*
+
+# Security stuff
+twillio_creds.sh
```
acavalla, line 8: I didn't get this far so might be correct but looks like a typo of twilio

lib/takeaway.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def create_message
+    expected_time = Time.now + 3600
+    formatted_time = expected_time.strftime("%H:%M")
+
+    @api_client.messages.create(
+      body: "Thank you! Your order was placed and will be delivered before #{formatted_time}",
+      to: ENV["PHONE_NUMBER"],
+      from: "+44 7588 065563"
```
acavalla, line 88: If this is your phone number i don't think you want it on github lol

acavalla (commented):
This code is elegant, easy to understand and way beyond my capability levels so i'm afraid i can't offer you any hints - just adulation

.gitignore
KarstenFinlay: Yes, that is indeed a typo, will change that. Thanks!

lib/takeaway.rb
KarstenFinlay: No this is the Twilio phone number I bought with the trial account. I don't think it matters if people get a hold of that.

acavalla:
You should add in some instructions in the readme though because I've checked it out and I can't get it to work - I see the prompt is in the takeaway class but I don't know what arguments i need to initialize a new takeaway or menu so i can't try it out

KarstenFinlay:
As acavalla pointed out, I need to update my readme, as it doesn't have any instruction. Unfortunately you have to hard code your number, working on making it so that it is not hard coded. Also if you want to run the program just run the app by doing `ruby app.rb` in the takeaway challenge directory.

lib/takeaway.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def show_order
+    if @current_order != nil
+      total = @current_order.order_total
+      puts "Cost: £%<total>.2f" % { total: total }
+    else
+      puts "Order is empty"
+    end
+  end
+
```
acavalla, line 40: Could display a summary of the order as well as the price

PR #1824, opened 2021-01-17
RTurney takeaway challenge partial solution - user story 3 

RTurney:
Completed up to user story 3. 

Features: 
Can create a takeaway object > requires a menu object > menu requires a dictionary of available dishes
Can list the menu dishes from the takeaway
Can order meals from the menu 
Can see your current basket
Can see the order summary (dishes and sum total)

### DIFF
--- a/lib/basket.rb
+++ b/lib/basket.rb
@@ -1,0 +1,3 @@
+class Basket
+
+end
--- a/lib/basket.rb
+++ b/lib/basket.rb
@@ -1,3 +1,5 @@
+require_relative 'menu'
+
 class Basket
   attr_reader :contents
 
@@ -7,6 +9,18 @@ class Basket
 
   def add_dish(dish, quantity = 1)
     quantity.times { contents << dish }
+  end
+
+  def total(menu)
+    sum = 0
+    calculate_total(menu, sum)
+  end
+
+  private #-----------------------------
+
+  def calculate_total(menu, sum)
+    contents.each { |selection| sum += menu.dishes[selection.to_sym] }
+    sum
   end
 
 end
--- a/lib/basket.rb
+++ b/lib/basket.rb
@@ -16,6 +16,11 @@ class Basket
     calculate_total(menu)
   end
 
+  def order_summary(menu)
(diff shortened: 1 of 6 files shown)


lib/menu.rb
```diff
@@ -5,9 +5,7 @@ def initialize(dishes)
   end
 
   def display
-    dishes.map do |key, value|
-      "#{key}: £%.2f" % [value]
-    end.join(", \n")
+    dishes.map { |key, value| "#{key}: £%.2f" % [value] }.join(", \n")
```
PetarSimonovic, line 8: This seems like a very straightforward way to add the dishes (I used a completely separate class but think this is cleaner)

lib/basket.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'menu'
+
+class Basket
+  attr_reader :contents
+
+  def initialize
+    @contents = {}
+  end
+
+  def add_dish(dish, quantity = 1)
```
PetarSimonovic, line 10: Being able to specify quantity is useful. I meant to implement this but didn't get round to it

lib/menu.rb
```diff
@@ -0,0 +1,11 @@
+class Menu
+  attr_reader :dishes
+  def initialize(dishes)
+    @dishes = dishes
```
PetarSimonovic, line 4: It's great that the user can supply their own menu - it makes the code very flexible (I hard-coded my menu),

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'menu'
+describe Menu do
+
+  subject(:menu) { described_class.new(dishes) }
+  let(:dishes) do
```
PetarSimonovic, line 5: Good use of doubling. I should have done this!

PetarSimonovic (commented):
The code seems clean and clear to me. All of the methods are short and well-labelled so it's easy to follow. The RSpec tests are also clear, making good use of doubling.

I would update the README file and provide step-by-step instructions on how to use the software.

PR #1825, opened 2021-01-17
Hugh Takeaway Challenge

hacaravan:


README.md
```diff
@@ -14,6 +14,43 @@ Takeaway Challenge
 
  ```
 
+ Intro
```
HolsDuckett, line 17: This is great - love how its clearly laid out and explains your process

lib/menu.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def select_price(name)
+    check_non_existent(name)
+    get_price_from(name)
+  end
+
+  private
+
+  def check_already_existing(name)
```
HolsDuckett, line 30: Great use of SRP - each method is clear and concise

HolsDuckett (commented):
I think what would be best is to get a coach to review this

PR #1828, opened 2021-01-17
Katrina's Takeaway Challenge

CodeRed30:


JosephSpraggon (commented):
Really nice code! You have very DRY code and a good understanding of method extraction. I can't comment too much on the TWILO stuff as I didn't get a chance to look at that myself. Your tests are a little hard coded I can see but I don't have the greatest understanding of mocks, stubs and doubles myself yet, so I don't feel like I can help to improve them!

All round this looks like great work :)

PR #1830, opened 2021-01-17
Anna's takeaway challenge

acavalla:


lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+  def initialize
+    @menu = Menu.new
+    @basket = []
+  end
+
+  def show_menu
+    print menu.dishes
+  end
+
+  def add(like, count = 1)
```
KarstenFinlay, line 15: `like` is throwing me off a little, `like` is a dish in this context.

Potentially changing variable name `like` to `dish` for easier understanding.

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+    order_present?(like)
+    add_to_basket(like, count)
+    update_basket_total
+    confirmation_message(like, count)
+  end
+
+  def complete_order
+    raise "Sorry, incorrect total." if wrong_price?
```
KarstenFinlay, line 23: Just out of interest, did this ever fail?

lib/takeaway.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def wrong_price?
+    basket_total != basket.map { |x| x.values[0] * x.values[1] }.sum
+  end
+
+  def confirmation_message_2
+    puts "You have ordered:"
+    basket.map { |x| puts x.values[1].to_s + " " + x.keys[0].downcase + "(s)"}
+    puts "Total: £#{basket_total.to_i}"
+  end
```
KarstenFinlay, line 55: Really like this block of code. Something that I might use for mine.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect(subject.menu.dishes).to be_a Hash
+    end
+
+    it "should print dishes on show_menu" do
+      expect { subject.show_menu }.to output.to_stdout
+    end
+  end
+
+  context 'ordering food' do
+    it 'should only accept items from the menu' do
```
KarstenFinlay, line 17: Could use double instead of hard coding.

`let(:invalid_choice) { double("invalid_choice") }`

However, you won't really be changing the  `invalid_choice`, so maybe hard coding here is fine.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      expect { subject.add "Boiled goose" }.to raise_error("Sorry, no boiled goose.")
+    end
+
+    it 'should output confirmation' do
+      message = "3x plantastic burger(s) added to your basket."
+      expect { subject.add 'Plantastic burger', 3 }.to output(message).to_stdout
```
KarstenFinlay, line 23: Here we want to mock our order. If we were to change plantastic burgers because we can't produce them, then our tests would fail.

KarstenFinlay (commented):
Looks really good. I like that it is easy to read and understand. 

Keep me updated on how you get on using the Twilio API!

PR #1831, opened 2021-01-18
Partial solution - still to complete user story 4 and integrate Twilio/ENV

twigz826:


lib/order.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'menu'
+
+class Order
+
+  attr_reader :basket, :basket_total
```
day-katy, line 5: I'm looking through your code and can't find anything that uses :basket_total - is it something you're planning on using, or something that got deleted?

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'menu'
+
+describe Menu do
+
+  let(:menu) { described_class.new }
```
day-katy, line 5: Only flagging this because I want you to explain "described_class.new" to me!

spec/order_spec.rb
```diff
@@ -0,0 +1,36 @@
+require 'order'
+
+describe Order do
+
+  let(:order) { described_class.new }
+  let(:dish) { "Battered cod" }
+  let(:price) { 3.99 }
```
day-katy, line 7: So this is where I think they want us to use doubles, which is what I tried to do. So I THINK if you did let(:dish){double}, it SHOULD give you a double that you can then use in testing. Because then you're keeping the information from the menu class separate from the order class....it's just actually then making it work in testing that gets me!

day-katy (commented):
Super clear code!

PR #1832, opened 2021-01-18
Bon Appétit!

Mornevanzyl:


README.md
```diff
@@ -14,70 +13,78 @@ Takeaway Challenge
...
 
-* Advanced! (have a go if you're feeling adventurous):
-  * Implement the ability to place orders via text message.
+```ruby
+require './lib/dish'
+require './lib/menu'
+require './lib/order'
 
-* A free account on Twilio will only allow you to send texts to "verified" numbers. Use your mobile phone number, don't worry about the customer's mobile phone.
+# Substitute your own dishes below
```
jshields123, line 47: useful readme instructions for usability

README.md
```diff
@@ -14,70 +13,78 @@ Takeaway Challenge
...
+order.open
 
+# Follow menu-prompts to add dishes to your order. Confirm order and you'll receive a SMS order confirmation.
+
+# Bon Appétit!
+```
 
-In code review we'll be hoping to see:
+## Approach
+The chosen approach to this application challenge was to implement three classes that played together nicely, but could be deployed independantly and re-used in other apps.
```
jshields123, line 66: Nice and clear plan put in place before heading into the code.

spec/features_spec.rb
```diff
@@ -0,0 +1,38 @@
+describe 'User Stories' do
+
+    # As a customer
+    # So that I can check if I want to order something
```
jshields123, line 4: I like the use of these user story spec. Not everyone would do that but keeps the user stories in mind.

spec/order_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    it 'accepts a dish and adds it to the order' do
+        expect(order.add(:dish)).to eq :dish
+    end
+
+    it 'is initialized with a Menu ojbect as arugument' do
+        expect(order.menu).to eq :menu
+    end
+    
+    it 'allows the user to select a number of dishes from the menu' do
```
jshields123, line 17: This spec file isn't testing all the methods that are present in order.rb. Might be worth getting them in there to check that to code works.

lib/order.rb
```diff
@@ -0,0 +1,90 @@
...
+        print
+        puts "Enter 'YES' to confirm order, leave blank to cancel order"
+        STDIN.gets.chomp.upcase == 'YES'
+    end
+
+    def print
+        summarise_order
+        puts "Your order summary is as follows:"
+        puts "---------------------------------"
+        @items_detail.each do |key, item|
```
jshields123, line 46: nice presentation of the order, dont think you need the puts though I may be wrong.

lib/order.rb
```diff
@@ -0,0 +1,90 @@
...
+        @total = 0
+        items_counter.each do |key, qty|
+            dish_detail = lookup_dish_on_menu(key, qty)
+            @items_detail[key] = dish_detail
+            @total += dish_detail[:subtotal]
+        end
+        @total = @total.round(2)
+    end
+
+    def lookup_dish_on_menu(key, qty)
```
jshields123, line 66: Not sure if this method belongs in order, with dish and menu named in the method I would try and get this into one of those classes for SRP.

lib/order.rb
```diff
@@ -0,0 +1,90 @@
+require 'twilio-ruby'
+
+class Order
+
+    attr_reader :menu, :items
+
+    def initialize(menu)
+        @menu = menu
+        @items = []
+        @items_detail
```
jshields123, line 10: I think this should be defined as to what is it here and not in the method if I am correct. I think it is given an empty hash in summarise order.

jshields123 (commented):
Great work, has got a lot of usability and well thought out displays to the user. I love the sms functionality too.

README.md
Mornevanzyl: Thanks Jack!

PR #1834, opened 2021-01-18
Beers n pizzas! 

jshields123:


lib/order.rb
```diff
@@ -0,0 +1,26 @@
+require_relative 'menu'
+require_relative 'takeaway'
```
Mornevanzyl, line 2: I'm assuming this is for further development?

lib/order.rb
```diff
@@ -0,0 +1,26 @@
...
+
+attr_reader :order, :menu, :total_value
+
+  def initialize
+    @order = Hash.new(0)
+    @menu = Menu::DISHES
+    @total_value = 0
+  end
+
+  def add_to_order(dish, amount = 1)
```
Mornevanzyl, line 14: Is this amount the quantity of this dish added to order? If so, it might be an idea to call it something like quantity as it might be confused with the price of the dish? Just a thought.

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'order'
```
Mornevanzyl, line 1: Do you need this? Even tough the Order class isn't instantiated from this file?

Mornevanzyl (commented):
Excellent clean implementation of the requirement. Clear and informative README. Left a couple of comments, but the underlying bones look great.

lib/order.rb
jshields123: Yeah I had a go at changing a few things but reverted back to this structure. Left those in.

lib/order.rb
jshields123: Ah yes that is true, amount could be confused with value.

lib/menu.rb
jshields123: You are right this isnt needed also. Thanks!

PR #1836, opened 2021-01-18
DT weekend 2 challenge

ddottyler:


lib/send_sms.rb
```diff
@@ -0,0 +1,16 @@
+require 'twilio-ruby'
+
+time = Time.new + 3600
```
charliefischer, line 3: nice saving this as a variable - didn't think of this myself but much easier to read

lib/send_sms.rb
```diff
@@ -0,0 +1,16 @@
+require 'twilio-ruby'
+
+time = Time.new + 3600 
+
+account_sid = "AC7c0dd1eab28593f5ec87bb03c409e40b"
```
charliefischer, line 5: I'm sure you got the same email as me from github and twilio saying that it would stop this SID and token from working. Need to save them as environmental variables so that they are saved on your command line. This lets only you run them - or someone else can run this with their details as they will save their own details in their command line

charliefischer: https://www.twilio.com/docs/usage/secure-credentials

lib/send_sms.rb
```diff
@@ -0,0 +1,16 @@
...
+auth_token = "0b767001604b77a4e1c3e6585c909586"
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+from = '+447862152030' # Your Twilio number
+to = '+447725810503' # Your mobile phone number
+
+client.messages.create(
```
charliefischer, line 12: might be more readable to have this section as a new method - keeps your code as single responsibility

lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+    @selected_dishes = []
+    @total = []
+  end 
+
+  def add_dish(dish)
+    @selected_dishes << dish
+  end 
+
+  def total
+    @selected_dishes.each { |item| @total << Takeaway.new.menu[item] }
```
charliefischer, line 17: much neater than what I went for here - works much better saving the order as an array of the names and then when you need the total calling the menu using those names

lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+    @selected_dishes << dish
+  end 
+
+  def total
+    @selected_dishes.each { |item| @total << Takeaway.new.menu[item] } 
+    @total.sum
+  end 
+
+  def place_order
+    puts "Your order details are below:"
```
charliefischer, line 22: nice idea summing up the order before sending the text

spec/order_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'order'
+
+describe Order do 
+
+  test_menu = Takeaway.new.menu
```
charliefischer, line 5: I believe you're supposed to use doubles here to create a fake Takeaway.new but seems like it would be much more code than this so not sure..?

spec/order_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    expect(subject.selected_dishes).to include cheeseburger
+  end 
+
+  it 'should add up total cost of selected dishes' do
+    subject.add_dish("Cheeseburger")
+    subject.add_dish("Hot dog")
+    expect(subject.total).to eq 17
+  end 
+
+end
```
charliefischer, line 20: not sure what the test coverage score was on this but to get it higher you could test that @selected_dishes and @total are empty arrays on initializing?

PR #1837, opened 2021-01-18
Takeaway Challenge

PetarSimonovic:


README.md
```diff
@@ -14,70 +14,60 @@ Takeaway Challenge
...
 
-Notes on Test Coverage
-------------------
+```rb
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+takeaway.place_order("+441234567890")
+
+```
+We'll let you know the bill total and send a confirmation to your phone number
```
RTurney, line 73: This instructions were clear and concise. I was able to follow them easily. The only issue is that I received an error when trying to send the confirmation text. 
`Twilio::REST::RestError ([HTTP 404] 20404 : Unable to create record)
The requested resource /2010-04-01/Accounts//Messages.json was not found
https://www.twilio.com/docs/errors/20404`

lib/bill_calculator.rb
```diff
@@ -0,0 +1,24 @@
...
+  def bill(order)
+    calculate(order)
+    puts "Total order cost is £#{@cost.to_f}"
+  end
+
+  def calculate(order)
+    order.each do |item|
+      @cost += BigDecimal("#{@menu.dishes[item]}")
+    end
+  end
```
RTurney, line 22: Good srp here. 
I liked your approach to returning a float, I wish I had done it the same way. Much more readable

lib/menu.rb
```diff
@@ -0,0 +1,30 @@
...
+    dishes
+    @dishes.each { |item, price| puts "#{item} - £#{price}" }
+  end
+
+  def on_menu?(dish)
+    dishes
+    @dishes.include?(dish)
+  end
+
+end
```
RTurney, line 30: Good easily readable code.

lib/takeaway.rb
```diff
@@ -0,0 +1,36 @@
...
+    @get_bill = BillCalculator.new
+  end
+
+  def display_menu
+    @menu.show_menu
+  end
+
+  def select_dish(dish)
+    fail "Selection unavailable" if @menu.on_menu?(dish) == false
+
```
RTurney, line 20: I think you could refactor this to use an 'unless' instead, which would increase readability and not require the '== false' statement.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'takeaway'
+require 'menu'
+
+describe Takeaway do
+
+  let(:menu) { Menu.new }
```
RTurney, line 6: I might be wrong with this, but in order to separate your unit tests via class, you would probably want to read in a double here for the menu class rather than initializing a new class in your takeaway tests.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    # it "checks if the dish is on the menu" do
+    #   expect(subject.select_dish("Hello Meow")).to be_on_menu
+    # end
+
+    it "adds a dish to the order if it's on the menu" do
+      subject.select_dish("Hello Meow")
+      expect(subject.order).to include("Hello Meow")
+    end
+
```
RTurney, line 24: Good SRP-ing here. Removed tests that no longer need to be present when new features added.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'takeaway'
+require 'menu'
```
RTurney, line 2: Use a double rather than reading in a different class

lib/takeaway.rb
```diff
@@ -0,0 +1,36 @@
...
+    @menu = Menu.new
+    @order = []
+    @get_bill = BillCalculator.new
+  end
+
+  def display_menu
+    @menu.show_menu
+  end
+
+  def select_dish(dish)
```
RTurney, line 18: Might be useful to add a 'quantity' variable here so that user's can order more that one helping of a dish at a time.

RTurney (commented):
Overall looks really good, not much I can comment on overall. Good job Pete!

PR #1838, opened 2021-01-18
takeaway-challenge - Oscar

OscarB89:


spec/order_spec.rb
```diff
@@ -21,10 +21,12 @@
   end
 
   describe '#total' do
+
     it 'should return the correct total' do
```
MegGosnell, line 25: indentation

MegGosnell (commented):
Really nice code. Easy to read and understand. To improve your code i would suggest using doubles when beginning to write your specs. It will help to familiarise yourself with them and strengthen your understanding of Mocking in general!

PR #1839, opened 2021-01-18
Code Review Commit

JoshSinyor:
A pull request for the Code Review session, 18/01/21.


lib/order.rb
```diff
@@ -29,12 +31,13 @@ def show_basket
   end
```
louiswicks, line 31: Very clever start method, welcoming customer/user to the restaurant, I never thought about it like that and potentially something I could absorb into my code.

lib/basket.rb
```diff
@@ -8,14 +8,16 @@ def initialize
   end
 
   def add(item)
-    menu_item = Menu_Item.new(item)
-    @basket.push(menu_item)
-    @sum += menu_item.price.to_f
+    item = Menu_Item.new(item)
+    @basket.push(item)
```
louiswicks, line 12: Clever use of basket term, good visualization of an online takeaway app.

lib/printer.rb
```diff
@@ -8,16 +8,11 @@ def initialize(col_width)
...
   def print(input)
     @input = input
-    clone_input
     justify_input
     print_input
     clear_job
   end
 
-  def clone_input
-
```
louiswicks: Curious to understand what the clone_input method would do, or might be used for, is it needed?

louiswicks (commented):
I think you have put a lot of in-depth thought into this challenge, you have features and characteristics that I haven't used which I think make the functionality very smooth. Great stuff, hope you feel better soon!

PR #1841, opened 2021-01-18
Takeaway Challenge - Reeshul

reeshul:
Struggled to implement the Twilio API - last user story is incomplete.


GlykeriaStr (commented):
My comments are more questions than suggestions!

PR #1844, opened 2021-02-05
Attempt at Takeaway Challenge

GraemeHedges:
Current Issues
--------------

- The final user story regarding a text response is yet to be implemented
- Some of the outputs are not particularly easy to read and include object memory references, I did have methods for these but removed them due to testing/time constraints
- My tests are very poor, I have not properly utilised doubles, stubs or mocks and so the tests are
not independant

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,8 +60,7 @@ I have removed some of the nil returns for brevity but retained the important ou
 
 Original README below
 
-====================================================================================================
-
+================================================================================
 
 
 
--- a/README.md
+++ b/README.md
@@ -1,3 +1,71 @@
+Takeaway Challenge Attempt
+==========================
+
+The original README text is retained below.
+
+
+Current Issues
+--------------
+
+- The final user story regarding a text response is yet to be implemented.
+- Some of the outputs are not particularly easy to read and include object memory references, I did have methods for these but removed them due to testing/time constraints.
+- My tests are very poor, I have not properly utilised doubles, stubs or mocks and so the tests are
+not independant of external classes.
+
+
+Approach
+--------
+
+Divide responsibilities up between classes, keep methods short, use private methods when
+appropriate, use a strict TDD approach to develop classes and methods, make sure tests are independent, stick to the Single Responsibility Principle as much as possible.
+
+Initial concept
+---------------
+
+A Takeaway class handles the taking of orders, gets the details of ordered items from a menu
+then adds them to a basket where a price is calculated and given to the customer, when the order
+is complete, the Takeaway will instruct a messanger object to send a text to the customer confirming
+the order and the total price.
+
+Example outputs
+---------------
+
+````
+2.6.5 :001 > # create a new instance of takeaway
+ => nil 
+2.6.5 :002 > PapaGraemes = Takeaway.new
+ => #<Takeaway:0x00007fb7c285cfb0 @order=[], @menu=#<Menu:0x00007fb7c285cf60 @dishes=[{:name=>"Pizza", :price=>10}, {:name=>"Parmo", :price=>9}, {:name=>"Calzone", :price=>11}]>> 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,3 +1,70 @@
+Takeaway Challenge Attempt
+==========================
+
+The original README text is retained below.
+
+
+Current Issues
+--------------
+
+- The final user story regarding a text response is yet to be implemented.
```
HarryCarton, line 10: Implement the last use story

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,47 @@
+require "takeaway"
+
+describe Takeaway do
+  before :each do
+    @menu = double
+    allow(@menu).to receive(:dish).with("Pizza") { [{ name: "Pizza", price: 10 }] }
+  end
+
+  describe '#menu' do
```
HarryCarton, line 9: Test seems redundant. Supposed to be subject.menu eq(@menu) ??

HarryCarton (commented):
my expert opinion

lib/basket.rb
```diff
@@ -0,0 +1,21 @@
...
+  
+  def receipt(order)
+    order.each { |item| puts "#{item[:name]} - £#{item[:price]}" } 
+    puts "Total - £#{price(order)}"
+    @total
+  end
+
+  private
+
+  def price(order)
```
EdwardAndress, line 17: minor point: you can use `sum` on a hash :)

lib/menu.rb
```diff
@@ -0,0 +1,19 @@
+require './lib/takeaway.rb'
+
+class Menu
+
+  attr_reader :dishes
```
EdwardAndress, line 5: How could you make your menu more flexible and allow for different sets of dishes?

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
+require './lib/menu.rb'
+require './lib/basket.rb'
+
+class Takeaway
+  attr_reader :order, :menu
+
```
EdwardAndress, line 6: How could you make your takeaway more flexible and allow for different menus?

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
...
+  def initialize
+    @order = []
+    @menu = Menu.new
+  end
+
+  def order_dish(quantity = 1, item)
```
EdwardAndress, line 12: good use of the default param but does the order of the params matter?

EdwardAndress: looks like it doesn't, based on your tests! feel free to ignore the above :)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    it 'gets the full menu' do
+      expect(subject.menu).to eq subject.menu
+    end
+  end
+
+  describe '#order' do
+    it 'can recieve a single item order' do
```
EdwardAndress, line 16: When writing tests, it's often better to be really explicit about the expectations, even at the cost of some repetition.  Here it would be clearer to use a literal representation of what you expect `order_dish` to return so that the reader doesn't need to look up at the before block.

That's what I think anyway!  What do you think?

EdwardAndress:
Hey Graeme!  I disagree with your assertion that the tests `are very poor` because you've got working tests for all your public methods.  If you want to use doubles, you'll need to inject dependencies (like `Menu`) and that would also address the questions I raised around the flexibility of the application.

Overall I think it's a decent solution.  You've clearly learned the basics of TDD and are thinking about how to do it better.  You're also thinking about how to break your program down into multiple objects, with clear responsibilities.  On that note, you might find it interesting to imagine how far you can take single responsibility principle!  How many classes can you imagine using?  What is the cost of having more classes and when does it become too much?

PR #1845, opened 2021-02-06
Alistair Tait pull request for takeaway challenge

AlistairTait12:
My answer to takeaway challenge


lib/cart.rb
```diff
@@ -0,0 +1,14 @@
+class Cart
+  attr_reader :order
+  def initialize
+    @order = []
+  end
+
+  def receive_item(item)
+    @order.push(item)
+  end
+
```
EdwardAndress, line 10: You can use `sum` on a hash as well ;)

lib/dish.rb
```diff
@@ -0,0 +1,8 @@
+class Dish
```
EdwardAndress, line 1: cool that you created a class for this!  Good application of SRP

lib/menu.rb
```diff
@@ -0,0 +1,40 @@
+require_relative './viewer'
+require_relative './dish'
+require_relative './cart'
+require_relative './sms_sender'
+
+class Menu
+  attr_reader :dishes
+  def initialize(*dishes)
```
EdwardAndress, line 8: nice use of the splat operator

lib/menu.rb
```diff
@@ -0,0 +1,40 @@
+require_relative './viewer'
+require_relative './dish'
+require_relative './cart'
+require_relative './sms_sender'
+
```
EdwardAndress, line 5: This class has a few dependencies – could you inject them to make the solution more flexible and easier to test?

spec/cart_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'cart'
+require 'dish'
+require 'viewer'
+require 'menu'
+
```
EdwardAndress, line 5: This is fine, in that it works, but it would be more idiomatic to use `let` statements

spec/cart_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+meat_feast = Dish.new("Meat Feast", 950)
+four_seasons = Dish.new("Four Seasons", 650)
+veggie = Dish.new("Veggie", 700)
+margherita = Dish.new("Margherita", 500)
+
+cart = Cart.new
+
```
EdwardAndress, line 12: and you can do this in a `before` block

EdwardAndress: but perhaps I am misunderstanding your intention here, since you use a before block below...

spec/cart_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+describe Cart do
+  describe '@order' do
+    it 'can hold the order of the customer' do
+      expect(cart.order).to eq [meat_feast, four_seasons, veggie]
+    end
+  end
+
+  describe '#receive_item' do
+    before(:each) do
```
EdwardAndress, line 25: I see you've removed some duplication here using the before block.  It's fine here but if you had another 5 test cases for the method, the before block would be a little way away from where the tests are executed and that indirection could be costly.  I'm not saying don't do it – just highlighting the trade off.  Some people will be OK with a certain amount of repetition in the tests if it brings greater clarity.

spec/sms_sender_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'sms_sender'
+
```
EdwardAndress, line 2: What happens if you break the SmsSender methods in some way – do these tests fail?

EdwardAndress:
This is a really nice solution, Alistair!  I particularly like the way that you've thought hard about the responsibility of each class and broken the program down into small units.  You could improve it by doing some dependency injection and you might also want to revisit your testing of the SmsSender.

PR #1846, opened 2021-02-07
takeaway challenge up to user story 3

vinc3108:


lib/checkout.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def order_checkout(order)
+    order.each { |item| puts "Dish: #{item[:dish]} Price: £#{item[:price]}" }
+    puts " Total to pay - £#{order_total(order)}"
+    @total
+  end
+  
+  private
+
```
EdwardAndress, line 16: You can use `sum`, which takes a block :) 

`order.sum {|item| item[:price]}`

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
+class Menu
+
+  attr_reader :menu
+
```
EdwardAndress, line 4: How could you create a menu that is customisable?

lib/order.rb
```diff
@@ -0,0 +1,30 @@
+require 'menu'
+# require 'total'
+
+class Order
+
```
EdwardAndress, line 5: How could you create an Order class that works with different menus?

lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+  
+  attr_reader :order, :menu
+
+  def initialize
+    @order = []
+    @menu = Menu.new
+  end
+
+  def order_dish(quantity = 1, dish)
+    raise "This dish is not on the menu, please choose another one!" unless dish_on_menu?(dish)
```
EdwardAndress, line 14: Nice!

lib/order.rb
```diff
@@ -0,0 +1,30 @@
...
+    
+    quantity.times { @order << @menu.select(dish) }
+    @order
+  end
+
```
EdwardAndress, line 19: Nice application of SRP

spec/order_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'order'
+
+RSpec.describe Order do
+  
+  before :each do
```
EdwardAndress, line 5: This is great, in that it works, so well done!  It would be more idiomatic to use `let`, rather than an instance var, however.

EdwardAndress:
Hey Vince!  This is a nice solution :). I like that you created a Checkout class.  As a thought experiment you could see how far you can take SRP before it becomes silly.

The main thing to focus on is dependency injection, which would make your solution more flexible.

PR #1847, opened 2021-02-08
Harry.C takeaway challenge 

HarryCarton:


lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    @order_list.each do |item|
+      puts "#{item.keys.first} : £%.2f" % item.values.first
+      total += item.values.first
+    end
+    puts "Total : £%.2f" % total
+  end
+
+  def interactive_menu
+    while true
+      item_name = gets.chomp
```
GraemeHedges, line 30: It might be a good to have a guard statement here to check the item is on the menu

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+  end
+
+  def print_current_order
+    total = 0
+    @order_list.each do |item|
+      puts "#{item.keys.first} : £%.2f" % item.values.first
+      total += item.values.first
+    end
+    puts "Total : £%.2f" % total
```
GraemeHedges, line 25: It could be worth moving this to another class to handle the money related behaviours of the order

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def interactive_menu
+    while true
```
GraemeHedges, line 29: This is a really cool idea, I like this a lot

lib/order.rb
```diff
@@ -0,0 +1,49 @@
+require_relative './takeaway'
+
+class Order 
+  attr_reader :order_list, :takeaway
+
+  def initialize(takeaway = Takeaway.new)
```
EdwardAndress, line 6: nice use of dependency injection

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    @order_list = []
+    @takeaway = takeaway
+  end
+
```
EdwardAndress, line 10: I wonder if you could neaten this up using`Array#find`.

lib/order.rb
EdwardAndress: I agree with @GraemeHedges – I'd separate things that are involved in printing / displaying from the business logic.

lib/order.rb
EdwardAndress: Looks like he checks to see if the item is on the menu up above.

This is turning into a review review :)

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+  describe '#add_by_name' do
+    it 'adds an item price hash to order_list by item name' do
+      subject.add_by_name('Mixed Starter')
+      expect(subject.order_list).to include({ 'Mixed Starter' => 7 })
+    end
+  end
+  describe '#interactive_menu' do
+    it 'takes inputs from user and if matches item name adds to order_list' do
+      allow(subject).to receive(:gets).and_return("Yuk Sung\n", "stop\n")
```
EdwardAndress, line 17: nice stub!

spec/order_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'order'
+
```
EdwardAndress, line 2: Since you're using dependency injection, you could double the takeaway in these tests.

EdwardAndress:
Lots of great stuff going on here, Harry – your use of stubs and dependency injection in particular.  Since we're focussed on OOP here, I'd recommend you reflect on the current responsibilities of each class and make a plan as to how you might break some of them down. See how far you can go!  Don't feel obliged to actually implement the plan though :)

PR #1848, opened 2021-02-08
Takeaway Challenge

harrythomson1:
I completed all user stories in this challenge except the stretch goal.

I really struggled with this challenge mainly with the tests because I wasn't sure how to test for what I wanted so I think I've hard coded a lot of the tests so that they pass which I know isn't good practice.

Also, it says I don't have 100% test coverage because I've not tested my private text method but this was tested within another method. I am aware though that the test I've used for this method is poor.

The test function will not work anymore as I foolishly uploaded it with my Twilio details, they have since been disabled and you could fill in with your own Twilio account if you wanted to test the functionality.

I have created two classes here because I was trying to implement OOP to the best of my knowledge but I can tell the one class is almost entirely dependant on the other which would cause me headaches if this program was much bigger and needed to be changed.

Overall I'm happy with how I managed to complete all the user stories and implement the outgoing test functionality but I'm not happy at all with my tests and know there's room for vast improvement.

Thank you for looking at my code!

### DIFF
--- a/lib/current_order.rb
+++ b/lib/current_order.rb
@@ -1,0 +1,30 @@
+class CurrentOrder
+
+  def initialize
+    @order = []
+  end
+
+  def add_item(item)
+    @order << item
+  end
+
+  def total
+    final_price = 0
+    @order.each do |n| 
+      final_price += n[:price]
+    end
+    show_order
+    final_price
+  end
+
+  def clear_order
+    @order = []
+  end
+
+  private
+
+  def show_order
+    puts @order
+  end
+
+end
--- a/lib/current_order.rb
+++ b/lib/current_order.rb
@@ -1,5 +1,7 @@
 class CurrentOrder
 
+  attr_reader :order
+
   def initialize
     @order = []
   end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,55 @@
+require_relative 'current_order'
+require 'twilio-ruby'
+
+class Takeaway
+
+  attr_reader :menu
+
+  def initialize
+    @menu = [{ item: 'Mega Breakfast', price: 6.75 },
+              { item: 'Full English Breakfast', price: 6.65 },
+              { item: 'Light Breakfast', price: 4.55 },
+              { item: 'Veggie Breakfast', price: 5.95 },
+              { item: 'Breakfast Special', price: 6.85 },
+              { item: 'Cheeseburger', price: 7.25 },
+              { item: 'BBQ Burger', price: 7.85 },
+              { item: 'Mega Burger', price: 8.75 },
+              { item: 'Vegatarian Burger', price: 5.95 }]
+    @current_order = CurrentOrder.new
+  end
+
+  def choose_meal(meal)
+    @current_order.add_item(@menu.find { |n| n[:item] == meal })
+  end
+
+  # def see_order
+  #   @current_order.order
+  # end
+
+  def confirm_order
+    @current_order.total
+  end
+
+  def pay_for_order
+    send_message
+    @current_order.clear_order
+  end
+
+ private
+
+  def send_message
+    account_sid = 'AC4386d4aa8e4ba2fbc1025275b943c9c4'
+    auth_token = '4cb57adcbb47f5ac9e66a9cffc8c7f30'
+    client = Twilio::REST::Client.new(account_sid, auth_token)
+
+    from = '+447782549143' # Your Twilio number
+    to = '+447453297338' # Your mobile phone number
+
+    client.messages.create(
+      from: from,
+      to: to,
+      body: "Thank you! Your order was placed and will be delivered before #{Time.new.hour + 1}:#{Time.new.min}"
+      )
+  end
+
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -22,10 +22,6 @@ class Takeaway
     @current_order.add_item(@menu.find { |n| n[:item] == meal })
   end
 
-  # def see_order
-  #   @current_order.order
-  # end
-
   def confirm_order
     @current_order.total
   end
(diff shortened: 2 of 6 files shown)


EdwardAndress:
Hey Harry!  There's lots of good stuff going on here and, based on your comments above, it seems like you've got some clear ideas of how to improve it, which is a really great position in which to be.  I'll leave some comments against the code in-line.  Please DM me if you have any questions.

lib/current_order.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize
+    @order = []
+  end
+
+  def add_item(item)
+    @order << item
+  end
+
+  def total
```
EdwardAndress, line 13: Very minor thing: `#sum` can be used here to get the total price in one line...

`@order.sum { |item| item[:price] }`

What do you think?  Concision does not always equal quality...

lib/current_order.rb
```diff
@@ -0,0 +1,32 @@
...
+    final_price
+  end
+
+  def clear_order
+    @order = []
+  end
+
+  private
+
+  def show_order
```
EdwardAndress, line 28: Good call to use a separate method for this.  The next step would be to move it into a new class and leave your `CurrentOrder` to focus on the business logic of orders.

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
+require_relative 'current_order'
+require 'twilio-ruby'
+
+class Takeaway
+
+  attr_reader :menu
+
```
EdwardAndress, line 7: How could you make the Takeaway more flexible?  I.e. How could you allow for different menus?

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    @current_order.total
+  end
+
+  def pay_for_order
+    send_message
+    @current_order.clear_order
+  end
+
+ private
+
```
EdwardAndress, line 35: Good job implementing text messaging!

EdwardAndress: Did you consider creating a separate class for this?  Maybe a `Notification` class?

spec/current_order_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  let(:menu_item) { double }
+
+  describe '#add_item' do
+    it 'adds an item to the current oder' do
+      subject.add_item(menu_item)
+      expect(subject.order).to eq([menu_item])
+    end
+  end
+
+  describe '#total' do
```
EdwardAndress, line 15: I was a bit confused by this because `TEST_ORDER` is defined in another file.  I suspect your intention is to avoid duplication, which is good, but the cost is indirection and increased cognitive load.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'takeaway'
+require 'current_order'
+
+describe Takeaway do
+
+  TEST_MEAL = 'Mega Breakfast'
```
EdwardAndress, line 6: This is fine, but it would be more idiomatic to use

`let(:test_meal) { 'Mega Breakfast }`

same for the other constant

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  TEST_ORDER = { item: "Mega Breakfast", price: 6.75 }
+  let(:send_message) { double }
+
+  describe '#menu' do
+    it 'returns a list of menu' do
+      expect(subject.menu).to eq(subject.menu)
```
EdwardAndress, line 12: Could this test ever fail?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'takeaway'
+require 'current_order'
+
+describe Takeaway do
+
+  TEST_MEAL = 'Mega Breakfast'
+  TEST_ORDER = { item: "Mega Breakfast", price: 6.75 }
+  let(:send_message) { double }
```
EdwardAndress, line 8: What happens if you delete this line?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    it 'Confirms the order and gives the correct price.' do
+      subject.choose_meal(TEST_MEAL)
+      subject.choose_meal(TEST_MEAL)
+      expect(subject.confirm_order).to eq(13.5)
+    end
+  end
+
+  describe '#pay_for_order' do
+    it 'Clears a customers order and sends a text' do
+      allow(subject).to receive(:send_message) { nil }
```
EdwardAndress, line 32: Good use of a stub!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      subject.choose_meal(TEST_MEAL)
+      subject.choose_meal(TEST_MEAL)
+      expect(subject.pay_for_order).to eq([])
+    end
+  end
+end
```
EdwardAndress, line 38: So, the hardest part of this exercise is testing the text message sending stuff, which you've side-stepped by making it a private method :)

I'll give you some tips on that anyway...

```
client.messages.create(
      from: from,
      to: to,
      body: "Thank you! Your order was placed and will be delivered before #{Time.new.hour + 1}:#{Time.new.min}"
      )
```

So you have a `client` that responds to `messages` and returns _something_ that responds to `create`.  To make this work, you need to be calling `create` with some sensible params.  Here's how you might create some doubles for this test...

```
let(:messages)  { double :twilio_messages }
let(:client)          { double :twilio_client, messages: messages }
```
Then in your test, you'd use a stub to declare that you expect `messages` to receive `create` with the correct params...

```
expect(messages).to receive(:create).with(
      from: from,
      to: to,
      body: "Thank you! Your order was placed and will be delivered before #{Time.new.hour + 1}:#{Time.new.min}"
)
```

EdwardAndress:
So, to summarise, I think the main thing to reflect on is SRP (single responsibility principle) since we were focussed on OOP.  Hope that helps!

PR #1849, opened 2021-02-14
Takeaway challenge - Iain

iainaitken:
See the README for an explanation of the program and reflection on what I think could be improved.


README.md
```diff
@@ -1,5 +1,117 @@
...
+
+The user places their order with the place method, which invokes the Texter class to create and send the text message.
+
+I have not used an attribute reader, to help maintain encapsulation, and the Dish, Calculator and Texter classes are injected into the menu object on initialisation to keep the classes isolated.
+
+#### Dish class
+
+This class stores information about dishes, including the name, price and whether it is available.
+
+Attributes are accessible only through methods, which each return a duplicate of the attribute rather than the attribute itself, to help maintain encapsulation.
```
tatsiana-makers, line 92: I see an attribute reader and these public methods as achieving equal levels of [encapsulation](https://www.geeksforgeeks.org/ruby-encapsulation/) in the class since instance variables are only accessible by the public attribute readers and can't be changed by other classes (since they would only be attribute readers and not accessors). What aspect of encapsulation do you see as being strengthened by returning duplicates of the attributes?

lib/menu.rb
```diff
@@ -0,0 +1,55 @@
...
+    @list.each do |item|
+      if item.name == dish.downcase
+        return :unavailable if item.available == "false"
+        return item
+      end
+    end
+    return :not_on_menu
+  end
+
+  private
```
tatsiana-makers, line 25: Nice use of private methods!

lib/menu.rb
```diff
@@ -0,0 +1,55 @@
+require_relative 'dish'
+
+class Menu
+
+  def initialize(dish_class = Dish)
+    @dish_class = dish_class
+    @list = []
```
tatsiana-makers, line 7: Something like `dishes` is a more descriptive name for this instance variable

lib/menu.rb
```diff
@@ -0,0 +1,55 @@
...
+    load_menu
+  end
+
+  def show
+    print_menu
+  end
+
+  def check(dish)
```
tatsiana-makers, line 15: Having the `check` method return either a custom symbol like `:unavailable` or the Dish instance makes code that calls this method a bit challenging to reason about since anyone who uses this method will have to remember that different types will be returned depending on the dish.

Rather than having one method do both of:
1. checking if the dish is available on the menu
2. finding the corresponding Dish instance

we could divide these responsibilities into two methods so that:
1. returns a symbol such as `:available`, `:unavailable`, or `:not_on_menu`
2. returns a Dish instance or `nil` if that dish isn't on the menu

README.md
```diff
@@ -1,5 +1,117 @@
...
+
+Possible further refactoring:
+
+* Remove hard-coding of values in the test suite.
+* Refactor order.add to only take a single variable, the name of the dish.
+* Refactor order.add to defend against edge cases, such as users entering the name of a dish in capital letters.
+* Consider all classes in light of SRP.
+* Check test suite to ensure it is DRY.
+* Expanding test coverage of the texter class - I don't really know how to test that properly.
+
```
tatsiana-makers, line 114: Really cool to read about your approach to the challenge!

lib/order.rb
```diff
@@ -0,0 +1,50 @@
+class Order
+
+  def initialize(dish_class = Dish, calc_class = Calculator, text_class = Texter)
```
tatsiana-makers, line 3: Nice use of dependency injection!

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+describe Order do
+  let(:dish_double) { double(:dish, :name => "pizza", :price => 10) }
+  let(:dish_class_double) { double(:dish_class, :new => dish_double) }
+  let(:calc_double) { double(:calc, :total => 20) }
+  let(:calc_class_double) { double(:calc_class, :new => calc_double) }
+
+  let(:text_double) { double(:text, :send => :message_sent) }
+  let(:text_class_double) { double(:text_class, :new => text_double) }
+  let(:menu) { double(:menu, :check => dish_double) }
+  subject { described_class.new(dish_class_double, calc_class_double, text_class_double) }
```
tatsiana-makers, line 12: Nice job isolating these unit tests from the actual Dish, Calculator and Texter classes

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        menu_2 = double(:menu, :check => :not_on_menu)
+        expect { subject.add(menu_2, "pizza") }.to output(/I'm sorry, that dish is not on our menu/).to_stdout
+      end
+    end
+    context 'when dish unavailable' do
+      it 'user will see error' do
+        menu_3 = double(:menu_3, :check => :unavailable)
+        expect { subject.add(menu_3, "pizza") }.to output(/I'm sorry, that dish is currently unavailable/).to_stdout
+      end
+    end
```
tatsiana-makers, line 39: We should also check that `add` works correctly in the good case scenario that a dish is available and on the menu

tatsiana-makers (commented):
Thanks for asking for a review! This is a nice go at the challenge, let me know if you have any questions about the comments I've left 🙂

PR #1850, opened 2021-02-14
AJ's takeaway-challenge

AJ8GH:
User stories all complete :)
positives:
+ completed the task and met the requirements of all user stories
+ implemented twilio API sms messaging successfully
+ kept the design simple and minimalistic
+ synced up with Travis CI and Coveralls for build-passing and test-coverage stats on repo readme 

what I'd do with more time:
- create a meal class or struct object
- build a text-to-order feature
- explore ways to simplify the codebase and improve readability further


lib/customer_service.rb
```diff
@@ -0,0 +1,33 @@
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load
+
+class CustomerService
+  def confirmation_text
+    create_client.messages.create(
+    from: ENV['TWILIO_PHONE_NUMBER'],
+    to:   ENV['MOBILE_PHONE_NUMBER'],
+    body: ["Thanks for your order! 🍕 It's being prepared ",
```
jshorns, line 10: I enjoy the presence of the emoji here

lib/customer_service.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def create_client
+    client_class::REST::Client.new(
+      ENV['TWILIO_ACCOUNT_SID'],
+      ENV['TWILIO_AUTH_TOKEN']
+    )
+  end
+
+  def delivery_time
```
jshorns, line 30: I like the integration of the time here  as well.

lib/display.rb
```diff
@@ -0,0 +1,42 @@
+class Display
+  CONFIRMATION_PROMPT   = 'Confirm order? Hit ⏎ to confirm, c to cancel'
+  INVALID_MEAL_MESSAGE  = ' is not a meal. Please make a valid choice'
+  MEAL_ADD_CONFIRMATION = ' has been added to your order 🍕'
```
jshorns, line 4: Really good to have an extra step to confirm the order, and a confirmation message that an item had been added.

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+class Menu
+  attr_reader :meals
```
jshorns, line 2: This is one where I guess I went beyond the user stories and was thinking about how the API would look from the restaurant side too, so had made the menu able to be amended. If the code was going to be shared across restaurants, you also wouldn't want it to have a default @meals attribute. Obviously this did mean every time I tested my code in irb I had to create new menus, which was a bit long!

lib/restaurant.rb
```diff
@@ -0,0 +1,35 @@
+class Restaurant
+  def view_menu
+    display.view_menu(menu)
+  end
+
+  def add(meal)
+    choice = meal.downcase.to_sym
```
jshorns, line 7: What happens to strings with spaces in them when converted .to_sym? Just wondering how this would impact your code if there was a menu item with two words. like a quattro formaggi (I have probably spelt this wrong)

lib/restaurant.rb
```diff
@@ -0,0 +1,35 @@
...
+    display.confirm_order
+    order.clear
+    customer_service.confirmation_text
+  end
+
+  private
+
+  attr_reader :menu, :order, :display, :customer_service
+
+  def initialize(args = {})
```
jshorns, line 29: not 100% sure what's going on here - is this saying that the argument will be a hash with  a customer service object, a display object, and a menu and meals object stored into those keys, which will then be stored to these instance variables? Could you not just have these as 3 arguments? I suppose I'm not quite following where the hash comes in. Also, do you want to be creating a new menu each time? Again, this introduces the possibility of the pizza menu getting added by mistake, when what you'd really want is to be adding a specific menu that a restaurant had created. Though I get that for the purposes of this challenge and testing it it equates to the same thing.

spec/customer_service_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  let(:messages) { double(:messages, create: 'text has been sent') }
+  subject        { described_class.new(twilio) }
+
+  describe 'confirmation_text' do
+    it 'sends customers texts' do
+      allow(subject).to receive(:create_client) { client }
+      expect(subject.confirmation_text).to eq 'text has been sent'
+    end
+
+    it 'calls create client method on self' do
```
jshorns, line 13: impressed that you got round to actually testing twilio itself, I did not reach that point!

spec/display_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  describe '#view_order' do
+    let(:order) { [{ bufalina: 13 }, { diavola: 14 }] }
+
+    it 'displays the order in nice format' do
+      expect(subject.view_order(order)).to include("Diavola: £14.00")
+    end
+
+    it 'displays the total in nice format' do
+      expect(subject.view_order(order)).to include("Total: £27.00")
```
jshorns, line 16: these tests are good, I was having a lot of trouble trying to get rspec to test specific strings so this is v useful to see!

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
+describe Menu do
+  describe '#meals' do
+    context 'when initialized without arguments' do
```
jshorns, line 3: again - I'm not sure you want Menu to have a default list of meals.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+
+  describe '#view_menu' do
+    it 'displays the meals and prices in nice format' do
+      expect(subject.view_menu).to eq menu_output
+    end
+  end
+
+  describe '#add' do
+    it 'adds meal to order' do
+      subject.add(:marinara)
```
jshorns, line 49: this isn't testing whether your downcase.to_sym works, since you're passing the argument as a downcase symbol anyway.

AJ8GH: Yeah you're right

README.md
```diff
@@ -1,5 +1,177 @@
...
+- Keeping things relatively simple from a design and readability point of view
+- Applying encapsulation, SRP, dependency injection
+
+### What to do differently
+
+- Create a meal class or struct
+- Build in text-ordering capability
+- Even more simplicity and readability
+- Using doubles and stubs to mock Twilio but still test the texting methods effectively
+
```
tatsiana-makers, line 174: Super nice readme! ✨

lib/customer_service.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  attr_reader :client, :client_class
+
+  def initialize(client_class = Twilio)
+    @client_class = client_class
+  end
+
+  def create_client
+    client_class::REST::Client.new(
+      ENV['TWILIO_ACCOUNT_SID'],
```
tatsiana-makers, line 25: Really cool use of environment variables!

lib/display.rb
```diff
@@ -0,0 +1,42 @@
...
+    "#{meal}#{INVALID_MEAL_MESSAGE}"
+  end
+
+  private
+
+  def total(order)
+    order.map(&:values).map(&:pop).sum
+  end
+
+  def format(amount)
```
tatsiana-makers, line 39: This is a nice detail

lib/restaurant.rb
tatsiana-makers, line 29: What is the reason for having the initialize method be private?

spec/customer_service_spec.rb
```diff
@@ -0,0 +1,26 @@
+describe CustomerService do
+  let(:twilio)   { class_double(Twilio, :twilio) }
+  let(:client)   { double(:client, messages: messages) }
+  let(:messages) { double(:messages, create: 'text has been sent') }
+  subject        { described_class.new(twilio) }
```
tatsiana-makers, line 5: Nice use of dependency injection to decouple your classes from their dependencies and allow you to mock Twilio here

spec/display_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        allow(STDOUT).to receive(:puts) { nil }
+        allow(subject).to receive(:gets) { "\n" }
+        expect { subject.confirm_order }.to output(confirmation).to_stdout
+      end
+    end
+  end
+
+  describe '#invalid_meal' do
+    it 'tells customer choice is invalid' do
+      invalid = "chips#{described_class::INVALID_MEAL_MESSAGE}"
```
tatsiana-makers, line 52: Rather than mimicking what the body of the method looks like, in tests we want to be as explicit and human readable as possible for what the expected output should be.

For example, let's say I wrote a method `sum` which takes two values as input and returns the sum of them. In my tests, I shouldn't have something that looks like:
```
def add(x,y)
    x + y
end
describe "#sum"
    it "adds two numbers together correctly" do
        @total = 3 + 4
        expect(sum(3,4)).to eq @total
    end
end
```

Because you're just recreating the logic of the code which risks the test passing even if there is a bug in the code as this bug will exist in both the original method and the expected output of the test. For example, the following would pass because I've mimicked the bug in my test, even though my logic is incorrect.
```
# actual code
def sum(x,y)
    x - y
end
# tests
describe "#sum"
    it "adds two numbers together correctly" do
        @total = 3 - 4
        expect(sum(3,4)).to eq @total
    end
end
```
To avoid this, the test for `sum` should look like, this way I am not mimicking the code in the test and I will catch any bugs.
```
describe "#sum"
    it "adds two numbers together correctly" do
        expect(sum(3,4)).to eq 7
    end
end
```

tatsiana-makers (commented):
This is a good go at the weekend challenge, thanks for asking for feedback! Let me know if you have any questions about the comments I've left (:

lib/restaurant.rb
AJ8GH: it's something that an Exercism mentor told me - that if I have private methods in a class I should put initialize there too, else it's misleading. Is that something you agree with?

lib/restaurant.rb
AJ8GH: Yes you are right. it's a detail I purposely side stepped by having one word pizzas only :)

not too much code to make it multi-word compatible so probably should have implemented it

lib/menu.rb
AJ8GH: I agree. I still like the idea of having a default menu stored as a constant - for ease of loading, and it can be initialized with different menus. But in hindsight I would add more flexibility by letting the menu be customisable by the restaurant through adding and removing dishes.

lib/restaurant.rb
AJ8GH: I used a hash because with 3 arguments to initialize, there's a serious risk of forgetting arguments or giving them in the wrong order. the hash solves that issue. the conditional assignment `|| =` means any argument left out becomes the default value. It's probably something that I could change in the design on my app. Restaurant is reliant on 3 classes so maybe a change in design would mean this set up isn't necessary and the app would be more flexible

spec/display_spec.rb
AJ8GH: so instead of using the variable name `"#{described_class::INVALID_MEAL_MESSAGE}"`
it's preferable to hardcode the desired output?

lib/restaurant.rb
tatsiana-makers: Interesting! I found [this stackoverflow](https://stackoverflow.com/a/36491066) answer interesting. It looks like you don't need to put the initialize method under the private heading for it to still be private. According to [this style guide](https://github.com/rubocop-hq/ruby-style-guide#classes--modules) the initialize method should go towards the top of the class and not under the private heading

spec/display_spec.rb
tatsiana-makers: Yes, hardcording it would make it more readable. It would also mean that the test would fail if anyone changed the constant's value which I think is a nice side effect of this change, as it will make people more conscious of changing the constant's value

lib/restaurant.rb
AJ8GH: Thanks that's good to know! I was previously putting at the top of the class as I knew it was  private by default but was told it's best practice to explicitly show it's private. This makes more sense to me though, having it at the top between class methods and instance methods. Thanks for the tip 🙂

PR #1851, opened 2021-02-14
Takeaway challenge - week 2

sandyMax974:
I couldn't complete the last user story -
```
 As a customer
```


lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+  def add(index, quantity = 1, order_id = @my_order)
+    add_items(index, quantity, order_id)
+    add_prices(index, quantity, order_id)
+  end
+
+  def basket(order_id = @my_order)
+    order_id.print_order
+  end
+
+private
```
kjjwhitlock, line 40: good use of private methods!

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def browse
+    @menu.each_with_index do |dish, index|
+      puts "#{index + 1} - #{dish[:item]} £#{dish[:price]}"
+    end
+  end
+
+  def start_order
+    @my_order = Order.new
```
kjjwhitlock, line 28: good division of classes in separate files

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    expect(subject).to respond_to(:menu)
+  end
+
+  describe '#browse' do
+    it 'prints a list of all available dishes with prices' do
+      expect { subject.browse }.to output.to_stdout
+    end
+  end
+
+  describe '#start_order' do
```
kjjwhitlock, line 15: this is excellent

PR #1852, opened 2021-02-14
TakeAway challenge

a-nawrocka:


jai-jk (commented):
Very clean code and clear tests, follows OOP principles perfectly. I could easily come into this codebase as a team member and work with it!

Only improvement I could suggest is removing comments/pseudocode from messenger file.

PR #1853, opened 2021-02-14
takeaway challenge

jshorns:


Gemfile
```diff
@@ -5,4 +5,10 @@ group :test do
   gem 'rubocop', '0.79.0'
   gem 'simplecov', require: false, group: :test
   gem 'simplecov-console', require: false, group: :test
+
+end
+
+group :comms do
+  gem 'twilio-ruby'
+  
 end
```
AJ8GH, line 14: nice use of comms group for gem file. I didn't know about that one

lib/dish.rb
```diff
@@ -0,0 +1,10 @@
+class Dish
+
+  attr_reader :name, :price
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+  end
+
+end
```
AJ8GH, line 10: lovely SRP class

lib/menu.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'dish_lister'
+
+class Menu
+  include DishLister
+  attr_reader :dishes, :title
+
```
AJ8GH, line 6: really nice short methods and encapsulation of private methods

lib/order.rb
```diff
@@ -0,0 +1,57 @@
...
+require_relative 'twilio_adapter'
+
+class Order
+  include DishLister
+
+  attr_reader :menu, :dishes, :complete, :balance
+  def initialize(menu, twilio_adapter, customer_number)
+    @twilio_adapter = twilio_adapter
+    @customer_number = customer_number
+    @menu = menu
```
AJ8GH, line 13: nice dependency injection

lib/order.rb
```diff
@@ -0,0 +1,57 @@
...
+  def complete_order
+    @complete = true
+    @message = "Thank you! Your order from #{@menu.title} for:\n" +
+    list_dishes +
+    "\nat a total of #{total} has been placed and will be delivered as soon as possible."
+  end
+
+  def get_balance
+    @balance = @dishes.map { |dish| dish.price }.inject(0, :+).round(2)
+  end
```
AJ8GH, line 55: I think 0 is the default value with inject and so doesn't need to be specified

AJ8GH: also consider:
`dishes.map(&:price)`

lib/order.rb
```diff
@@ -0,0 +1,57 @@
...
+  def add_items(*args)
+    order_closed_error
+    new_dishes = args.map { |item| @menu.pick(item) }
+    add_dishes(new_dishes)
+    get_balance
+  end
+
+  def total
+    print_money(@balance)
+  end
```
AJ8GH, line 28: supposedly using the attr_reader is preferred over the instance variable once it's defined.
e.g.
`print_money(balance)` rather than `print_money(@balance)`

lib/twilio_adapter.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def send_message(customer_number, body)
+    @client.messages.create(
+      from: ENV['TWILIO_PHONE_NUMBER'],
+      to: customer_number,
+      body: body
+    )
+  end
+
+end
```
AJ8GH, line 19: another great 'skinny' class :)

spec/menu_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+  let(:dish1) { double(:dish, :name => "Baked Potato", :price => 3.25) }
+  let(:dish2) { double(:dish, :name => "Spaghetti Bolognese", :price => 4.70) }
+  let(:dish3) { double(:dish, :name => "Ham Sandwich", :price => 2.65) }
+  let(:dish4) { double(:dish, :name => "Tartiflette", :price => 6.80) }
+  let(:test_menu) { described_class.new("Fake Ass Restaurant") }
+
+  describe '#initialize' do
+    it 'has an empty array of dishes' do
+      expect(test_menu.dishes).to eq []
+    end
```
AJ8GH, line 13: seems to be testing state rather than behaviour - I defo recommend checking out the practical on this topic
essentially it advises avoiding tests on internal implementation and focusing on desired behaviour. so here instead of checking that it is an array, you could test:
`expect(test_menu.dishes).to be_empty`

then if you want to change your implentation internally for whatever reason, say to use a hash, your test still works

spec/order_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    it 'throws an error if the order has already been finalized' do
+      test_order.finalize
+      expect { test_order.add_items(dish2) }.to raise_error "this order is closed"
+    end
+    it 'adds the price of the added items to the balance' do
+      expect { test_order.add_items(dish1, dish1) }.to change { test_order.balance }.by(dish1.price * 2)
+    end
+
```
AJ8GH, line 42: overall really like the readability of your tests, it's easy to follow and see what they are doing

spec/twilio_adapter_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'twilio_adapter'
+
+describe TwilioAdapter do
+  
+  it { is_expected.to respond_to :send_message }
+
+end
```
AJ8GH, line 7: could maybe try and apply doubles and stubs here for more tests, although I struggled with deciding on the best way to go about this and not have vacuous tests

spec/menu_spec.rb
```diff
@@ -0,0 +1,52 @@
+require 'menu'
+
+describe Menu do
+  let(:dish1) { double(:dish, :name => "Baked Potato", :price => 3.25) }
+  let(:dish2) { double(:dish, :name => "Spaghetti Bolognese", :price => 4.70) }
+  let(:dish3) { double(:dish, :name => "Ham Sandwich", :price => 2.65) }
+  let(:dish4) { double(:dish, :name => "Tartiflette", :price => 6.80) }
+  let(:test_menu) { described_class.new("Fake Ass Restaurant") }
```
AJ8GH, line 8: nice

AJ8GH (commented):
great work :) 
- I like how your classes are well defined with a clear responsibility
- your methods are short
- you've used encapsulation and dependency injection well
- The only place in your code I had to do a bit of mental gymnastics to follow was `Order`, perhaps some more time for refactoring would have sorted that
-  overall  really well designed and in line with the brief

PR #1855, opened 2021-02-14
Week 2 Takeaway challenge submission

kerisic:
Submission satisfies all user stories. There may be more tests needed for edge cases.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,8 @@
 source 'https://rubygems.org'
 
-gem "twilio-ruby"
 gem "dotenv"
 gem "sinatra"
+gem "twilio-ruby"
 
 group :test do
   gem 'rspec'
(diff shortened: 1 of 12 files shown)


ianmcnicholas:
Really like your cohesion.  Several classes splits up their responsibilities really well, with the overall takeout class controlling everything.  Also, this means that in your tests you only really have to use doubles in the takeout_spec.rb file.

Decent amount of edge cases thought of - fail message if dish ins’t available etc.

Unsure about the method that collects the time.  When I ran this in IRB it would give me a time 40 minutes from now.

In the pull request, it would be good to see some comments that the reviewer can pick up on straight away before even getting to the README.  E.g. is there anything specific you'd like them to view.

Really like how much you got into the texting app, to provide the user a guide on how to use it.

PR #1857, opened 2021-02-14
TakeAway - program functioning

charlierdm:
Readme and tests need updating. Program required refining.


lib/takeaway.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @basket = []
+  end
+
+  def order(item, quantity = 1)
+    quantity.times do
+      @menu.list.each do |dish|
+        @basket << dish if dish[:dish] == item
```
jgumoes, line 15: this is way less complicated than what I did

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  it 'lets the user check the total order' do
+    expect(subject).to respond_to(:view_order)
+  end
+
+  it 'the order starts at £0' do
+    expect(subject.check_total).to eq(0)
+  end
+
+  it 'sends a text message' do
```
jgumoes, line 25: a better test might be to give subject a sms class double, and expecting Sms.new to be called

jgumoes: also this saves you credit

PR #1859, opened 2021-02-14
Takeaway Challenge

peter-james-allen:
Complete
Covers all user stories

Unfinished
Twilio credentials are not hidden in environment variables
Testing coverage

To Improve
Forgot to commit as I went along
Program needs simplifying
Code needs to be refactored
Classes need to be looked at to make sure they follow OOP

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+This is my final submission for the Makers Academy Week 2 Challenge
+
 Takeaway Challenge
 ==================
 ```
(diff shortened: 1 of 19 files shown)


tatsiana-makers (commented):
This is a good go at the Takeaway challenge with some nice usage of dependency injection to decouple the classes!Let me know if you have any questions about the comments I've left (:

PR #1861, opened 2021-02-15
My attempt at the Takeaway Challenge

anioji94:
I wasn't able to get my program to send a confirmation text to my phone.
Although, I could get Twilio to send it through a local ruby file.
Unfortunately I couldn't get them to talk to eachother!


lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+
+		def view_order
+			# should return a list of all the items with price and finally the total
+				@cart.each { |item| puts "#{item.name}, £#{item.price}" }
+				puts "Your Total is: £#{total}"
+		end
+
+		def confirm_order
+				@cart.clear
+				# somehow call twilio to send a message#
```
JLBrigham, line 27: i think to call twilio you just need to so sms = SMS.new

PR #1862, opened 2021-02-15
my take-out joint

jgumoes:


charlierdm:
Here's my thoughts after reviewing your code.

- The first thing that stands out is the readme is the original one. You've named your readme MYREADME so it doesn't appear on the main page in your repo, It'd be worth changing that.

- Testing is quite basic currently only checking if classes respond to methods. 

- Nice use of descriptions in the menu and the messaging is clear for the user.

- (# send confirmation text to customer). I can see this as a note but can't see the text being built into the class, looks like Twilio isn't in there currently. 

- Overall a good clear layout and separation between classes.

PR #1865, opened 2021-02-15
Takeaway Challenge

hannahdesmond:


README.md
```diff
@@ -69,15 +59,8 @@ I would like to receive a text such as "Thank you! Your order was placed and wil
...
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
-
-Notes on Test Coverage
-------------------
+Progress 
+--------
+I have implemented the first three of the four user stories.
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) when you run your tests.
+One particular issue was how to ensure that there were always two digits after the decimal place when displaying the price. I ended up adding a string of '0' to pass the tests, which works for my menu but not if the price in pence is not multiples of ten. I tried looking at BigDecimal but it didn't give me the output I wanted.
```
tatsiana-makers, line 66: This shows a way of doing string formatting in ruby: https://www.rubyguides.com/2012/01/ruby-string-formatting/

README.md
```diff
@@ -14,19 +14,9 @@ Takeaway Challenge
```
tatsiana-makers, line 14: With regards to a README, the things a reviewer would love to see are:
- instructions on how to install the code
- instructions on how to run the code (a nice addition could be a code snippet or screenshot of example usage of the code)
- instructions on how to run the tests

Some nice to haves are:
- diagrams which represent the program. Perhaps in your planning you made class or sequence diagrams. These are nice to include as the reviewer will then have an idea of what the program looks like before diving into Ruby syntax
- descriptions of how you approached the challenge
- changes you would make if you had more time

lib/takeaway.rb
```diff
@@ -0,0 +1,39 @@
...
+    @basket = []
+  end
+
+  def add_to_basket(item, quantity)
+    self.menu.map do |food, price|
+      if item == food
+        @basket.push("#{item} x #{quantity.to_s}: £")
+        @basket.push((price * quantity))
+      end
+    end
```
tatsiana-makers, line 21: This block of code:
1. finds the price of the given item
2. pushes two strings into the `@basket` instance variable.


For 1:
Hashes have a nice way of doing lookup using the square bracket syntax: `self.menu[item]` will give you back the corresponding price of the item

tatsiana-makers: Currently the `@basket` is something that stores strings which will be printed out to the user, which means that we're coupling the storing of items to how we want to present them to the user later on. The basket is currently doing two jobs:
- storing the items we want
- caring about how those items will be later printed to the user

Rather than storing strings, we could store dishes that the user ordered. Then we could have a method which translated this list of dishes into the format that we wanted to show the user. This would mean that the two concerns are separated.

lib/checkout.rb
```diff
@@ -0,0 +1,16 @@
+require_relative './takeaway.rb'
+
+class Checkout
```
tatsiana-makers, line 3: This class currently isn't in use. What is the motivation for this class? It currently formats a number into a string, which is something that can be done using the method shown in the link above so perhaps we don't need a separate class for this? What do you think?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+    describe '#show_basket' do
+      it 'prints the basket as a string' do
+        expect(@takeaway.show_basket).to eq('bibimbap x 3: £31.50')
+      end
+    end
+
+    describe '#total' do
+      it 'gives the total sum of the basket' do
+        expect { @takeaway.order_total }.to output("£31.50\n").to_stdout
```
tatsiana-makers, line 21: Nice testing of printing to stdout!

README.md
```diff
@@ -69,15 +59,8 @@ I would like to receive a text such as "Thank you! Your order was placed and wil
...
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
-
-Notes on Test Coverage
-------------------
+Progress 
+--------
+I have implemented the first three of the four user stories.
```
tatsiana-makers, line 64: https://github.com/makersacademy/course/blob/master/pills/levels_of_stubbing.md is a pill about mocking 3rd party APIs 🙂

tatsiana-makers (commented):
Thanks for asking for review! I've left a few comments about your code. Let me know if you have questions (:

PR #1870, opened 2021-03-13
Takeaway Challlenge

eatruman:
Up to user story 3 completed, user story 4 I couldn't figure out how to use the environment variables so the text message function is not working.


spec/menu_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'menu'
+
+describe Menu do 
+    # User Story 1: As a customer
+    # So that I can check if I want to order something
+    # I would like to see a list of dishes with prices
```
willspencer16, line 6: I like this use of comments. I feel it serves as a good reminder of what exactly you want the test to do.

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+                "hawaiian" => 13,
+                "vegetarian" => 13,
+                "meat" => 15 }
+  end 
+
+  def print_menu
+    @pizzas.each do |key, value|
+        puts "#{key} : #{value}"
+    end  
+  end
```
willspencer16, line 16: Really like this code - easy to read and understand.

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        it "responds to add with one argument" do
+            expect(subject).to respond_to(:add).with(1).argument
+        end 
+       it "adds a menu item to the list" do 
+        expect(subject.add("cheese")).to eq([["cheese", 10]])
+       end 
+       it "adds several menu items to the list" do 
+        subject.add("cheese")
+        expect(subject.add("pepperoni")).to eq([["cheese", 10], ["pepperoni", 12]])
+       end
```
willspencer16, line 25: As discussed, I think we both could have used a double for this test.

PR #1871, opened 2021-03-14
takeaway challenge - lou

loushark:
notes:
* menu is currently hardcoded to one menu list that initializes with every new restaurant instance
* set up to send texts with Twilio
* twilio keys and numbers protected with dotenv

more notes on README.md


README.md
```diff
@@ -46,38 +33,68 @@ As a customer
 So that I am reassured that my order will be delivered on time
 I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
 ```
+----
+
+#### Domain Model
```
TataSher, line 38: Very clear structure how you set it up!

README.md
```diff
@@ -46,38 +33,68 @@ As a customer
...
+#### IRB testing
 
-> :warning: **WARNING:** think twice before you push your **mobile number** or **Twilio API Key** to a public space like GitHub :eyes:
->
-> :key: Now is a great time to think about security and how you can keep your private information secret. You might want to explore environment variables.
+require './lib/restaurant.rb'
 
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
+##### create a restaurant instance:  
+sparkleBurgers = Restaurant.new("Sparkle Burgers LTD")
```
TataSher, line 78: You added a full name of the restaurant as a string which  will be a very useful feature

README.md
```diff
@@ -46,38 +33,68 @@ As a customer
...
+sparkleBurgers.view_menu  
+      => "Sparkle Burgers LTD's Menu:\n, 1: Cheese burger, £10\n, 2: Cheese and bacon burger, £12\n, 3: Pulled pork burger, £15\n, 4: Fries, £6\n, 5: Chilli fries, £8\n, 6: Onion rings, £7\n, 7: Coke 330ml, £2\n, 8: Fanta 330ml,£2\n, 9: Brew Dog Vegabond 4.5%, £5"
 
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
+#####  add to order:  
+sparkleBurgers.choose_dish(2)
```
TataSher, line 89: Adding numbers of dishes rather than names can be more convenient. Maybe including the quantity as a second argument could be faster for the user

lib/order.rb
```diff
@@ -0,0 +1,84 @@
+require_relative 'text'
+
+class Order
+
+  attr_reader :number, :basket, :total, :history
+
+  def initialize
+    @number = 1
```
TataSher, line 8: Unique order number can help to track orders which will make things so much easier for the takeaway

lib/order.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  def add_to_basket(dish)
+    @basket << dish
+    total_so_far
+  end
+
+# viewing basket with the total cost of basket
+  def view_basket
+    @print = ["Viewing basket for order number #{@number}:\n"]
```
TataSher, line 22: This method looks a bit too lengthy. Maybe it can be refactored or split into 2?

lib/order.rb
```diff
@@ -0,0 +1,84 @@
...
+      @basket.each do |dish|
+        @print << "#{dish[0]} £#{dish[1]}\n"
+        num += 1
+      end
+    end
+    @print << "Total cost of order: £#{@total}"
+    @print.join(", ").chomp
+  end
+
+  # order cannot be completed if the total is not correct
```
TataSher, line 34: All the notes make it so much easier to understand the intension of the code

spec/order_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    it "adds to the history when complete_order" do
+      order001.complete_order
+      expect(order001.history).to eq([{ :"Order #1, completed on #{Time.now}:" => ["Cheese Burger", 10, "Fries", 6] }])
+    end
+
+    it "doesn't have a history unless first order completes" do
+      expect(order001.history).to eq []
+    end
+
+    it "has two history records when two orders are completed" do
```
TataSher, line 104: Amazing that you incorporated the order history feature

TataSher (commented):
I think it's great how you created an Order, Restaurant and Text class. They seem to be very cohesive. It is very visible that it was test driven, as tests seem to cover all the functionality of the code. I thought it's amazing that you also added a unique order number for every order and found a way to generate it within the initialized method. The only thing I would maybe add is for a user to add quantity of dishes when they order for many people.  Lots of additional cool features, overall very impressive!

README.md
loushark: Totally agree here! I would love to improve this so that both a quantity can be added, and also that if the same dish was added again later in the order, it would also just update the quantity rather than add another to the list

lib/order.rb
loushark: Agreed! I wasn't sure i liked how lengthy this was. Either of your suggestions are really useful here

spec/order_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+      allow(Text).to receive(:new).and_return(no_text)
+      allow(no_text).to receive(:send_text)
+    end
+
+    it "confirms the order is complete and will be delivered an hour from now" do
+      expect(order001.complete_order).to eq("Thank you! Your order was placed and will be delivered before #{(Time.now + 60 * 60).strftime("%I:%M %p")}")
+    end
+  end
+
+  describe "#order_history" do
```
siellsiell, line 83: Testing private methods in your unit tests can be an indication that you're violating encapsulation (i.e. testing the internals of the class). In this case it might indicate instead that order history is something that should be kept in a different class (separation of concerns) and be part of the public interface of that class so that you can test it.

spec/order_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    it "has two history records when two orders are completed" do
+      order001.complete_order
+      order001.add_to_basket(cheeseBurger)
+      order001.add_to_basket(fries)
+      order001.complete_order
+      expect(order001.history.length).to eq(2)
+    end
+
+    describe "#reset_order" do
```
siellsiell, line 112: I think a better name for this group of tests is `#complete_order` as this is really the method your are testing -- how it resets the order is an implementation detail.

siellsiell (commented):
Hey Lou, thanks for asking for a review!

I like that you've separated your code into Restaurant and Order classes. You could improve a little on how these classes work together to make the experience more seamless for the user as well as improve encapsulation -- at the moment, your tests assume a little too much about the internal workings of your classes.
You've also made really good use of mocking!

loushark:
@siellsiell thanks so much for reviewing this challenge for me! These are some really helpful notes for me to think about and improve on in my future code. I really appreciate how clearly explained your comments are <3

siellsiell:
> @siellsiell thanks so much for reviewing this challenge for me! These are some really helpful notes for me to think about and improve on in my future code. I really appreciate how clearly explained your comments are <3

You're very welcome! Glad my comments were useful :)

PR #1876, opened 2021-03-14
Partial submission

inbarvernia:
Fulfilled first three user stories, didn't get to last one. Some methods are doing too much and need to be split up, but I ran out of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,13 +36,61 @@ I would like to be able to select some number of several available dishes
 - I created an Order class, which takes an instance of the Menu class as an argument upon initializing (defaulting to Menu.new if none is provided), and which has an `@items` attribute, which initializes as an empty array.
 - I defined the `#add_item` method, which takes two arguments: dish name, and quantity. If no quantity is provided, it defaults to 1. Calling this method adds dishes from the Menu class's list of dishes (that is, a hash representing a dish) to the order's `@items` array, adding to the hash the key `:quantity`, with its value set to the quantity specified (or to 1 if no quantity is specified).
 - At present, this doesn't respond to edge cases such as users trying to call this method with a dish that's not on the menu (including typos).
+- Of course, for a real takeaway app, I would also want to give users the option of removing items or updating quantities through the introduction of additional methods.
 
 ```
 As a customer
 So that I can verify that my order is correct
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -14,70 +14,83 @@ Takeaway Challenge
```
cspoppuppy, line 14: Good details of README :)

lib/menu.rb
```diff
@@ -0,0 +1,18 @@
+class Menu
```
cspoppuppy, line 1: Good use of hash!

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+    @items = []
+    @total = 0
+    @menu = menu
+    @menu_items = menu.dishes
+  end
+
+  def add_item(dish, quantity = 1)
+    # This method is currently doing too much and needs refactoring
+    item = @menu_items.select { |menu_item| menu_item[:name] == dish }
+    item = item.first
```
cspoppuppy, line 14: can refactor into one line 
item = @menu_items.select { |menu_item| menu_item[:name] == dish }.first

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+    item[:quantity] = quantity
```
cspoppuppy, line 15: Can cause issue if you add the same item twice, the previous quantity will get overwritten

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+      statement += "#{item[:quantity]} * #{item[:name]} (unit price #{item[:price]} GBP,
+      total price #{item[:price] * item[:quantity]} GBP)\n"
+    end
+    statement += "----------\nTotal: #{@total}\n"
+    print statement
+  end
+
+  private
+  def update_total(item)
+    @total += (item[:price] * item[:quantity])
```
cspoppuppy, line 34: Like the total price!

cspoppuppy (commented):
Well written README, like the design of code and functionality. A couple of issues raised in the code lines

lib/order.rb
inbarvernia: Yes, if I had more time I would have added some kind of conditional to check if this dish is already inside the order's `@items` array and if so, only updating quantity instead of adding a new item.

PR #1877, opened 2021-03-14
takeaway challenge

fg24davies:
I couldn't seem to initialise a Text object in my Order - it was making all my tests fail and I'm not sure why.
Therefore I was not able to test the texting functionality and so the last user story is not covered.
There was no checking to see if an item was available (i.e. on the menu) which I would like to add in the future. 
Layout of printing should be made nicer.
Readme yet to be updated.

### DIFF
--- a/lib/text.rb
+++ b/lib/text.rb
@@ -1,0 +1,22 @@
+require 'rubygems'
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load
+
+class Text
+  
+  def initialize 
+    account_sid = ENV["TWILIO_ACCOUNT_SID"]
+    auth_token = ENV["TWILIO_AUTH_TOKEN"]
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+  end 
+
+  def send_sms(total, number)
+    @client.messages.create(
+      from: ENV["MY_TWILIO_NUMBER"],
+      to: number,
+      body: "Thank you! Your order was placed and will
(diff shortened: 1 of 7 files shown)


lib/text.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def send_sms(total, number)
+    @client.messages.create(
+      from: ENV["MY_TWILIO_NUMBER"],
+      to: number,
+      body: "Thank you! Your order was placed and will
+      be delivered #{Time.now + 60 * 60}.")
+  end 
+
+end
```
jackmcc08, line 22: Twilio code looks good, I like how you have send SMS with arguments which will allow you to input different numbers. 

Does dotenv gem allow you to call on the project specific environment methods with ENV[]?

spec/order_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      subject.select_dish(dish)
+      expect { subject.see_order }.to output("Your order:\n1 x Hash Brown...£1.00\nTotal = £1.00\n").to_stdout 
+    end 
+  end 
+  
+  describe '#text_confirmation' do
+    it { is_expected.to respond_to(:text_confirmation).with(1).argument }
+    it 'sends a text confirmation' do 
+      allow(text).to receive(:send_sms).with("£1", "1111111111")
+    end
```
jackmcc08, line 67: What were you looking to implement here?

PR #1878, opened 2021-03-14
Takeaway Weekend Challenge by Jack M.

jackmcc08:
See README update for context and explanation. 

Met user stories through creation of 5 classes and implemented all 4 user stories. Also implemented TWILIO funcitonality.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,4 +8,5 @@ group :test do
   gem 'simplecov', require: false, group: :test
   gem 'simplecov-console', require: false, group: :test
   gem 'timecop'
+  gem 'rake'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -5,4 +5,5 @@ group :test do
   gem 'rubocop', '0.79.0'
   gem 'simplecov', require: false, group: :test
   gem 'simplecov-console', require: false, group: :test
+  gem 'timecop'
 end
--- a/Gemfile
+++ b/Gemfile
@@ -3,10 +3,10 @@ source 'https://rubygems.org'
 gem 'twilio-ruby', '~> 5.48.0'
 
 group :test do
+  gem 'rake'
   gem 'rspec'
   gem 'rubocop', '0.79.0'
-  gem 'simplecov', require: false, group: :test
-  gem 'simplecov-console', require: false, group: :test
-  gem 'timecop'
-  gem 'rake'
+  gem 'simplecov', require: false
+  gem 'simplecov-console', require: false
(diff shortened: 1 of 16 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,65 @@
+require_relative 'customer.rb'
+require_relative 'order.rb'
+require_relative 'menu.rb'
+require_relative 'send_text.rb'
+
+class Takeaway
+  def initialize(menu = Menu.new, order_class = Order, text = SendText.new)
```
fg24davies, line 7: Really nice of extraction of methods into other classes so that your takeaway class worked well and clearly in IRB

lib/order.rb
```diff
@@ -0,0 +1,85 @@
...
+    total_cost(@current_order)
+  end
+
+  def empty?
+    @current_order.empty?
+  end
+
+  private
+
+  def find_menu_item_by_number(item_number)
```
fg24davies, line 43: I really liked finding menu item by number idea

Gemfile
```diff
@@ -1,8 +1,11 @@
 source 'https://rubygems.org'
 
+gem 'twilio-ruby', '~> 5.48.0'
```
fg24davies, line 3: Noticed that this is in a different place to mine, why did you add it there?

(We talked about thiS earlier!)

fg24davies (commented):
I ran all the RSpec which passed and rubocop too, for which there were no errors :)
Overall I was really impressed with how you tackled the challenge. There were so many extra features which in real life are so obvious but not necessarily given in a user story and you managed to tie these altogether for a very realistic app. Apologies I couldn't get more technical on the code, however I could follow it and it felt readable for the many things that you had incorporated. It's also given me lots of things to think about.
Nice work on the twilio and getting it to work!

Gemfile
jackmcc08: Let's discuss this in our pairing session - I am not sure to be honest what the converntion is 😃

lib/order.rb
```diff
@@ -0,0 +1,85 @@
...
+    @menu = menu
+  end
+
+  def add_to_order(item_number, quantity)
+    item_order = { find_menu_item_by_number(item_number) => quantity }
+    @current_order << item_order
+    item_order
+  end
+
+  def remove_from_order(item_number, quantity)
```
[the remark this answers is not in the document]
jackmcc08: good point, definitely something for me to watch out for.

PR #1879, opened 2021-03-14
Partial complete of user story 3

crotchetycrow:
A real shame I couldn't complete this one or get to the later tasks with Twilio. I got stuck with calculating my hash, I attempted to create a separate class to store my hash to try and make things easier but I could not resolve. Will have to attempt at a later date.


lib/order.rb
```diff
@@ -0,0 +1,22 @@
+require './lib/takeaway'
+
+
+class Order
+
+  attr_reader :basket, :total
+
+  def initialize
```
dmgreaves, line 8: Good plan of initialising an empty basket to add items to.

lib/takeaway.rb
```diff
@@ -0,0 +1,14 @@
+class Takeaway
+
+  def initialize(menu = @menu)
```
dmgreaves, line 3: Could you create your list of dishes in initialise as an array or hash? Good explanation of your thought process in the readme though.

lib/takeaway.rb
```diff
@@ -0,0 +1,14 @@
...
+    @menu = menu
+  end
+
+  def list_menu
```
dmgreaves, line 7: The list_menu method you have could remain in place and still print out the list of dishes.

spec/order_takeaway_spec.rb
```diff
@@ -0,0 +1,19 @@
+require './lib/order'
+require './lib/takeaway'
+
+describe Order do
+  let(:test_order) { Order.new }
```
dmgreaves, line 5: Good set up of a test_order to use in your Spec file.

spec/order_takeaway_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  describe '#add_item' do
+    it 'adds item to basket' do
+      expect(test_order.add_item("quesidilla",1)).to eq 1
```
dmgreaves, line 9: Could you create a second test to check that the basket also contains "quesidilla" as well as quantity 1.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/takeaway'
+
+describe Takeaway do
+  let (:test_menu) { described_class.new }
+
+  describe 'takeaway_menu' do
+    it 'prints a default list of dishes with prices' do
```
dmgreaves, line 7: Good approach to test the menu dishes. It sounds like this is the one you couldn't get working according to your readme. Could you use a "before do" block or maybe set up another let/allow at the top of the test?

dmgreaves (commented):
Good approach to the challenge overall. Created separate classes for separate tasks, good testing process, and thorough explanation in the readme.

PR #1880, opened 2021-03-15
takeaway challenge

taran314:
Did not have time for text messaging feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,3 +101,4 @@ Notes
 - add implementation for extra menus (drinks)
 - subclasses starter, main, desert, drinks and polymorphic
   functions
+- add prep time instance variable for dishes, create a function which calculates wait time.
--- a/README.md
+++ b/README.md
@@ -76,7 +76,10 @@ see_menu - lists out dishes with prices
 
 US2
 select_dishes - select and store dishes
+getting user input - mock this
+Edge case - input does not match case
 dish owns available
+- restaurant method to say dish is out of stock.
 
 US3
 check_total - sum of prices of selected dishes
@@ -93,3 +96,8 @@ Menu owns MAX_WAITING_TIME
 Twilio API
   gem file
   set phone number to environment variable
+
+Notes
+- add implementation for extra menus (drinks)
+- subclasses starter, main, desert, drinks and polymorphic
+  functions
--- a/README.md
+++ b/README.md
@@ -66,18 +66,30 @@ I would like to receive a text such as "Thank you! Your order was placed and wil
 >
 > :key: Now is a great time to think about security and how you can keep your private information secret. You might want to explore environment variables.
 
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
+User class
+Owns menu variable 
 
+US1
+Dish owns price
+Menu class which creates and stores dishes?
+see_menu - lists out dishes with prices
 
-In code review we'll be hoping to see:
+US2
+select_dishes - select and store dishes
+dish owns available
 
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+US3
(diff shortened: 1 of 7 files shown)


lib/dish.rb
```diff
@@ -0,0 +1,14 @@
+class Dish
+  attr_reader :name, :price, :available
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+    @available = true
```
Emmapr123, line 7: Oh I like this idea!

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'restaurant'
+
+describe Restaurant do
+  let(:restaurant) { Restaurant.new }
```
Emmapr123, line 4: I think you could have made a double here because you initialize the restaurant with a new instance of the menu class. So you could first double that menu, and make the restaurant with a menu double here - I found that the easiest way to do it, but only last night so it isn't in my code either.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  describe 'display_menu' do
+    it 'displays a menu of one dish to the user' do
+      # not sure if I should have used a double here
+      # dish = double(name: "Pizza", price: "5")
+      # menu = double(menu: [dish])
```
Emmapr123, line 10: that way (look at the previous comment) this would have been a clean double.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      expect(restaurant.display_menu).to eq("Pizza: 5\nPasta: 4\n")
+    end
+  end
+
+  describe 'add_to_order' do
+    it 'stores selected available dishes from the menu' do
+      dish1 = double(name: "Pizza", price: "5", available: true)
+      dish2 = double(name: "Pasta", price: "4", available: true)
+      menu = double(dishes: [dish1, dish2])
+      restaurant = Restaurant.new(menu)
```
Emmapr123, line 26: ahhh So you do use that double here! Noice

lib/dish.rb
```diff
@@ -0,0 +1,14 @@
+class Dish
```
Emmapr123, line 1: Nice simple class, has everything in it that it needs, and nothing more

lib/menu.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def format_dishes(dishes)
+    dishes.map do |dish|
+      "#{dish.name}: #{dish.price}\n"
+    end
+  end
+
+  def calculate_total(dishes)
+    prices = dishes.map do |dish|
+      dish.price.to_i
```
Emmapr123, line 36: don't you give the price as an integer already?

lib/restaurant.rb
```diff
@@ -0,0 +1,55 @@
...
+  def dish_unavailable(name)
+    selection = select_dish(name)
+    return "No dish found" if selection.empty?
+
+    selection.first.not_available
+  end
+
+  def order_complete
+    time = Time.now + MAXIMUM_WAIT_TIME
+    "Thank you! Your order was placed and " +
```
Emmapr123, line 45: nice use of adding the hour

lib/restaurant.rb
```diff
@@ -0,0 +1,55 @@
...
+    "will be delivered before #{time.hour}:#{time.min}"
```
Emmapr123, line 46: Why didnt you use twilio?

README.md
```diff
@@ -66,18 +66,39 @@ I would like to receive a text such as "Thank you! Your order was placed and wil
...
+
+US4
+send_text - sends a text message (string) "Thank you! Your order was placed and will be delivered before #{time.now + MAX_WAITING_TIME}"
+Menu owns MAX_WAITING_TIME
+
+Twilio API
+  gem file
+  set phone number to environment variable
+
+Notes
```
Emmapr123, line 100: I like your notes! Good Ideas, I feel like I need to do this too instead of doing them straight away and having to hurry through the user stories after

Emmapr123 (commented):
Nice code! I like your approach

PR #1881, opened 2021-03-15
takeaway challenge

ralphbartley:


lib/menu.rb
```diff
@@ -0,0 +1,14 @@
...
+    @dishes = [Dish.new("Scrambled Eggs", 5),
+                Dish.new("Steak", 15),
+                 Dish.new("Hot Sauce", 3),
+                  Dish.new("Hash Browns", 5)]
+  end
+
+  def list_dishes
+    @dishes
+  end
+
```
louballantyne, line 13: This is a nice way of storing your menu. Perhaps when you list dishes, you could format the currency so that it displays as e.g. £5.00? Might it also be useful to make the menu private so that a user can't edit the prices?

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def initialize
+    @order = Order.new
+    @menu = Menu.new
+  end
+
+  def select_dish(dish)
+    @order.dishes.push(dish)
+  end
+
```
louballantyne, line 15: Is the user able to edit their order (e.g. delete dishes) while they're compiling it? Can they view it before they submit it?

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+    delivery_confirmation
+  end
+
+  def delivery_confirmation
+    status = send_message
+    delivery_message = "Thank you! Your order was placed and will be delivered before #{@time}"
+    raise "Delivery confirmation error" if status == "failed"
+    return delivery_message if status != nil
+  end
+
```
louballantyne, line 27: I love the status idea.

spec/order_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  it 'starts with an empty collection' do
+    expect(subject.dishes).to be_empty
+  end
+
+  it 'records the time an order is to be delivered' do
+    subject.complete
+    expect(subject.delivery_time).to be_a_kind_of(String)
+  end
+
```
louballantyne, line 14: Does this test whether the time is recorded, or whether something is recorded?

spec/order_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    subject.dishes.push(Dish.new("name", 5))
+    subject.dishes.push(Dish.new("name", 5))
+    expect(subject.verify_total).to eq(10)
+  end
+
+  it 'has a delivery time when the order is complete' do
+    subject.complete
+    expect(subject.delivery_time).to_not eq(nil)
+  end
+
```
louballantyne, line 25: Does this verify a time or does it verify that the variable holds something?

louballantyne (commented):
Your code is organised well and is generally terse. Well done :)

PR #1882, opened 2021-03-15
Takeaway-challenge first 3 user stories complete

willspencer16:


lib/order.rb
```diff
@@ -0,0 +1,18 @@
...
+class Order
+	attr_reader :basket, :total
+
+	def initialize(menu = Menu.new)
+		@basket = []
+		@menu = menu
+	end
+
+	def add(dish_number)
+		@menu.dishes.each { |item| @basket << item if item[:dish_number] == dish_number }
```
eatruman, line 12: I did something similar here. I think a good next step for both of us would be to get it to raise an error if the dish doesn't exist in the menu.

spec/menu_spec.rb
```diff
@@ -0,0 +1,12 @@
+require './lib/menu.rb'
+
+describe Menu do
+	it 'includes dishes with prices' do
+		expect(subject.dishes).to be_an(Array)
+		expect((subject.dishes).all? { |item| item.has_key?(:price)}).to be_truthy
```
eatruman, line 6: Why have you used "be_truthy" here? I'm not totally clear on what "truthy" does?

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
+require './lib/order.rb'
+
+describe Order do
+	it 'has somewhere to store dishes chosen by user' do
+		expect(subject.basket).to eq([])
+	end
+
+	it 'is able to add a dish to the basket' do
+		subject.add(2)
```
eatruman, line 9: Again, I did something similar here but going forward I think we could both probably use mocking for a test like this so that if the menu changes the testing of the add method will still pass.

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+		expect(subject.basket).to eq([{:dish_number => 2, :dish => 'Beef', :price => 10}])
+	end
+
+	it 'is able to add multiple dishes to the basket' do
+		subject.add(3)
```
eatruman, line 14: Another example of where we probably could have used mocking?

eatruman (commented):
Hi Will! I think we had very similar ideas for our code and testing. I found your code easy to follow and your tests really clear and easy to understand. I think something to work on (and I need to do the same) is to use mocking to test behaviour of the method rather than state. For example, if you used mocking in your "add" or "total" tests the tests would still work even if, in the future, the menu or prices were changed.

PR #1883, opened 2021-03-15
Takeaway order

dmgreaves:


lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+def add_to_order(item)
+  item.each do |key, value|
+  @order[key] = value
+  end
+end
+
+def display_order
+  p @order
+  p @order.length
+end
```
crotchetycrow, line 21: Methods are really concise and tidy. Perhaps the order class is doing a bit too much - a Menu class to deal with display_menu and your dishes, having order deal with adding and checking out

crotchetycrow, line 21: Really nit-picky but, your code is good and I don't have much to say!! I believe it is best practice to have defined indentation

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+  account_sid = ''
+  auth_token = ''
+  client = Twilio::REST::Client.new(account_sid, auth_token)
+  from = '' # Your Twilio number
+  to = '' # Your mobile phone number
+  client.messages.create(
+  from: from,
+  to: to,
+  body: "Thank you! Your order was placed and will be delivered before 18:52")
+end
```
crotchetycrow, line 34: I can't comment on this as I didn't get far, sorry!

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'order'
+
+describe Order do
+  it { is_expected.to respond_to :display_menu}
+
+  it { is_expected.to respond_to :add_to_order}
```
crotchetycrow, line 6: You might need a more detailed test for this particular method, it passes but doesn't describe what it does or what you expect it to do.

crotchetycrow (commented):
Overall I think you produced a really tidy piece of code. It's easy to read and understand and doesn't look like you overcomplicated things. There isn't much for me to add other than writing some detailed tests rather than having  respond_to.

Its a bit nit-picky but I don't know what I'm doing so what do I know! :)

PR #1886, opened 2021-03-15
RaRaRamen Take Away

TataSher:
Still need to make MessageSender class work with Twilio API. Currently it's empty


lib/take_away.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def basket_message
+    "#{@quantity}x #{@dish}(s) added to your basket."
+  end
+  
+  def basket_summary
+    @basket.map do |dish, quantity|
+      if @menu.has_key?(dish)
+        "#{quantity}x #{dish}(s) = £#{@menu[dish] * quantity}"
```
loushark, line 31: Love the addition of the quantity and the calculation of that

lib/take_away.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def total
+    @total = @basket.map do |dish, quantity|
+      @menu[dish] * quantity
+    end
+    @total.sum
+  end
+
+  def check_out(payment)
```
loushark, line 43: Would the user need to add the payment total into the argument? 
Could the argument be pulled from the basket summary automatically when the use decides to checkout?

lib/menu.rb
```diff
@@ -0,0 +1,6 @@
+class Menu
```
loushark, line 1: Love that the menu is a separate class from the take-away. This could be built on to add different styles of menu for different new restaurant instances. nice!

spec/take_away_spec.rb
```diff
@@ -0,0 +1,54 @@
+require './lib/take_away.rb'
+
+describe TakeAway do
+  let(:menu) { { 'Shio Ramen' => 5.50, 'Shoyu Ramen' => 6 } }
```
loushark, line 4: could you test on the real instance of menu here?

lib/take_away.rb
```diff
@@ -0,0 +1,51 @@
...
+    @menu = menu
+    @basket = {}
+    @message_sender = message_sender
+  end
+
+  def read_menu
+    @menu
+  end
+
+  def order(dish, quantity = 1)
```
loushark, line 17: Could Order be a different class that your takeaway can use?

lib/take_away.rb
loushark: When running the code in irb the total cost misses the last zero e,g £15.5 instead of £15.50. For user readability could you adjust the code to ensure this last digit always shows?

lib/take_away.rb
```diff
@@ -0,0 +1,51 @@
...
+  
+  def basket_summary
+    @basket.map do |dish, quantity|
+      if @menu.has_key?(dish)
+        "#{quantity}x #{dish}(s) = £#{@menu[dish] * quantity}"
+      end
+    end
+  end
+
+  def total
```
loushark, line 36: Do you have a way to check that this total is correct against the what the user sees in the basket total?  perhaps it could throw an error if the total is ever incorrect and write a guard test for this event

lib/take_away.rb
```diff
@@ -0,0 +1,51 @@
...
+    @total = @basket.map do |dish, quantity|
+      @menu[dish] * quantity
+    end
+    @total.sum
+  end
+
+  def check_out(payment)
+    return nil unless @total.sum == payment
+
+    t = Time.now
```
loushark, line 46: Could the Time lines here be refactored to one line? or perhaps moved into a new method to use SRP?

loushark (commented):
I wonder if you could use more real testing with the real instances? e.g. testing if the read_menu shows the menu, this could actually be tested on the real instance to ensure the right menu from Menu.new has been initialized with TakeAway.new?

Your code is really easy to read and the naming conventions of your methods are super clear, telling you exactly what the method is supposed to do.
Your write up on IRB was really helpful at giving a good summary of what your program does. 

Great work Nata!! I loved reading your code!

lib/take_away.rb
TataSher: Could do! Don't really know why I added it in the first place

PR #1887, opened 2021-03-15
Menu, Order and Dish class commit

Yates101:
I'm aware the dish class seems pretty superfluous at this point - would probably refactor that out for an array of hashes. And didn't get down to integrating the twilio gem.
also have a couple of tests that I couldn't quite isolate what I wanted to be testing, but I think with the elimination of the dish class, they would of been fine

### DIFF
--- a/lib/dish.rb
+++ b/lib/dish.rb
@@ -1,0 +1,8 @@
+class Dish
+  attr_reader :name, :price
+
+  def initialize(name, price)
+    @name = name
+    @price = price.to_f
+  end
+end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,24 @@
+require_relative './dish.rb'
+require_relative './order.rb'
+
+class Menu
+  attr_reader :menu_list
+
+  def initialize
+    @dishes = []
+  end
+
+  def add_dish(dish)
+    @dishes << dish
+  end
+
+  def show_menu
+    @dishes.each do |dish|
+      puts "#{dish.name} - #{dish.price}"
+    end
+  end
+
+  def create_order(*dish)
+    Order.new(*dish)
+  end
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,38 @@
+require_relative './dish.rb'
+require 'rubygems'
+require 'twilio-ruby'
+
+# put your own credentials here
+account_sid = 'AC8987a086c237f27afe114b616e04d0eb'
+auth_token = 'd7ccfed6ac5e24e0d862d78a056b1092'
+
+# set up a client to talk to the Twilio REST API
+@client = Twilio::REST::Client.new account_sid, auth_token
+
+class Order
+  attr_reader :dishes_ordered
+
+  def initialize(*dishes)
+    @dishes_ordered = dishes
+    @order_total = total
+  end
+
+  def add
+
+  def total
+    total = []
+    @dishes_ordered.each do |dish|
+      total << dish.price
+    end
+    total.sum
+  end
+
+  def check_order
+    print_order = "Your order: "
+    @dishes_ordered.each do |dish|
+      print_order << "#{dish.name} - £#{dish.price}, "
+    end
+    print_order << "Total: £#{@order_total}"
+    print_order
+  end
+end
--- a/spec/dish_spec.rb
+++ b/spec/dish_spec.rb
@@ -1,0 +1,8 @@
+require 'dish'
+
+describe Dish do
+  subject { Dish.new("Test Dish", 10) }
+
+  it { is_expected.to respond_to(:name) }
+  it { is_expected.to respond_to(:price) }
+end
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,0 +1,33 @@
+require 'menu'
+
+describe Menu do
+  let(:dish1) { double :dish1, name: "Dish 1", price: 5 }
+  let(:dish2) { double :dish2, name: "Dish 2", price: 10 }
+
+  it { is_expected.to respond_to(:menu_list) }
+  
+  it 'stores dishes to the menu' do
+    subject.add_dish(dish1)
+
+    expect(subject.menu_list).to include(dish1)
+  end
+
+  context 'given a list of dishes' do
+    before do
+      subject.add_dish(dish1)
+      subject.add_dish(dish2)
+    end
+
+    it 'shows me a list of all the dishes' do
+      expect(subject.show_menu).to include(dish1, dish2)
+    end
+
+    # it 'shows me the name and price of each dish' do
+    #   expect(subject.show_menu).to eq("#{dish1.name} - #{dish1.price}\n#{dish2.name} - #{dish2.price}")
+    # end
+
+    it 'creates a new order' do
+      expect(subject.create_order(dish1, dish2)).to be_an Order
+    end
+  end
+end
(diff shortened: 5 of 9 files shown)


lib/dish.rb
```diff
@@ -0,0 +1,8 @@
+class Dish
+  attr_reader :name, :price
+
+  def initialize(name, price)
+    @name = name
+    @price = price.to_f
+  end
+end
```
mstrivens, line 8: This looks nice and simple. It's easily readable and sticks to SRP.

mstrivens: Does this mean you can have one dish on the menu? Or can you add multiple?

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+  attr_reader :menu_list
+
+  def initialize
+    @dishes = []
+  end
+
+  def add_dish(dish)
+    @dishes << dish
+  end
+
```
mstrivens, line 14: This is a good way of adding dishes to the order. Should it be linked to the dishes class?

mstrivens: I think was confused. Does the require relative mean you don't need to initialize a new instance of Dish?

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+  def show_menu
+    @dishes.each do |dish|
+      puts "#{dish.name} - #{dish.price}"
+    end
+  end
+
```
mstrivens, line 20: This looks good.  I can imagine how it would print out.

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
...
+  def create_order(*dish)
+    Order.new(*dish)
+  end
+end
```
mstrivens, line 24: Nice, this makes sense. It looks like it adds any dishes in the argument to a new instance of the Order class

lib/order.rb
```diff
@@ -0,0 +1,38 @@
+require_relative './dish.rb'
+require 'rubygems'
+require 'twilio-ruby'
+
+# put your own credentials here
+account_sid = 'AC8987a086c237f27afe114b616e04d0eb'
+auth_token = 'd7ccfed6ac5e24e0d862d78a056b1092'
+
```
mstrivens, line 8: I think you need to hide these in env variable as twilio will automatically change you auth_token when you upload it onto github because it's exposed to hackers

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+# set up a client to talk to the Twilio REST API
+@client = Twilio::REST::Client.new account_sid, auth_token
+
+class Order
+  attr_reader :dishes_ordered
+
+  def initialize(*dishes)
+    @dishes_ordered = dishes
+    @order_total = total
```
mstrivens, line 17: I like that this isn't dependent on instances from another class. I don't know what works better in terms of dependency but this seems simpler

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def add
+
+  def total
+    total = []
+    @dishes_ordered.each do |dish|
+      total << dish.price
+    end
+    total.sum
+  end
```
mstrivens, line 28: This is good, it shows better understanding than what i tried

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def check_order
+    print_order = "Your order: "
+    @dishes_ordered.each do |dish|
+      print_order << "#{dish.name} - £#{dish.price}, "
+    end
+    print_order << "Total: £#{@order_total}"
+    print_order
+  end
+end
```
mstrivens, line 38: Good. The classes look good, understandable and DRY

spec/dish_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'dish'
+
+describe Dish do
+  subject { Dish.new("Test Dish", 10) }
+
+  it { is_expected.to respond_to(:name) }
+  it { is_expected.to respond_to(:price) }
+end
```
mstrivens, line 8: Does this test if there is a value assigned to :name and :price like line 4 is indicating or that the subject is aware of the attribute?

spec/menu_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'menu'
+
+describe Menu do
+  let(:dish1) { double :dish1, name: "Dish 1", price: 5 }
+  let(:dish2) { double :dish2, name: "Dish 2", price: 10 }
```
mstrivens, line 5: good use of let

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
+require_relative './dish.rb'
+require_relative './order.rb'
+
+class Menu
+  attr_reader :menu_list
```
mstrivens, line 5: I'm a bit confused as to what menu_list is. I can't see another instance variable with the same name

spec/menu_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  it { is_expected.to respond_to(:menu_list) }
+  
+  it 'stores dishes to the menu' do
+    subject.add_dish(dish1)
+
+    expect(subject.menu_list).to include(dish1)
```
mstrivens, line 12: I couldn't find a method or variable with the name menu_list so not sure what it's testing. Maybe the name of the method was changed?

spec/menu_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  context 'given a list of dishes' do
```
mstrivens, line 15: nice use of context

spec/menu_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    it 'shows me a list of all the dishes' do
+      expect(subject.show_menu).to include(dish1, dish2)
+    end
+
+    # it 'shows me the name and price of each dish' do
+    #   expect(subject.show_menu).to eq("#{dish1.name} - #{dish1.price}\n#{dish2.name} - #{dish2.price}")
+    # end
+
+    it 'creates a new order' do
+      expect(subject.create_order(dish1, dish2)).to be_an Order
```
mstrivens, line 30: This looks like a good way of writing this. I haven't seen a test for an instance of a class written like this before

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  let(:dish1) { double :dish1, name: "Dish 1", price: 5 }
+  let(:dish2) { double :dish2, name: "Dish 2", price: 10 }
+
+  it { is_expected.to respond_to(:dishes_ordered) }
+
+  context 'when passed 2 dishes' do
+    subject { Order.new(dish1, dish2) }
+
+    it 'saves both dishes in the order' do
+      expect(subject.dishes_ordered).to include(dish1, dish2)
```
mstrivens, line 13: This makes sense and good use of doubles

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    it 'calculates a total from the list of dishes' do
+      expect(subject.total).to eq(dish1.price + dish2.price)
+    end
+
+    it 'customer can check the total calculation and view their order' do
+      expect(subject.check_order).to include(dish1.price.to_s, dish2.price.to_s, subject.total.to_s)
+    end
```
mstrivens, line 22: This looks like a good test

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+end
```
mstrivens, line 24: This code looks good. Simple and non-repetitive and looks like it effectively does what it's meant to do. I don't understand enough to know the best way to set up dependencies, another way to do it might have been to have a class to bring the other classes together rather than have the classes passing information to each other directly.

PR #1889, opened 2021-03-22
Completed the 4 main user stories, refactored slightly and attempted to implement API (but not complete)

Leghnider:


lib/menu.rb
```diff
@@ -0,0 +1,23 @@
...
+            "Bruschetta" => 3,
+            "Burrata" => 4.5,
+            "Lasagne" => 10,
+            "Gelato" => 2.5,
+            "Tiramisu" => 4.99
+        }
+    end
+
+    def show_menu
+        @dishes.map { |key, value| "#{key}, £#{value}"}
```
Dominic-Taylor-Dev, line 17: Maybe could have been more descriptive with the variable names in 'map' e.g. | menu_item, price | but think this still does exactly the right job...:)

lib/message.rb
```diff
@@ -0,0 +1,35 @@
+require 'twilio-ruby'
+# require 'dotenv'
```
Dominic-Taylor-Dev, line 2: Think you were very much on the right lines with this. The 'staff' answer also uses this dotenv approach to access environmental variables (rather than needed to have them hard-coded into this file). Just needed an extra couple of steps to make it work, I guess...?

lib/message.rb
```diff
@@ -0,0 +1,35 @@
...
+    def initialize(twilio = Twilio::REST::Client.new(ENV['TWILIO_ACCOUNT_SID'], ENV['TWILIO_AUTH_TOKEN']))
+        @twilio = twilio
+    end
+
+    def send_sms(phone_number)
+        time = Time.now
+        delivery_time = Time.now + 1*60*60
+        @twilio.messages.create({
+            from: ENV['TWILIO_PHONE'],
+            to: phone_number
```
Dominic-Taylor-Dev, line 29: I think I missed the point at which you initialized the variable for recipient's phone number...? I can see it's being passed as an argument to this method, but where does that come from? Where do you call this method (presumably from another class but I can't find it)?

lib/message.rb
```diff
@@ -0,0 +1,35 @@
...
+
+class Message
+    attr_reader :twilio
+    def initialize(twilio = Twilio::REST::Client.new(ENV['TWILIO_ACCOUNT_SID'], ENV['TWILIO_AUTH_TOKEN']))
+        @twilio = twilio
+    end
+
+    def send_sms(phone_number)
+        time = Time.now
+        delivery_time = Time.now + 1*60*60
```
Dominic-Taylor-Dev, line 26: 1*60*60 == 60*60?

lib/order.rb
```diff
@@ -0,0 +1,32 @@
+require_relative 'menu'
+
+class Order
+    attr_reader :dishes, :order
+    
+    def initialize
+        @order = Hash.new(0)
+        @total = 0
+        @dishes = Menu.new.dishes
+        @time = (Time.now + 1*60*60).strftime("%I:%M%p")
```
Dominic-Taylor-Dev, line 10: Your code totally fine as far as I can see, but you know my feelings about 12 vs 24 hour clock ;)

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def add(dish, quantity)
+        fail "This order isn't on the menu!" unless dishes.has_key?(dish)
+        @order[dish] += quantity
+    end
+
+    def show_order 
+        @order
+    end
```
Dominic-Taylor-Dev, line 20: Arguably asking the class to do more than one thing here (same with the complete method below).

Gemfile
```diff
@@ -6,3 +6,5 @@ group :test do
   gem 'simplecov', require: false, group: :test
   gem 'simplecov-console', require: false, group: :test
 end
+
+gem 'twilio-ruby'
```
Dominic-Taylor-Dev, line 10: GENERAL COMMENT ON THE WHOLE THING: 10x better than what I did so take my comments with a pinch of salt. I think some of them are also just where you were still work in progress on implementing the SMS feature. Good job, though :)

lib/message.rb
```diff
@@ -0,0 +1,35 @@
+require 'twilio-ruby'
+# require 'dotenv'
+# require 'order'
+
+account_sid = '***'
+auth_token = '***'
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+# from = '+447403930982' # Your Twilio number
```
Dominic-Taylor-Dev, line 9: Think you could ultimately have had this as ENV["PHONE"] when uncommented and used somewhere

lib/message.rb
```diff
@@ -0,0 +1,35 @@
...
+
+# client.messages.create(
+# from: from,
+# to: to,
+# body: "Your order is being prepared! It should be with you in "
+# )
+
+class Message
+    attr_reader :twilio
+    def initialize(twilio = Twilio::REST::Client.new(ENV['TWILIO_ACCOUNT_SID'], ENV['TWILIO_AUTH_TOKEN']))
```
Dominic-Taylor-Dev, line 20: nice!!

lib/message.rb
```diff
@@ -0,0 +1,35 @@
...
+        @twilio = twilio
+    end
+
+    def send_sms(phone_number)
+        time = Time.now
```
Dominic-Taylor-Dev, line 25: You already have an instance variable for time elsewhere (in 'Order'), so I guess it depends how you're chaining classes and methods, but you might not need the repetition if you end up instantiating 'Message' from within another class. I wasn't quite sure about the way you were going to do things, but I thought you might have been planning to instantiate message inside the 'complete' method of the Order class anyway.

PR #1896, opened 2021-04-11
Takeaway challenge

fizzy-fifs:


fizzy-fifs:
User story 1:

        As a customer

I completed this story by creating a dish class and a takeaway class with a menu method.

User story 2:

        As a customer
I got stuck on this story. I created an order class with a 'add_to_order()' method that was meant to read from 'Takeway.menu' and add to an empty 'order_list' hash based on the user's input (i.e. argument passed through the add_to_order method), but was unsuccessful in my attempts.

User stories 3&4 have not been attempted as at yet.

lib/dishes.rb
```diff
@@ -0,0 +1,25 @@
+class Dishes
```
tremella, line 1: If this class is making a single dish, it might be clearer for a reader if you call it Dish instead of Dishes - because Dishes makes it sound like this class might represent a menu.

lib/dishes.rb
```diff
@@ -0,0 +1,25 @@
...
+    if availability == "yes"
+      @available = true
+    elsif availability == "no"
+      @available = false
+    else
+      @available
+    end
+  end
+
+  def list_available_dishes
```
tremella, line 21: Does a single dish need to know about a list_of_available dishes? - that feels like a job for a higher up class. Was this function moved and accidentally left over? I can see the Takeaway class has an attribute for this.

lib/dishes.rb
```diff
@@ -0,0 +1,25 @@
+class Dishes
+  attr_reader :name, :price, :available, :list_available_dishes
+
+  def initialize(name, price, available = true)
+    $name = name
```
tremella, line 5: I'm not familiar with using a global variable in this context, what's the plan for it?

lib/order.rb
```diff
@@ -0,0 +1,17 @@
+require_relative 'dishes'
+require_relative 'takeaway'
+class Order
+  attr_reader :order_list
+
+  def initialize(order_list = {})
+    @order_list = order_list
+  end
+
+  def add_to_order(add)
```
tremella, line 10: I'm not totally sure how this function is meant to work, maybe a comment or some info in the Readme would be handy for reviewers?

tremella (commented):
I kind of had a hard time understanding the code - partially due to some syntax/style I don't know, but mostly because I think we approached this problem in very different ways. I think submitting a readme and a domain model would be really useful from a reviewer perspective, and being careful with naming things. 

That said, your tests seemed well-implemented, readable and tidy.

PR #1897, opened 2021-04-11
Tiffany's Takeaway Challenge

tiffanyvallo:


Domain_Model.md
```diff
@@ -0,0 +1,60 @@
+## Domain Model
```
phileeep, line 1: Very nice use of the Domain model and formatting clearly with the actions vs properties

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
```
phileeep, line 1: Good work on splitting takeaway into two separate classes.

Domain_Model.md
```diff
@@ -0,0 +1,60 @@
...
+ 
+| Class                             | Order                           | 
+| --------------------------------- | --------------------            |
+| Properties(instance variables)    | @total                          | 
+| Actions(methods)                  | basket, add_to_basket, checkout | 
+
+| Class                         | Text              | 
+| ------------------------------| ---------         | 
+| Properties(instance variables)| @time             | 
+| Actions(methods)              | text_confirmation |
```
phileeep, line 60: Possibly good to expand this with how the program works in your Readme

spec/menu_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  it "stores the menu as a hash" do
+    expect(subject.dish).to be_an_instance_of Hash
+  end
+
+  it "responds to the show menu method" do
+    expect(subject).to respond_to(:show_menu)
+  end
+
+  it "can retrun the each dish with it's corresponding price as a string" do
```
phileeep, line 15: small typo

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  attr_reader :order, :menu, :total
+
+  def initialize
+    @order = []
+    @total = 0
+    @menu = Menu.new
+  end
+
+  def add_to_basket(food, quantity = 1)
```
phileeep, line 13: Good use of SRP and keeping the code concise and default value for quantity.

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+    quantity.times { @order << food }
+  end
+
+  def checkout
+    @order.map do |food|
+      @total += @menu.dish[food] 
+    end
+    "Your order comes to £#{@total}"
+  end
+
```
phileeep, line 23: Missing user story of user receiving a text

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+  def add_to_basket(food, quantity = 1)
+    quantity.times { @order << food }
+  end
+
+  def checkout
+    @order.map do |food|
+      @total += @menu.dish[food] 
+    end
```
phileeep, line 20: In the rubric &  user story 3 they suggested adding a set so the use could check it is the right amount

PR #1899, opened 2021-04-11
Incomplete Weekend Challenge 

C-A-Tech:


lib/menu.rb
```diff
@@ -0,0 +1,18 @@
...
+      "sandwhich" => 3.20, 
+      "shepherd's pie" => 4.50, 
+      "pot roast and gravy" => 6.50, 
+      "macaroni and cheese" => 4.75
+    }
+  end 
+
+  def see_menu
+    @available_dishes.each do |dish, price|
+      p "#{dish}: #{price}"
```
kerrimcm, line 15: Might be nice to include the £ sign inside your output here

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def add(request)
+    fail 'That item is not available' if check_menu(request) == false
+
+    @order_placed << request
+    add_to_balance(request)
+  end
+
+  def check_menu(request)
+    menu = Menu.new
```
kerrimcm, line 19: Instead of instantiating new Menu class each time, could try initialise it and call on methods stored inside the Menu class i.e. check_menu moved into Menu class

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+class Order
+  attr_reader :order_placed, :balance
+
+  def initialize
+    @order_placed = []
+    @balance = 0
+  end
+
+  def add(request)
+    fail 'That item is not available' if check_menu(request) == false
```
kerrimcm, line 12: check_menu could be changed to a boolean output, so something like check_menu? so it will already evaluate to true/false removing need for '== false'

lib/order.rb
```diff
@@ -0,0 +1,37 @@
...
+    @balance += menu.available_dishes[request]
+  end
+
+  def check_balance
+    menu = Menu.new
+    statement = ""
+    @order_placed.each do |item| 
+      statement += "#{item}: #{menu.available_dishes[item]}, " 
+    end
+    p "#{statement}Total: #{balance}"
```
kerrimcm, line 34: This is a really cool way of doing this!

spec/order_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  it 'initializes with an empty order' do
+    expect(subject.order_placed).to eq [] 
+  end
+
+  describe "#add" do
+
+    it 'adds available dishes to order' do
+      # need to input a double here so test is unaffected by menu class
+      subject.add("sandwhich")
+      expect(subject.order_placed[0]).to eq("sandwhich")
```
kerrimcm, line 14: Could change order placed into a hash so you can store the item ordered as well as the quantity - so let's say they want three sandwiches so it would be something like `expect(subject.order_placed).to eq({sandwich => 3})` and `.add `would take two arguments

spec/order_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    it 'checks if requested item is on menu' do
+      # need to input a double here so test is unaffected by menu class
+      expect(subject.check_menu("sandwhich")).to eq true
+    end
```
kerrimcm, line 20: Again, as above, this is something I would make the Menu class responsible for

spec/order_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  describe "#balance" do
+
+    it 'calculates charge' do
+      subject.add("sandwhich")
+      expect(subject.balance).to eq 3.20
+    end    
+    
+    it 'allows customer to check total' do
+      order = Order.new
```
kerrimcm, line 37: could replace this with subject which already evaluates to `Order.new`

kerrimcm (commented):
Really nice clear code! Could make the Menu class responsible for a little bit more but overall it was very clear understanding your intentions with each line 😊

PR #1908, opened 2021-04-12
Pull request for my takeaway challenge

tobyjessup:


spec/takeaway_spec.rb
```diff
@@ -1,14 +1,23 @@
...
+  context 'menu' do
+    it 'shows the menu including dishes and prices' do
+      expect(takeaway.provide_menu).to eq(menu_output)
     end
-end
\ No newline at end of file
+  end
+
+  it 'can order some dishes' do
+    expect(order).to receive(:add).twice
```
noarfarber, line 19: not familiar with the "twice" method, but I'm guessing you might want to add dishes three times for example, would it still work for that purpose?

lib/menu.rb
```diff
@@ -1,13 +1,14 @@
...
-            "%s £%.2f" % [dish.to_s.capitalize, price]
-        end.join(", ")
-    end
-end
\ No newline at end of file
+  def print
+    dishes.map do |dish, price|
+      "%s £%.2f" % [dish.to_s.capitalize, price]
+    end.join(", ")
+  end
```
noarfarber, line 13: I like the way you decided to print your menu.

PR #1910, opened 2021-05-07
Finished the challenge

LCzoboriek:


lib/Order.rb
```diff
@@ -0,0 +1,46 @@
+require_relative 'menu.rb'
+
+# This new class is going to create an order
+class Order
+
+  attr_reader :dishes
+
+  def initialize
+    @dishes = []
+    @menu = Menu.new
```
ChescaRobertson, line 10: Good that an instance of menu class is initialised here, keeps code DRY, stops you having to initialise every time you use Menu within this Order class

spec/dishes_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'menu'
+
+describe Menu do
+  
+  subject(:menu) { described_class.new }  
+    
+  describe '#list_dishes' do
```
ChescaRobertson, line 7: This is a good starting unit test but could maybe be taken out as further tests implicitly check this anyway

spec/order_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'order.rb'
+
+describe Order do
+  subject(:order) { described_class.new }
+  let(:menu) { double() }
```
ChescaRobertson, line 5: Nice that you've started to make a double of menu but I can't see where it is used within the tests? Perhaps you were getting to this later?

lib/Order.rb
```diff
@@ -0,0 +1,46 @@
...
+    @complete = false
+  end
+
+  def add_dish(name, amount = 1)
+    @name = name
+    @amount = amount
+    fail "This dish isn't available, please run menu.list_dishes to see the dishes" unless available
+
+    @dishes << { :name => name, :amount => amount }
```
ChescaRobertson, line 19: Great use of arrays and hashes, really clear and simple

EdwardAndress: If you use a symbol as a hash key you can use this syntax

`{name: name, amount: amount}`

It's more idiomatic that way.

lib/Order.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def available
```
EdwardAndress, line 22: The convention here would be for the method to end with a `?` to indicate that it returns a bool.  It also reads like a question, which is nice, don't you think?

lib/Order.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def complete
+    puts "Your total order today is £#{total_price}"
+    @complete = true
+    message
+  end
+
+  def message
+    raise "Sorry your order isn't complete" unless @complete == true
```
EdwardAndress, line 39: Nice guard clause here.  You could also make the message method private to prevent it being called from outside the class.

spec/order_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      order.add_dish("Curry", 1)
+      expect(order.total_price).to eq 13
+    end
+  end
+
+  context '#confirmation_of_order' do
+    it 'checks that a confirmation message is sent out after completing the order total' do
+      order.add_dish("Pizza", 2)
+      order.add_dish("Curry", 1)
+      order.complete
```
EdwardAndress, line 31: here you could do expect `{order.complete}.not_to raise_error` since message is called in complete.

EdwardAndress:
Nice solution, Luke!  I've left a few very minor comments in line against the code.

lib/Order.rb
LCzoboriek: Yeh that would have been more applicable, will use that next time

PR #1912, opened 2021-05-08
Becca's Takeaway challenge

BecaLParker:
Thanks for reviewing - in particular I'd love feedback to help me implement the stubbing properly for the SMS API testing, as I did not achieve that (despite many tries).


lib/restaurant.rb
```diff
@@ -0,0 +1,66 @@
...
+    @sender = sender
+    @basket = {}
+  end
+
+  def show_menu
+    menu.list
+  end
+
+  def order(dish, portions = 1)
+    fail "Sorry, #{dish} aren't on the menu" unless menu.list.include?(dish)
```
GoWebMe2020, line 19: I like the fail that you have added here.

lib/send_sms.rb
```diff
@@ -0,0 +1,21 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+
+class SendSMS
+  def initialize
+    account_sid = ENV['TWILIO_SID']
```
GoWebMe2020, line 6: Great idea using external variables to store your account details. I am curious to know what dependency you used for this? Was it dotenv?

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'restaurant.rb'
+require 'menu.rb'
+
+describe Restaurant do
+  subject { 
+    menu = double(Menu.new)
```
GoWebMe2020, line 6: I may be incorrect, but I was under the impression that we were not supposed to rely on the other classes during testing. In this case, you are referring directly to the Menu class from the restaurant class.

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  describe '#basket_summary' do
+    it 'outputs subtotals for each of the dishes ordered' do
+      subject.order("fries", 5)
+      subject.order("burger", 5)
+      expect(subject.basket_summary).to eq "5 portions of fries @ £1.50 each = £7.50, 5 portions of burger @ £2.00 each = £10.00"
+    end
+  end
+
+  # describe '#delivery_time' do
```
GoWebMe2020, line 47: I had just as much difficulty testing the delivery time.

GoWebMe2020 (commented):
With the little I know so far, I think this is very written. I have actually learned a thing or two from this.

lib/send_sms.rb
BecaLParker: Thanks! I used [this model from the precourse](https://makersacademy.teachable.com/courses/256825/lectures/3989201):

Put your secret key in an environment variable and then read it from your code when you need it. For example:

`export SECRET_KEY=12345abcde`

Then, in your Ruby code, read the value:

`secret_key = ENV['SECRET_KEY']`

This way you'll be able to share your code with anyone and upload it to a GitHub repository without compromising your keys.

lib/restaurant.rb
BecaLParker: Thanks - it covers an edgecace so that the customer can't get too carried away making an invalid order and break the 'total' / 'basket_summary' functions.

PR #1920, opened 2021-05-10
Takeaway Challenge - Gunel MC

GunelMC:
I would really appreciate feedback on how to extract my order class into a separate basket class, and how to stub Twilio message functionality - this was one of my blockers during the challenge( please see pending Twilio message test and another block commits for further details). In general, any feedback on MockException Errors and stubbing would be helpful! 

I also struggled to access menu.price in my order file to calculate the total price, despite requiring the file and making menu instance variable accessible as an attribute reader. In the end, I solved the problem using the local variable to access the prices of meals, however, I would appreciate feedback on what was wrong - I recorded the error on my errors branch. 

I would also appreciate a review of my UML diagram. 

Thanks a lot!


README.md
```diff
@@ -47,37 +38,62 @@ So that I am reassured that my order will be delivered on time
...
+1. Open ~/.zshrc or ~/.profile depend on your Mac OS version
 
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/master/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+2. Append in the file:
+   export TWILIO*ACCOUNT_SID="ACCOUNT SID"
+   export TWILIO_AUTH_TOKEN="AUTH TOKEN"
+   export TWILIO_NUMBER="PHONE NUMBER"
+   export USER_MOBILE="Verified Caller"
```
rdupplaw, line 73: Can't they just make a .env file containing this?

lib/order.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'menu'
```
rdupplaw, line 1: Do you need to require menu here if you're using dependency injection on line 5?

lib/order.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'menu' 
+class Order
+  attr_reader :basket
+
+  def initialize(menu)
+    @menu = menu
+    @basket = {}
```
rdupplaw, line 7: Storing basket items as a hash with their quantity is good, I didn't think to do this

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'menu'
+
+describe Menu do
+
+  let(:dishes) do
+    described_class::DEFAULT_MENU.map { |dish, price| "#{dish.to_s.capitalize}: £#{'%.2f' % price}" }.join(',')
+  end
```
rdupplaw, line 7: I don't think this is a good test. This is the same code as in your Menu class' #view_menu method. You're testing the code against itself. I think it would be better to literally write out what you expect this code to return.

spec/message_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'message'
+require 'dotenv'
+require 'twilio-ruby'
+Dotenv.load
+
+
+describe Message do
+  let(:client) { double :client, messages: messages }
+  let(:messages) { double :messages }
```
rdupplaw, line 9: You have initialized and assigned messages after it was used in client on line 8

PR #1925, opened 2021-05-10
Jay Mullen

Jaikido007:


spec/menu_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'menu'
+require 'order'
+
+describe Menu do
+# dependence injection. Inject dishes hash to menu.
+  subject(:menu) { described_class.new(dishes) }
```
lkilk, line 6: like the use of described_class.new here, makes the tests very neat

spec/sms_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'sms'
+require 'takeaway'
+
+describe SMS do
+  subject(:sms) { described_class.new(config, client: client) }
+
```
lkilk, line 6: like the use of doubles and stubs in the tests, makes the tests very concise

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+require_relative './order.rb'
+
+class Menu
+  attr_reader :dishes
+
+  def initialize(dishes)
+    @dishes = dishes
+  end
+
+# setting float to 2.decimal places (Synytax from 'O'Riley - Head First Ruby' book)
```
lkilk, line 10: very nice, no idea how that works but i'm impressed

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+  def print
+    dishes.map do |title, price|
+      "%s £%.2f" % [title.to_s.capitalize, price]
```
EdwardAndress, line 13: Nice!

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+    end.join(", ")
+  end
+# Look at ! in greater detail
+  def has_dish?(dish)
```
EdwardAndress, line 17: Nice and concise.

Another option would be

`dishes.keys.include? dish`

Which do you prefer?

There's probably a dozen other options as well! :)

lib/order.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def add(dish, quantity)
+    fail NoDishError, "#{dish.capitalize} is not on the menu!" unless menu.has_dish?(dish)
+
+    dishes[dish] = quantity
+  end
+
+  def total
+    item_totals.inject(:+)
```
EdwardAndress, line 17: Nice use of inject!

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
+require_relative './order.rb'
+require_relative './sms.rb'
+
+class Takeaway
+
+  def initialize(menu:, order: nil, sms: nil)
+    @menu = menu
+    @order = order
```
EdwardAndress, line 8: How could you handle multiple live orders?

spec/menu_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  let(:dishes) do
```
EdwardAndress, line 8: This structure for dishes is neat.  An alternative...

```
[
{name: 'lasagne', price: 8.99}
]
```

Here, the keys are consistent and that is sometimes useful when you're iterating.

spec/sms_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  let(:client) { double(:client, messages: messages) }
+  let(:messages) { double(:messages) }
+# implementing Twilio: https://rails.devcamp.com/trails/rest-microservices/campsites/sms-messages/guides/creating-a-rspec-stub-to-mimic-sending-sms-messages
+  let(:config) do
+    {
```
EdwardAndress, line 11: To avoid uploading things that you might wish to keep private: https://www.rubyguides.com/2019/01/ruby-environment-variables/

EdwardAndress:
This is a really great solution, Jay, with lots of nice little touches that show you're reading around and implementing the things you're discovering.

PR #1926, opened 2021-05-10
Takeaway challenge

mvgame74:


lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+    puts "M E N U"
+    puts "-------"
+    @menu.each do |key, value|
+      puts "#{key}:  GBP #{value}"
+    end
+  end
+end
+
+menu = Menu.new
+menu.display_menu
```
BecaLParker, line 34: I love the attention you've given to to format of the menu output  👍🏻 
I was easily able to reproduce that by requiring the file in irb.

lib/order.rb
```diff
@@ -0,0 +1,18 @@
+require 'menu'
+
+class Order
+  attr_reader :menu
+  
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @order =[]
+  end
```
BecaLParker, line 9: Good idea to use dependency injection to get the menu class and order class interacting. 👍🏻 

This will help later down the line when you want to give your takeaway_spec all the information it would get from the menu class - without it having to ask the menu class (through mocking/doubles).

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+      "Seaweed Soup" => 7.99,
+      "Bone Marrow Soup" => 6.99,
+      "Haemul Pajeon" => 9.49,
+      "Tteokbokki" => 8.49,
+      "Japchae" => 10.99,
+      "Kimchi Jjigae" => 11.99,
+      "Jjampong with Noodles" => 13.99,
+      "Yangnum Chicken" => 12.99,
+      "Ganjang Chicken" => 12.99,
+      "Beef Bulgogi Bibimbap" => 11.99}
```
BecaLParker, line 21: A hash structure seems sensible - you'll be able to easily look up prices when you come to implement the 'display total' functionality 👍🏻

BecaLParker (commented):
See specific comments too, but overall:

👍🏻  You've covered the first user story with thought given to dependency injection, data structures, and clear naming - all of which will help with the later user stories.

As you work on the other user stories, consider updating the README explaining your implementation (eg how to replicate the interaction in irb) so that users can easily see what commands to input (if any).

PR #1927, opened 2021-05-10
takeawaychallengenotests

JamesMcMull:


lib/order.rb
```diff
@@ -0,0 +1,50 @@
+require 'lib/restaurant.rb'
+require 'lib/messaging.rb'
+
+class Order
+
+  def initialize()
```
charlie-casserly, line 6:
```suggestion
  def initialize(restaurant = Restaurant.new)
    @restaurant = restaurant
  end
  
  def get_menu
    @restaurant.get_menu
  end
```
I think this is a more flexible way of coding your get_menu function. It will allow you to test order.get_menu in isolation by initializing an Order object with a mock of Restaurant instead of a hard-coded Restaurant.new

charlie-casserly (commented):
Nice solution! Looking forward to pairing later.

PR #1928, opened 2021-06-06
Zsuzsanna's takeaway challenge

MrsVer:


bin/takeaway.rb
```diff
@@ -0,0 +1,39 @@
+#! /usr/bin/env ruby
+
+require 'dotenv'
+require "./lib/takeaway"
+require "./lib/menu"
+
+Dotenv.load
```
merrynhr, line 7: Really great way to test your code! I will try this out in the future!

lib/sms.rb
```diff
@@ -0,0 +1,31 @@
+require "twilio-ruby"
+
+class SMS
+
+  TIME_FORMAT = "%H:%M"
```
merrynhr, line 5: Really clever :)

merrynhr (commented):
Really clear code

PR #1930, opened 2021-06-06
Craig's Takeaway Attempt

craigea92:


lib/takeaway.rb
```diff
@@ -0,0 +1,27 @@
+class Takeaway
+
+  attr_reader :menu, :orders
+
+  def initialize
+    @orders = []
+  end
+
+  def take_order(order)
+    @new = order
```
frank-mck, line 10: This @new variable could be a bit confusing. I would suggest @new_order :)

frank-mck (commented):
Well written code and tests :)

PR #1932, opened 2021-06-06
Jenny's takeaway challenge

jen0828:
Please see my README (README_JM.md). Thanks

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,9 @@
+class Menu
+  
+  attr_accessor :dish, :price
+
+  def initialize(dish,price = 1)
+    @dish = dish
+    @price = price
+  end
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,32 @@
+require_relative 'menu'
+
+class Order
+
+  attr_reader :basket, :count, :total
+
+  def initialize
+    @basket = { }
+    @count = 0
+    @total = 0
+  end
+
+  def add_to_basket (dish,amount)
+    @basket[dish] = amount
+  end
+
+  def current_basket
+    @basket
+  end
+
+  def count_item
+    basket.each { |dish, amount| @count += amount }
+    @count
+  end
+
+  def total_in_basket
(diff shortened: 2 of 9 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,9 @@
+class Menu
+  
+  attr_accessor :dish, :price
+
+  def initialize(dish,price = 1)
+    @dish = dish
+    @price = price
+  end
```
coo990, line 8: Nice short class, succinct with the Menu

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+    basket.each { |dish, amount| @count += amount }
+    @count
+  end
+
+  def total_in_basket
+    basket.each { |dish, amount| @total += (dish.price * amount)}
+    @total
+  end
+end
+
```
coo990, line 31: Good encapsulation of class

spec/confirmation_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'confirmation'
+
+describe Confirmation do
+  subject(:confirmation) { described_class.new }
+  let(:payment) { :payment }
+
+  it 'raises an error if the money does not match the total' do
+    expect { subject.checkout(payment) }.to raise_error(RuntimeError,'something went wrong, try again')
```
coo990, line 8: Clear message when raising an error, is `RuntimeError` needed in the parameter?

coo990 (commented):
I really liked how clear each class does, not all in one place, same with the spec files as well. Nice Jenny!

spec/confirmation_spec.rb
jen0828: This syntax can be used to specify the error, so I tried it. 
Example:  expect { raise "oops" }.to raise_error(RuntimeError, "oops")

lib/menu.rb
jen0828: Thanks, Chuka! Now I'm looking at your Takeaway Class & menu set up, I think I need to work on this further, perhaps with a new Class, so thank you!

PR #1933, opened 2021-06-06
Franks Takeaway Challenge

frank-mck:


lib/orderpizza.rb
```diff
@@ -0,0 +1,27 @@
+require './lib/menu'
+
+class Orderpizza < Menu
```
craigea92, line 3: I really like the use of class inheritance here for Orderpizza < Menu.

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
+
+
```
craigea92, line 2: I can read your code and understand what each piece is doing, I really like it.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'orderpizza'
```
craigea92, line 1: Good use of TDD. Tests are clear and make sense.

PR #1939, opened 2021-06-07
Tarun's Pull Request

TarunTheo13:
- Did not use doubles
- Did not use TWILIO API

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,8 @@
+Menu = {
+  burger: 9,
+  tacos: 8.5,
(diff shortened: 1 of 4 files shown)


OptimusWall-E:
def add_item(food, quantity = 1)
    quantity.times { @order << food.to_sym }
  end

Appreciations
this is an elegant method. The default value combines well with quantity.times method

Suggestions
to someone coming newly to the code, add_item may sound like either adding to Menu or to the order. It's tricky because 'order' doesn't have loads of synonyms

PR #1941, opened 2021-06-07
Pull request for Takeaway Challenge

mikey-roberts:


Ashley-Slaney (commented):
Appreciations 

- Clear code (readable and clear/short methods)
- Good use of RSpec tests


Suggestions

- Could commit more often rather than implementing many things at once so you can easily go back to an older version without losing much code if an issue arrises and it would make each commit easier to read

PR #1942, opened 2021-06-07
Misha's pull request

OptimusWall-E:


TarunTheo13:
Appreciations:
- Comments at the start explaining thought process
- Your creation of a hash within an array - really like this, think it has good implications for scaling up the program
- Easy to read code, made it work not using excess lines

Suggestions:
- Delete redundant comments, will make your code easier to read
- Write more tests
- Write a README
- Change prices of the menu items so it is clear they are prices and not the number of the item
- receipt method is currently not working on the version I have tried from your github, but think you have updated it since

PR #1943, opened 2021-06-11
first commit

EmilyFitzs:


EdwardAndress:
Hey Emily, this is a really nice solution – great work!!  I'm going to leave a few comments in line against the code.

EdwardAndress:
Re: Twilio – it's possible to get the text to send without buying a Twilio number but the error messages and UI can be unhelpful sometimes.  I was chatting to Juel about this and she said that Twilio provides a trial number that is hidden behind a button..?

lib/send_sms.rb
```diff
@@ -0,0 +1,29 @@
+require 'twilio-ruby'
```
EdwardAndress, line 1: It was a good call to encapsulate this in it's own class 👍

lib/takeaway.rb
```diff
@@ -0,0 +1,22 @@
...
+  def initialize
+  @order = []
+  end
+
+  def add_dish(dish)
+    fail "Dish not on menu!" if !@@menu.dishes.key?(dish)
+    @order << dish
+  end
+
+  def total
```
EdwardAndress, line 17: This is nice and neat.  Since it's possible to pass a block to sum, there might be a more concise way of writing this though.

How about...

```ruby
@order.sum { |dish| @@menu.dishes[dish] }
```

EdwardAndress: Which do you prefer in terms of readability?

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'menu'
+
+describe Menu do
+  
+  let(:dishes) { { vegan_curry: 13,
```
EdwardAndress, line 5: I'm not sure you need this hash of dishes here.

EdwardAndress: Maybe you were going to use it to test `subject.dishes`?

lib/send_sms.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class SMS
+  attr_reader :client, :config
+  TIME_FORMAT = "%H:%M"
```
EdwardAndress, line 5: Nice use of a constant for the time format!  This really helps readability.

lib/takeaway.rb
```diff
@@ -0,0 +1,22 @@
+require_relative 'menu'
+
+class Takeaway
+  attr_reader :order
+
+  @@menu = Menu.new
```
EdwardAndress, line 6: At the moment, the Menu is hard coded so all Takeaways will have the same Menu.  Can you think of how to make this more flexible such that each Takeaway you initialize can have a different Menu?

PR #1944, opened 2021-06-14
Irina's Takeaway Challenge

IrinaGC:


lib/list.rb
```diff
@@ -0,0 +1,31 @@
...
+      "Sandwich (Cold)" => 5.90,
+      "Sandwich (Hot)" => 8.90,
+      "Fresh Juices" => 4.70,
+      "Pastries" => 1.80
+    }
+  end
+
+end
+
+#  @menu = [
```
EdwardAndress, line 18: It's great that you tried two different data structures! :)

lib/order.rb
```diff
@@ -0,0 +1,46 @@
+require_relative "list"
+
+class Order
+
+  attr_reader :selection, :list, :total
+
+  def initialize(list = List.new)
```
EdwardAndress, line 7: Nice use of dependency injection!

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+    dishes.each do |key, value|
+      @selection << dish if key == dish
+    end
+    @selection if @selection.length > 0
+
+
+  end
+
+  def total
+
```
EdwardAndress, line 39: Did you know that `sum` can take a block? Can you think of a way to do that which would simplify things here?

lib/send_sms.rb
```diff
@@ -0,0 +1,14 @@
+require 'twilio-ruby'
```
EdwardAndress, line 1: It's good that you have this in a separate file. You could go further and wrap it all inside a new class.

spec/dishes_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "list"
+
+describe List do
+
+  it "diplayed with prices to be seen" do
+    expect( subject.menu ).to be_an_instance_of Hash
```
EdwardAndress, line 6: Here you could write a test that ensures the default menu is correct, though it would break when the menu changes.

spec/order_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+describe Order do
+  
+  order = Order.new
+
+  it "can see the list with dishes and prices" do
+    expect(order.see_menu).to be_an_instance_of Hash
+  end
+
+  it "can select dishes" do
+    menu = double('menu', :dishes => ["Soft Drink" ,
```
EdwardAndress, line 12: What happens if you delete this line?

EdwardAndress:
Nice work Irina.  There are a few places where the solution could be improved a little but I think you've done a good job of covering the first few user stories and have met the learning objectives. There are some comments in line against your code.  Let me know if you have any questions.

PR #1947, opened 2021-06-22
takeaway challenge

EdwardAndress:
opened by eddie

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,12 @@
+class Takeaway
+
+  attr_reader :finished_order
+
+  def initialize
+    @finished_order = {}
+  end
+
+  def add_to_order(dish, quantity)
+    @finished_order[dish] = quantity
+  end
+end
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,0 +1,49 @@
+require "menu"
+
+describe Menu do
+  subject(:menu){ Menu.new }
+
+  describe "initialize" do
+
+    it "creates an instance of a menu" do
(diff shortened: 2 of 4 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,12 @@
+class Takeaway
+
+  attr_reader :finished_order
+
+  def initialize
```
EdwardAndress, line 5: I like the use of a hash for this.

spec/menu_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(subject.dishes).to be_empty
+    end
+  end
+
+  describe "#add_to_menu" do
+
+    it { is_expected.to respond_to(:add_to_menu).with(2).arguments }
+
+    it "adds dishes using dish names and prices" do
+      expect { subject.add_to_menu("Burger", 4) }.to change { subject.dishes.length }.by(1)
```
EdwardAndress, line 26: This is a nice test!  It's quite advanced in terms of syntax.

EdwardAndress: Do you know _why_ it's a good approach?

EdwardAndress:
Hello! This is a good solution overall and there are a couple of really nice tests – I highlighted one of them with a comment.  The main thing to reflect on is how the responsibilities are divided up between classes.  At the moment, the Menu class has a method called `check_out` but should checking out really be something that the menu does?

Perhaps as a follow on exercise, without redoing the code, you could think about how to better divide the responsibilities.  Only if you have time and would find it useful though.

Please DM me on Slack if you have any questions.

PR #1949, opened 2021-06-22
added list of dishes with prices

EdwardAndress:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "takeaway"
+
+describe Takeaway do
+  subject(:takeaway) { described_class.new(menu: menu) }
+  
+  let(:menu) { double(:menu, print: print_menu) }
```
EdwardAndress, line 6: Nice use of a double

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "menu"
```
EdwardAndress, line 1: The two tests in here are nice!

EdwardAndress:
What you have done here is pretty solid so there's not much to say in terms of feedback.  Nice work so far!

PR #1953, opened 2021-07-11
Izzy Takeaway Challenge

isobelpepp:


JEC1100:
Well Done Isobel — just initial reaction from looking over files changes in your pull request — what beautiful code!

JEC1100:
* I particularly like the logic in your commits, rather than just satisfying broad user story steps, you have taken a simple approach to what you want each little piece of code to satisfy across a more broken-down version of the user journey.

* Again, this helps me understand what your program is doing but also how one might approach this challenge or one like it.

* I was also impressed from both your approach to this challenge and an aesthetic point of view, how you used simple integers (over floats) for menu pricing. I would always prefer to order a dish that was labelled 11 over one that was 10.95 but it has the built-in benefit here of avoiding unnecessary coding clutter and trivial challenges which might detract from the core.

* Your RSpec test “It” statements strike that great delicate balance between being both concise and informative: like a window into your though process for what it is you want the corresponding production code to do for you and the user.

JEC1100:
* The only minor suggestions I have, really: 

        * to consider injecting your menu content, instead of hardcoding it, but other than for future scalability of this program you are building, I’m not sure what other benefits doing so would afford. 

        * I think you should call your Takeaway “Eats Squisito!”

JEC1100:
One question: 

* are the comments left in your text_spec.rb file to help you, the programmer, follow the steps you were executing, or are they there to aid me, the reader, to follow what is happening? Either way, they achieved the former and that latter (so, thank you).

JEC1100:
This program really flows. It is super human-readable, informing the reader of it what the program does before they open a readme — and it helped me understand the challenge better too. I look forward to ordering spag bol & tiramisu from it soon!

Bravissima!

PR #1954, opened 2021-07-11
completed 3/4 user stories

senoodle:
missing the text confirmation part.


lib/menu.rb
```diff
@@ -0,0 +1,27 @@
...
+  attr_reader :menu
+
+  THE_MENU = {'Chicken Tacos' => 6,
+    'Burrito' => 8,
+    'Enchiladas' => 7,
+    'Quesadillas' => 8,
+    'Nachos' => 5,
+    'Sangria' => 3,
+    'Water' => 2
+  }
```
AureateGG, line 11: So much smarter to just create a hash than trying to actually build a menu containing dish objects :D

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def display_menu
+   @menu.print 
+  end
+
+  def add_dish(dish, quantity = 1)
+    raise 'Sorry, this item is not on our menu.' if on_menu?(dish)
+    quantity.times { @order << find_dish(dish) }
+  end
+
```
AureateGG, line 20: default quantity also smart haven't thought about it :)

AureateGG (commented):
so much cleaner than mine :D

PR #1958, opened 2021-07-12
Takeaway Challenge

BrettKarlson:
Still working on Twilio, and last user story

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ I would like to see a list of dishes with prices
 
 * Class Diagram:
  
- |     takeaway        |
+ |     Takeaway        |
  | :-------------------: |
  |   + menu: hash        |
  |   + display_menu()     |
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -14,34 +14,89 @@ Takeaway Challenge
...
 * Write a Takeaway program with the following user stories:
 
 ```
 As a customer
+```
+
+
+|   Objects |   Messages|
```
pav0107, line 29: I like how you've laid your thinking out with object, messages and diagrams

lib/takeaway.rb
```diff
@@ -0,0 +1,25 @@
+require_relative 'menu'
+require_relative 'basket'
+
+class Takeaway
+  attr_reader :dishes, :basket
+
```
pav0107, line 6: lovely clean code. Very easy to understand

PR #1960, opened 2021-07-12
James' Takeaway Challenge

JEC1100:
User story four and all related sms/text message functionality is a work in progress only (hence some commented out code).
See updated README.md for further clarifications.

### DIFF
--- a/lib/README(updated).me
+++ b/lib/README(updated).me
@@ -1,0 +1,111 @@
+additions to README.md file:
+
+Implemented first three user stories
+Attempted RSpec tests and produciton code for user story four
+Gem related practical steps to be completed in order to implement user story four functionality -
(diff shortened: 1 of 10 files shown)


isobelpepp:
Lovely code!! I particularly like how you have made a takeaway class that brings everything together, your use of doubles is very impressive and your code is clear and concise!

isobelpepp:
A couple of small suggestions:
- The indentation seems to be 4 spaces instead of 2, I found running rubocop in irb really useful for correcting my syntax
- Line 25 of takeaway.rb - attr_readers tend to go at the top of the class and from my understanding if they are made private they are essentially redundant because the point of them is to be able to access the instance variable from outside the code? (but correct me if I’m wrong!)

isobelpepp:
And one question:

- I can’t figure out if you use your Menu class elsewhere in your code? In your initialize method in takeaway you use (menu:) as an argument - does this syntax bring in the Menu class and allow you to use its methods?

PR #1962, opened 2021-07-12
pull_one

AureateGG:
-can create menu
-can create a dish, set name and price and add it to the menu
-can initialize cart
-can initialize order_item

stuck on the part where I want to check and copy a dish object into an order_item and push it into the cart because the @current_menu is either being returned as nil or empty :S


lib/dish.rb
```diff
@@ -0,0 +1,19 @@
+
+class Dish
```
senoodle, line 2: This is so different to what I've done! You've taken an interesting approach with the different classes :)

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
+
+
+class Menu
```
senoodle, line 3: I think you were supposed to add a hash of items from a menu and display them (user story #1: I would like to see a list of dishes with prices) so maybe if you added that, it could really help with your testing!

lib/cart.rb
```diff
@@ -0,0 +1,15 @@
+require_relative "menu.rb"
+
+class Cart
```
senoodle, line 3: The add_to_cart method in this class could be moved to the Order class - but this is a bit subjective, so I'm not sure if my advice is the best here!

lib/order_item.rb
```diff
@@ -0,0 +1,8 @@
+class Order_item
```
senoodle, line 1: Similar to my comment on the Cart class - I think you could merge a lot of these methods and group them up in a single class (Just from my notes from last week: Classes - describes what the object type will be [person/place/thing] and methods - do things [action/experience/condition]). I think for the Cart and Order item' classes , the methods belong together :)

spec/cart_spec.rb
```diff
@@ -0,0 +1,36 @@
+require "cart"
+
+
+
+describe "Order" do
```
senoodle, line 5: I really like the layout for your testing methods - they're super tidy and really easy to read!

spec/menu_spec.rb
```diff
@@ -0,0 +1,63 @@
+require "menu"
+require "dish"
+require "cart"
+
+
+describe "Menu" do
```
senoodle, line 6: Again - really well organised code!

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+    @menu_name = name
+    @current_menu = []
+    end
+
+    def add(dish)
+        fail "Dont add doubles to the Menu" if @current_menu.include?(dish) 
+        @current_menu.push(dish)
+    end
+
+    def remove(dish)
```
senoodle, line 17: This is pretty cool! I didn't think to add a delete function - this is smart :D

PR #1963, opened 2021-07-30
Takeaway Challenge

hjdalton:
Progress and notes in the README

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,30 @@
+# require_relative 'menu'
+
+class Takeaway
+  attr_reader :order, :total
+
+  def initialize
+    @order = []
+    @total = 0
+  end
+  
+  def show_menu
+    menu.display_menu
+  end
+
+  def select_meal(choice)
+    @order.push(choice)
+  end
+
+  def order_total
+    @total
+  end
+
+  # def calculate_fee
+  #   # @menu.each do |key, value| 
+  #   #   if key.include? @order
+  #   #   @total = value.sum
+  #   #   else 
+  #   #   end
+  #   end
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -3,14 +3,14 @@
 class Takeaway
   attr_reader :order, :total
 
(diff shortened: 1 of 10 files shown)


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,33 @@
+require 'takeaway'
+
+describe Takeaway do
+  # describe '#show_menu' do
+  #   # subject(:menu) { Menu.}
```
siellsiell, line 5: The reason using `subject` probably didn't work is that it's an RSpec that refers to the class mentioned on line 3 after the `describe` keyword (so that would be `Takeaway`) in this case. 

What you want here is to create a menu [double](https://github.com/makersacademy/course/blob/main/pills/doubles.md#doubles-mocks-stubs-spies-and-fakes) to represent an instance of the Menu class. The simplest way of creating a double is like this:
```
my_double = double()
```
You can also give your double a name to make it easier to identify in any test output

```
my_double = double(:menu_double)
```

You'll also have tell the double that it should respond to the `show_menu` method and return something. The examples on this page might help: https://relishapp.com/rspec/rspec-mocks/v/3-10/docs/basics/partial-test-doubles#only-the-specified-methods-are-redefined/

Then you'd have to pass your double into the `initialize` method of your Takeaway.

lib/takeaway.rb
```diff
@@ -0,0 +1,31 @@
+# require_relative 'menu'
+
+class Takeaway
+  attr_reader :order, :total
+
+  def initialize(meals:)
+    @order = []
+    @total = 0
+    @menu = meals
+  end
```
siellsiell, line 10: Nice use of dependency injection to make your Takeaway class be able to use your Menu class!

siellsiell (commented):
Hey Heather, thanks for having a go at this challenge and submitting your solution!

I can see that you thought about your object-oriented design and I like that you have separated the menu and ordering into its own class. You ran into some trouble with making them work together and test them. I left a comment with some tips/hints.

Let me know if anything is unclear! Happy to answer any questions.

PR #1964, opened 2021-07-30
Tyler Takeaway Challenge

tylercws:


lib/order.rb
```diff
@@ -0,0 +1,21 @@
...
+    @basket = []
+    @total = 0
+  end
+
+  def order(item, amount=1)
+    #read_menu.has_key?(item)
+    chosen_item = @menu.menu.select { |k,v| k == item }
+
+    amount.times{@basket.push(chosen_item)}
+    amount.times{@total += chosen_item.values[0]}
```
JessDevOps, line 16: great order method, really easy to read and make sense of

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+ end
+
+
+  it "has a list of items with prices" do
+   expect(subject.menu)
+  end
+
+  it "prints a list of items with prices" do
+   expect(subject.print)
+ end
```
siellsiell, line 27: These two tests don't make any assertions so they don't test what is listed in the description of the test.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let(:print_menu) {{"Vanilla Milkshake" => 3.99}}
+
+  ITEM = "Vanilla Milkshake"
+
+  it 'should have an empty basket' do
+    expect(takeaway.basket).to be_empty
+  end
+
+  it "shows the menu with the dishes and prices" do
+    expect(takeaway.print_menu)
```
siellsiell, line 15: same here. You need to follow up `expect` with `.to` and an Rspec matcher to make an assertion about the output.

lib/send_sms.rb
```diff
@@ -0,0 +1,20 @@
+require 'twilio-ruby'
+require 'time'
+class Messenger
+  def initialize()
+    @account_sid =
+    @auth_token = ''
+    @client = Twilio::REST::Client.new(@account_sid, @auth_token)
+    @from = '+447445034137' # Your Twilio number
```
siellsiell, line 8: It's best to avoid putting sensitive info like phone numbers in source code that you upload to the internet as it can easily get into the wrong hands! You can use something called an environment variable to avoid having to do that. https://www.rubyguides.com/2019/01/ruby-environment-variables/

lib/takeaway.rb
```diff
@@ -0,0 +1,29 @@
+require "./lib/send_sms"
+require "./lib/order"
+
+class TakeAway
+  def initialize(order=Order.new,menu=Menu.new)
+    @order = order
+    @menu = menu
```
siellsiell, line 7: Nice use of dependency injection!

siellsiell (commented):
Hey Tyler, thanks for submitting your solution! 

I'm really impressed with your object-oriented design. You separated responsibilities between your classes very cleanly. Your code is also easy to follow.

I saw that you didn't quite manage to test all of the functionality. If you ran into any blockers while test driving this that you still feel would be useful to get feedback/help on, let me know.

PR #1965, opened 2021-08-01
Hamid

p4r7hum4n:


siellsiell (commented):
Hey Hamid, thanks for submitting your solution! Nice work on getting all of the user stories implemented 💥 

We actually already discussed a lot of this on our call a few weeks ago so hopefully my comments on the rest of this PR will sound familiar! The main things you could do to improve your code would be:
- use SRP to split this code into smaller classes that are each responsible for a single job and make the code easier to test and stub
- Representing your menu data in a more convenient way so that you can make your filtering code less complicated.

You already had some great ideas on how to do that on our call and I left you some more detailed comments below. Let me know if you'd like some more guidance around any of these topics!

PR #1966, opened 2021-08-02
Ben Davis - Takeaway Challenge

ben-davis-devops:


lib/order.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  end
+
+  def view_basket
+    @basket
+  end
+
+  def add_dish(dish, number)
+    # fail "We're out of #{dish}" unless menu.dish?(dish)
+    @basket[dish] += number
```
NazirShaikh9093, line 16: A very clever method to add to the basket hash! I was confused at first because what if the dish doesn't exist yet how would it append the number to the value pair? But this is smart! It creates the new value pair 'dish => number' if the dish key doesn't exist yet and if it does and the user wants to add more of that dish to the basket it will append on the number of times they want it again!

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+      "squid" => 5,
+      "pork" => 8,
+      "ginger" => 11,
+      "king" => 12,
+      "chicken" => 10
+    }
+  end
+
+  def display
+    @menu_items.map { |dish, price| "#{dish}: £#{price}" }.join(", ")
```
NazirShaikh9093, line 15: A clever way of displaying the menu! Using the map method means that it will create a new array and not affect the original, while also not having the be stored anywhere because Ruby returns the last line of code in a method.

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
+class Menu 
+  attr_reader :menu_items
+  def initialize
+    @menu_items = {
```
Jestfer, line 4: This is fine and what was expected from the challenge solution. However, I'd like you to think about something: at the moment, the Menu is hardcoded so all Takeaways will have the same Menu. Can you think of how to make this more flexible such that each Takeaway you initialize can have a different Menu?

lib/order.rb
Jestfer: Interesting! This is a clever method that solves the typical scenario of a group of friends ordering food: "Let's order 1 🍕 , no... let's order 2, not..., let's make it 3!" The final value of items to be ordered will be updated based on the last request. This is all fine.

However, take a look at your menu (`@menu_items` under your Menu class). What would happen if I try to order a salad for example? e.g. `add_dish("salad", 1)`. What would happen after I run that method with these values on an instance of the Order Class?

lib/order.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def total_order
+    total = 0
+    @basket.each do |dish, number|
+    total += (number * @menu.menu_items[dish])
```
Jestfer, line 22: This is great! Just one thing: indenting the code inside blocks will make your life much easier moving forward in terms of readability.

Jestfer:
Nice work Ben. There are a few places where the solution could be improved a little but I think you've done a good job of covering the first few user stories and have met the learning objectives. There are some comments in line against your code for you to think about them.

PR #1967, opened 2021-08-02
takeaway challenge - danyaal 

ibnasif:


lib/menu.rb
```diff
@@ -0,0 +1,30 @@
+class Menu
+
+  attr_accessor :dishes
+
+  def initialize(dishes)
```
Jestfer, line 5: I love this approach. This way, you're not limited to always having the same menu in every instance of your Menu class, and you can instantiate it to whatever menu you want :)

lib/menu.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def print
+    @dishes.map do |dish, price|
+      "#{dish}: £#{price}"
+
+    end.join(", ")
+    
+  end
+
+  def has_dish?(dish)
```
Jestfer, line 18: Really good extracting this check to its own method and also in a Ruby conventional way.

lib/order.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def initialize(menu)
+    @menu = menu
+    @ordered_dishes = {}
+    @separate_totals = []
+  end
+
+    
+  def add(dish, quantity)
+    raise "#{dish} is not on the menu." unless menu.has_dish?(dish)
```
Jestfer, line 14: Nice use of a guard clause!

PR #1968, opened 2021-08-02
pull request JD

JessDevOps:


lib/menu.rb
```diff
@@ -0,0 +1,14 @@
+class Menu
+
+  attr_reader :food
+
+  def initialize(food)
+    @food = food
+  end
+
+  def list
+    food.map do |dish, price|
```
tylercws, line 10: I love the way you have done this, able to change a hash table to a nice formatted friendly list for the people who are running this to look at!

spec/menu_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'menu.rb'
+
+describe Menu do
+
+  subject(:menu) { described_class.new(food) }
+  let(:food) do
```
tylercws, line 6: Nice use of let method, to single out testing of the menu class so that it only relies on the test and non changing functions within the main program. Might be worth looking further into this with the use of a double, for example
`allow(food).to receive(:menu).and_return({ tofu_ramen: 3.99, veggie_curry: 2.99 })`

lib/menu.rb
```diff
@@ -0,0 +1,14 @@
+class Menu
+
+  attr_reader :food
+
+  def initialize(food)
+    @food = food
```
Jestfer, line 6: I love this approach. This way, you're not limited to always having the same menu in every instance of your Menu class, and you can instantiate it to whatever menu you want :)

Jestfer:
Hey Jessica, this is a good start and it seems you were about to start working on the second user story and on the right track for that. Please reach out to your coach for the week if you have any questions about end of module challenges.

Here's a video of a former colleague doing the same challenge that I hope is helpful: https://www.youtube.com/watch?v=mgiJKdH9x8c&ab_channel=SamMorgan

PR #1969, opened 2021-08-02
Individual challenge

ramaguire:


lib/takeaway.rb
```diff
@@ -0,0 +1,43 @@
...
+    @order = []
+    @menu = MENU
+    @order_total = 0
+  end
+
+  def read_menu
+    MENU
+  end
+
+  def select(pick)
```
EdwardAndress, line 22: I suspect that you're creating a nested array here because `select` returns an array?  Try using `find` instead :)

lib/takeaway.rb
```diff
@@ -0,0 +1,43 @@
...
+    }
+    p @order_total
+  end
+
+  def check_total
+    total
+    "Your order was #{@order}. Your total was #{@order_total}."
+  end
+
+  @client.messages.create(
```
EdwardAndress, line 38: Nice work with the Twilio gem! 👍

EdwardAndress: But I suspect that you would like to know how to stop the tests from sending you texts ;)

EdwardAndress:
This is a really good effort.  The next thing to do (if you want more OOP practice) would be to think about how you can break it up into multiple classes and then get them to work together, with test coverage.  It would feel contrived because there's perhaps not enough code to justify that but it would allow you to practice some more OOP stuff.  You could, for example, have an order class and a menu class.  What do you think?

If you have any questions, please drop me a a line on Slack, as I might not check back here again any time soon.

lib/takeaway.rb
EdwardAndress: To achieve this, you'd need to use dependency injection and then, instead of injecting the real client, in your test you can use a double.

EdwardAndress: The first part would look like this...

```ruby

class Takeaway

  attr_reader :MENU, :order
  MENU = { "brocolli" => 3, "risotto" => 10, "bread" => 5 }

  def initialize(twilio_client)
    @order = []
    @menu = MENU
    @order_total = 0
    @client = client
  end
```

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'takeaway'
+
+describe Takeaway do
```
EdwardAndress, line 3: Nice clear tests

PR #1970, opened 2021-08-02
Takeaway Challenge

manumoualek:


spec/menu_spec.rb
```diff
@@ -0,0 +1,23 @@
+require "takeaway"
+require "menu"
+require "order"
+
+describe Menu do
```
EdwardAndress, line 5: nice use of `subject`

EdwardAndress:
Hello!  You've made a good start here but it looks like you got a bit stuck when it came to placing orders.  Is that correct?  I'm going to leave some comments in line against the code.  Please DM me on Slack if you have any questions because I might not look back here anytime soon :)

lib/takeaway.rb
```diff
@@ -0,0 +1,22 @@
...
+  
+  def view_menu
+    list = "**"
+    @menu.dish_list.each do|dish|
+      list += " #{dish[:dish]} - £#{dish[:price]} **"
+    end
+    list
+  end
+  
+  def order(dish, number)
```
EdwardAndress, line 20: So, what would go here?  You want to add a dish and quantity to an oder, right? So you'll need an empty order to start adding things to.  I think the intention is that this `order` method is called multiple times to create the full order, so you'll need the initial empty order to be assigned outside the method (otherwise it will be emptied every time you call the `order` method).

You could create a new empty order in `initialize` and then add things to it in `order`.

The next thing to consider is what information does the order need to contain?  You'll need the name of the dish, the price and the quantity.  It'll be easier to calculate the total if, instead of having something like

3 x Margarita, 5

you have 

Margarita, 5
Margarita, 5
Margarita, 5

So let's aim to achieve that!

In which case, your `order` method will need to do something like this...

```ruby
number.times do
  @order << dish
end
```

That will add the dish `number` times.  But you won't have the price.  So you'd need to use the `dish` to either retrieve the price from the menu or just get the whole item from the menu and add that.

```ruby
def order(dish, number)
  number.times do
    @order << menu.dish_list.find {| menu_dish | menu_dish[:dish] == dish }
  end
end
```

EdwardAndress: At this point, you might start to notice that we're using the term `dish` a lot and it gets confusing.  So one thing to watch out for here is your choice of names.  Using `name` instead of `dish` could help a little..

```ruby
{name: "Margarita", price: 5}
```

PR #1971, opened 2021-08-02
Michael Poole

0308199710520:


lib/basket.rb
```diff
@@ -0,0 +1,31 @@
+class Basket
```
EdwardAndress, line 1: Very nice!

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+    "Kebab" => { name: "Kebab", price: 5 } }
+
+  def initialize(menu: DEFAULT_MENU)
+    @menu = menu
+    @basket = Basket.new
+  end
+  
+  def menu
+    string = ""
+    @menu.keys.each do |item|
```
EdwardAndress, line 19: I think this would also work...

```ruby
@menu.each do |item|
  string += "#{item[:name]} : £#{item[:price]}\n"
 end
```
Which version do you prefer, and why?

EdwardAndress: Ah, no, it that would require your menu to be an array of hashes, rather than a hash of hashes!

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+      string += "#{@menu[item][:name]} : £#{@menu[item][:price]}\n"
+    end
+    return string
+  end
+
+  def addBasket(food)
```
EdwardAndress, line 25: OK, and your hash of hashes allows you to select the correct menu item in this nice way

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+    food = @menu[food]
+    @basket.add(food)
+  end
+
+  def viewBasket
```
EdwardAndress, line 30: nice use of delegation

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,24 @@
+require_relative "../lib/takeaway.rb"
+
+describe Takeaway do
+  describe "#printer" do
```
EdwardAndress, line 4: did you change the method name from `printer` to `menu` at some point?

EdwardAndress:
Hey Michael, this is a really nice solution!  I've left some comments in line against the code and will add a few more now.  If you have questions for me, please send them by DM on Slack rather than asking them here as I might not look back here soon!

One minor thing which is general, rather than about any specific line, is that you're writing Ruby a bit like Javascript :) ... nothing to be concerned about but it's worth being aware that different idioms exist.  So, for example, in Ruby we tend not to use `return` unless it's really needed.  We also use snake_case for methods and variables rather than camelCase.

These small things don't matter so much right now but when you join a team on placement you'll need to pick up the style of the team and use whatever idioms exist in their language of choice.

spec/basket_spec.rb
```diff
@@ -0,0 +1,28 @@
+require_relative "../lib/basket.rb"
+
+describe Basket do
+  describe "#add" do
+    it "should add an item from the menu to the basket" do
+      basket = Basket.new
```
EdwardAndress, line 6: Do you know how to replace `Basket.new` with `subject`?

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :menu
+
+  DEFAULT_MENU = {
+    "Burger" => { name: "Burger", price: 3 },       
+    "Chips" => { name: "Chips", price: 1 },
+    "Kebab" => { name: "Kebab", price: 5 } }
+
+  def initialize(menu: DEFAULT_MENU)
+    @menu = menu
+    @basket = Basket.new
```
EdwardAndress, line 14: Do you know how to avoid hard coding your Basket class here?  If you can achieve that, you could then use a double for the instance of Basket in your tests.

lib/takeaway.rb
```diff
@@ -0,0 +1,34 @@
+require_relative "../lib/basket.rb"
+
+class Takeaway
+
+  attr_reader :menu
```
EdwardAndress, line 5: This `attr_reader` will create a method called `menu` which, I think, is then overwritten by the `menu` method defined below.

lib/basket.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def viewBasket
+    return menu_string_constructor(@basket)
+  end
+
+  def basket
+    return @basket
+  end
+
+  def reciept
```
EdwardAndress, line 18: Nice use of `inject`.  It's not the most intuitive method to use.  I think you could also use `sum`, which can take a block and will sum together all the return values of the block (without the need for an accumulator).

EdwardAndress:
OK so, I think the two main areas for extension would be

* Using doubles to isolate your tests
* Really pushing SRP (single responsibility principle) and creating more classes.  For example, you could create Printer class that is used to print the menu and anything else that is displayed to the customer.  In an exercise like this, it gets quite contrived very quickly but it's all about OOD and putting into practice those principles (rather than building something in a pragmatic way).

Really nice work though!!

PR #1972, opened 2021-08-08
My attempt at the Takeaway weekend challenge

ashtroboy79:
I believe this fulfils the user stories to the best of my ability. I could do with some guidance on how to deal with mocking and stubbing, and about annotated token (like %<foo>s) over unannotated tokens (like %s) usage.


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  it "has items in the menu" do
+    expect(subject.menu).not_to be_empty
+  end
+
+  it "displays the menu" do
+    expect(subject.show_menu).to eq subject.menu
+  end
+
+  it "allows items to be ordered" do
+    subject.add_item("Samosa")
```
zb-coder, line 13: I like how easy your tests are to read and understand.

lib/dishes.rb
```diff
@@ -0,0 +1,8 @@
+class Dishes
```
zb-coder, line 1: Why did you make a separate dishes class?

lib/takeaway.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def total
+    puts "£#{'%.2f' % @order.total}"
+  end
+
+  def summary
+    puts @order.summary
+  end
+
```
zb-coder, line 32: can you talk me through this part?

leoht (commented):
Great job! I left a few comments here and there about minor things to improve the code, and on the design of how classes interact together :)

PR #1973, opened 2021-08-08
Week 2 weekend challenge - NC

Nicola-Carroll:
**I didn't complete this challenge.**

Things to complete/improve:

1. Use stubs/mocks for the Twilio API so that I can finish writing tests for `order.rb` (I didn't use TDD here and my test coverage is low because I was blocked by this)
2. The `Order` class is dependent on the `menu` argument passed to be formatted as an array of hashes in the format `{dish_name => price}` which is too restrictive
3. The `print_menu` method in the `Menu` just prints a hard coded list of dishes and doesn't depend on the items passed initialised in the menu
4. `Order` class is dependent on `Menu` and `ConfirmationText` class (didn't use dependency injection because I'm confused about how to write tests with this)
5. `Order` class not very cohesive, intended on splitting it down into more classes, for e.g. one to deal with customer comms/interface stuff
6. `menu_spec` test started failing because it was printing £ as some unicode character thing 🤷‍♀️


lib/confirmation_text.rb
```diff
@@ -0,0 +1,33 @@
+require 'twilio-ruby'
+require 'dotenv'
+Dotenv.load
+
+class ConfirmationText
+
+  current_time = Time.now
+  delivery_time = (current_time + 3600).strftime("%H:%M")
+
+  DEFAULT_FROM = ENV['FROM']
```
CorinneBosch, line 10: I really like the use of constants here!

lib/menu.rb
```diff
@@ -0,0 +1,44 @@
...
+    {"Salad" => 3},
+    {"Tiramisu" => 6},
+    {"Chocolate fudge cake" => 7}
+  ]
+
+  def initialize(items=DEFAULT_ITEMS)
+    @items = items
+  end
+ 
+  def print_menu
```
CorinneBosch, line 21: Working with an additional key value pair might have helped you with this!

lib/order.rb
```diff
@@ -0,0 +1,60 @@
...
+    @menu.print_menu
+    order_instructions
+    select_order
+    cost
+  end
+
+
+  private
+
+  def order_instructions
```
CorinneBosch, line 32: maybe break it down into 2 strings as the line exceeds the 80 chars limit

lib/menu.rb
CorinneBosch: you could have saved it as constant too, to have it available in your spec test

PR #1981, opened 2021-08-09
Takeaway Challenge

PaulRainsford:


spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  subject(:menu) { described_class.new }
+  let(:dishes) { double :dishes }
+  let(:dishes) do {
+    "tacos" => 5,
+    "burrito" => 5,
+    "quesadilla" => 7,
+    "enchiladas" => 7
+  }
+  end
+
```
yousufmohamed92, line 13: You've added lets for dishes twice.

lib/order.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def initialize
+    @selection = Hash.new(0)
+    @menu = Menu.new
+  end
+
+  def add_order(dish, quantity)
+    @selection[dish] += quantity
+  end
+
```
yousufmohamed92, line 15: I'm not sure this will add to the hash like you want.

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+      }
+  end
+
+  def print_menu
+    @dishes.map { |dish, price|
+      "#{dish.capitalize}: £#{price}"
+    }.join(', ')
+  end
+
+end
```
yousufmohamed92, line 20: Nice, neat class.

lib/order.rb
```diff
@@ -0,0 +1,23 @@
...
+  def total
+    order_total = 0
+    @selection.each do |dish, quantity|
```
yousufmohamed92, line 18: Maybe set order_total as attribute?

lib/order.rb
```diff
@@ -0,0 +1,23 @@
...
+    @selection = Hash.new(0)
+    @menu = Menu.new
+  end
+
+  def add_order(dish, quantity)
+    @selection[dish] += quantity
+  end
+
+  def total
+    order_total = 0
```
yousufmohamed92, line 17: Maybe set order_total as an attribute?

PR #1982, opened 2021-08-09
SandCream Delivery Service

CorinneBosch:


lib/happiness_delivered.rb
```diff
@@ -0,0 +1,111 @@
...
+    lines
+    puts 'Welcome to Happiness Delivered! Please select from' 
+    puts 'the following menu options to place your order.'
+    lines
+  end
+
+  def order
+    puts 'Please enter the menu item number you wish'
+    puts 'to add to your basket or enter 999 to exit.'
+    x = $stdin.gets.chomp.to_i
```
Nicola-Carroll, line 92: Might be better to use a more descriptive variable name here, like `user_input` or something, as opposed to `x`

lib/happiness_delivered.rb
```diff
@@ -0,0 +1,111 @@
...
+  end
+
+  def welcome_note
+    lines
+    puts 'Welcome to Happiness Delivered! Please select from' 
+    puts 'the following menu options to place your order.'
+    lines
+  end
+
+  def order
```
Nicola-Carroll, line 89: Potentially, you could have a separate method for laying out the interface instructions for the user and call it here, so that in time you could refactor the instructions or reuse them elsewhere if other features were added

lib/sandcream.rb
```diff
@@ -0,0 +1,50 @@
...
+  attr_reader :sando_menu, :icecream_menu
+
+  SANDO = [
+    { sando: '1. Submarine Club - Tuna Salad', price: 9.50 },
+    { sando: '2. Clubster Cluck - Grilled Chicken', price: 9.50 },
+    { sando: '3. Shrimpy Sea Chanty - Prawn Cocktail', price: 10.50 },
+    { sando: '4. Sweet Home Avocado - All Veggie', price: 9.00 },
+    { sando: '5. Big City Life - Smoked Salmon & Cream Cheese', price: 10.50 },
+    { sando: '6. Bologna Baloney- Pistachio Mortadella', price: 11.00 },
+    { sando: '7. Egg Slut Sando - Egg Salad', price: 8.50 }
```
Nicola-Carroll, line 11: 😂 love number 7

spec/happiness_delivered_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    it 'basket is empty' do
+      expect(subject.basket).to be_empty
+    end
+
+    it 'order_total is empty' do
+      expect(subject.order_total).to be_empty
+    end
+  end
+
```
Nicola-Carroll, line 30: I'm not sure if all the above tests are really necessary at this point, and that they might be too granular. If you test the behaviour of your methods properly then in theory those tests should also test everything above as part of them. Although this is something I need to work on myself so who knows 👀

Nicola-Carroll (commented):
Love the menu and how much detail you've put into the user experience!

lib/happiness_delivered.rb
```diff
@@ -0,0 +1,111 @@
+# require 'customer_sms'
+require_relative 'sandcream'
+
+class HappinessDelivered
```
Nicola-Carroll, line 4: I think Class names should be slightly more descriptive maybe, I'm not fully sure what to expect of this class (what it does/how to use it) based on the name. This is just your interactive menu class right>

PR #1983, opened 2021-08-09
takeaway-challenge

eballantine:


lib/order.rb
```diff
@@ -0,0 +1,25 @@
+# Responsible for collating the order and cost
+
```
gjstirling, line 2: Clear responsibilities, nice !

lib/restaurant.rb
```diff
@@ -0,0 +1,71 @@
...
+      ["Sushi", 7],
+      ["Burger", 6],
+      ["Fries", 3],
+      ["Noodle Soup", 10],
+      ["Bahn Mi", 9.25]
+    ]
+    @orders_history = []
+    @current_order = Order.new
+  end
+
```
gjstirling, line 19: Would a hash have been easier to use? Key/value

PR #1984, opened 2021-08-09
PR for review

Battery0:
PR for review

### DIFF
--- a/lib/food_order.rb
+++ b/lib/food_order.rb
@@ -12,12 +12,9 @@ class FoodOrder
   def add_item(food, number)
     raise "This menu item does not exist" if @menu.menu[food].nil?
 
-    menu_item = @menu.menu.select do |key, _value|
-      key == food
-    end
+    menu_item = @menu.menu.select { |key, _value| key == food }
     number.times { @food_order << menu_item }
+    @food_order
   end
-
-  # deny access to their total amount 
 
 end
--- a/lib/food_order.rb
+++ b/lib/food_order.rb
@@ -1,0 +1,23 @@
+require_relative './takeaway_menu'
+
+class FoodOrder
+
+  attr_reader :food_order
+
+  def initialize
+    @food_order = []
+    @menu = Menu.new
+  end
+
+  def add_item(food, number)
+    raise "This menu item does not exist" if @menu.menu[food].nil?
+
+    menu_item = @menu.menu.select do |key, _value|
+      key == food
+    end
+    number.times { @food_order << menu_item }
+  end
+
+  # deny access to their total amount 
+
+end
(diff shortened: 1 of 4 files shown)


lib/food_order.rb
```diff
@@ -0,0 +1,30 @@
+require_relative './takeaway_menu'
+
+class FoodOrder
+
+  attr_reader :food_order
```
leoht, line 5: It might be good to rename the attribute to something else than `food_order`, to avoid mixing up with the class name

lib/food_order.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def order_sum
+    order_total = 0
+    @food_order.each do |order_item|
+      order_item.each do |_key, value|
+        order_total += value
+      end
+    end
+    puts "The cost of your order is £%.2f" % order_total
```
leoht, line 27: It'd be better to return the value, rather than printing a message including it — this way the method `order_sum` only deals with data, and only returns a useful result (the total sum), and printing it could be done in some other place in the code — remember separation of concerns also mean that one class dealing with data such as `FoodOrder` shouldn't be also responsible for printing/displaying something

lib/takeaway_menu.rb
```diff
@@ -0,0 +1,22 @@
+class Menu
+
+  def menu
```
leoht, line 3: You could simply have one public method `menu` returning the hash directly, delegating to a private method is not necessary

spec/food_order_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'food_order'
+
+describe FoodOrder do
+
```
leoht, line 4: Because the class `FoodOrder` is depending on `Menu`, it should be isolated from the Menu class in those unit tests (using a double, for example). To do this you'll need to:
 * inject the `@menu` attribute in the initialize method, rather than instantiating directly the Menu class in it: `@menu = Menu.new`
 * explicitly define the RSpec's subject so you can create the instance of `FoodOrder` and pass the Menu double in it
 * you'll know if your class is isolated correctly if you can "break" the `Menu` class (maybe by commenting out its method), but the tests for `FoodOrder` still pass

leoht (commented):
That's a good start! I left a few comments around code design and isolating classes in your tests — some user stories are missing, but you can work on those later if you want :)

lib/food_order.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  attr_reader :food_order
+
+  def initialize
+    @food_order = []
+    @menu = Menu.new
+  end
+
+  def add_item(food, number)
+    raise "This menu item does not exist" if @menu.menu[food].nil?
```
leoht, line 13: That's a good use of defensive programming to ensure we do correct things :)

PR #1987, opened 2021-08-09
attempt at takeaway challenge

suhani-zaman:


lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+        end
+      end
+    end
+    def total_price
+        total = 0
+        @items.each do |item, qty|
+            price = @price[item]
+            qty_price = price*qty
+            total += qty_price
+        end
```
edpackard, line 35: I might be reading this wrong, but don't lines 32-34 here basically do this:`total += @price[item]*qty`

edpackard (commented):
This is a logical start to the project, and there is some good practice here, including the use of dependency injection, and clear method/variable labels. Your code is neat and easy to follow.

I've offered some thoughts on the code itself, but I would also say that you seem to have written some methods without writing tests for them first? I know how tempting it is to dive in and get coding, but you'll improve your rspec (and tdd) skills if you write a clear test before coding each method. Let me know if you pick up this project again, as I'd be interested to see what direction you take it - and good luck implementing the Twilio component, it's really good fun!

PR #1992, opened 2021-08-09
Beckie Takeaway Challenge

beckiedoyle:


lib/order.rb
```diff
@@ -0,0 +1,31 @@
...
+    items.each do |item|
+      if @menu.has_key?(item)
+        @my_order.push(item)
+      else
+        raise "that's not on the menu!"
+      end
+    end
+  end
+
+  def calculate_total
```
EdwardAndress, line 24: Nice work!

You might be interested to know that `sum` can take a block so don't need to use `each`.

```ruby
@my_order.sum {| item| menu[item] }
```

spec/menu_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'menu.rb'
+
+describe Menu do
+
+  describe "#see_menu" do
+    it "shows me the menu options and their prices" do
+      menu = Menu.new
+      expect(menu.see_menu).to eq(@menu)
```
EdwardAndress, line 8: Does this test pass? 🤔

EdwardAndress: I'd expect `@menu` to be `nil` in this scope.

spec/order_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'order'
+
+describe Order do
+
+  describe "#order_meal" do
+    it "lets me select multiple dishes" do
+      expect(subject.order_meal("big mac", "fries")).to eq([ "big mac", "fries" ])
+    end
+    it "gives an error if the item isn't on the menu" do
+      expect{subject.order_meal("apple")}.to raise_error "that's not on the menu!"
```
EdwardAndress, line 10: Nice test for error raising.  Do you know why the syntax has to be different here? I.e. Why the curly braces are needed?  Send over your answer on Slack, if you have one or would like to know the answer from me :)

spec/order_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'order'
+
+describe Order do
+
+  describe "#order_meal" do
+    it "lets me select multiple dishes" do
+      expect(subject.order_meal("big mac", "fries")).to eq([ "big mac", "fries" ])
```
EdwardAndress, line 7: The thing to watch out for here is that you're testing the right / important things.  For methods like `order_meal` we probably care less about their return value than we do about what they accomplish behind the scenes.  So a better test would make sure that `order_meal` correctly updates the order, perhaps.

EdwardAndress:
Good effort Beckie!  I can see that you're picking things up well.  There are some comments left in line against your solution.  Please DM me on Slack if you have any questions as I might not check back here.

PR #1994, opened 2021-09-05
takeaway weekend challenge

dhopz:


lib/order.rb
```diff
@@ -0,0 +1,62 @@
...
+    
+    def order_confirmation
+        confirmation_list = []
+        item_sub_total = @order_list.group_by(&:itself).map{|k, v| k.merge(count: v.length)} 
+        item_sub_total.each do |x|
+            sub_total = x[:price].to_i * x[:count].to_i
+            confirmation_list << "#{x[:count]} x #{x[:name]} £#{sub_total}"
+        end
+        confirmation_list
+    end
```
dhopz, line 60: I didn't need to do this, thought it would be interesting to try and add

PR #1995, opened 2021-09-05
Takeaway challenge, Week 2

lildann:
Summary
-----

**For the Restaurant**
* Dish
  * A dish is created with a name and a price. 
  * Prices are input as a float with two decimal points.
  * Dishes can be itemised as a hash (itemise method).

* Menu
  * Dishes (itemised as a hash) can be added to the menu (add_to_menu(dish) method)
  * Menu can be viewed (view_menu method)
  * DEFAULT_MENU constant supplied, access with access_menu method
  * Added a banner for the takeaway cafe: "*** TDD Greasy Spoon ***"

**For the Customer**
* Order
  * An order is initialized with an empty my_order array 
  * Dishes can be added one by one to the order, receives name and quantity as arguments (add_dish_to_order(name, quantity) method)
  * Cost of order calculated (order_total method)
  * Added view_order method for customer which includes cost of item 
  * Added how_much method to output total order cost

Notes
-----

 * Disappointed I couldn't figure out Twilio. I added too many extras and ran out of time (- should have kept to the brief!)
 * Would have liked to load an external CSV file for the menu, and output that in the terminal.
 * Now realise my rspec tests all evaluate state not behaviour! 
 * Noticed that `puts` works better in irb for view_order and view_menu but breaks my tests as nil becomes output (now learned I could have used: change .to output.to_stdout)
 * Also encountered problem with decimal points - should be limited 2dp throughout - any suggestions for overcoming this when using numbers as currency?


README_LD.md
```diff
@@ -0,0 +1,74 @@
...
+           //   | ===  |   /        \
+       .:'//.   \ \=|   \ /  .:'':./    )))))
+      :' // ':   \ \ ''..'--:'-.. ':
+      '. '' .'    \:.....:--'.-'' .'
+       ':..:'                ':..:'
+
+ ```
+
+Summary
+-----
```
Jessocxz98, line 18: Your read me is extremely detailed and clear, which helps a lot with understanding what you have done and why. I like that you have summarised what each class does and the methods, variables etc it contains.

lib/dish.rb
```diff
@@ -0,0 +1,14 @@
...
+  attr_reader :name, :price, :hash
+
+  def initialize(name = "Hash brown", price = 0.99)
+    @name = name
+    @price = price
+    @hash = {}
+  end
+
+  def itemise
+    @hash[@name] = @price.round(2)
```
Jessocxz98, line 11: This .round method is clever, I ended up using integers as I found working with floats difficult.

README_LD.md
```diff
@@ -0,0 +1,74 @@
...
+  * Dishes can be added one by one to the order, receives name and quantity as arguments (add_dish_to_order(name, quantity) method)
+  * Cost of order calculated (order_total method)
+  * Added view_order method for customer which includes cost of item 
+  * Added how_much method to output total order cost
+
+
+Notes
+-----
+
+ * Disappointed I couldn't figure out Twilio. It was asking me to pay for the number?
```
Jessocxz98, line 44: When you signed up to the free trial it gives you credit of around £11, you can use this to 'buy' a twilio number for about 80p a month - so doesn't actually charge you! You then use this as the number you send the sms message from.

lib/dish.rb
```diff
@@ -0,0 +1,14 @@
+class Dish
+  attr_reader :name, :price, :hash
+
+  def initialize(name = "Hash brown", price = 0.99)
+    @name = name
+    @price = price
+    @hash = {}
```
leoht, line 7: I'm not sure I understand why `@hash` is there — it you need to transform the object into a key/value pair (what `itemise` is doing), you could maybe do it in the class that will call `Dish` and transform the list of dishes into a hash. But it somehow feels weird to have this hash into the Dish class, since a single Dish instance will translate only into one key/value pair (let me know if that's not clear!)

leoht: Or if you want to keep the same behaviour, you could simply by rewriting the method `itemise` like this:
```ruby
def itemise
   { @name => @price.round(2) }
end
```

so you don't need to `@hash` attribute at all

lib/menu.rb
```diff
@@ -0,0 +1,44 @@
...
+  def initialize
+    @list_of_dishes = []
+  end
+
+  def add_to_menu(dish)
+    @list_of_dishes.push(dish.itemise)
+  end
+
+  def access_menu
+    @list_of_dishes = DEFAULT_MENU
```
leoht, line 13: When I'm reading this method, it feels like it's "resetting" the list of dishes to a default one — so maybe a better name for it could be "reset_menu" or "initialize_default_menu"?

lib/menu.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def view_menu
+    access_menu if @list_of_dishes.empty?
+    banner_display
```
leoht, line 18: If we want to strictly respect the SRP, we ideally don't want to have things that deal with "presentation" (printing/display, etc) in the same class or component that deals with data. So maybe it'd be good to move the methods `menu_display` and `banner_display` into their own class (e.g `MenuDisplay`), and this class instance could be dependency-injected into `Menu` (then we could pass the list of dishes in arguments to those methods, if needed).

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+  attr_reader :my_order 
+
+  def initialize
+    @my_order = []
+  end
+
+  def add_dish_to_order(name, quantity)
+    @name, @quantity = name, quantity
+    @my_order << { @name => @quantity }
+    return "#{quantity} x #{name} added to your order"
```
leoht, line 11: It's usually better to not return "user-friendly" message strings as return values of a method that operates on data (again, to respect the SRP, but it's also easier to make mistakes if we'd need to verify/test this string message later). What you could do is return the dish that was added (or perhaps just `true` or `false` for a result, etc)

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  def view_order
+    order_display
+  end
+
+  def order_total
+    sum = 0
+    menu = Menu.new.access_menu
```
leoht, line 20: We have a dependency on `Menu` in this class, so when testing, we'd ideally want to isolate our `Order` class from it. This can be done if we dependency-inject the `Menu` instance into Order, rather than creating it here - it would look like this:
```
class Order
     @menu = menu
    # ...
```

Then you could use a double in your tests to pass it instead of the "real" Menu instance, in order to mock its behaviour

lib/order.rb
```diff
@@ -0,0 +1,55 @@
...
+        menu.each do |dish| 
+          dish.each { |item, price| sum += (price * quantity) if name == item } 
+        end
+      end
+    end
+    sum.round(2)
+  end
+
+  def how_much
+    return "Customer to pay £#{order_total}"
```
leoht, line 32: Same remark as above: ideally this should be in a class dedicated to "display" or print messages to the user in the terminal

spec/dish_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    end
+
+    it 'has a #price' do 
+      expect(subject.price).to eq(0.99)
+    end
+  end
+
+  context 'menu format' do
+    it 'combines name and price in a hash' do 
+      expect(subject.itemise).to eq({ "Hash brown" => 0.99 })
```
leoht, line 17: I'd also try a similar test but using custom values from the name and price, rather than just default ones (to catch potential bugs where default values are not overwritten with given ones - that's neatpicking though!)

leoht (commented):
Great job! The code is really clear to read :) I mainly left comments about a few improvements here and there. There's still a bit more to do to fully respect proper separation of concerns — notably between methods dealing with data and other methods taking care of printing and displaying messages on the terminal. Feel free to improve on it later when you'll have time and space for it, and let me know if you have any question on the feedback, or if something isn't clear!

PR #1998, opened 2021-09-05
Week 2: Takeaway Challenge

leahcolleen:


jordanveness (commented):
refactored shows clear variable names provides better insight into what the code is doing

jordanveness (commented):
easy to follow, is there a way to not have the spec file as lengthy?

PR #2002, opened 2021-09-06
takeaway challenge

florence-bain:


lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
...
+  end 
+
+  def menu
+  index = 1
+    @pizzas.each do |pizza|
+      puts "#{index}: #{pizza[:name]} (#{pizza[:price]})"
+    end 
+  end 
+
+  def create_order(order = Order.new)
```
mandyvuong, line 25: Good use of separating classes and naming methods/variables are clear

lib/takeaway.rb
```diff
@@ -0,0 +1,48 @@
...
+    @order = order
+  end 
+
+  def add_order(user_order)
```
mandyvuong, line 29: There are some Order methods in the Takeaway class - you can try class extraction to move the Order methods into the Class methods or rename the Order methods to avoid more than one description in a class

PR #2004, opened 2021-09-06
Weekend Challenge Week 2

Maldorana:


spec/send_text_spec.rb
```diff
@@ -0,0 +1,5 @@
+describe SendText do
+  it { is_expected.to respond_to(:go_send) }
```
eoinmakers, line 2: For testing the third part API, there's some good reading in the takeaway challenge review doc that you can see here: https://github.com/Maldorana/takeaway-challenge/blob/main/docs/review.md#step-2-tests-and-_specrb-files

You only need to go into as much detail as is in the review doc for this - so testing that the correct methods are called, but not actually calling the API each time the test is run.

eoinmakers (commented):
Well done on your implementation of the functionality of each user story. I think your domain modelling is well done, and you made good choices about class responsibilities. 

There are maybe opportunities to take more advantage of the idea of 'Dish' being an object represented in your program, in order to simplify further certain methods that interact with the dishes later on.

PR #2009, opened 2021-09-06
Takeaway challenge 

InfobyAdrienne:
User Story 1 complete

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,13 @@
+class Menu
+  attr_reader :dishes
+
+  def initialize(dishes)
+    @dishes = dishes
+  end
+
+  def print
+    dishes.map do |title, price|
+      "%s £%.2f" % [title.to_s.capitalize, price]
+    end.join(", ")
+  end
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,13 @@
+class Takeaway
+  def initialize(menu:)
+    @menu = menu
+  end
+
+  def print_menu
+    menu.print
+  end
+
+  private
+
+  attr_reader :menu
+end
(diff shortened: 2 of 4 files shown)


InfobyAdrienne:
Completed User Story 1 only. Please work from  'User Story 1: Further edits'. 

When running the test in rspec there are 0 failures, however when running the irb test and calling 'menu.printed_menu' it prints the menu out in the format I want it to first and then also prints out the whole hash again. I'm not sure if this duplication is correct.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
...
+      "Fish dog" => 11,
+      "Halloumi fritter burger" => 8,
+      "Sloppy Joe dog" => 9,
+      "Truffle fries" => 4,
+      "Mac & Cheese" => 5,
+      "Corn ribs" => 4,
+    }
+  end
+
+  def printed_menu
```
leoht, line 17: For your question about why this method prints "twice" when run in IRB, that's because IRB will print out the returned value of the method you're calling. So two things end being printed:
 * what is printed by `puts`
 * the return value of `printed_menu` - since calling `.each` will return the value of the array it's called on (after executing the block for each item), you end up having the `menu` array returned by the whole method

leoht: Maybe a way to make this "cleaner" could be to build a return string to return, rather than directly printing using `puts`:
```ruby
menu.each do |dish, price|
      output += "#{dish}: £#{price}"
 end
```

You can then simply decide to "print" (puts) this value or do something else with it, but that will be the job of the caller, rather than of this method

leoht (commented):
That's a good start! You can come back to it later if you want to make further progress on it if you'd like, to implement the `Takeaway` class, etc :) I've left a comment about the `printed_menu` method behaviour

PR #2011, opened 2021-09-06
Jakes takeaway challange

Jjake540:


lib/restaurant.rb
```diff
@@ -0,0 +1,17 @@
...
+class Restaurant
+
+  attr_reader :menu
+
+  def initialize(menu = nil)
+    @menu = Menu.new(menu)
+  end
+
+  def view_menu
+    @menu.display_menu
```
Legal-Dan, line 13: Is display_menu defined anywhere? I can't see it. If you just want to print out the menu contents, as it is in the attribute reader I believe you could just do yourrestaurantinstance.menu

spec/menu_spec.rb
```diff
@@ -0,0 +1,27 @@
+require './lib/menu'
+
+describe Menu do
+
+  let(:subject) { Menu.new(nil)}
+
+  describe '#initialize' do
+    it 'initializes with the menu' do
+      expect(subject.menu).to include('Hallumi fritter burger' => 8)
```
Legal-Dan, line 9: I like the use of to include. That's better than how I would have thought to do it.

lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+  
+  attr_reader :menu
+
+  def initialize(menu)
+    @menu = menu
+    list_dishes
+  end
+
+  def list_dishes
+    @menu = MENU if @menu == nil
```
Legal-Dan, line 23: I don't really understand why we are requiring a menu parameter with a default of NIL as well as having a constant menu that overwrites this. Was this one of the requirements that you could put in a different menu on initialising? If so it might be worth seeing if there is a way to simplify this.

PR #2012, opened 2021-09-06
For code review

rhianeKobar:
```
As a customer

As a customer

As a customer
```

The above user stories have been covered, I'm not confident on how thoroughly they have been implemented. I'd especially like some comments on my tests because I'm not sure I'm covering all that I should in them.


spec/order_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+	it 'should have an item in the order hash' do
+		order = Order.new
+		pizza = Restaurant.new
+		pizza.add("margarita", 10)
+		pizza.add("pepperoni", 12)
+		pizza.add("hawaiian", 12)
+		order.select(pizza,"margarita")
+		expect(order.my_order).not_to be_empty
+	end
```
dhopz, line 19: I thought a 'double' may be a good option here because you would be calling from another class the "Dish", and then possibly not have to repeat.

lib/order.rb
```diff
@@ -0,0 +1,33 @@
...
+			if key == dish
+				if @my_order.has_key? key
+					nuVal = (@my_order.fetch(key)) + value
+					@my_order.store(key,nuVal)
+				else
+					@my_order.store(key,value)
+				end
+			end
+		}
+	end
```
dhopz, line 21: I'm not sure, my though process would be select an item from the menu, then shovel << to the my_order hash. Maybe then you wouldn't have to use .each?

dhopz (commented):
My first code review, please don't read too much into it. I like your approach!

spec/order_spec.rb
rhianeKobar: I've been having trouble understanding how to use doubles tbh. But after viewing your code, I think I understand them bit better.

spec/order_spec.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/order'
+require './lib/restaurant'
+
+describe "Order" do
+
```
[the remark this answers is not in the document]
rhianeKobar: ahhhh, I kept seeing that being implemented in example but I kept going around it instead of using it.

lib/order.rb
rhianeKobar: I don't think you can use << on a hash, I'm pretty sure that's just for arrays. Why would you advise against using .each?

PR #2013, opened 2021-09-06
Takeaway challenge

borisl16:


lib/order.rb
```diff
@@ -0,0 +1,24 @@
+require 'order'
+require 'menu'
+
+class Order
+
+    attr_reader :dishes
+
+    def initialize(menu)
+        @dishes = {}
```
TomisinJ, line 9: good use of a hash to store dishes on the menu

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+    
+    attr_reader :dishes
+
+    def initialize(dishes)
+        @dishes = dishes
+    end
+
+    def print
+        dishes.map do |title, price|
+        "%s £%.2f" % [title.to_s.capitalize, price]
```
TomisinJ, line 12: Readability may be increased if string interpolation was used here instead of format strings

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    it 'contain a list of dishes with prices' do
+        expect(menu.dishes).to eq(dishes) 
+    end
+
+    it 'shows a list of dishes with prices' do
+        printed_menu = "Beef £3.99, Fish £4.50, Sausage £1.99"
+        expect(menu.print).to eq(printed_menu)
+    end
+    
+end
```
TomisinJ, line 24: Tests are written clearly and are all relevant to the Menu class

lib/takeaway.rb
```diff
@@ -0,0 +1,25 @@
...
+    def place_order(dishes)
+        dishes.each do |dish, quantity|
+        order.add(dish, quantity)
+        end
+            
+    end
+
+    private
+    
+    attr_reader :menu, :order
```
TomisinJ, line 23: Code is easy to understand and good use of the attribute reader

PR #2016, opened 2021-10-02
Takeaway Challenge - September 2021 Tigers

FrancineBlanc:
This is my approach to the takeaway challenge, which covers all of the user stories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,20 +14,9 @@ Takeaway Challenge
 
  ```
 
-Instructions
--------
-
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
-
-Task
+## Introduction
 -----
-
-* Fork this repo
-* Run the command 'bundle' in the project directory to ensure you have all the gems
-* Write a Takeaway program with the following user stories:
+This program implements software using Ruby with tests using RSpec, in order to create a takeaway app that satisfies the following user stories:
 
 ```
 As a customer
@@ -47,37 +36,39 @@ So that I am reassured that my order will be delivered on time
 I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
 ```
 
-* Hints on functionality to implement:
-  * Ensure you have a list of dishes with prices
-  * The text should state that the order was placed successfully and that it will be delivered 1 hour from now, e.g. "Thank you! Your order was placed and will be delivered before 18:52".
-  * The text sending functionality should be implemented using Twilio API. You'll need to register for it. It’s free.
-  * Use the twilio-ruby gem to access the API
-  * Use the Gemfile to manage your gems
-  * Make sure that your Takeaway is thoroughly tested and that you use mocks and/or stubs, as necessary to not to send texts when your tests are run
-  * However, if your Takeaway is loaded into IRB and the order is placed, the text should actually be sent
-  * Note that you can only send texts in the same country as you have your account. I.e. if you have a UK account you can only send to UK numbers.
+## Installation
 
-* Advanced! (have a go if you're feeling adventurous):
-  * Implement the ability to place orders via text message.
+After forking and cloning this repo, you will need to run gem install bundler. After installation use the bundle command.
+```bash
+gem install bundler
+```
 
-* A free account on Twilio will only allow you to send texts to "verified" numbers. Use your mobile phone number, don't worry about the customer's mobile phone.
+```bash
+bundle
+```
 
-> :warning: **WARNING:** think twice before you push your **mobile number** or **Twilio API Key** to a public space like GitHub :eyes:
->
-> :key: Now is a great time to think about security and how you can keep your private information secret. You might want to explore environment variables.
+## Usage
 
-* Finally submit a pull request before Monday at 9am with your solution or partial solution.  However much or little amount of code you wrote please please please submit a pull request before Monday at 9am
-
-
-In code review we'll be hoping to see:
-
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
-
-Notes on Test Coverage
-------------------
-
-You can see your [test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) when you run your tests.
+To run this program, please launch `irb` froom the terminal. An example snippet of functionality is below:
+```ruby
+$ irb
+irb(main):001> require "./lib/food_order.rb"
+=> true
+irb(main):002> require "./lib/menu.rb"
+=> true
+irb(main):003> require "./lib/text_client.rb"
+=> true
+irb(main):004> food_order = FoodOrder.new
+=> 
+<FoodOrder:0x00007fb47296fea0
+... 
+irb(main):005> food_order.add_dish("fried plantain")
+=> [{"fried plantain"=>2.0}]
+irb(main):006> food_order.add_dish("curried goat")
+=> [{"fried plantain"=>2.0}, {"curried goat"=>7.5}]
+irb(main):007> food_order.total
+=> "Your total order is £9.50"
+irb(main):008> food_order.place_order
+=> "Order successfully placed!"
+```
+The user will receive a text message stating: "Thank you! Your order was placed and will be delivered before [TIME]"
--- a/lib/food_order.rb
+++ b/lib/food_order.rb
@@ -16,4 +16,8 @@ class FoodOrder
     basket.each { |item| item.each_value { |cost| sum += cost } }
     sum
   end
+
+  def correct_total?(price)
+    total == price
+  end
 end
--- a/lib/food_order.rb
+++ b/lib/food_order.rb
@@ -6,9 +6,14 @@ class FoodOrder
     @takeaway_menu = takeaway_menu
   end
 
-  #TODO return error if a dish on the menu is not selected
   def add_dish(dish)
-    selection = takeaway_menu.menu.select { |key, value| key == dish.to_sym }
+    selection = takeaway_menu.menu.select { |key, _value| key == dish.to_sym }
     basket << selection
+  end
+
+  def total
+    sum = 0
+    basket.each { |item| item.each_value { |cost| sum += cost } }
+    sum
   end
 end
--- a/lib/food_order.rb
+++ b/lib/food_order.rb
@@ -1,13 +1,14 @@
 class FoodOrder
-  attr_accessor :basket
+  attr_accessor :basket, :takeaway_menu
 
-  def initialize
+  def initialize(takeaway_menu = Menu.new)
     @basket = []
+    @takeaway_menu = takeaway_menu
   end
 
+  #TODO return error if a dish on the menu is not selected
   def add_dish(dish)
-    new_menu = Menu.new
-    selection = new_menu.menu.select { |key, value| key == dish.to_sym }
+    selection = takeaway_menu.menu.select { |key, value| key == dish.to_sym }
     basket << selection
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -47,37 +36,39 @@ So that I am reassured that my order will be delivered on time
...
-
-Notes on Test Coverage
-------------------
+```bash
+bundle
+```
 
-You can see your [test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) when you run your tests.
+## Usage
+
```
maishachowdhury, line 51: Good use of irb to show how program works to others

lib/food_order.rb
```diff
@@ -0,0 +1,30 @@
+class FoodOrder
```
maishachowdhury, line 1: Good and consistent formatting throughout

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
```
maishachowdhury, line 1: Clear classes for each process, compliant with separation of concerns, ease of debugging and simplicity

lib/message.rb
```diff
@@ -0,0 +1,6 @@
+class Message
```
maishachowdhury, line 1: Encapsulation achieved with separate classes

lib/text_client.rb
```diff
@@ -0,0 +1,22 @@
+require "twilio-ruby"
```
maishachowdhury, line 1: Good use of API

lib/text_client.rb
```diff
@@ -0,0 +1,22 @@
...
+    @message_formatter = message_formatter
+    @account_sid = ENV["TWILIO_ACCOUNT_SID"]
+    @auth_token = ENV["TWILIO_AUTH_TOKEN"] 
+    @client = Twilio::REST::Client.new(@account_sid, @auth_token)
+  end
+  
+  def send_text
+    message = @client.messages.create(
+      body: @message_formatter.format_message,
+      to: ENV["TWILIO_TO_NUMBER"],
```
maishachowdhury, line 18: Privacy achieved

spec/food_order_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "food_order"
+require "menu"
+require "text_client"
+
+describe FoodOrder do
```
maishachowdhury, line 5: Variables named appropriately

spec/food_order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  let(:dish_list) { { "test dish1" => 6.00, "test dish2" => 4.00, "test dish3" => 4.00 } }
```
maishachowdhury, line 7: Good use of doubles

spec/food_order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  let(:menu_double) { instance_double(Menu, menu: dish_list) }
+
+  let(:formatted_time) { (Time.now + 3600).strftime("%k:%M") }
+  let(:formatted_confirmation) { "Thank you! Your order was placed and will be delivered before #{formatted_time}" }
+
+  let(:text_client_double) { instance_double(TextClient) }
+
+  subject { described_class.new(menu_double, text_client_double) }
+
+  context "placing an order" do
```
maishachowdhury, line 17: clear and simple code complies with DRY principles

spec/spec_helper.rb
```diff
@@ -4,7 +4,7 @@
 SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
```
maishachowdhury, line 4: Good use of planning to make dubugging easier

lib/text_client.rb
```diff
@@ -0,0 +1,22 @@
...
+  attr_reader :message_formatter
+
+  def initialize(message_formatter = Message.new)
+    @message_formatter = message_formatter
+    @account_sid = ENV["TWILIO_ACCOUNT_SID"]
+    @auth_token = ENV["TWILIO_AUTH_TOKEN"] 
+    @client = Twilio::REST::Client.new(@account_sid, @auth_token)
+  end
+  
+  def send_text
```
maishachowdhury, line 15: Good way to achieve abstraction you have successfully built send_text and have hid this from the users perspective. For a user ordering they simply receive a text without doing anything. - You interact with your order by using only a few buttons. What’s going on under the hood? You don’t have to know — implementation details are hidden. You only need to know a short set of actions. 

ref: https://www.freecodecamp.org/news/object-oriented-programming-concepts-21bb035f7260/

lib/food_order.rb
```diff
@@ -0,0 +1,30 @@
...
+  attr_accessor :basket, :takeaway_menu, :total_order
```
maishachowdhury, line 2: It would be easy to build inheritance in your code. For example, we can have different types of menus with the menu class

maishachowdhury (commented):
Completed review. Well done!

lib/food_order.rb
```diff
@@ -0,0 +1,30 @@
...
+    @basket = []
+    @takeaway_menu = takeaway_menu
+    @message = message
+    @total_order = 0
+  end
+
+  def add_dish(dish)
+    selection = takeaway_menu.menu.select { |key, _value| key == dish }
+    basket << selection
+  end
```
alicelieutier, line 15: Here you are iterating through the menu instance variable of a different class. Whenever you reach into another class instance variable that makes your code more coupled. What if you wanted to change the way the menu is stored? You would have to change the FoodOrder class too. Instead, if you need something from a different class, it's good to create a method for it. Here you could have a method on `Menu` that returns a selection from a dish.
eg

in `FoodOrder`
```rb
def add_dish(dish)
    selection = takeaway_menu.get_selection(dish)
    basket << selection
end
```

in `Menu`
```rb
def get_selection(dish)
    @menu.select { |key, _value| key == dish }
end
```
This way, if you change the way `Menu` works later, as long as you keep the method `get_selection` working the same, then the FoodOrder should still work.
Decoupling like this will also make testing in isolation a lot easier as you just have to stub the return of the `get_selection` method instead of stubbing a full menu.

lib/food_order.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def total
+    basket.each { |item| item.each_value { |cost| self.total_order += cost } }
+    "Your total order is £%0.2f" % [total_order]
+  end
```
alicelieutier, line 20: I would avoid storing the total in an instance variable (remove `@total_order`). Storing it causes duplicated information. If you have a basket stored already, the total can always be calculated on the fly based on that basket. Not storing duplicated information avoids bugs when information is out of sync (eg the user expects the order_total to be up to date once they've added something to their basket but it is not). 
Make total_order be a method that returns the calculated total everytime.

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+      "fried plantain" => 2.00,
+      "mac and cheese" => 4.00,
+      "rice and peas" => 3.00,
+      "plain rice" => 2.50,
+    }
+  end
+
+  def show_menu
+    menu
+  end
```
alicelieutier, line 19: Should this have some formatting to it?

lib/text_client.rb
```diff
@@ -0,0 +1,22 @@
...
+    message = @client.messages.create(
+      body: @message_formatter.format_message,
+      to: ENV["TWILIO_TO_NUMBER"],
+      from: ENV["TWILIO_FROM_NUMBER"])
+    "Order successfully placed!"
```
alicelieutier, line 20: This string could also live in the Message class, this way this class doesn't have to know what type of message it's sending, and could be reused for al sorts of messages (like an error message, for example)

spec/food_order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+require "menu"
+require "text_client"
+
+describe FoodOrder do
+
+  let(:dish_list) { { "test dish1" => 6.00, "test dish2" => 4.00, "test dish3" => 4.00 } } 
+  let(:menu_double) { instance_double(Menu, menu: dish_list) }
+
+  let(:formatted_time) { (Time.now + 3600).strftime("%k:%M") }
+  let(:formatted_confirmation) { "Thank you! Your order was placed and will be delivered before #{formatted_time}" }
```
alicelieutier, line 11: You can use the `timecop` gem to stub the time in your tests instead of relying on `Time.now`.
For example, you could freeze the time to Oct 1st 2pm, and expect the confirmation to be:
"Thank you! Your order was placed and will be delivered before 15:00". It's always better in your tests to control the environment as much as possible.

spec/food_order_spec.rb
alicelieutier: Actually, DRY is less important for tests - readability of individual tests is the most important.

spec/food_order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  context "placing an order" do
+    before(:example) do
+      subject.add_dish("test dish1")
+      subject.add_dish("test dish3")
+    end
+
+    describe "#add_dish" do
+      it "adds dishes to the customer's basket" do
+        expect(subject.basket).to eq [{ "test dish1" => 6.00 }, { "test dish3" => 4.00 }]
```
alicelieutier, line 25: Should the user have direct access to the basket instance variable? Or should they have access to a method that prints the current basket? In the first case, that fixes the implementation into only one way. In the second, your tests could be based on the behaviour or printing a specific string, and it leaves you with more flexibility on how to change the implementation after.

spec/food_order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    end
+
+    describe "#total" do
+      it "returns the total sum of all orders in the basket" do
+        expect(subject.total).to eq "Your total order is £%0.2f" % [10]
```
alicelieutier, line 31: Why not write:
`expect(subject.total).to eq "Your total order is £10:00"`
It's best to test against hard coded strings (the opposite of code..). This is because as a tester, you know exactly how you want the string to look like, but you shouldn't have to ask yourself how to write the code to achieve it.

spec/food_order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    end
+
+    describe "#check_total" do
+      it "returns true if #total matches the sum of the dishes in the basket" do
+        price = 10.00
+        subject.total_order = 10.00
```
alicelieutier, line 38: You are stubbing the class under test here which is a bad practice. If your total was calculated based on the basket, you would not need this line.

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
+require "menu"
+
+describe Menu do
+
+  subject { described_class.new }
+  
+  let(:menu) { { test_dish1: 7.50, test_dish2: 6.00, test_dish3: 4.00 } }
+  
+  before(:example) do
+    allow(subject).to receive(:menu).and_return(menu)
```
alicelieutier, line 10: Stubbing the class under test is never a good idea. You could use dependency injection here.. But then I'm wondering what you want from the class.

spec/menu_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  describe "#menu" do
+    it "has a list of dishes with prices" do
+      expect(subject.menu).to include(test_dish1: 7.50)
+    end
+  end
+
+  describe "#show_menu" do
+    it "displays the menu" do
+      expect(subject.show_menu).to eq test_dish1: 7.50, test_dish2: 6.00, test_dish3: 4.00
+    end
```
alicelieutier, line 22: Right now the menu class is only storage, no logic, so there's no real point in writing tests for it. I would add logic to it though. Then tests would be necessary.

spec/message_spec.rb
```diff
@@ -0,0 +1,9 @@
+require "message"
+
+describe Message do
+
+  it "writes a confirmation of order message" do
+    delivery_time = (Time.now + 3600).strftime("%k:%M") 
+    expect(subject.format_message).to eq "Thank you! Your order was placed and will be delivered before #{delivery_time}"
```
alicelieutier, line 7: Same comment as above about `timecop`

alicelieutier (commented):
Great: You code looks super clean and readable. Short methods, good use of guard clauses.
Your tests are pretty good overall (organised, clear, well isolated) but with a few possible improvements.
To improve: I think you can improve the design by making is less coupled, and removing duplicated information. This will also make the tests easier to write. Your tests should be more hardcoded and focus on behaviour rather than state.

My comments below will point out some examples.

PR #2019, opened 2021-10-03
tom_kellett_weekend_submission

t-kellett:
Testing is woefully sparse - I really struggled with the syntax of the doubles on user input and so resorted to brute-force testing in irb - this was successful so I am confident the spec is satisfied, but testing is not great.

I am conscious that my flow for handling incorrect input is also a little sketchy (e.g. if you try to add a dish, and add it 4 times, you only get told the dish does not exist after you have selected how many times you want to add it) but this can all be cleaned up in time.

What this highlighted is I need to understand the syntax of how to test user input as I could not grasp it over the weekend.


lib/order.rb
```diff
@@ -0,0 +1,20 @@
+require 'dotenv/load'
```
ConorButler, line 1: Nice use of dotenv to load the environment variables.

lib/takeaway.rb
```diff
@@ -0,0 +1,86 @@
...
+    @order.send_confirmation
+    @order = @order_class.new
+    puts "Thanks, you should receive a text confirmation, feel free to start another order"
+  end
+
+  def serve_customer
+    loop do
+      puts "Please enter the number of the option you would like to select from the following: "
+      print_options
+      interface(valid_input(STDIN.gets.chomp))
```
ConorButler, line 45: I like how you check the user input, this is very readable.

ConorButler (commented):
The code is very readable and all methods are simple. The interface method is a bit long but I'm not sure how that can be shortened. I think the way you check the parameters and user input is good as it will prevent problems and potentially some edge cases. Regarding the tests, some of the stubbing is being repeated, and it might be better to avoid relying on user input.

PR #2021, opened 2021-10-03
Takeaway Challenge - Bernard Leanse

bernardleanse:


lib/twilio.rb
```diff
@@ -0,0 +1,24 @@
+require 'twilio-ruby'
+
+class TwilioSender
+
+	def initialize
+
+		@client = Twilio::REST::Client.new(
```
arthurfincham, line 7: Nice work with the environment variables - took me a while to do

spec/menu_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'menu'
+
+describe Menu do
+
+  it { is_expected.to respond_to(:dishes) }
+
+  let(:menu) { Menu.new }
+  let(:dish1) { double(:dish, name: "Chicken Tikka", price: 5.00 ) }
```
arthurfincham, line 8: Is this how you give a double attributes? Looks nice!

PR #2024, opened 2021-10-03
Takeaway challenge

AJOsmaston:
My solution to the Makers Academy Week 2 Takeaway Challenge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
-Takeaway Challenge
+Anthony's Takeaway Challenge
+=================
+[Readme adapted from here](https://github.com/makersacademy/takeaway-challenge)
 ==================
 ```
                             _________
               r==           |       |
-           _  //            |  M.A. |   ))))
-          |_)//(''''':      |       |
+           _  //            |  A.O. |   ))))
+          |_)//(''''':      |  M.A. |
             //  \_____:_____.-------D     )))))
            //   | ===  |   /        \
        .:'//.   \ \=|   \ /  .:'':./    )))))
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -14,70 +16,87 @@ Takeaway Challenge
...
+What I would like to implement in the future  
+-----  
+
+* Attempt the advanced challenge:  
+`Implement the ability to place orders via text message.`
+* Better display of information (e.g. the list of menu items)
+* Adding a default menu display (perhaps loading from an external csv file)
+
+Features
+-----
```
charlie83xt, line 99: Great job, your README file facilitates the reviewing. It pretty much walks the viewer through the whole implementation which makes easier to understand the code.

lib/item.rb
```diff
@@ -0,0 +1,15 @@
+class Item
+
+  attr_reader :name, :price
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+    @avaialble = true
```
charlie83xt, line 8: Cool, so the takeaway can remove those dishes that are no longer available.

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
+require_relative 'menu'
+require_relative 'twiliosms'
+
+class Takeaway
```
charlie83xt, line 4: Great application of SRP in your class methods, they literally do one single thing each

README.md
```diff
@@ -14,70 +16,87 @@ Takeaway Challenge
...
-* The code is elegant: every class has a clear responsibility, methods are short etc.
-
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
-
-Notes on Test Coverage
-------------------
-
-You can see your [test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) when you run your tests.
+* All tests pass - 15 examples, 0 failures
+* [Test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md): COVERAGE: 100.00% -- 156/156 lines in 8 files
```
charlie83xt, line 102: Impressive! 100% coverage.. Well done.

lib/twiliosms.rb
```diff
@@ -0,0 +1,32 @@
+require 'twilio-ruby'
+require 'date'
+
+class TwilioSMS
```
charlie83xt, line 4: Your code and tests look highly professional. Amazing job.

charlie83xt (commented):
Unbelievable Job.

lib/item.rb
alicelieutier: Currently it doesn't seem like there'a method to make items unavailable, which renders this attribute less useful

alicelieutier: Typo - `avaialble` -> `available`. You can use a spell checker extension on your code editor to find these.

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  attr_reader :list
+
+  def initialize
+    @list = []
+  end
+
+  def add(name, price, item = Item.new(name, price))
+    @list << item
+  end
```
alicelieutier, line 14: Are you adding the item as a parameter for mocking in your tests? It feels a bit out of place here, you may want to consider doing dependency injection of the class Item in your constructor.

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+    @choice = []
+  end
+
+  def menu
+    @menu_object.list
+  end
+
+  def add_item(name, cost)
+    @menu_object.add(name, cost)
+  end
```
alicelieutier, line 18: Should the user of the programme really have access to adding items to the menu? It maybe be better to not have this in here.

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def display_basket
+    @choice
+  end
+
+  def sub_total
+    @cost = INITIAL_COST
+    add_prices
+    "£#{@cost.round(2)}"
+  end
```
alicelieutier, line 34: It would be better if you didn't store the @cost as an instance variable, as it duplicates information that already exists in the choice array. Here you are only using it between this method and add_prices. It would be a lot better to calculate on the fly rather than store in the variable as a buffer:

```rb
def sub_total
    cost = add_prices
    "£#{cost.round(2)}"
end

def add_prices(initial_cost = INITIAL_COST)
    cost = initial_cost
    @choice.each do |item|
        cost += convert_price(item)
    end
    cost
end
```

lib/takeaway.rb
```diff
@@ -0,0 +1,51 @@
...
+  private
+
+  def add_prices
+    @choice.each do |item|
+      @cost += convert_price(item)
+    end
+  end
+
+  def convert_price(item)
+    item.price.gsub("£","").to_f
```
alicelieutier, line 49: You store the price as a string, and have to convert it back as a number. It's not ideal, better to store it as a number, and then only making it a string when you need to display it.

lib/twiliosms.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def send_sms(from = ENV["TWILIO_MOBILE"])
+    client = Twilio::REST::Client.new(@account_sid, @auth_token)
+
+    to = ENV["PERSONAL_MOBILE"]
+
+    client.messages.create(
+    from: from,
+    to: to,
+    body: "Thank you! Your order was placed and will be delivered before #{@delivery_time}"
```
alicelieutier, line 28: indent

lib/twiliosms.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(account_sid = ENV["TWILIO_ACCOUNT_SID"], auth_token = ENV["TWILIO_AUTH_TOKEN"])
+    @account_sid = account_sid
+    @auth_token = auth_token
+
+    @delivery_time = Time.now.strftime("%H:%M").split(":")
+    @delivery_time[0] = @delivery_time[0].to_i
+    @delivery_time[0] += DELIVERY_TIME
+    @delivery_time[0] = @delivery_time[0].to_s
+    @delivery_time = @delivery_time.join(":")
```
alicelieutier, line 17: I'm not sure it makes sense that the TwillioSMS class is the one calculating the delivery time. Maybe the class that calls `send_sms` could pass in a delivery time instead?

alicelieutier: Or maybe it's just the class that is not well name, and this should be called `DeliveryNotifier` or something.

spec/item_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'item'
+
+describe Item do
```
alicelieutier, line 3: This class is mostly just storing things, not really having any behaviour. I wouldn't test it unless it had behaviour.

You could add a behaviour of storing numbers but having a display method that displays the nice formatted prices, and then that would be a behaviour to test for.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  let(:name_two) { "name two" }
+  let(:price_two) { "£0.02" }
+  let(:item_one) { double :item, :name => name_one, :price => price_one, :available => true }
+  let(:item_two) { double :item, :name => name_two, :price => price_two, :avaialble => true }
+
+  
+  describe '#menu' do
+    it "can be accessed" do
+      expect(shop).to respond_to(:menu)
+    end
```
alicelieutier, line 17: Remove these - respond_to matchers are only useful during code writing.. and even then I'd skip them.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+
+    it "displays to the customer" do
+      expect(menu).to receive(:list)
+
+      shop.menu
+    end
```
alicelieutier, line 23: Good test!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  describe '#add_item' do
+    before do
+      expect(menu).to receive(:add)
```
alicelieutier, line 28: You may want to add some argument matchers https://relishapp.com/rspec/rspec-mocks/docs/setting-constraints/matching-arguments

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,92 @@
+require 'takeaway'
+
+describe Takeaway do
+  let(:menu) { double :menu }
+  let(:shop) { described_class.new(menu) }
+  let(:name_one) { "name one" }
+  let(:price_one) { "£0.01" }
+  let(:name_two) { "name two" }
+  let(:price_two) { "£0.02" }
```
alicelieutier, line 9: These make the tests harder to read IMO, being up there. Nothing wrong with having hard coded strings in your tests. It may even be better.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      item_one_cost = price_one
+
+      shop.add_item(item_one_name, item_one_cost)
+    end
+  end
+
+  describe '#pick' do
+    before do
+      expect(menu).to receive(:list).and_return [item_one]
+      expect(item_one).to receive(:available?).and_return true
```
alicelieutier, line 41: Are you expecting or allowing? Or both at the same time? It feels like you are testing more than one behaviour here.

alicelieutier: Actually - you should not use expectations in before, so these are likely allows.

spec/twiliosms_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  let(:twilio) { TwilioSMS.new(test_sid, test_key) }
+
+  let(:time_plus_one) do
+    time = Time.now.strftime("%H:%M")
+    time_plus_one = Time.now.strftime("%H:%M").split(":")
+    time_plus_one[0] = time_plus_one[0].to_i
+    time_plus_one[0] += TwilioSMS::DELIVERY_TIME
+    time_plus_one[0] = time_plus_one[0].to_s
+    time_plus_one = time_plus_one.join(":")
+  end
```
alicelieutier, line 17: This is way too much logic to have in a test. What if that logic is wrong?
Use timecop and hardcoded strings instead.
For example, you can freeze the time using timecop to be 1pm, and then use `"Thank you! Your order was placed and will be delivered before 14:00"` in your expectation.

spec/twiliosms_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  it 'calculates current time + 1 hours' do
+    expect(twilio.delivery_time).to eq time_plus_one
+  end
```
alicelieutier, line 21: `expect(twilio.delivery_time).to eq"10:00"`
Use timecop again, try and freeze time at different hours.

alicelieutier (commented):
You code works well, it's easy to understand and clean.Your tests are good overall.

Some things to improve:
Be careful to not have so much logic in your tests.
Storing prices as strings is never a good idea.
You have two classes that have behaviours, and two that are mostly storage, it could be more balanced.

PR #2025, opened 2021-10-03
Delia Bute_Takeaway Challenge

delexii:


lib/menu.rb
```diff
@@ -16,7 +16,7 @@ def initialize
   def see_menu
     @list.each_with_index do |(item, price), index|
       @range[index + 1] = item
-     "#{index + 1} #{item} = £#{price}"
+      puts "#{index + 1} #{item} = £#{price}"
```
Joseph112358, line 19: Good

lib/takeaway.rb
```diff
@@ -31,24 +28,46 @@ def user_order
       user_options
     when 2
       add_items
-      user_options
+    when 3
+      view_basket
+    when 4
+      send_text
+    when 5
+      exit
```
Joseph112358, line 37: Good, clear, concise code

lib/takeaway.rb
```diff
@@ -11,16 +11,13 @@ def initialize(menu = Menu.new)
```
Joseph112358, line 11: Good use of dependency injection above.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "takeaway"
+
+RSpec.describe Takeaway do
+  let(:takeaway) { Takeaway.new }
+  let(:option) { double :option, :user_options => 1 }
```
Joseph112358: Good use of double here.

PR #2030, opened 2021-10-03
Takeaway

emanfolo:


README.md
```diff
@@ -1,83 +1,38 @@
...
-
-You can see your [test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) when you run your tests.
+Upon initializing a instance of the Restaurant class, a user is able to create orders and receive a text message to their phone upon completing their order. 
+
+Users can also view a menu, and there are internal checks taken when an order is completed to ensure the total was correctly calculated.
+
+I've included an exerpt of the code's functionality in irb below. In this script you can see the error message which arises when a user tries to order something not on the menu, and you can see how a user is able to add items to an order and complete the order which then sends a Twilio text to their phone. 
+
+I was able to complete the functionality of the code to a good level, I had problems with the testing however and I need to look into how to properly test the SMS sending. I've added a fakesender which means that everytime Restaurant.complete_order is run on rspec you do not receive a text, however my testing is not extensive as I would like.
+
```
jtc27, line 8: Good work Emmanuel!  Your use of arrays was well done in the project

README.md
```diff
@@ -1,83 +1,38 @@
...
+   #<Dish:0x0000000138e7c2a8 @name="fries", @price=2.99>,
+   #<Dish:0x0000000138e7c208 @name="fries", @price=2.99>,
+   #<Dish:0x0000000138e7c168 @name="fries", @price=2.99>],
+ @menu={:fries=>2.99, :burger=>4.99, :coke=>1.99, :fanta=>1.49, :pizza=>10.49, :lasagna=>5.99, :sprite=>1.49, :nachos=>3.49, :popcorn=>1.49},
+ @total=13.96> 
+
+3.0.2 :011 > mcdonalds.complete_order
+ => <Twilio.Api.V2010.MessageInstance body: Sent from your Twilio trial account - Thank you! Your order was placed and will be delivered before 22:45 num_segments: 1 direction: outbound-api from: +447897017954 to: +447956093846 date_updated: 2021-10-03 20:45:28 +0000 price:  error_message:  uri: /2010-04-01/Accounts/AC89329d06bcc6a8d0c85b99f198937bc1/Messages/SMb4925b496a194bad81422495d59490e8.json account_sid: AC89329d06bcc6a8d0c85b99f198937bc1 num_media: 0 status: queued messaging_service_sid:  sid: SMb4925b496a194bad81422495d59490e8 date_sent:  date_created: 2021-10-03 20:45:28 +0000 error_code:  price_unit: USD api_version: 2010-04-01 subresource_uris: {"media"=>"/2010-04-01/Accounts/AC89329d06bcc6a8d0c85b99f198937bc1/Messages/SMb4925b496a194bad81422495d59490e8/Media.json"}>
```
jtc27, line 38: Remember to delete your twilio and phone information, it is included here

docs/Untitled Diagram.drawio
```diff
@@ -0,0 +1,62 @@
...
+          <mxGeometry width="120" relative="1" as="geometry">
+            <mxPoint x="560" y="420" as="sourcePoint" />
+            <mxPoint x="680" y="420" as="targetPoint" />
+            <Array as="points" />
+          </mxGeometry>
+        </mxCell>
+      </root>
+    </mxGraphModel>
+  </diagram>
+</mxfile>
```
jtc27, line 62: If you want to include this diagram in the readme, that would be cool

lib/fakesender.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def messages
+    self
+  end
+  
+  def create(from:, to:, body:)
+    @messages << Message.new(from: from, to: to, body: body)
+  end
+end
```
jtc27, line 20: Very cool and creative idea.  I suppose this is supposed to test to see if the text messenger works?

lib/order.rb
```diff
@@ -0,0 +1,58 @@
...
+      coke: 1.99,
+      fanta: 1.49,
+      pizza: 10.49,
+      lasagna: 5.99,
+      sprite: 1.49,
+      nachos: 3.49,
+      popcorn: 1.49,
+  }.freeze
+
+end
```
jtc27, line 58: You could put MENU in a separate class, that way it's easier to modify

lib/restaurant.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def start_order
+    @order = @order_class.new
+  end
+
+  def scrap_order
+    @order = @order_class.new
+  end
+
```
jtc27, line 27: @order = nil might be more useful, for **def initialize** and **def scrap_order**

lib/restaurant.rb
```diff
@@ -0,0 +1,56 @@
...
+    @text = @text_class.new
+    @text = @text.text(order)
+  end 
+
+  def send_text(_text)
+    @sender.send_text(@text)
+  end
+  
+
+end
```
jtc27, line 56: In def initialize you have many (8) variables.  Some refactoring might help reduce their number

lib/sender.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :client
+
+  def initialize
+    @client = Twilio::REST::Client.new(ENV["TWILIO_ACCOUNT_SID"], ENV["TWILIO_AUTH_TOKEN"])
+  end 
+
+  def send_text(text)
+    twilio_text(text)
+  end
+
```
jtc27, line 16: you can probably initialize this

lib/order.rb
```diff
@@ -0,0 +1,58 @@
...
+  end 
+
+  def add_dish(item, quantity = 1) 
+    
+    quantity.times do
+      @dish = dish_creator(item)
+      accumulator(@dish)
+      @dishes << @dish
+      @dish = nil
+    end
```
JohnForster, line 20: I don't think you need `@dish` to be an instance variable here. You can return it from dish_creator and use it as a local variable instead.

```rb
      dish = dish_creator(item)
      accumulator(dish)
      @dishes << dish
```

Wherever possible, you should avoid storing data in instance variables.

lib/order.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def menu_listing
+    puts "Menu listed below:" 
+    listing = MENU.collect { |key, value| "#{key}: £#{value}" }
+    listing.each { |item| puts item } 
+  end
+
+private
```
JohnForster, line 29: Nice separation of public and private methods 👍

lib/order.rb
```diff
@@ -0,0 +1,58 @@
+class Order
+
+  attr_reader :dishes, :total, :calc, :menu
+
+  def initialize(dish_class = Dish)
+    @dish_class = dish_class
+    @dishes = []
+    @dish = nil
+    @total = 0
```
JohnForster, line 9: Similar to my comment below, it is better to avoid duplicating state. What I mean by this is that `total` is the sum of all the dishes in the order, so it is better to calculate that when needed, than store it separately in its own variable. Instead, it is better to create a total method which calculates the total from the stored dishes.

lib/restaurant.rb
```diff
@@ -0,0 +1,56 @@
...
+  attr_accessor :calc
+
+  def initialize(order_class = Order, text_class = Text, sender_class = Sender)
+    @order_class = order_class
+    @order_history = []
+    @order = @order_class.new
+    @calc = 0
+    @text_class = text_class
+    @text = nil
+    @sender_class = sender_class
```
JohnForster, line 12: `@text_class` and `@sender_class` are only being used within the initialize function, so probably don't need to be stored in an instance variable.

lib/text.rb
```diff
@@ -0,0 +1,21 @@
...
+    delivery_time
+    "Thank you! Your order was placed and will be delivered before #{time}"
+  end
+
+  private 
+
+  def delivery_time
+    @t = @t.new + 1 * 60 * 60
+    @time = @t.strftime("%H:%M")
+  end
```
JohnForster, line 20: Again, it is better to return a value than to store it in an instance variable.

```rb
def text
  delivery_time = calculate_delivery_time(Time.new)
  "Thank you! Your order was placed and will be delivered before #{delivery_time}"
end
```

This way you can remove both the `@t` and `@time` instance variables.

lib/dish.rb
```diff
@@ -0,0 +1,12 @@
...
+  attr_reader :name, :price
+
+  def initialize(name, price)
+    @name = name 
+    @price = price
+  end
+
+  
+
+end
```
JohnForster, line 12: This class is very simple, could you use a hash instead?

PR #2031, opened 2021-10-03
Sergei Palmer Takeaway Pull Request

s-palmer:
Known Issues:

- Tests incomplete
- User input can currently break functionality if integer outside of menu options is selected or if text string is added


Gemfile
```diff
@@ -11,3 +11,7 @@ end
 group :development, :test do
   gem 'rubocop', '1.20'
 end
+
+gem 'dotenv'
```
Matt-Warnock, line 15: nice :-)

lib/menu.rb
```diff
@@ -0,0 +1,75 @@
...
+  def initialize
+    @menu = [{ :name => "Risotto", :price => 7.99 },
+      { :name => "Large Pizza", :price => 12.99 },
+      { :name => "Fried Chicken", :price => 6.99 },
+      { :name => "Green Salad", :price => 5.99 },
+      { :name => "Pasta", :price => 8.99 },
+      { :name => "Vegetable Curry", :price => 6.99 },
+      { :name => "Rice", :price => 1.99 },
+      { :name => "Steamed Vegetables", :price => 1.99 },
+      { :name => "Acai Bowl", :price => 3.99 }]
```
Matt-Warnock, line 16: Very nice and tidy. I prefer to put long arrays in this format:
```menu = [
{ name:  'Risotto', price: 7.99 }
ect...
]```

File: lib/order.rb [hunk 3]
```diff
@@ -0,0 +1,33 @@
...
+  def send_sms
+    account_sid = ENV["TWILIO_ACCOUNT_SID"]
+    auth_token = ENV["TWILIO_AUTH_TOKEN"]
+    @client = Twilio::REST::Client.new account_sid, auth_token
+
+    message = @client.messages.create(
+      body: "Thanks for your order! It will be delivered in 45 minutes at #{@time.strftime("%H:%M")}.",
+      to: ENV["MY_PHONE_NUMBER"],
+      from: ENV["TWILIO_MAGIC_NUMBER"])
+  end
```
Matt-Warnock, line 31: Very nice. Maybe, having delivery_time return the time string would keep the time responsibility away from this method.
Like:
```
 def delivery_time
    time = Time.new + 45 * 60
    time.strftime("%H:%M")
  end
```

lib/send_sms.rb
```diff
@@ -0,0 +1,15 @@
...
+@client = Twilio::REST::Client.new @account_sid, @auth_token
+
+message = @client.messages.create(
+  body: "Thanks for your order! It will be delivered at approximately #{@time.strftime("%H:%M")}.",
+  to: ENV["MY_PHONE_NUMBER"],
+  from: ENV["TWILIO_MAGIC_NUMBER"])
+
+def delivery_time
+  @time = Time.new + 45 * 60
+end
```
Matt-Warnock, line 15: I'm guessing you used this as a note pad and its not part of the production code.

lib/order.rb
s-palmer: Good idea. Would make it easier passing the time between methods

lib/menu.rb
s-palmer: Yeah definitely something for my refactoring pass

lib/menu.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  def order_food
+    puts "Enter the number of the menu item to add to your order."
+    puts "To finish your order, hit return twice."
+  end
+
+  def input_order
+    start_new_order
+    input = STDIN.gets.chomp
+    # if input =~ /^-?[0-9]+$/
```
Matt-Warnock, line 48: Ok, so I think I see what maybe you had issues with.
Using regex to vaidate the input, I think is the right way. From my understanding, the best way to do this is:

`if input.match(^\d$)` this would return true on strings that only have 1 number (between 0-9)

you would then need to setup a loop if you wanted to give the user the chance to add a valid input

lib/menu.rb
```diff
@@ -0,0 +1,75 @@
...
+  def display_menu
+    menu.each_with_index do |item, index|
+      puts "#{index + 1}. #{item[:name]}, Price: £#{item[:price]}"
+    end
+  end
+
+  def display_order
+    order_total = 0
+    puts "Order summary"
+    @new_order.orders.each do |item|
```
Matt-Warnock, line 28: I would consider give item -1 semantic value by puting it into a var like:
`user_choice = item - 1`

spec/menu_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+  end
+
+  describe "#start_new_order" do
+    it 'creates a new Order' do
+      expect(menu.start_new_order).to be_an_instance_of Order
+    end
+  end
+
+  describe "#display_order" do
```
Matt-Warnock, line 25: Umm, missing test

lib/send_sms.rb
s-palmer: Will eventually be used to call the sms function from its own class, but for now I've just added this straight to the order.rb file.

spec/menu_spec.rb
s-palmer: You'll find a lot of this haha

lib/menu.rb
```diff
@@ -0,0 +1,79 @@
...
+    puts "#{@new_order.orders.count} items in order"
+  end
+  
+  def add_to_order(item)
+    @new_order.orders << item.to_i
+  end
+
+  def start_new_order
+    @new_order = Order.new
+  end
```
JohnForster, line 66: Does a menu have an order? I don't think that `@new_order` should belong to the menu class. Maybe have a think about how you could restructure this.

PR #2034, opened 2021-10-03
Weekend challenge #2, submission via pull request

blue-luke:
See README2.md for info specific to my code


README2.md
```diff
@@ -0,0 +1,57 @@
+Takeaway Challenge
```
Kntanos, line 1: Wow. Very detailed. Good job!

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def add_dish(dish)
+    @order.append(dish.name => dish.price)
+  end
+
+  def contents
+    @order
+  end
+
+  def total
```
Kntanos, line 23: Maybe the Order class is starting to do too many things? Could Total be a separate class?

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class Order
+
+  attr_reader :menu
+
+  def initialize
+    @order = []
+  end
+
+  def add_menu(menu)
```
Kntanos, line 11: If I understand correctly, you have this method so that you can have access to the dishes. Could you not initialise the class with a menu and inject the dependency there? Given that you can't have an order without a menu, it would save you from having to call add_menu.

lib/quickstart.rb
```diff
@@ -0,0 +1,5 @@
+require 'sinatra'
```
Kntanos, line 1: Did you code this in? I am not sure what it is doing here.

lib/order.rb
```diff
@@ -0,0 +1,36 @@
+require './lib/send_sms'
+
+class Order
```
Kntanos, line 3: Overall the methods are impressively clear, concise and easy to follow.

docs/automate_feature_test.rb
```diff
@@ -0,0 +1,21 @@
+require './lib/menu'
```
Kntanos, line 1: Very clever time saving!

lib/order.rb
```diff
@@ -0,0 +1,36 @@
...
+  def total
+    total = 0
+    @order.each do |x|
+      total += x.values.sum
+    end
+    total
+  end
+
+  def place_order
+    sms = SMS.new
```
leoht, line 32: The `SMS` instance could be dependency-injected in this class, rather than directly instantiated here  — it'll make it easier to mock in the tests

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  describe '#add_dish' do
+    it 'responds to add_dish' do
+      menu = Menu.new
+      expect(menu).to respond_to(:add_dish)
+    end
+    it 'adds a dish' do
+      menu = Menu.new
+      d1 = double('dish')
+      allow(d1).to receive_messages(
+      :name => "Chicken Korma",
```
leoht, line 21: Minor — the indentation style seems slightly off in some places, did you try running rubocop over those files? 🤔

lib/send_sms.rb
```diff
@@ -0,0 +1,30 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'dotenv/load'
+require 'twilio-ruby'
+
+class SMS
+
+  attr_reader :account_sid, :auth_token, :from, :to, :time
+
+  def initialize
+    @account_sid = ENV["ACCOUNT_SID"]
```
leoht, line 10: Good use of environment variables :) but it'd be even better if those values were injected via the constructor, to avoid `SMS` internally relying on the environment, but leaving this decision to the caller upstream:

```ruby
SMS.new(
  ENV["ACCOUNT_SID"],
  ENV["AUTH_TOKEN"],
  ENV["TWILIO_MOBILE"],
  ENV["MY_MOBILE"]
)
```

spec/menu_spec.rb
```diff
@@ -0,0 +1,29 @@
+require './lib/menu'
+
+describe Menu do
+
+  describe '#initialize' do
+    it 'is initialized with an empty hash' do
+      menu = Menu.new
+      expect(menu.contents).to be_kind_of(Hash)
```
leoht, line 8: Minor — this test could be changed/simplified to `expect(menu.contents).to eq({})` so it checks the type and *also* that the hash is empty

lib/send_sms.rb
```diff
@@ -0,0 +1,30 @@
...
+    @client = Twilio::REST::Client.new(@account_sid, @auth_token)
+    @time = Time.new + 3600
+    @delivery_time = @time.strftime("%k:%M")
+    @from = ENV["TWILIO_MOBILE"] # Your Twilio number
+    @to = ENV["MY_MOBILE"] # Your mobile phone number
+  end
+
+  def send_sms
+    @client.messages.create(
+    from: @from,
```
leoht, line 21: Same remark as below re: indentation

lib/quickstart.rb
```diff
@@ -0,0 +1,5 @@
...
+
+get '/' do
```
leoht, line 3: Is that a wandering sinatra app in this project? You could probably delete it if it's not used :)

leoht (commented):
Great work! Overall the code is super clear, well named and code design is good too. I just had a few comments about minor things such to simplify the code and formatting, etc. Let me know if you've got any further question :)

PR #2035, opened 2021-10-03
Leo Brown takeaway-challenge

LeoBrownPhonic:
The program is finished in that it works and can text a phone but does not do so in test mode.  Instead it uses the Twilio test function.  To enable both sets of Twilio functionality - acc_sid, authorisation token and send/receive phone numbers are added along with their Twilio test equivalents in sms_sender.rb and sms_sender_spec.rb.  I have removed these for privacy so the test will fail without these.

There is still some duplication in tests that could be further DRY'ed given more time.

Some tests are fairly light or inflexible in their implementation and could be improved.

Dishes could be added as a class.

I have not added the ability to order via text.

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,6 +1,6 @@
 require 'twilio-ruby'
-require 'sms_sender'
-require 'menu'
+require_relative 'sms_sender'
+require_relative 'menu'
 
 class Order
   attr_reader :menu, :order, :total, :delivery_message
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -6,22 +6,17 @@ class Order
   attr_reader :menu, :order, :total, :delivery_message
 
   def initialize(sms_sender: sms_sender = SmsSender.new, menu: menu = Menu.new)
-    @list = {
-      1 => "cod £5",
-      2 => "fishcake £4",
-      3 => "chips £3"
-    }
-    @menu = menu.list
+    @menu = menu
     @order = []
     @sms_sender = sms_sender
   end
 
   def see_menu
-    @menu
+    @menu.list
   end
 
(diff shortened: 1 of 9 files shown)


spec/sms_sender_spec.rb
```diff
@@ -1,18 +1,14 @@
...
     it "Sends an SMS with the delivery message" do
-      tw_testacc_sid = "ACccb2dbca9aea7869f0ab7bbbdca1df82"
-      tw_test_token = "2bf2f033df709c2079d8ee1182a16315"
-      tw_testnum = "+15005550006"
+      tw_testacc_sid = "ACccb2dbca9aea7869f0ab7bbbdca1df82" # Enter test acc_sid here
+      tw_test_token = "2bf2f033df709c2079d8ee1182a16315" # Enter test authorisation token here
+      tw_testnum = "+15005550006" # This is Twilio's magic phone  number for test
       sms_sender = SmsSender.new
       expect(sms_sender.send_text("test", tw_testacc_sid, tw_test_token, tw_testnum)).to be_truthy
     end
```
ewa-j, line 14: Good choice on separating text message class and order class, it would make it much easier to change API in the future.

lib/order.rb
```diff
@@ -33,7 +33,7 @@ def confirm_order
 
   def time_plus_1hour
     t = Time.new
-    hour_plus1 = ((t.strftime("%H").to_i) + 1)
+    hour_plus1 = (t.strftime("%H").to_i + 1)
     t.strftime("at #{hour_plus1}:%M%p")
   end
```
ewa-j, line 39: Incrementing time by one hour seems a bit overcomplicated, could be done in shorter way.

PR #2036, opened 2021-10-04
Take-away challenge- User stories 1-3 complete, User stop 4 incomplete

Anamaria-Mocanu:


spec/order_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    it "calculates the total of the order" do
+      allow(dish).to receive(:available?).and_return(true)
+      order.select_meal(1)
+      order.select_meal(2)
+      expect(order.total_order).to eq (7.50)
+    end
+  end
+  describe "#confirm" do
+    it "changes the send_order variable to true" do
+      expect(order.confirm).to eq true
```
feedXia, line 25: Maybe could use the one liner syntax:
```
it { should be_true }
    # or
it { is_expected.to be_true }
```
here :)

feedXia (commented):
Hey Ana!
Thanks for letting me review your code ☺️.
It was so interesting to see how you approached & implemented the user stories. 

I especially like how you used the !'s as a boolean, the ternary operators & how you included the ability to check if something is not on the menu.

I left a few feedback suggestions, feel free to use/discard/let me know what you think!

PR #2037, opened 2021-10-04
Takeaway Challenge 

feedXia:
### User Stories

- [x] Customers can check a list of dishes with prices
- [x] Customers can select the meal they want by selecting some number of several available dishes
- [x] Customers can verify their order is correct by checking that the total they have been given matches the sum of the various dishes in their order
- [x] Customer can receive a text stating that the order was placed successfully and that it will be delivered 1 hour from now such as _"Thank you! Your order was placed and will be delivered before 18:52"_ after order is placed

### Project Miscellaneous

- [x] Plan & draw a diagram of my domain model
- [x] Use the twilio-ruby gem to access the API
- [x] Use the Gemfile to manage your gems
- [x] Make sure that your Takeaway is thoroughly tested and that you use mocks and/or stubs, as necessary to not to send texts when your tests are run
- [ ] However, if your Takeaway is loaded into IRB and the order is placed, the text should actually be sent
- [x] All tests passing
- [x] > 95% Test Coverage
- [x] Make the code elegant: every class has a clear responsibility, methods are short etc.
- [ ] (Advanced) Implement the ability to place orders via text message.
- [x] Ensure private information such as **mobile number** or **Twilio API Key** is kept secret: e.g. use environment variables
- [x] Submit pull request


lib/meal.rb
```diff
@@ -0,0 +1,16 @@
+class Meal
+  attr_accessor :dishes, :menu
+
+  def initialize(menu_class = Menu)
```
Anamaria-Mocanu, line 4: Did not know that you can do the initialise like that, looks so much cleaner

lib/menu.rb
```diff
@@ -0,0 +1,24 @@
+class Menu
+  attr_accessor :menu
```
Anamaria-Mocanu, line 2: would it be better to use attr_reader in here, so that the @menu cannot be altered somewhere else in the code?

spec/confirmation_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  describe "#text" do
+    it "Sends text stating order was placed successfully & will be delivered 1hr from now" do
+      expect(subject.text).to eq "Thank you! Your order was placed and will be delivered before #{delivery_time}"
+    end
+  end
+
+  describe "#create_message" do
+    it "Sends text with order confirmation & delivery time" do
+      p subject.create_message
```
Anamaria-Mocanu, line 33: you missed taking out the p here

spec/meal_spec.rb
```diff
@@ -0,0 +1,29 @@
+require_relative "../lib/meal.rb"
+require_relative "../lib/menu.rb"
+
+RSpec.describe Meal do
+  let(:customer_order) { [:avocado_maki, :katsu_curry, :miso_soup] }
+  describe "#order" do
+    it "stores a dish" do
```
Anamaria-Mocanu, line 7: brilliant way to check the instance variables

spec/meal_spec.rb
```diff
@@ -0,0 +1,29 @@
+require_relative "../lib/meal.rb"
+require_relative "../lib/menu.rb"
```
Anamaria-Mocanu, line 2: requiring the menu in here might not be necessary, tried a couple of times to use the tests with only the require to the class I am testing, and it seemed to work. I think maybe RSpec knows to look in the lib folder and find what it requires, if you have a bit of time, maybe take it out and let me know if it still works, really curious about this.

PR #2038, opened 2021-10-04
takeaway_challenge - Arthur Fincham

arthurfincham:


lib/menu.rb
```diff
@@ -0,0 +1,30 @@
...
+    Dish.new('Mushroom Pizza', 9),
+    Dish.new('American Hot Pizza', 11.5),
+    Dish.new('Salsiaccia Pizza', 10.5),
+    Dish.new('Truffle Pizza', 11),
+    Dish.new('Chorizo Arancini', 6.5),
+    Dish.new('Veg Arancini', 5.5),
+    Dish.new('Tiramusi', 5),
+    Dish.new('Peroni', 3),
+    Dish.new('Nastro Azzuro', 3)
+  ].freeze
```
bernardleanse, line 17: Nice touch to have immutable menu tied to the menu class

lib/order.rb
```diff
@@ -0,0 +1,83 @@
...
+    @items = []
+    @dishes = Menu::DISHES
+    @total = 0
+  end
+
+  # Order an item with quantity
+  def order(item, quantity = 1)
+    check_item(item)
+    item_quant = [item, quantity]
+    @items << item_quant
```
bernardleanse, line 23: could refactor this into sub private methods potentially

lib/order.rb
```diff
@@ -0,0 +1,83 @@
...
+  def calc_price
+    @items.each do |item|
+      @dishes.each do |dish|
+        if item[0] == dish.name
+          @total += (dish.price * item[1])
+        end
+      end
+    end
+  end
+
```
bernardleanse, line 47: maybe theres a simpler way to do this? maybe items could contain dish objects with a tied quantity instance variable? to be fair I tried to do this but couldn't get it implemented that way. not sure if this is much of an issue anyway

lib/order.rb
```diff
@@ -0,0 +1,83 @@
...
+      false
+      raise('We are unable to find that item.')
+    else
+      true
+    end
+  end
+
+  def complete_order
+    @items.empty? ? "There are no items in your basket." : send_text
+  end
```
bernardleanse, line 66: nice concise one liner method

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+      subject.order('Veg Arancini', 2)
+      expect { subject.list }.to output(
+  <<-menu
+2x Peroni
+2x Veg Arancini
+Your total today is £23.0
+menu
+      ).to_stdout
+    end
+  end
```
bernardleanse, line 36: is there a way to have £23.00 instead of £23.0? had an issue myself making it show two decimal places, couldn't sort it.

bernardleanse (commented):
Well structured and readable code, perhaps a few methods could be simpler.

lib/order.rb
bernardleanse, line 66: write test

PR #2039, opened 2021-10-04
Ewa Jablonska takeaway-challenge

ewa-j:


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,61 @@
+require "./lib/takeaway"
+require "./lib/menu"
```
LeoBrownPhonic, line 2: This test is not independent of the menu class - I don't think tests should require class files other than the one they are treating.  A double of menu should be used.

lib/takeaway.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def total
+    @basket.each { |item, quantity| @total += @menu.menu[item] * quantity }
+    @total
+  end
+
+  def checkout(amount)
+    raise "Incorrect total order value" unless amount == total
+    time = (Time.now + 3600).strftime("%H:%M")
```
LeoBrownPhonic, line 32: This is a much neater way of increasing the time by 1 hour than mine

lib/takeaway.rb
```diff
@@ -0,0 +1,52 @@
...
+    message = "Thank you! Your order was placed and will be delivered before #{time}"
+    send_message(message)
+  end
+
+  def confirm_basket(dish, quantity)
+    print "#{quantity} x #{dish} has been added to your basket."
+  end
+
+  def send_message(message)
```
LeoBrownPhonic, line 41: The text sender should be separated into a separate class.

PR #2040, opened 2021-10-04
Takeaway Challenge - Amin

amin-newcastle:
I spent more time planning then got stuck with testing so didn't manage to complete much.


README.md
```diff
@@ -1,83 +1,14 @@
...
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
+# Takeaway Challenge
 
- ```
+Develop a food ordering program in Ruby to fulfill the specific user stories, testimg with RSpec.
 
-Instructions
--------
+Click [here](https://github.com/amin-newcastle/takeaway-challenge/blob/2fc2cbba04b98697bc79fd071c572a1a54659654/domain_modelling_takeaway_challenge.pdf) to view my domain modeling pdf
```
BethLambertMatthews, line 5: I really liked reading through this PDF. The user stories table and diagram was very clear.

lib/customers.rb
```diff
@@ -0,0 +1,12 @@
+class Customers
```
BethLambertMatthews, line 1: Would Customer maybe be a clearer name for the Class? Then when you're making an instance, it is one new customer (not a group).

lib/customers.rb
```diff
@@ -0,0 +1,12 @@
...
+    attr_accessor :a_customer
+    
+    def initialize
+        @a_customer = {}
```
BethLambertMatthews, line 5: Does the curly brackets mean you are creating a hash? I always forget this syntax and use [ ] for an array. Thank you for the reminder that { } is an option.

lib/customers.rb
```diff
@@ -0,0 +1,12 @@
...
+    end
+
+    def create_customer(customer_name, mobile_number)
+        a_customer[customer_name] = mobile_number
```
BethLambertMatthews, line 9: I'm not sure what this code means - does it set the mobile_number as the value, and the customer_name is the key?

lib/dishes.rb
```diff
@@ -0,0 +1,16 @@
+class Dishes
+    attr_accessor :my_menu
+
+    def initialize
```
BethLambertMatthews, line 4: I like the idea of using the my_menu hash to store information.

spec/customer_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'customers'
+
+describe Customers do
+
+    let(:my_customer) { Customers.new }
+
+    describe "#create_customer" do
+        it "Should add a new customer with a name and number" do
+            mobile_number = 7999985111
+            expect(my_customer.create_customer("Joe Bloggs", mobile_number)).to eq mobile_number
```
BethLambertMatthews, line 10: Why does creating a new customer only return the mobile number? Could it return a hash where name: "Joe Bloggs" and mobile_number: = 7999985111

spec/dishes_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'dishes'
+
+describe Dishes do
+
+    let(:my_dishes) { Dishes.new }
+
+    describe "#create_menu" do
+        it "Should add a new dish to the menu with a name and price" do
+            price = 8.95
```
BethLambertMatthews, line 9: Would it be clearer to write name = "Chicken Tikka" here and pass both variables in below?

PR #2041, opened 2021-10-04
Takeaway challenge

Eagle555:


lib/takeaway.rb
```diff
@@ -0,0 +1,35 @@
+class Takeaway
+  def initialize
+    @message = []
+  end
+  
+  def notification(deliver)
+    @message << "Thank you! Your order was placed on #{Time.now.strftime("%d/%m/%Y %H:%M")} and will be delivered before #{(Time.now + 1 * 60 * 60).strftime("%H:%M")}"
```
TTurvey, line 7: In order to cut down on the amount of code within this method you could potentially have a separate method which is responsible for returning the expected delivery time.

lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+class Menu
+  
+  def initialize
```
TTurvey, line 3: By including the menu hash in the initialize method, any changes to the menu within a Menu class instance would not be saved when the code was loaded up again. You would have to manually enter the changes into the menu hash for them to be permanent.

TTurvey (changes requested):

PR #2043, opened 2021-10-04
Submit Takeaway Challenge

pzrsa:
This PR contains my submission for the Takeaway Challenge.

I have fulfilled 3/4 user stories:

- [x] I would like to see a list of dishes with prices
- [x] I would like to be able to select some number of several available dishes
- [x] I would like to check that the total I have been given matches the sum of the various dishes in my order
- [ ] I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered

### DIFF
--- a/lib/dish.rb
+++ b/lib/dish.rb
@@ -1,0 +1,2 @@
+class Dish
+end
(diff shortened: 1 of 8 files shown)


ranjotsd (commented):
The code is well formatted, easy to follow and split into related classes. Use of mocking is great.
The Takeaway class seems unnecessary and would be easier to understand with a README.

PR #2045, opened 2021-10-04
Takeaway Challenge - Doug Kerins

dpkerins:


README.md
```diff
@@ -1,83 +1,11 @@
...
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
+My code accomplishes all the user stories and sends a text via Twilio on confirmation. 
 
- ```
+Process -->
 
-Instructions
--------
+I knew that I wanted the order class to be dependent on the menu and the menu on order and order on dish, so I wrote each class in reverse order beginning with the dish. I injected dependency as I went and used mocking to isolate tests in rspec (or to the best that I could)
```
Langan84, line 5: This is great idea to work in reverse order and helps keep things clear & simple.

lib/dish.rb
```diff
@@ -0,0 +1,7 @@
+class Dish
+  attr_reader :name, :price
+  def initialize(name, price)
+    @name = name
+    @price = price
+  end
+end
```
Langan84, line 7: Really easy to understand what this class is doing.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :selected_items
+  def initialize(menu_class = Menu, dish_class = Dish, confirmation_class = Confirmation)
+    @selected_items = []
+    @menu = menu_class.new([dish_class.new("chips", 1.50), dish_class.new("nuggets", 2.00),
+    dish_class.new("big mac", 3.50), dish_class.new("apple pie", 0.75)])
+    @total = 0
+    @confirmation_class = confirmation_class
+  end
+  def display_menu
+    @menu.dishes.each_with_index { |dish, index| puts "#{index + 1}. #{dish.name} £#{dish.price}"}
```
Langan84, line 11: Great idea to add an index to iterate through the dishes. Rather than try and use a hash.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+  def add_order_item(dish_number, quantity = 1)
+    raise ("This is not a menu item! Please pick something from our menu!") unless dish_number.is_a?(Integer)
+    raise ("This is not a menu item! Please pick something from our menu!") unless dish_number <= @menu.dishes.length
+    dish_index = dish_number - 1
+    quantity.times {@selected_items << @menu.dishes[dish_index]}
+  end
+  def check_total
+    raise ("You haven't ordered anything yet!") if empty_order?
+    @selected_items.each{ |item| @total += item.price}
```
Langan84, line 21: Code is DRY and all methods/classes follow the SRP

Langan84 (commented):
The code is clear and very easy to understand, code is DRY could find no repetition. All methods had one job and all classes followed SRP. A job well done!

lib/confirmation.rb
```diff
@@ -0,0 +1,27 @@
...
+AUTH_TOKEN = ENV['TWILIO_TOKEN']
+RESTAURANT_PHONE = ENV['TWILIO_PHONE']
+CUSTOMER_PHONE = ENV['MY_PHONE']
+
+class Confirmation
+  def initialize(order_time = Time.now)
+    @order_time = order_time
+    @delivery_time = order_time + 3600
+    @delivery_time_hour = @delivery_time.hour
+    @delivery_time_minute = @delivery_time.min
```
JohnForster, line 13: You should try to keep as little in state as possible. Considering the `delivery_time` variables all are derived from `@order_time` it's better to just store `@order_time` and calculate the others when you need them, rather than calculating at the top and storing in an instance variable.

lib/confirmation.rb
```diff
@@ -0,0 +1,27 @@
...
+    "Thank you! Your order was placed and will be delivered before #{@delivery_time_hour}:#{@delivery_time_minute}"
+  end
+  def send_confirmation
+    client = Twilio::REST::Client.new(ACCOUNT_SID, AUTH_TOKEN)
+    
+    client.messages.create(
+    from: RESTAURANT_PHONE,
+    to: CUSTOMER_PHONE,
+    body: content
+    )
```
JohnForster, line 25: Be careful of your indentation. You can use a linter like rubocop to make sure that your indentation is always correct.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
+class Order
+  attr_reader :selected_items
+  def initialize(menu_class = Menu, dish_class = Dish, confirmation_class = Confirmation)
+    @selected_items = []
+    @menu = menu_class.new([dish_class.new("chips", 1.50), dish_class.new("nuggets", 2.00),
+    dish_class.new("big mac", 3.50), dish_class.new("apple pie", 0.75)])
```
JohnForster, line 6: Could you store this data somewhere else? The contents of this menu shouldn't be stored in the Order class.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+    @total = 0
```
JohnForster, line 7: Considering the total is derived from the dishes in the order, it's better to calculate it when needed, rather than store it in an instance variable. Duplicating state like this can cause bugs if they get out of sync, so you should try to stick to a single source of truth.

lib/order.rb
JohnForster: Does an order need to instantiate it's own menu? It might be an idea to pass in a menu instance here, instead of the menu class. Same goes for Dishes.

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+    @confirmation_class = confirmation_class
+  end
+  def display_menu
+    @menu.dishes.each_with_index { |dish, index| puts "#{index + 1}. #{dish.name} £#{dish.price}"}
+  end
```
JohnForster, line 12: Should the Order decide how to format the Menu, or should this be part of the Menu class?

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+    raise ("This is not a menu item! Please pick something from our menu!") unless dish_number.is_a?(Integer)
+    raise ("This is not a menu item! Please pick something from our menu!") unless dish_number <= @menu.dishes.length
+    dish_index = dish_number - 1
+    quantity.times {@selected_items << @menu.dishes[dish_index]}
+  end
+  def check_total
+    raise ("You haven't ordered anything yet!") if empty_order?
+    @selected_items.each{ |item| @total += item.price}
+    @total
+  end
```
JohnForster, line 23: I've mentioned it above, but you shouldn't use an instance variable here for total

lib/order.rb
```diff
@@ -0,0 +1,32 @@
...
+  def submit_order
+    check_total
+    confirmation = @confirmation_class.new()
+    confirmation.send_confirmation
+  end
+  def empty_order?
+    @selected_items.empty? ? true : false
+  end
+end
```
JohnForster, line 32: It would be good to make some of the methods in this file private.

spec/confirmation_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "confirmation"
+
+describe "Confirmation" do
+  describe ".initialize" do
+    it "should create a confirmation text detailing when the cusotmer should receive their food by" do
+      order_time = Time.new(2021, 10, 01, 12, 27)
+      # order_time = Time.now
+      puts order_time
+      confirmation = Confirmation.new(order_time)
+      expect(confirmation.content).to eq("Thank you! Your order was placed and will be delivered before 13:27")
```
JohnForster, line 10: No other class is accessing this `.content` reader, so you shouldn't be testing it. You want to be testing _behaviour_ not _state_. This means that you only want to be testing the inputs and outputs of the interface to this class.

spec/menu_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "menu"
+describe Menu do
+  describe ".initialize" do
+    it "should create a new Menu instance with a dishes array" do
+      menu = Menu.new([])
+      expected_result = []
```
JohnForster, line 6: This variable is unused

PR #2047, opened 2021-10-04
Takeaway Challenge

AishaDorsett:
Completed user stories except number 3. Will need to implement a better function that checks the total price matches the sum of the items requested. I made the Twilio texts work when the credentials were hard coded but when I had to change to env variables, it stopped working. Need to work on this too.


lib/menu.rb
```diff
@@ -2,12 +2,13 @@ class Menu
...
 
   def available_items
-    @items = {:chips => 3, :rice => 2, :lamb_shish => 10, :chicken_shish => 10, :lamb_doner => 7, :soft_drink => 2, :nil => 0}
+    @items = { :chips => 3, :rice => 2, :lamb_shish => 10,
+       :chicken_shish => 10, :lamb_doner => 7, :soft_drink => 2, :nil => 0 }
   end
 
   def display_menu
     available_items.each do |food, price|
       puts "#{food}: £#{price}.00"
```
nejlalowe, line 11: smart way of showing the prices! Do you think you could implement functionalist to this to enable the number to be a float if it was eg 10.50?

spec/order_spec.rb
```diff
@@ -4,7 +4,6 @@
 RSpec.describe Order do
   describe '#place order' do 
     it 'places order' do
-      menu = Menu.new
       expect(subject.place_order(:chips, :lamb_shish, :soft_drink)).to eq [3,10,2]
```
nejlalowe, line 7: really like that your place order returns the cost! Could customers place their order as a string?

spec/order_spec.rb
```diff
@@ -19,10 +18,10 @@
 
   describe '#order_confirmation' do
     it 'confirms the order' do
-      menu = Menu.new
       subject.place_order(:chips, :lamb_shish, :nil)
```
nejlalowe, line 21: Great workaround with ":nil"! Could you make an array or somewhere to put the order they choose so that you can add more food items than three?

spec/delivery_spec.rb
```diff
@@ -1,9 +1,11 @@
...
 require 'delivery'
 
-describe Delivery do
-  it 'states order delivery time' do
-    time = (Time.now + 1*60*60).strftime("%H:%M")
-    expect(subject.delivery_time).to eq "You will recieve your order at or before #{time}."
+RSpec.describe Delivery do
+  describe '#delivery_time' do
+    it 'states order delivery time' do
+      time = (Time.now + 1 * 60 * 60).strftime("%H:%M")
```
nejlalowe, line 7: This is awesome! I wouldn't have known how to implement the time to be an hour from now, its cool to see!

lib/order.rb
```diff
@@ -9,16 +9,14 @@ def place_order(item1 = nil, item2 = nil, item3 = nil)
...
   def order_total
-    total = @item_prices.inject(0, :+)
+    @item_prices.inject(0, :+)
   end
   
   def order_confirmation
     @selected_items.delete(:nil)
-    p selected_items
-    p "Thank you for placing your order at Aisha's Bistro. You have ordered: #{selected_items.keys}. Your total is £#{order_total}."
+    p "Thank you for placing your order at Aisha's Bistro.
```
nejlalowe, line 19: Have to give some praise to the personal touch in the naming here!

PR #2048, opened 2021-10-04
Take-away Joseph 

Joseph112358:
Finished take away (although I haven't resolved all test dependencies )

### DIFF
--- a/lib/customer.rb
+++ b/lib/customer.rb
@@ -1,6 +1,7 @@
 require_relative 'take_away'
 class Customer
   attr_reader :m_takeaway
+  attr_reader :basket
   def initialize 
     @basket = []
     @total = 0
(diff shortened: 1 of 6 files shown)


lib/customer.rb
```diff
@@ -0,0 +1,50 @@
...
+    @m_takeaway = m_takeaway
+    @sms = sms
+    
+  end
+
+  def readmenu
+    m_takeaway.list_menu
+  end
+
+  def add_item(item)
```
delexii, line 21: I really like how you've constructed the add and remove methods using simple syntax, instead of looping through the menu hash

lib/send-sms.rb
```diff
@@ -0,0 +1,13 @@
+require("bundler")
```
delexii, line 1: This is a really neat way of nesting an API within a method. I couldn't really understand it, so I've just used the standard twilio ruby file. That's great that you've linked it so well with your customer.rb

spec/take_away_spec.rb
```diff
@@ -0,0 +1,60 @@
+require_relative "../lib/customer.rb"
+
+
+describe Takeaway do
```
delexii, line 4: The tests are concise. One small suggestion would be to split the rspec files per class

PR #2049, opened 2021-10-04
Takwaway challenge

KirstieCat:
Incomplete

I was unable to complete user story 2

### DIFF
--- a/lib/take_away.rb
+++ b/lib/take_away.rb
@@ -2,11 +2,12 @@ require './lib/menu'
 
 class TakeAway
 
-attr_accessor :menu, :basket
+attr_accessor :menu, :basket, :total
 
   def initialize
     @menu = Menu.new
     @basket = []
+    @total = 0
   end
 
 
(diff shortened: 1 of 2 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,27 @@
...
+#   dishes = 
+#         {
+#               "Margherita Pizza": 8.99},
+#       #   {name: "Salami Pizza", price: 9.99},
+#       #   {name: "Hawaiian Pizza", price: 9.99},
+#       #   {name: "Chips", price: 4.99},
+#   }
+#   end
+
+  def print_dishes
```
dpkerins, line 18: I like the way you kept these methods in the separate classes and called them in the takeaway class. I think I should have done something more similar to this.

lib/take_away.rb
```diff
@@ -0,0 +1,21 @@
...
+    @menu = menu
+    @order = order
+  end
+
+  def print_menu
+    menu.print_dishes
+  end
+  
+  # def order_items(dishes)
+  #   dishes.each |item, quantity|
```
dpkerins, line 16: I think here you want to access each dish with the each and then access the item and quantity from the hash at each index and do the order.add

spec/take_away_spec.rb
```diff
@@ -0,0 +1,15 @@
+
+require "take_away"
+
+describe TakeAway do
+  subject(:take_away) { described_class.new(menu: menu) }
```
dpkerins, line 5: you might need to switch the order these are in -- I had that problem because I was putting a double in a double in a double and creating them in the wrong order

PR #2050, opened 2021-10-04
Takeaway challenge

marazzo:
I left the tests out as I was focusing on dependency injection

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -3,14 +3,33 @@ class Menu
   attr_reader :menu
 
   def initialize
-    @menu = []
+    @menu = [
+      { name: "beef burger", price: 5 }, 
+      { name: "chicken burger", price: 5 }, 
+      { name: "chicken popcorn", price: 3 },
+      { name: "wrap", price: 5 },
+      { name: "chips", price: 2 },
+    ]
   end
 
+# add dish to the menu attribute
   def add_dish(name, price)
-    @menu << {name => price}
+    @menu << { name: name, price: price }
   end 
 
+# view the menu
+  def view_menu
+    puts "~~~~That Chikkin' Shop~~~~"
+    @menu.each do |dish|
+      puts "#{dish[:name].capitalize} - £#{dish[:price]}."
+    end
+  end
+
+# this method doesn't work - it's meant to remove dish from menu
   # def remove_dish(name)
-  #   @menu.delete(name)
+  #   @menu.each do|dish|
+  #     dish.delete(name)
+  #   end
   # end
+
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,17 @@
+class Menu
+
+  attr_reader :menu, :dish
+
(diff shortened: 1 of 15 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+
+# view the menu
+  def view_menu
+    puts "~~~~That Chikkin' Shop~~~~"
+    @menu.each do |dish|
+      puts "#{dish[:name].capitalize} - £#{dish[:price]}."
+    end
+  end
+
+# this method doesn't work - it's meant to remove dish from menu
```
starf0g, line 28: Do you know why it doesn't work?

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize
+    @menu = [
+      { name: "beef burger", price: 5 }, 
+      { name: "chicken burger", price: 5 }, 
+      { name: "chicken popcorn", price: 3 },
+      { name: "wrap", price: 5 },
+      { name: "chips", price: 2 },
+    ]
+  end
+
```
starf0g, line 14: Nice simple methods here

lib/order.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :ordered_items, :menu
+
+  def initialize(menu_class = Menu)
+    @menu = menu_class.new
+    @ordered_items = []
+  end
+
+  # adds an item to order
+  def add_item(chosen_dish)
+    # raise "The item is not on the menu." if !@menu.menu.has_value(chosen_dish)
```
starf0g, line 12: Does this work?

lib/send_text.rb
```diff
@@ -0,0 +1,18 @@
+require 'twilio-ruby'
+
+class SendText
+  def send_text
+    time = Time.new
+    account_sid = "ACad78197ffa4e768e01d488004100210e"
```
starf0g, line 6: I can show you how to use dotenv to store these. You should remove these from your code and push again.

lib/takeaway.rb
```diff
@@ -0,0 +1,22 @@
+class Takeaway
```
starf0g, line 1: I like how short each method is, makes it easy to understand

spec/menu_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'menu'
+
+describe Menu do
+
+  it "can add a dish to menu array" do
+    subject.add_dish("fish and chips", 5)
+    expect(subject.menu.length).to eq 6
```
starf0g, line 7: This is hard to understand due to the magic number. One way you could improve this is to allow your Menu class to be passed a hash of menu items when its initialised. You can then create a test hash of items that you can refer to during testing.

starf0g (commented):
Overall I like how you've used multiple classes each with simple methods, it makes it readable and easy to understand. 
The areas I'd focus on to improve are:
- using env vars to store your credentials
- the unit testing for some of the classes

PR #2054, opened 2021-10-04
Takeaway weekend challenge

Kntanos:
For code review


lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+require 'csv'
+
+class Menu < Array
+
+  def initialize
+    load_menu
+  end
+
+  def load_menu
+    CSV.foreach("menu.csv", headers: true, header_converters: :symbol, converters: :numeric) do |row|
```
blue-luke, line 10: I like this way of loading the menu, as it implies that it can be done by someone non-technical and then loaded into the programme, which is probably how it would work in real life

.gitignore
```diff
@@ -3,3 +3,5 @@
 
 # Local cache of Rubocop remote config
 .rubocop-*
+
+.env
```
blue-luke, line 7: I like your use of dotenv and .env to store confidential variables

lib/checkout.rb
```diff
@@ -0,0 +1,25 @@
+require 'dotenv/load'
+require 'twilio-ruby'
+
+class Checkout
```
blue-luke, line 4: Perhaps there is a better name for this class, as this class only deals with sending a text message? A later class might deal with actual checkout, which would involve payment.

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+require 'csv'
+
+class Menu < Array
+
+  def initialize
```
blue-luke, line 5: I also like how you've split this into an initialize and a further function, it separates responsibilities clearly

lib/order.rb
```diff
@@ -0,0 +1,23 @@
+require_relative 'menu'
+require_relative 'total'
+
+
+class Order
+  attr_reader :menu, :total
+
+  def initialize(menu = Menu.new, total = Total.new)
```
blue-luke, line 8: It looks like you've used dependency injection here, nice!

lib/order.rb
```diff
@@ -0,0 +1,23 @@
...
+  attr_reader :menu, :total
+
+  def initialize(menu = Menu.new, total = Total.new)
+    @menu = menu
+    @total = total
+  end
+
+  def add_to_order(dish, quantity)
+    menu.each do |item|
+      total.selection << item.merge({:quantity => quantity}) if item.value?(dish)
```
blue-luke, line 15: I found it a bit hard to parse this line and figure out what the code is doing. You could consider refactoring it to make it more readable

myREADME.md
```diff
@@ -0,0 +1,14 @@
+Run
+```ruby
+$ ruby -r ./lib/order
+```
+
+The test menu used in development is:
```
blue-luke, line 6: Mentioning the test menu is useful. I also would've like to have seen more information about your approach to this task, difficulties, etc

spec/menu_spec.rb
```diff
@@ -0,0 +1 @@
+require 'menu'
```
blue-luke, line 1: As you've already mentioned you didn't have time to complete some of the tests.

spec/order_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'order'
+
+describe Order do
+
+  describe "#initialize" do
+    it "should initialize with a menu array" do
```
blue-luke, line 6: I like this test, I think it uses dependency injection well

spec/order_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'order'
+
+describe Order do
+
+  describe "#initialize" do
+    it "should initialize with a menu array" do
+      menu_double = double :menu, new: Array 
+      order = Order.new
```
blue-luke, line 8: I've had another look at this, and I'm not sure you've passed in the double to override the default argument. Wouldn't we need to say?:
order = Order.new(menu_double)

Kntanos: Right... it does look strange, but the test fails any other way I tried. I wonder if it is to do with the dependency injection initializing a menu by default?

PR #2057, opened 2021-10-06
Weekend Takeaway challenge for coaches code review

Kntanos:
I didn't have enough time to complete the challenge and I particularly struggled with Rspec, because not knowing how to test was keeping me from writing the code. So I didn't follow the process.

### DIFF
--- a/lib/checkout.rb
+++ b/lib/checkout.rb
@@ -2,6 +2,8 @@ require 'dotenv/load'
 require 'twilio-ruby'
 
 class Checkout
+  # possibly unfortunate name for the class, but I would have expected it to also handle
+  # the payment and not just send an SMS
 
   def initialize
     @time = Time.now + 3600
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -19,5 +19,9 @@ class Order
   def request_bill
     total.bill
   end
+
+  def request_receipt
+    total.receipt
+  end
 end  
 
--- a/lib/total.rb
+++ b/lib/total.rb
@@ -17,11 +17,8 @@ class Total
   end
 
   def receipt
-    puts "Your Receipt"
-    selection.each do |item|
-      puts "#{item[:dish]} - quantity:#{item[:quantity]} - £#{item[:price]}"
-     end 
-    puts "Total #{bill}"
+    selection
+  # not entirely sure what the user story required
   end
 
   def pay
--- a/spec/checkout_spec.rb
+++ b/spec/checkout_spec.rb
@@ -1,0 +1,11 @@
+require 'checkout'
+
+describe Checkout do
+  describe "#send_notification" do
+    it "should send an SMS with the delivery time window" do
+      checkout = Checkout.new
+      # expect(checkout.send_notification).to do something with Twilio
+      # no idea how to test this
+    end
+  end
+end
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,1 +1,3 @@
 require 'menu'
+
+# ?? What do I test ?
--- a/spec/order_spec.rb
+++ b/spec/order_spec.rb
@@ -1,16 +1,30 @@
 require 'order'
 
 describe Order do
-
   describe "#initialize" do
-    it "should initialize with a menu array" do
-      menu_double = double :menu, new: Array 
-      order = Order.new
-      expect(order.menu).to be_a Array
+    # I would normally think this test as unecessary, but in this case I want to make sure that
+    # if the input method for the menu changes, it muct still return an array. Is that correct?
+    let(:menu) {[]}
+    it "should initialize with a menu array" do      
+      expect(subject.menu).to be_a Array 
     end
   end 
 
   describe "#add_to_order" do
-    # no idea how to test this or what to test
+    # is this test insane?
+    it "should add the quantity ordered to the correct dish" do
+      menu = [{:dish => "Burgers", :price => 10}]
+      selection = []
+      @order = Order.new(menu)
+      @order.add_to_order("Burgers", 5)
+      total = Total.new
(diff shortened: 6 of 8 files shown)


lib/order.rb
```diff
@@ -0,0 +1,27 @@
+require_relative 'menu'
+require_relative 'total'
+
+
+class Order
+  attr_reader :menu, :total
+
+  def initialize(menu = Menu.new, total = Total.new)
```
leoht, line 8: I'd say the `Checkout` should also be a direct dependency of `Order` rather than of `Total` — maybe the naming could be changed as well to make the intent clearer. I'd suggest renaming `Order` to `OrderBuilder` maybe, and `Total` to `Order` (since it represents the *actual* order selection)

lib/total.rb
```diff
@@ -0,0 +1,27 @@
...
+    sum = []
+    selection.each do |item|
+      sum << item[:price].to_f * item[:quantity]
+    end
+    sum.inject(:+)
+  end
+
+  def receipt
+    selection
+  # not entirely sure what the user story required
```
leoht, line 21: It's probably fine to just leave out the `.selection` method to get the order list — if we wanted to have some printed and formatted receipt, it would probably become the job of another class (since this would deal with presenting/displaying data to the user, which is a different concern)

spec/checkout_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'checkout'
+
+describe Checkout do
+  describe "#send_notification" do
+    it "should send an SMS with the delivery time window" do
+      checkout = Checkout.new
+      # expect(checkout.send_notification).to do something with Twilio
```
leoht, line 7: By digging a bit into the Twilio library, you could have something like...
```ruby
expect_any_instance_of(Twilio::REST::Api::V2010::AccountContext::MessageList).to receive(:send).with({ from: expected_from, to: expected_to, body: expected_body })
```

to assert that the library is called with the right parameters when `checkout.send_notification` is called

spec/menu_spec.rb
```diff
@@ -0,0 +1,3 @@
+require 'menu'
+
+# ?? What do I test ?
```
leoht, line 3: I guess a good thing to test here would be the loading of a menu array from a CSV file, e.g:

```ruby
menu = Menu.new("test_menu.csv")
expect(menu).to eq(["Burger", "Pasta", "Salad"])
```

spec/order_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'order'
+
+describe Order do
+  describe "#initialize" do
+    # I would normally think this test as unecessary, but in this case I want to make sure that
+    # if the input method for the menu changes, it muct still return an array. Is that correct?
```
leoht, line 6: That's correct, but in that case you'd need to inject a mock or double of `Menu` into the order, and then try this test with different cases from what `Menu` data is:
```ruby
menu = Menu.new("test_menu.csv")
order = Order.new(menu)

# expect that it should be an array containing the menu items

menu = Menu.new("wrong_menu.csv")
order = Order.new(menu)

# expect menu should default to an empty array, even if the input CSV data was wrong...
```

but obviously this depends on how you'd like your program to behave in case the input data is wrong — since that's not specified in the instructions, that's up to you!

spec/order_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    let(:menu) {[]}
+    it "should initialize with a menu array" do      
+      expect(subject.menu).to be_a Array 
+    end
+  end 
+
+  describe "#add_to_order" do
+    # is this test insane?
+    it "should add the quantity ordered to the correct dish" do
+      menu = [{:dish => "Burgers", :price => 10}]
```
leoht, line 16: The test is definitely sane, but ideally the instance of `Total` should be mocked/stubbed and injected into `Order` via the `.new` method — right now, there's a `Total.new` instantiated at line 20, but it's never used, since the instance used is the default one (the default argument from the `initialize` definition, because `Order.new` is not being called with it)

spec/order_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+      @order = Order.new(menu)
+      @order.add_to_order("Burgers", 5)
+      total = Total.new
+      expect(@order.total.selection).to eq([{:dish => "Burgers", :price => 10, :quantity => 5}]) 
+    end
+  end
+
+  describe "#request_bill" do
+    it "should request the bill" do # is testing this necessary?
+
```
leoht, line 27: Yep! Since it's a public method — it should return the total value as a number :)

spec/total_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'total'
+
+describe Total do
+  describe "#bill" do
+    # can't think of a better way to test this
+    it "should sum up the total cost of the order" do
+      @total = Total.new
+      allow(@total).to receive(:selection).and_return(
```
leoht, line 8: So here we want to test the behaviour of the class `Total` itself, so we don't want to mock it (it's not a dependency of our subject, but the test subject itself). Therefore the whole `allow(...)` stubbing is not needed, we should instead call methods or attributes of `Total` directly to setup the necessary state/context (= the selection hash)

lib/checkout.rb
```diff
@@ -0,0 +1,27 @@
...
+class Checkout
+  # possibly unfortunate name for the class, but I would have expected it to also handle
+  # the payment and not just send an SMS
+
+  def initialize
+    @time = Time.now + 3600
+  end
+
+  def send_notification
+    account_sid = ENV['SID'] # Your Test Account SID from www.twilio.com/console/settings
```
leoht, line 13: Nice use of ENV variables :) to make it even better, those values could be passed when initialising the `Checkout` instance

PR #2073, opened 2021-10-31
Take-Away challenge

JacobCH93:
So I got to practice my RSpec and coding in general, also leaned how to use dependency injection. I have also added a Twilio link. However, I feel this is partially incomplete as I am not happy with the end result.

### DIFF
--- a/lib/Menu.rb
+++ b/lib/Menu.rb
@@ -1,0 +1,19 @@
+require 'menu'
+
+class Menu
+  attr_reader :food_list, :order
+  def initialize
+    @order = []
+    @chicken = { 'chicken' => 15 }
+    @fish = { 'fish' => 10 }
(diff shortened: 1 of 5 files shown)


lib/Menu.rb
```diff
@@ -0,0 +1,19 @@
+require 'menu'
+
+class Menu
+  attr_reader :food_list, :order
+  def initialize
+    @order = []
+    @chicken = { 'chicken' => 15 }
+    @fish = { 'fish' => 10 }
```
doinyne, line 8: Im thinking but not 100% sure how to write. If this was a hash then maybe a block like |item, price| for line 12 on the order class would be able to iterate through the menu. again not 100% sure because thats where i stopped on my repo haha

PR #2074, opened 2021-10-31
takeaway-challenge Ryan Grimes pull request

RPGrimes:
All four user stories have been implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,11 @@
-Takeaway Challenge
+Takeaway Challenge :hamburger: :pizza: :fries:
 ==================
+Weekend challenge for weekend 2 of the Makers bootcamp. This task was completed independently.
+
+The challenge was to create a command line based application which allows a customer to order food from a menu. Upon completing the order a text message is sent to the customer to inform them their delivery will be with them 1 hour after ordering.
+
+This application was completed with TDD to 96% coverage.
+
 ```
                             _________
               r==           |       |
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -3,7 +3,13 @@ class Menu
   attr_reader :dishes
 
   def initialize
-    @dishes = {}
+    @dishes = {
+      pizza: 10,
+      burger: 7,
+      paella: 15,
+      carrot: 1,
+      cake: 3
+    }
   end
 
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,6 +1,7 @@
+
 class Menu
   
-  attr_reader :dishes, :selected_dishes
+  attr_reader :dishes
 
   def initialize
     @dishes = [
@@ -10,8 +11,6 @@ class Menu
       {name: 'Salad', price: 6},
       {name: 'Paella', price: 12}
     ]
-
-    @selected_dishes = [] #unsure of I'll keep this... 
   end
 
   def view_menu
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,15 +1,20 @@
 class Menu
   
-  attr_reader :dishes
+  attr_reader :dishes, :selected_dishes
 
   def initialize
-    @dishes = {
-      pizza: 10,
-      burger: 7,
-      paella: 15,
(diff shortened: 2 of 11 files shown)


README.md
```diff
@@ -1,5 +1,11 @@
-Takeaway Challenge
+Takeaway Challenge :hamburger: :pizza: :fries:
```
Cloudy1986, line 1: I think you could add some more information in your readme. For example, you could include:

-  an explanation of the approach you took for each user story
- an example feature test in irb for each user story
- any reflections from completing the challenge e.g What was difficult? What gaps in your learning did you notice?

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+
+class Menu
+  
+  attr_reader :dishes
+
+  def initialize
```
Cloudy1986, line 6: Nice approach for initialising a menu : )

lib/order.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'menu'
+require_relative 'text_message'
+
+class Order
+  attr_reader :ordered_items, :menu, :total, :text_message
+
+  def initialize
```
Cloudy1986, line 7: Again nice approach here for initialising an order : )

spec/menu_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'menu'
+
+describe Menu do
+  let(:dish) { { name:'Burger', price: 8 } }
```
Cloudy1986, line 4: I wonder if this could be a double rather than a variable?

spec/order_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'order'
+
+describe Order do
+  let(:menu) { Menu.new }
+  let(:dish) { { name:'Burger', price: 8 } }
```
Cloudy1986, line 5: I wonder if this could be a double rather than a variable?

Cloudy1986 (commented):
I've added some comments. Happy to discuss : )

README.md
RPGrimes: Totally agree that the README.md could use more detail. I'll update shortly

spec/menu_spec.rb
RPGrimes: Good question. When we review the code let's give it a go and see what happens!

spec/order_spec.rb
RPGrimes: Good question. When we review the code let's give it a go and see what happens!

spec/text_message_spec.rb
```diff
@@ -0,0 +1,6 @@
+require 'text_message'
+
+describe TextMessage do
```
eoinmakers, line 3: I wouldn't expect you to test an external framework, specifically, but maybe just to test your expectations of what you're getting from this framework. 

For example, if you're expecting a method called send_text in Order to call @text_message.send_message(total_cost), you would test this in order_spec.rb. 

eg. 
```
it 'sends a text to the user when they place their order' do
     expect(text_message).to receive(:send_message).with(1).argument
     order.send_text
end
```

It may seem quite simple, but it's testing the expected behaviour: this method is not responsible for the actual sending of the text, but it's responsible for _calling the method that does so._

There's one caveat to the above test being possible, and that caveat is that it requires us to be able to pass in a version of @text_message that we can set this 'expectation' on. Currently, your test would ignore our above 'expect', as it gets to line 11 of Order and it creates a new instance by running @text_message = TextMessage.now. 

We can get around this with dependency injection. This would mean refactoring your initialize method to look like this: 

```
def initialize(text_message = TextMessage.new)
     @ordered_items = []
     @menu = Menu.new
     @total = 0
     @text_message = text_message
end
```

This means that in your test, you would write it like this: 

```
it 'sends a text to the user when they place their order' do
     text_message = double
     allow(text_message).to receive(:send_message).with(1).argument

     order = Order.new(text_message)

     expect(text_message).to receive(:send_message).with(1).argument
     order.send_text
end
```

PR #2075, opened 2021-10-31
Week 2 Weekend Challenge

DoodleDeBug:
I attempted all 4 user stories below: 

```
As a customer

As a customer

As a customer

As a customer
```
I did not implement the last one using Twilio, rather I represented the text as a simple return message when the order method is called.


lib/takeaway.rb
```diff
@@ -0,0 +1,35 @@
+class Takeaway
+
+  attr_reader :menu, :my_order
+
```
sigii1992, line 4: I think the naming is really good since it is easy to understand what you are trying to implement with your code.
I would suggest to have different class for Menu in separate file since it would be easier to navigate if your Takeaway class would grow and have more functionality in the future. This keeps your project better oragnized and makes it easier for someone else to navigate your structure.
https://sourcemaking.com/refactoring/extract-class

PR #2076, opened 2021-10-31
Ginny - Takeaway challenge pull request

ginnyamazed:
getting approximately 85% test coverage.
4 user story tasks completed.
See notes at bottom of README file to see where I struggled and spent too much time.  It also shows desires if more time was available.
I spent approximately 2 hrs on Friday, 5 hrs on Saturday and approximately 6 hours Sunday.  
- If I had more time I would:
 ** work to reduce how many times the tests ask for select_items to make the RSpec easier to go through
 ** I would also separate out classes


lib/dish.rb
```diff
@@ -0,0 +1,14 @@
+class Dish
```
Zimmja, line 1: More a thought on naming classes... would this be better named as Menu? Feels like the function is about displaying several dishes, not just one dish

lib/order.rb
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  def select_items
+    puts "Please select your desired dishes based on Item number from the menu below"
+    puts "To proceed to your total, please press 'Enter' twice" "\n\n"
+    puts "------------Available Dishes------------"
+    @available_dishes.each do |menu|
+      puts "Item  #{menu[:item_number]}: #{menu[:dish]}  Price: £ #{menu[:price]}"
+      end
+      selection = STDIN.gets.chomp
```
Zimmja, line 37: Really nice to implement it with some actual input too, good job!

lib/order.rb
```diff
@@ -0,0 +1,116 @@
...
+    return order_sum
+  end
+
+  def order_options
+    print_menu
+    process(STDIN.gets.chomp)
+  end
+
+  def process(selection)
+    # 3. options to re-select items, submit order, or cancel
```
Zimmja, line 71: Very clear how this works, good job

lib/sms.rb
```diff
@@ -0,0 +1,20 @@
+class SMS
+
+  # Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+  require 'twilio-ruby'
+
+  def send_sms
```
Zimmja, line 6: Great use of environment variables

spec/sms_spec.rb
```diff
@@ -0,0 +1,8 @@
+require_relative '../lib/sms.rb'
+
+describe SMS do
```
Zimmja, line 3: Well done on testing your SMS, I never got round to that myself!

available_dishes.csv
```diff
@@ -0,0 +1,8 @@
+Tikka Masala,11.00,1
+Cheeseburger,11.50,2
+Hamburger,10.50,3
+Margarita Pizza,10.00,4
+Pepperoni Pizza,11.50,5
+Veggie Pizza,11.00,6
+Messy Fries,9.50,7
+Chips,4.50,8
```
eoinmakers, line 8: This approach feels a little overcomplicated - writing to / reading a CSV is perhaps a little beyond the scope of this challenge. It's encouraged to keep things as a simple as possible in your implementations, and you could define each of these dishes as an instance of Dish, for example, stored in an array in a 'Menu' object.

lib/dish.rb
eoinmakers: I agree with Will's feedback here! I would think of this object as more of a 'Menu' as its responsibility is handling multiple Dish items.

lib/order.rb
```diff
@@ -0,0 +1,116 @@
...
+      while !selection.empty? do
+        selection = selection.to_i
+        @selected_dishes << @available_dishes[selection-1]
+        selection = STDIN.gets.chomp
+      end
+
+  end
+
+  def order_sum
+    #TODO This is ugly and I would love to find another/better way to get the price to be total-able
```
eoinmakers, line 47: There is an example of a quite tidy approach here, that uses an array of items rather than a hash, that you might find interesting: https://github.com/makersacademy/takeaway-challenge/blob/main/docs/review.md#use-reduce-to-aggregate-over-a-collection-eg-when-calculating-the-total

This also ties into my previous comment about potentially storing Dishes / Items as an object instead.

lib/order.rb
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  def submit_order
+    puts 'Order has been submitted, you will receive a text confirmation shortly'
+    sms = SMS.new
+    sms.send_sms
+    exit
+  end
+
+  #And end process that puts all the pieces into running order would be a good implementation, however, for this
```
eoinmakers, line 100: It's best to remove any unfinished code / unused whitespace for tidiness. I primarily see notes for other devs used in the context of // TODO: ____________, rather than leaving unfinished methods in.

spec/sms_spec.rb
```diff
@@ -0,0 +1,8 @@
...
+  it 'sends a confirmation text to the user' do
```
eoinmakers, line 4: Useful reading around this kind of test here: https://github.com/makersacademy/takeaway-challenge/blob/main/docs/review.md#tests-should-test-real-behaviours-not-stubs

You want to test the behaviour of the method under test, rather than accidentally stubbing that behaviour, and testing the stub itself. 

To implement a stub for the messenger for your tests, you'll need dependency injection: https://github.com/makersacademy/takeaway-challenge/blob/main/docs/review.md#appropriate-use-of-dependency-injection

I'll share some more resources on this with you on Slack. :)

PR #2080, opened 2021-10-31
Takeaway Javascript PR

emanfolo:


README.md
```diff
@@ -1,83 +1,7 @@
...
-      '. '' .'    \:.....:--'.-'' .'
-       ':..:'                ':..:'
+Takeaway Challenge redone in Javascript. 
 
- ```
+Code is unfinished as I have not fully implemented the front end pages, I have yet to persist user log-in, and I still need to store completed orders in the orders DB. 
 
-Instructions
--------
+If you want to take a look at the site, run 'node server.js' in the terminal after installing all the dependencies in the package.json file.
```
t-kellett, line 5: When I follow these instructions, I get `error: role "emmanuel" does not exist` and I can't run the page

db/databaseconnection.js
```diff
@@ -0,0 +1,29 @@
+const { Client } = require('pg')
```
t-kellett, line 1: would be interested to know why the Client is in { } as this is not something I am familiar with

db/databaseconnection.js
```diff
@@ -0,0 +1,29 @@
...
+const client = new Client({
+  user: 'emmanuel',
```
t-kellett, line 3: I think this is where the issue is, on my computer there is no emmanuel - I think you would need to have an env variable here perhaps that takes the current user? I'm not sure though

src/order.test.js
```diff
@@ -0,0 +1,41 @@
...
+describe('order', ()=> {
+  describe('getTotal', ()=> {
+    it('returns a float', ()=>{
+      expect(typeof(order1.getTotal()) == 'number').toBe(true)
+    })
+    it('returns the total price of all dishes in the dishes array', ()=> {
+      order1.dishes = []
+      order1.addItem(friesDouble)
+      order1.addItem(burgerDouble)
+      expect(order1.getTotal()).toBe(1.99+3.99)
```
t-kellett, line 16: good use of testing behaviour over state

src/restaurant.js
```diff
@@ -0,0 +1,39 @@
...
+  listMenu() {
+    return this.menu.items
+  }
+
+  addItem(food) {
+    this.currentOrder.dishes.push(food)
+  }
+
+  completeOrder(){
+    if (this.currentOrder.dishes.length == 0)
```
t-kellett, line 25: does this need curly braces?

src/menu.js
```diff
@@ -0,0 +1,71 @@
+class Menu {
+  constructor(){
+    this.items = {
+      fries: {
+        name: 'Fries', price: 1.99, id: 1
```
t-kellett, line 5: If your table serialises the IDs, but you provide hard-coded IDs in the constructor - do you not run into issues with IDs already existing?

src/restaurant.test.js
```diff
@@ -0,0 +1,47 @@
...
+beforeEach(() => {
+  restaurant.clearOrder();
+});
+
+const restaurant = new Restaurant("Dishoom", menu, orderDouble)
+
+describe('restaurant', ()=> {
+  describe('listMenu', ()=> {
+    it('returns a hash object', ()=> {
+      expect(typeof(restaurant.listMenu()) == 'object').toBe(true)
```
t-kellett, line 25: this is an example of testing 'state' that I think is generally to be avoided. Worth asking coaches I think, but when I did something similar Leo mentioned that testing state rather than behaviour.

t-kellett (commented):
There's so much code here! Properly impressive to have got through all of this in a weekend. I think it would be even better if I could run it but there's clearly a lot of very nicely written code. Hard to give too much constructive criticism as it is still in progress, but the one about testing behaviour over state is probably an easy one to correct.

PR #2082, opened 2021-10-31
Takeaway Challenge for Code Review

katemyoung:


tatiananantes (commented):
I really like how you structure your readme files and I believe it will be a great asset during the interview stage as you document the challenges, learnings and improvements in such a clear way.

spec/dishes_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'dishes'
+describe Dishes do
+  let(:menu1) { Dishes.new ({ dish: 'Pizza', price: 5 }) }
```
JohnForster, line 3: This line indicates to me that maybe your Dishes class would be better named Menu?

README.md
```diff
@@ -47,37 +59,41 @@ So that I am reassured that my order will be delivered on time
...
+- Taking the time to feature test in irb before, during and after. 
 
-In code review we'll be hoping to see:
+## Questions
 
-* All tests passing
-* High [Test coverage](https://github.com/makersacademy/course/blob/main/pills/test_coverage.md) (>95% is good)
-* The code is elegant: every class has a clear responsibility, methods are short etc.
+- I'm not sure I have initialized the dishes class in the most efficient way. How to initalize a hash automatically?
+- I wonder if a separate Dish class would be more SRP.
```
JohnForster, line 91: It's ok to use a simple {:dish, :price} hash as a dish, for now.

README.md
```diff
@@ -47,37 +59,41 @@ So that I am reassured that my order will be delivered on time
...
+- I created an 'add' method which was not specified in the user stories, which I feel is not very TDD. Could I have mocked this instead for my test set up?
```
JohnForster, line 92: There's nothing wrong with the `add` method, and it does make your code more flexible, as your menu is not hard-coded. But you're right in that it wasn't specified in the user stories, so you are writing more code than is necessary. That's not necessarily a bad thing, if you're doing it to make your code more extendable in the future, but that is a decision that should be made purposefully.

JohnForster (commented):
A good start!

I've had a look at your domain model, and while it's very good, there are a couple of things missing. The story that says that the user wants to "select some dishes" has something implicit behind it,  which is that selected dishes will be added to the order. This probably means that you would need some method on your Order class to interface with the Dishes class so that they can be added to the order. 

 Good work on the testing, you are testing the behaviour of the methods rather than the state of the class!
 
 Your methods are clean and well named, this seems like a good beginning!

README.md
katemyoung: Thank you

spec/dishes_spec.rb
katemyoung: I was also thinking that after the code review, thanks for the confirmation.

README.md
katemyoung: Yes, and I feel the time I spend creating that method meant I didn't have the time to progress to other steps that exercised learning objectives making classes work together, which would have been useful.

PR #2088, opened 2021-11-01
Attempt at TakeAway Challenge

tatiananantes:
Completed Stories:
As a customer

As a customer

As a customer

TO COMPLETE:
As a customer
I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordere

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,20 +1,20 @@
 class Menu
-    attr_reader :dishes
+  attr_reader :dishes 
 
-    def initialize
-        @dishes = {
-          "calzone" => 5,
-          "pizza" => 8,
-          "mozzarella sticks" => 3,
-          "pasta" => 6,
-          "gelato" => 4
-        }
-    end
+  def initialize 
+    @dishes = {
+      "calzone" => 5,
+      "pizza" => 8,
+      "mozzarella sticks" => 3,
+      "pasta" => 6,
+      "gelato" => 4
+      }
+  end
     
-    def view_menu
-        @dishes.map { | dish, price |
-          "#{dish.capitalize}: £#{price}"
-        }.join(", ")
-    end
+  def view_menu
+    @dishes.map { | dish, price |
+      "#{dish.capitalize}: £#{price}" } 
+      .join(", ")
+  end
  
 end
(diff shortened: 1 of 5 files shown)


katemyoung (commented):
All the Rspec tests pass and there is 100% test coverage.

spec/menu_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      calzone: 5,
+      pizza: 8,
+      mozzarella: 3,
+      pasta: 6,
+      gelato: 4
+    }
+  end
+  
+  describe '#initialize' do
+    it 'has a list of dishes with its prices' do
```
katemyoung, line 16: Specs are well structured and easily readable.

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+require_relative '../lib/order'
+class Menu
+  attr_reader :dishes
+
+  def initialize
```
katemyoung, line 5: Code is clearly structured and easy to read. I was able to run it in irb.

spec/order_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+describe Order do
+  subject(:order) { described_class.new }
+  let(:basket) { { pasta: 6,  gelato: 4 } }
+  
+  # it { is_expected.to respond_to(:add_to_order).with(1).argument }
+  
+  describe "#add_to_order" do
+    it 'adds a dish to the order' do
+      order.add_to_order(:pasta, 6)
```
katemyoung, line 12: Suggestion: If the sample code for these examples do not need to be differnet, you could considered using a before block to minimise duplication of test set up code.

lib/order.rb
```diff
@@ -0,0 +1,26 @@
...
+class Order
+  attr_reader :basket, :menu, :dishes
+
+  def initialize
+    @basket = Hash.new(0)
+    @order_total = 0 
+    @menu = Menu.new
+  end
+
+  def add_to_order(dish, quantity)
```
katemyoung, line 12: Suggestion: It took me a couple of tries to realise I needed to pass the dish name as a symbol not a string in irb, so this could be something to look at. Maybe consider including a note or example in the README to specify how to order.

katemyoung (commented):
All necessary gems included in Gemfile.

PR #2089, opened 2021-11-01
Dean Lewis takeaway challenage

doinyne:


lib/menu.rb
```diff
@@ -0,0 +1,19 @@
+require_relative 'order'
+
+class Menu
+  attr_reader :food
+
+  def initialize
+    @food = [
+      { name: "Cod", price: 6 },
```
JacobCH93, line 8: Maybe make the foods different prices so it's easier to RSpec and see errors.

spec/order_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'order'
+
+describe Order do
+  it 'selects food from menu' do
+    order = Order.new
+    expect(order.current_selected_food).to eq nil
+  end
+  it 'puts food into ordered food array' do
+    order = Order.new
```
JacobCH93, line 9: Good RSpec, maybe also have a test where you check items are going into the array.

lib/menu.rb
```diff
@@ -2,18 +2,18 @@ class Menu
...
-    "Plaice": 6, 
-    "Chips": 2, 
-    "Sausage": 3 }
+    @menu = [
+      { Cod: 6,
+      Haddock: 6, 
+      Plaice: 6, 
+      Chips: 2, 
+      Sausage: 3 }
+    ]
```
ayseldilan, line 13: really good styling of the menu!

JacobCH93 (commented):
Good code, some solid ideas. Could maybe use a few more RSpec tests.

ayseldilan (commented):
Good start to the takeaway challenge. Would love to see the end result using twilio.

PR #2091, opened 2021-11-01
User story 1

Luke-The-Duke:
Could only manage the first user story, but will work on it more tonight.


smoni19 (commented):
Easily understandable variable & method names.
Operations are broken down logically so the code is easy to follow.

PR #2097, opened 2021-11-27
takeaway-challenge

sherwingp:


spec/order_spec.rb
```diff
@@ -7,7 +7,7 @@
   describe '#add' do
     it 'adds items to order' do
       subject.add(item)
-      expect(subject.items.include(item)).to be_truthy
+      expect(subject.items.include?(item)).to be_truthy
```
s-dousse: thoughtful and meticulous implementation of each feature!

lib/menu.rb
```diff
@@ -1,7 +1,7 @@
 class Menu 
   attr_reader :dishes
   def initialize
-    @dishes = {}
+    @dishes = []
```
s-dousse: How did you come to this conclusion?

spec/order_spec.rb
```diff
@@ -2,30 +2,34 @@
 
 describe Order do
   TEST_PRICE = 8
```
s-dousse: Nice, I now see why the Test price was at the beginning of the described block as it's being reused a couple of times!

lib/order.rb
```diff
@@ -13,6 +17,31 @@ def summary
...
+    message = @client.messages.create(   
+      body: message,
+      messaging_service_sid: 'MGc6ed9a1f11afc694b64b2ce6341d0019',      
+      to: ENV['SECRET_NUMBER'] 
+    ) 
+
+    puts message.sid
+  end
+
+  private
```
s-dousse, line 43: good use of private method!

lib/menu.rb
sherwingp: I initially used one big hash for the entire menu, but then it felt overly complicated having to first index the array by key, it also made adding menu items more difficult as I had to hard code menu item numbers. The array implementation makes it more scalable. If, for example, you wanted to make a method that adds or removes dishes to the menu, it is simple and easy to implement through array manipulation.

Using an array with hashes made more sense to me because I could use array indexes as menu numbers, filling it with hashes with key : value as dish : price.

lib/order.rb
```diff
@@ -23,7 +23,7 @@ def total
 
   def complete
     calculate_total
-    send_text("Thank you for your order: £#{@total}")
+    send_text("Thank you! Your order was placed and will be delivered before #{(Time.now + (60 * 60)).strftime("%k:%M")}.")
```
s-dousse, line 26: great that you had a step in between and added the time option once everything worked well together :)

PR #2099, opened 2021-11-28
Weekend Challenge pull request

chris-clement:


lib/text.rb
```diff
@@ -0,0 +1,26 @@
...
+      to: ENV['CUSTOMER_PHONE_NUMBER'],    # Replace with your phone number
+      from: ENV['TWILIO_PHONE_NUMBER'])  # Use this Magic Number for creating SMS
+    message.sid
+  end
+
+  private
+
+  def delivery_time
+    time = Time.now + 60 * 60
+    time.strftime("%H:%M")
```
FrancescoCollotto, line 24: looks cool! can you explain?

spec/order_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+    it 'has an empty order basket' do 
+      expect(subject.basket).to be_empty
+    end
+  end
+
+  describe '#order' do 
+    context 'item is on the menu' do 
+      it 'adds the item to the basket' do 
+        subject.order(good_item)
+        expect(subject.basket.find { |item| item[good_item] }).to_not be_nil
```
FrancescoCollotto, line 22: nice test!!

spec/order_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  describe '#confirm_order' do 
+    context 'basket is empty' do 
+      it 'raises an error' do 
+        expect { subject.confirm_order }.to raise_error 'Basket is empty, cannot confirm order'
+      end
+    end
+    context 'basket is not empty' do
+      it 'confirms order' do 
+        subject.order(good_item)
+        expect { subject.confirm_order }.not_to raise_error 'Basket is empty, cannot confirm order'
```
FrancescoCollotto, line 57: do you also need this? since you already check if an error is raised before?

planning_notes.md
```diff
@@ -0,0 +1,99 @@
+# User Story 1
```
FrancescoCollotto, line 1: nice detailed README description of your approach

FrancescoCollotto (commented):
Nice clean implementation!

lib/text.rb
chris-clement: Sure thing, it is a way of formatting the time instead of having it in the long default format which would include the day, month, year which i thought was unnecessary. strftime is a built in ruby method which formats the time into a string where if you pass the argument "%H:%M" it gives you a time like "14:43"

spec/order_spec.rb
chris-clement: good point! I think what I am trying to achieve here is:

- first test checks that if you confirm order when the basket is empty you get an error. 
- second test checks that if you confirm the order when the basket is not empty you dont also get an error. 

So I wanted to check that an error isnt always raised but I see you point that maybe it is too much.

chris-clement (commented):
Thank you for your review. Some answers to your questions have been added.

PR #2100, opened 2021-11-28
Take-away weekend challenge

FrancescoCollotto:
### TAKE_AWAY CHALLENGE

Approach:
Following the SRP principle I have implemented 3 different classes:

- SmsSender: to take care of API call to Twilio
- Menu: responsible for listing and selecting the available dishes
- TakeAway: the "main" class to add a dish to the cart, listing the menu, showing the total price and place an order. all of this making use of the 2 previous classes through class injection


lib/take_away.rb
```diff
@@ -0,0 +1,48 @@
...
+    @menu.list_dishes
+  end
+
+  def add_to_cart(name, quantity = 1)
+    new_dish = @menu.select_dish(name, quantity)
+    update_dish_quantity(new_dish) || @cart << new_dish
+  end
+
+  def update_dish_quantity(new_dish)
+    edit = false
```
chris-clement, line 23: I'm not clear on what this edit variable is for? Could a more descriptive word be used perhaps?

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
+class Menu
+  def initialize
+    @dishes = [
```
chris-clement, line 3: I wonder if storing the prices as floats rather than strings would be preferable. I can see below you then have to convert these into numbers to get the total cost of the order. If this program were to get more complicated it might be more flexible if these were stored as floats.

lib/send_sms.rb
```diff
@@ -0,0 +1,24 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+require 'dotenv/load'
+
+class SmsSender
+  ACCOUNT_SID = ENV['TWILIO_ACCOUNT_SID']
```
chris-clement, line 6: Good use of environment variables. Clearly named so I understand what they are.

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      expect(pizza[:quantity]).to eq 1
+    end
+
+    it "raise an error if dish not in the menu" do
+      expect { menu.select_dish("pineapple") }.to raise_error("dish not available")
+    end
+
+  end
+
+end
```
chris-clement, line 26: Test are nice and clear. I can easily follow what is being tested.

spec/take_away_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      menu = double("menu", select_dish: dish)
+      sms = double("sms", send_sms: double("message", body: "Thank you! Your order was placed"))
+      take_away = TakeAway.new(menu: menu, sender: sms)
+      take_away.add_to_cart(double())
+      expect(take_away.place_order).to eq("Thank you! Your order was placed")
+    end
+
+  end
+
+end
```
chris-clement, line 68: I think these tests would benefit with some 'before' statements to keep you code DRY. The set up on some of the test are doing the same thing e.g. setting the double dish = double("dish") and menu = double("menu").

lib/send_sms.rb
```diff
@@ -0,0 +1,24 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
```
chris-clement, line 1: Is there no send_sms_spec.rb file?

lib/take_away.rb
```diff
@@ -0,0 +1,48 @@
...
+    @menu = menu
+    @sender = sender
+    @cart = []
+  end
+
+  def menu
+    @menu.list_dishes
+  end
+
+  def add_to_cart(name, quantity = 1)
```
chris-clement, line 17: Good use of single responsibility principle. You are splitting out each component into its own method.

lib/take_away.rb
```diff
@@ -0,0 +1,48 @@
...
+      if item[:dish] == new_dish[:dish]
+        item[:quantity] += new_dish[:quantity]
+        edit = true
+      end
+    end
+    edit
+  end
+
+  def total_price
+    total = @cart.reduce(0) do |memo, item|
```
chris-clement, line 34: what does memo represent here?

spec/menu_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'menu'
+
+describe Menu do
+  let(:menu) { Menu.new }
```
chris-clement, line 4: I think rather than this you could also use 'subject' instead of 'menu' and remove this line.

Up to you what you prefer :)

chris-clement (commented):
Good work! Some comments have been added for you to look at

PR #2102, opened 2021-11-28
Takeaway challenge

EviePalaiochorinou:
Only managed to complete the first two user stories

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,30 @@
+class Menu
+
+    attr_reader :dishes
+    def initialize(dishes)
+        @dishes = []
+        dishes.each do |dish|
+            @dishes << dish
+        end
+        # @dishes = dishes
+    end
+
+    def list
+        menu = ""
+        @dishes.each do |dish|
+            line = "#{dish.name}: #{dish.price}\n"
+            menu += line
+        end
+        return menu.strip()
+    end
+
+    def exists(other_dish)
+        @dishes.each do |dish|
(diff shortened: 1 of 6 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,30 @@
...
+    attr_reader :dishes
+    def initialize(dishes)
+        @dishes = []
+        dishes.each do |dish|
+            @dishes << dish
+        end
+        # @dishes = dishes
+    end
+
+    def list
```
giorgigutsaevi, line 12: Really thoughtful, that you have a #list method on the Menu object!

lib/order.rb
```diff
@@ -0,0 +1,13 @@
+class Order
+
+    attr_reader :menu, :myorder
+    def initialize(menu)
+        @menu = menu
+        @myorder = {}
+    end
+
+    def add_to_order(dish, amount)
+        raise "Dish not in menu" if !@menu.exists(dish)
```
giorgigutsaevi, line 10: well done on using a guard condition to raise order if dish does NOT exist!

spec/order_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'order'
+require 'menu'
+
+describe Order do
+
+    before do
```
giorgigutsaevi, line 6: Using before really shows what you are trying to achieve. Well done!

giorgigutsaevi (commented):
review for the takeaway challenge

PR #2103, opened 2021-11-28
pull request for code review

Alaka-K:


lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
+class Takeaway
+  attr_reader :menu, :orders
+
+  def initialize
+    @menu = { rice: 3, chicken: 2, beans: 3, yam: 2, fish: 4, noodles: 2 }
+    @orders = []
+    @sum = 0
+  end
+
+  def add_orders(choice)
```
msc49, line 10: nice!

msc49, line 10: also maybe a fail method if the choice is not in your menu then it will raise an error before pushing it to orders?

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
...
+    @orders.push(choice)  
+  end
+
+  def print_menu
+    @menu.each { |key,value| puts "#{key} => £#{value}" }
+  end
+
+  def print_orders
```
msc49, line 18: nice!

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
...
+        end
+      end
+    end
+  end
+
+  def orders_total
+    @orders.each { |choice| @sum += @menu[choice.to_sym].to_i }
+    return "£#{@sum}"
+  end 
+end
```
msc49, line 32: next time a module/class on Twilio to send order messages may be useful. It was difficult to set up but the documentation has a step by step guide :)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  
+  describe "#print_menu" do
+    it "displays menu" do
+      list_of_dishes = subject.print_menu
+      expect(subject.print_menu).to eq(list_of_dishes)
+    end
+  end
+
+  describe "#add_orders" do
+    it { is_expected.to respond_to(:add_orders).with(1).argument }
```
msc49, line 27: i like the way you can use rspec syntax in the right places

lib/takeaway.rb
```diff
@@ -0,0 +1,32 @@
+class Takeaway
```
leoht, line 1: This class `Takeaway` could probably benefit to be split into smaller classes (at least two) to respect what's called separation of concerns — right know this class is doing different things:
 * it holds the menu (with the instance variable `@menu`)
 * it can print the menu
 * it holds the order and can add something to it (`@orders` and method `add_orders`)
 * it can print the order content (and the total)

The easiest would be to start by extracting everything related to the menu (so the 2 first points) into a dedicated class (maybe called `Menu`), and this class would be responsible for storing the menu and printing it.

The `Takeaway` class would then become the "main" class of our program, and would hold an instance variable `@menu` that is an instance of the menu class — which means the initialize method would then look like this:

```ruby
def initialize
    @menu = Menu.new
    @orders = []
    @sum = 0
end
```

... then, when we need to access the menu (for example, in the method `print_orders`), we could use a method `menu_items` of the `Menu` class:

```ruby
def print_orders
  # @menu is not a hash anymore, it's an instance of the new Menu class
  @menu.menu_items.each do |key, value|
    # etc...
  end
end
```

If you're not sure where to start, try these steps:
  * create a new `Menu` class in another file
  * start copying the methods related to menu management into this class
  * remove them from the `Takeaway` class
  * make the necessary changes so `Takeaway` now *relies on the new `Menu` methods *

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'takeaway'
+describe Takeaway do
+  subject { Takeaway.new }
+  let(:dish) { double("Chicken") }
```
leoht, line 4: So it's not necessary to use doubles for simple strings (like `"Chicken"`), because we use doubles for objects on which we want to "mock" or "fake" method calls — in this file, you could simply replace the double by a plain string value:
```ruby
let(:dish) { "Chicken" }
```

For example, let's say you've extracted the `Menu` class (as explained in my other comment above) — you would need a new tests file for this new class (likely named `menu_spec.rb`), and then in the `Takeaway` tests, you would need to use a double to remove the direct dependency that `Takeaway` has on `Menu`. This would be a step further from extracting the `Menu` class, so work on this only once the Menu class has been extracted correctly :)

leoht, line 4: So it's not necessary to use doubles for simple strings (like `"Chicken"`), because we use doubles for objects on which we want to "mock" or "fake" method calls — in this file, you could simply replace the double by a plain string value:
```ruby
let(:dish) { "Chicken" }
```

For example, let's say you've extracted the `Menu` class (as explained in my other comment above) — you would need a new tests file for this new class (likely named `menu_spec.rb`), and then in the `Takeaway` tests, you would need to use a double to remove the direct dependency that `Takeaway` has on `Menu`. This would be a step further from extracting the `Menu` class, so work on this only once the Menu class has been extracted correctly :)

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:price) { double("£4") }
+  let(:choice) { double(:choice) } 
+
+  describe "#initialize" do
+    it "menu array" do
+      expect(subject.menu).to be_truthy
```
leoht, line 10: This test could probably be removed — you would use `be_truthy` to verify a value is "true" (if it'd be a boolean, in something like `expect(user.allowed?).to be_truthy`. Since `subject.menu` returns a hash, it's probably not the right RSpec matcher to express that `subject.menu` should be a hash

Maybe something better would be:
```ruby
expect(subject.menu).to be_a(Hash)
```

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    it "empty orders array" do
+      expect(subject.orders).to eq []
+    end
+ 
+  end
+  
+  describe "#print_menu" do
+    it "displays menu" do
+      list_of_dishes = subject.print_menu
```
leoht, line 21: This test is "circular", so it seems it's not actually testing anything — it would be the equivalent of writing something like:
```ruby
my_number = 10 
expect(my_number).to eq(10)
```

What you're looking for is to compare the result of `subject.print_menu` to an *expected* list of dishes, which it should be equal to:

```ruby
expected_list_of_dishes = { rice: 3, chicken: 2, beans: 3, yam: 2, fish: 4, noodles: 2 }
expect(subject.print_menu).to eq(expected_list_of_dishes)
```

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    it { is_expected.to respond_to(:add_orders).with(1).argument } 
+  end
+
+  it "add choice to orders array" do
+    subject.add_orders("chicken")
+    expect(subject.orders).to include("chicken")
+  end
+
+  it "sum of the order" do
+    subject.add_orders("rice")
```
leoht, line 36: This test is nicely laid out and follows well the "arrange, act, assert" steps :)

leoht, line 36: This test is nicely laid out and follows well the "arrange, act, assert" steps :)

PR #2104, opened 2021-11-28
TakeAway - Challenge

s-dousse:


README.md
```diff
@@ -81,3 +81,296 @@ Notes on Test Coverage
...
+So that I can verify that my order is correct
+I would like to check that the total I have been given matches the sum of the various dishes in my order
+
+As a customer
+So that I am reassured that my order will be delivered on time
+I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered
+```
+
+#Feature test in irb
+works well until I want to check out:
```
sherwingp, line 117: It is good that you are documenting the user experience and any issues in detail

lib/checkout.rb
```diff
@@ -0,0 +1,28 @@
+require_relative 'menu'
```
sherwingp, line 1: This class is implemented nicely, methods are short and it is clear what each one does and is responsible for

lib/menu.rb
```diff
@@ -0,0 +1,60 @@
+
+class Menu
+
+  attr_reader :basket, :receipt_list, :checkout
+
+  MEALS_LIST =
```
sherwingp, line 6: I think an array of hashes might have been a more suitable data structure, because instead of hardcoding menu numbers as keys you can take advantage of the array object's built in indexing

lib/menu.rb
```diff
@@ -0,0 +1,60 @@
...
+    @basket = []
+    @receipt_list = []
+  end
+
+  def print_menu
+    print MEALS_LIST
+  end
+
+  
+  def choose_meal(meal_number) # TODO : Has too many lines && NEEDS TESTING
```
sherwingp, line 32: As you know there are too many lines, maybe using a switch statement for the various states of @meal_number could make it more readable

lib/twilio_sms.rb
```diff
@@ -0,0 +1,44 @@
+# Download the helper library from https://www.twilio.com/docs/ruby/install
+require 'rubygems'
```
sherwingp, line 2: Good idea to extract all of the code for sending texts into a separate file

spec/checkout_spec.rb
```diff
@@ -0,0 +1,69 @@
+require 'checkout'
+#require 'menu'
+
+describe Checkout do
+   
+  describe "test view_basket" do
+    context "with 1 item in the basket" do
+      before do
+        # instantiate checkout with fake variable to not use the instance variables of the Menu class
+        # does it need to be a double as it can be passed directly? no need to mock a behaviour
```
sherwingp, line 10: I personally think it doesn't need to be a double because doing it this way still allows the test to be independent

README.md
```diff
@@ -81,3 +81,296 @@ Notes on Test Coverage
...
+
+```
+Swa@Swas-MacBook-Pro takeaway-challenge % irb -r ./lib/menu.rb
+3.0.2 :001 > menu = Menu.new
+       Welcome, this is our menu        
+Please enter the number of the meal you would like to add to the basket
+___________________________________________________________
+
+        1 to select Kimbab : £8
```
sherwingp, line 126: Very tasty menu

spec/menu_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it "raises an error if user chooses as number that doesn't correspond to an option" do
+      expect { menu.choose_meal(0 - Menu::MEALS_LIST.size) }.to raise_error "sorry we don't have this item"
+      expect { menu.choose_meal(Menu::MEALS_LIST.size + 2) }.to raise_error "sorry we don't have this item"
+    end
+
+    it "raises an erro if the user tries to checkout with nothing in his basket" do
+      menu = Menu.new #empty basket
+      expect { menu.choose_meal(Menu::MEALS_LIST.keys[-1]) }.to raise_error "sorry we don't have this item"
```
sherwingp, line 65: This should be checking if "sorry but your basket is empty" was raised

PR #2105, opened 2021-11-28
Takeaway challenge

TomasGarciaDev:
All the User stories done. Still, work needs to be done to protect sensitive data.


lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def check_order
+    @total = 0
+    @order.each do |item|
+      item.each do |item, price|
+        puts "#{item}: £#{price}"
+        @total += price
+      end
+    end
+    puts "Total: £#{@total}"
```
Hereloss, line 26: This method is fairly long - it might be worth splitting this and putting the total calculation in a new method

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
+require_relative 'order'
+
+class Menu
```
Hereloss, line 3: I like how this has it's own class - good idea!

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def submit_and_confirm
+    raise "Your order is empty" if @order.empty?
+    time = Time.new
+
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
+    client = Twilio::REST::Client.new(account_sid, auth_token)
```
Hereloss, line 35: Good to keep as environmental variables - keeps them secure!

spec/order_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      order = Order.new
+      expect { order.submit_and_confirm }.to raise_error("Your order is empty")
+    end
+
+  xit "send a message when order is confirmed" do
+      twilio_client = instance_double(Twilio::REST::Client)
+      allow(Twilio::REST::Client).to receive(:new).and_return twilio_client
+      fake_messages = double('Messages')
+      allow(twilio_client).to receive(:messages).and_return fake_messages
+      expect(fake_messages).to receive(:create).with(from: '+16516152835', to: '+447413342817', body: "Thank you! Your order was placed and will be delivered before #{(time + 3600).strftime('%H:%M')}")
```
Hereloss, line 48: If this double was done as a dependency injection instead (so the client is created in initialise), it might be slightly easier to mock this!

spec/order_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  describe "#submit_and_confirm" do
+    it "should respond to check_order" do
+      expect(Order.new).to respond_to(:submit_and_confirm)
+    end
+
+    it "should raise and error if no itmes has been addded to the order" do
+      order = Order.new
+      expect { order.submit_and_confirm }.to raise_error("Your order is empty")
+    end
```
Hereloss, line 41: Good to add edge case of empty order!

lib/menu.rb
```diff
@@ -0,0 +1,25 @@
...
+    ]
+  end
+
+  def print_menu
+    counter = 1 
+    @menu.each do |pizza|
+      pizza.each do |key, value| 
+        puts "#{counter}. #{key}: £#{value}"
+        counter += 1
+      end
```
Hereloss, line 22: Nice user friendly way to print the order out!

lib/order.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def initialize
+    @menu = Menu.new
+    # @menu.print_menu
+    @order = []
+  end
+
+  def add_items(item)
+    @order << @menu.menu[item - 1]
+  end
```
Hereloss, line 16: Clear way to add items with numbers - but what if they put the name instead of a number?

spec/menu_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+describe Menu do
+  it "should be able to create and instance of Menu" do
+    expect(Menu).to respond_to(:new)
+  end
+
+  subject(:menu) { described_class.new }
+  let(:print_menu) { "Margherita Pizza: £8" }
+  it "can print menu" do
+    expect(menu.print_menu).to eq(menu.menu)
+  end
```
Hereloss, line 12: I think this covers menu 100% - good job!

lib/order.rb
TomasGarciaDev: Great suggestion, you are right. It is quite long.

spec/order_spec.rb
TomasGarciaDev: Yeah, so that on yours. I completely missed that.

spec/order_spec.rb
TomasGarciaDev: Man, doubles and mocking are still getting me very confused.

PR #2106, opened 2021-11-28
week 2 takeaway challenge

sabrinalord:


spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+subject(:menu) { Menu.new(menu_items) }
+    let(:menu_items) do
+    {
+        pizza: 8, 
+        ramen: 6, 
+    }
+    end
+
+    it 'stores a list of dishes and their prices' do
+    expect(menu.menu_items).to eq(menu_items)
```
j-cort, line 13: clear and accurate unit tests!

spec/order_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'order'
+require 'takeaway'
+
+describe Order do
+    
+    # it 'stores my selected dish' do
```
j-cort, line 6: Seems like a good test. Why comment it out?

lib/takeaway.rb
```diff
@@ -0,0 +1,22 @@
...
+      @menu.print
+    end
+
+    def add_to_order(dish, quantity)
+      @order.my_dishes << {dish => quantity}
+      puts "#{quantity} #{dish} added to your order!"
+    end 
+
+end 
+
```
j-cort, line 22: I like that you used many different classes/module to separate the responsibilities of the program. I think this will make it easier to expand the functionality of the program in the future.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  describe '#add_to_order' do
+    let(:order) { double(Order.new)}
+    let(:dish) { "pizza" }
+    let(:quantity) {3}
+        it 'adds the dish and its quantity into my order' do 
+        subject.add_to_order(dish, quantity)
+        expect(subject.order.my_dishes).to include({dish => quantity })
+        end
+  end  
+end
```
j-cort, line 25: Great tests! How about also testing for any edge cases, e.g. if an item isn't on the menu., but a user tries to add it.

PR #2107, opened 2021-11-29
Weekend challenge 2 

tc1316:


spec/menu_spec.rb
```diff
@@ -0,0 +1,45 @@
+require 'menu'
+
+describe Menu do
+  let(:dish_price) { rand(1..5) }
+  let(:quantity) { rand(1..10) }
+  let(:dish_total) { dish_price * quantity }
+  let(:dish) { double("dish", :read_name => "dish", :read_price => dish_price, :amount => quantity, :dish_total => dish_total) }
+  let(:dish2) { double("dish2", :read_name => "dish2", :read_price => dish_price, :amount => quantity, :dish_total => dish_total) }
```
jessgordon, line 8: This is fantastic! Such a good use of let and doubles to test with flexibility rather than hardcoded dishes and prices.  I've learnt something here already! Nice one.

lib/dish.rb
```diff
@@ -0,0 +1,28 @@
+class Dish
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+  end
+
+  def read_name
+    @name.dup
```
jessgordon, line 9: I know you will have a good reason for using dup - if you have time would you be able to respond with your rationale? I'd love to learn why it's better than returning the original instance variable.

jessgordon (commented):
Your spec files read really clearly, fantastic use of doubles especially keeping things flexible using rand(). I'm sorry I haven't been able to spot any areas for improvement!

The only thing I think that could be useful may be a class diagram so present how the classes interact with each other for other devs to quickly become acquainted with your code.

PR #2108, opened 2021-11-29
Weekend challenge 2 - Chris Thomas

Hereloss:


lib/takeaway.rb
```diff
@@ -1,13 +1,16 @@
...
+require_relative 'twilioruby'
+require 'env'
 
 class Takeaway
 
   attr_reader :menu, :my_choices
 
   def initialize
     @menu_object = []
-    @menu = {"Bryani" => 7, "Curry" => 9, "Beef" => 10, "Shrimp" => 11, "Korma" => 4}
```
TomasGarciaDev: Maybe could be done as an array of hashes instead of just a hash. But this is just a suggestion.

lib/takeaway.rb
```diff
@@ -23,12 +26,12 @@ def see_menu
...
 
   def choose_dish(name)
-    unless @menu[name].nil?
-      puts "You've chosen #{name}!"
-      @my_choices << name
-    else
+    if @menu[name].nil?
       puts "That's not a menu option! We don't sell that"
       @my_choices
+    else
```
TomasGarciaDev, line 32: So, in the case of using an array of hashes you can 'push' the menu options as a number. But, again, just a different approach to the same result.

lib/takeaway.rb
```diff
@@ -50,6 +53,7 @@ def make_order
     return @final_order
```
TomasGarciaDev, line 53: There is no need for the return here, I think.

spec/twilioruby_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'twilioruby'
```
TomasGarciaDev, line 1: Amazing that you managed to write rspec for the 'text sms'. I didn't know how where to start with this one.

lib/takeaway.rb
```diff
@@ -23,12 +26,12 @@ def see_menu
   end
 
   def choose_dish(name)
-    unless @menu[name].nil?
-      puts "You've chosen #{name}!"
-      @my_choices << name
-    else
+    if @menu[name].nil?
       puts "That's not a menu option! We don't sell that"
```
TomasGarciaDev, line 30: Great that you looked for edge cases.

spec/twilioruby_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    
+    allow(twilio_client_class).to receive(:new).and_return(client)
+    allow(client).to receive(:messages).and_return(message_ready)
+    allow(message_ready).to receive(:create).and_return([to,body])
+
+    twilio_ruby = SendMeAText.new(twilio_client_class,"1234","1244")
+
+    expect(twilio_ruby.send_text("20 burgers")).to eq([to,body])
+  end
+end
```
TomasGarciaDev, line 21: Great job Chris

PR #2112, opened 2021-11-29
Riky - takeaway challenge

Riky5:
Unfortunatly I wasn't feeling well most of the weekend so I couldn't add more tests as intended to. Coverage: 96%


lib/dishes.rb
```diff
@@ -0,0 +1,20 @@
+class Dishes
+  attr_reader :dishes
+
+  def initialize
```
Fadaka, line 4: good use of initialized to set menu dishes

PR #2113, opened 2021-11-29
Jess Takeaway Challenge

jessgordon:


lib/order.rb
```diff
@@ -0,0 +1,72 @@
...
+    @summary = {}
+    @menu = Menu.new(restaurant)
+  end
+
+  def print_menu
+    puts @menu.list
+  end
+
+  def add(dish, quantity)
+    error_check_add_arguments(dish, quantity)
```
tc1316, line 19: Nice use of type checking

lib/order.rb
```diff
@@ -0,0 +1,72 @@
...
+    price = @menu.dishes[dish].to_f * quantity
+    add_to_summary(dish, quantity, price)
```
tc1316, line 21: Adding quantity and total price here makes a lot of sense (I didnt)

lib/menu.rb
```diff
@@ -0,0 +1,48 @@
+class Menu
+  attr_reader :dishes, :restaurant
+
+  @@restaurants = {
```
tc1316, line 4: Not entirely sure what the @@ is for

lib/order.rb
```diff
@@ -0,0 +1,72 @@
...
+    if @summary.has_key?(dish)
+      @summary[dish][:quantity] += quantity
+      @summary[dish][:price] += price
+    else
+      @summary[dish] = { quantity: quantity, price: price }
+    end
+  end
+
+  def calculate_total
+    @summary.map { |_, hash| hash[:price] }.sum
```
tc1316, line 54: Nice way of doing the sum in 1 line

lib/order.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  def send_text(number)
```
tc1316, line 57: Suggestion: Could put the text sending in its own class, as it draws upon a new Twilio class

PR #2114, opened 2021-11-29
Takeaway Challenge

christopherxwebb:
The project is currently incomplete.

There main areas to address are:

1. TESTING. So far, I've only used a few basic tests. I recognise that I did not use a TDD approach: I found it easier to write the code instead.
I will go back and write the tests accordingly.

2. CLASSES. I began with creating the 'menu.rb', which is a mistake I intend to rectify in order to adhere to the principle that dictates classes should be split up.
I intend to separate menu.rb from takeaway.rb (in which the actual ordering will take place) and the send_sms.rb. I would also like to make the takeaway.rb more interactive (use gets.chomp etc.).

3. SEND_SMS.RB. I need to implement this properly. Right now, whenever IRB is booted up, it sends a text. I need to create a method and link it to the takeaway.rb.

4. PRIVACY. I need to set up environment variables that will protect my Twilio credentials.


README.md
```diff
@@ -1,3 +1,20 @@
...
+1. TESTING. So far, I've only used a few basic tests. I recognise that I did not use a TDD approach: I found it easier to write the code instead.
+I will go back and write the tests accordingly.
+
+2. CLASSES. I began with creating the 'menu.rb', which is a mistake I intend to rectify in order to adhere to the principle that dictates classes should be split up.
+I intend to separate menu.rb from takeaway.rb (in which the actual ordering will take place) and the send_sms.rb. I would also like to make the takeaway.rb more interactive (use gets.chomp etc.).
+
+3. SEND_SMS.RB. I need to implement this properly. Right now, whenever IRB is booted up, it sends a text. I need to create a method and link it to the takeaway.rb.
+
+4. PRIVACY. I need to set up environment variables that will protect my Twilio credentials.
+
```
AlfonsoGhislieri, line 14: This sounds great! You seem to have a very clear idea of what still has to be implemented.

lib/menu.rb
```diff
@@ -0,0 +1,45 @@
...
+            "green_salad" => 9,
+            "rump_steak" => 20,
+            "tiramisu" => 25
+        }
+        @selection_of_dishes = []
+        #@texter = SendText.new
+    end
+
+    def show_menu
+        puts @menu
```
AlfonsoGhislieri, line 21: You could get rid of the attr_reader for menu since you have this method that allows the user to see it

lib/menu.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    def order(dish, quantity = 1) #
+        # For more interactivity, I could:
+        # puts "Which dish(es) would you like to order?"
+        # @selection_of_dishes = gets.chomp
+        puts "#{quantity} x #{dish}(s) added to order."
+        @selection_of_dishes << @menu.fetch(dish) * quantity
+    end
```
AlfonsoGhislieri, line 30: Good implementation, I would consider also maybe checking that the dish would be a valid dish?

lib/send_sms.rb
```diff
@@ -0,0 +1,15 @@
...
+client = Twilio::REST::Client.new(account_sid, auth_token)
+
+from = '+447782338793'
+to = '+44' 
+
+client.messages.create(
+from: from,
+to: to,
+body: "Thank you! Your order was placed and will be delivered before #{Time.now + 1*60*60}!"
+)
```
AlfonsoGhislieri, line 15: You can use the dotenv gem to take care of saving the confidential data

lib/menu.rb
```diff
@@ -0,0 +1,45 @@
...
+    def complete_order
+        puts "Thank you for your order. You will receive message confirmation shortly."
+        # return @send_message
+    end
+
+    # def send_sms(texter = @texter)
+    #     texter.send_sms
+    #   end
+
+end
```
AlfonsoGhislieri, line 45: You can remove puts from most of these, Ruby will automatically return the last line of code that a method runs.

lib/takeaway.rb
```diff
@@ -0,0 +1,16 @@
...
+#         "oysters" => 14,
+#         "fries" => 4,
+#         "green_salad" => 9,
+#         "rump_steak" => 20,
+#         "tiramisu" => 25
+#     }
+    
+#     @selection_of_dishes = []
+
+# end
```
AlfonsoGhislieri, line 16: Good idea to keep the menu as a separate class, this would definitely allow for more functionality than hard coding it in the initialize method

lib/menu.rb
christopherxwebb: Yeah, a really good point. I was aware that this relied heavily on good spelling, etc., on behalf of the user!

lib/takeaway.rb
christopherxwebb: Yeah, you're totally right. Thanks for all these super useful comments!

PR #2115, opened 2021-11-29
Takeaway Challenge

msc49:


lib/menu.rb
```diff
@@ -0,0 +1,45 @@
+class Menu
```
Alaka-K, line 1: your code is very readable.

Alaka-K: you work is extremely thorough. A lot of which you had implemented, I wouldn't have thought of. you must have a very good process in breaking the user stories. well done!

lib/menu.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  attr_reader :menu_hash
+  def initialize 
+    @menu_hash = {}
+  end
+
+  def add_to_menu(dish, price)
```
Alaka-K, line 8: good idea to have a fail clause! Didn't think of that

lib/menu.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def cost_specific_dish(dish)
+    @menu_hash.each do |food, price|
+      if dish.to_sym == food
+        puts "£#{price}"
+      end
+    end    
+  end
+
+  def generate_example_menu
```
Alaka-K, line 36: I was wondering why you didn't have a preset menu?

lib/menu.rb
```diff
@@ -0,0 +1,45 @@
...
+    @menu_hash.each do |food, _price|
+      if dish.to_sym == food
+        @menu_hash.delete(dish.to_sym)
+      end
+    end    
+
+  end
+
+  def show_menu
+    @menu_hash.each_with_index do |(dish, price), index|
```
Alaka-K, line 23: good use of each_with_index method!

lib/menu.rb
Alaka-K, line 36: I assume that this will be your present menu

Alaka-K (commented):
you did a great job!

PR #2116, opened 2021-11-29
Finished take away challenge

bryn9696:


lib/takeaway.rb
```diff
@@ -0,0 +1,58 @@
+require_relative 'dish.rb'
+require_relative 'text.rb'
+require 'twilio-ruby'
+
```
Inimesh, line 4: Overall code is very readable and concise! :)

lib/takeaway.rb
```diff
@@ -0,0 +1,58 @@
...
+class Takeaway
```
Inimesh, line 5: May be an idea to split Takeaway class into separate classes to have single responsibility

lib/takeaway.rb
```diff
@@ -0,0 +1,58 @@
+require_relative 'dish.rb'
+require_relative 'text.rb'
+require 'twilio-ruby'
+
+class Takeaway 
+  attr_reader :menu, :final_order, :texter
+  def initialize
```
Inimesh, line 7: Could be an idea to have default values passed to initialize as arguments?

lib/takeaway.rb
```diff
@@ -0,0 +1,58 @@
...
+    @texter = Text.new
+    @menu = {
+      "dish 1" => 5,
+      "dish 2" => 7,
+      "dish 3" => 9}
+    @final_order = []
+  end
+
+  def menu
```
Inimesh, line 16: A good, concise and readable method but could maybe have a more descriptive name seeing as it displays data; you also have an @menu attr_reader which might be confusing when calling Takeaway.menu

lib/takeaway.rb
```diff
@@ -0,0 +1,58 @@
...
+    @menu.each do |dish, price|
+      puts "#{dish} -- £#{price}"
+    end
+  end
+
+  def add_dish(name, price)
```
Inimesh, line 22: Method definitely would work in production but dependency injection might be a better way to implement the creating of a class instance when using rspec tests?

lib/takeaway.rb
```diff
@@ -0,0 +1,58 @@
...
+    dish = Dish.new(name, price)
+    @menu[dish.name] = dish.price
+  end
+
+  def selected_dishes(*order)
```
Inimesh, line 27: I like the use of '*args' with "splat" operator, keeps the method flexible and open for expansion!

lib/text.rb
```diff
@@ -0,0 +1,17 @@
+require 'twilio-ruby'
+
+class Text
+  def initialize
+    account_sid = 'AC5e4e1ddec6330a437df6f9f1345820a4'
```
Inimesh, line 5: Will want to be careful as I can see your unique codes on a public platform! May want to keep these hidden using environment variables!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'takeaway'
+require 'twilio-ruby'
+
+
+describe Takeaway do
```
Inimesh, line 5: Tests look good, but could add some `before do .. end` blocks to set up some of the tests to adhere to DRY principle and might be an idea to have the name of the method being tested in the `it` block description or nested `describe`/`context` blocks.

PR #2118, opened 2021-11-29
Take away 

Fadaka:
weekend challenge


Riky5 (commented):
Overall it does have most functionalities requested. It's missing the confirmation text message when an order is placed and maybe just a few more tests.

PR #2121, opened 2022-01-03
takeaway-challenge 

mikitaro-27:


lib/dish.rb
```diff
@@ -0,0 +1,8 @@
+class Dish
```
JohnForster, line 1: This class might be unnecessary, as it doesn't do anything that couldn't be done with a hash: `{ name: "Pizza", price: 7.00 }`

lib/order.rb
```diff
@@ -0,0 +1,54 @@
+require './lib/send_sms'
+
+class Order
+  include Message
```
JohnForster, line 4: Modules are generally used when you want to include extra functionality in a class. For example, now that you've included `Message`, `Order` has access to `@to`, `@from`, `@account_sid` etc.

Is this what you want? Does an Order need access to these variables? If not, you probably don't need to use a Module here, and are better off creating a Message class that can encapsulate this data.

Modules are generally used when you want to share a function between multiple classes.

lib/order.rb
```diff
@@ -0,0 +1,54 @@
...
+  end
+
+  def finish
+    Message.send_confirmation
+  end
+
+  def find_dish(chosen_dish)
+    availability = false
+    at_index = nil
+    @available_dishes.each_with_index { |dish, index|
```
JohnForster, line 45: `Array#find` might be a simpler way of doing this.

lib/takeaway.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+  
+  private
+
+  def load_menu(name)
+    path = "./lib/menus/#{name}.csv"
+    raise 'No menu to load' unless File.exist?(path)
+    csv_lines = File.open(path, 'r') { |file| file.readlines }
+    csv_lines.each { |csv_lines| name, price = csv_lines.split(','); add_dish(Dish.new(name, price.strip.to_i))  }
+  end
```
JohnForster, line 30: There is a lot of CSV logic here, and it creates a hard dependency between the Dish class and the Takeaway class. Could this be moved into a separate class?

spec/order_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+describe Order do
+  
+  let(:pizza) { double(name: 'pizza', price: 7) }
+  let(:pasta) { double(name: 'pasta', price: 15) }
+  let(:olives) { double(name: 'olives', price: 2) }
+  let(:menu_list) { [pizza, pasta, olives] }
+  let(:order) { Order.new(menu_list) }
+  
+  it 'has an empty order_list when instantiated' do
+    expect(order).to have_attributes(order_list: [])
```
JohnForster, line 12: This is testing state, not behaviour. You want to avoid tests that peek inside a class to see what the state of an instance variable is. Better to have one test that checks whether you can successfully add a dish, and then another test that uses use another method to check that both have worked.

Think back to the Piggy Bank example from Week 1. You can't see the coins inside the piggy bank without breaking it, so you can't test whether once you put a coin in, the piggy bank now contains that coin. But you _can_ test the behaviour of the class, ie. once you .shake it, it now goes "cling!" whereas before it returned nothing.

Similarly, you shouldn't be testing that calling methods alters the internal state of your order_list, you should only be testing the inputs and outputs of the class.

lib/takeaway.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def show_menu
+    dishes = []
+    @menu_list.each { |dish| dishes << "#{dish.name} - #{dish.price}" } 
+    dishes.join(', ')
+  end
+
+  def make_order
+    Order.new(@menu_list)
```
JohnForster, line 20: This is a hard dependency between your takeaway class and your Order class. Could you use dependency injection here? If it is tricky, maybe you need to think about the structure of your classes. For example, does an order need to know what menu was used to make it?

spec/order_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  
+    describe '#finish' do
+      before :each do
+        #prevents Twilio from sending a message when using rspec
+        allow(Message).to receive(:send_confirmation) { true }
+      end
+      
+      it 'sends an sms to confirm order' do
+        # does this need to be tested?
+      end
```
JohnForster, line 50: Hard to test at the moment because you're using a module, but testing this would make more sense if the sending logic was encapsulated in its own class.

JohnForster (commented):
Generally, a great solution. Be careful with variable names, and make sure you're testing _behaviour_ not _state_.

PR #2124, opened 2022-01-03
Weekend Challenge: Takeaway

hannahdesmond:
Completed weekend challenge. Please see FEEDBACK.md for desired feedback/pointers for review.


README.md
```diff
@@ -14,20 +14,11 @@ Takeaway Challenge
 
  ```
```
KiranKukar, line 15: General comments:

Your read me is really clear, all your tests are passing. when I ran restaurant.rb in IRB it worked well, however basket.rb threw some errors on total and print_order.

FEEDBACK.md
```diff
@@ -0,0 +1,23 @@
...
+
+* How is my mocking? 
+
+* Could I have used keyword arguments better to make my code more changeable later?
+
+* How would these classes work together as a program if they are so loosely coupled. Do I need a customer ordering interface to implement the public methods of my classes? Or should one of my classes be doing this already e.g. restaurant?
+
+* Am I accessinging my attributes correctly in my restaurant methods? Should I be using the method rather than the instance variable?
+
+* I initialise my restaurant class with a pre-set menu. Is there a better way of doing this, e.g. with optional arguments or keyword arguments?
```
eoinmakers, line 17: I think it's fine! You don't need to overdo it: we want to keep our implementations as simple and close to the user stories as possible.

FEEDBACK.md
```diff
@@ -0,0 +1,23 @@
...
+
+* In my tests, I have hard coded the costs of the menu. Is this reasonable or should I be using my doubles to do this?
```
eoinmakers, line 19: It's fine to hardcode values in your tests that are the 'expected' result / behaviour, as an anchor you'll compare the 'actual' result / behaviour to.

eoinmakers (commented):
Well done on your implementation, it's clear you put a lot of thought into the quality of your code.

lib/restaurant.rb
```diff
@@ -0,0 +1,36 @@
+require_relative 'basket'
+require_relative 'send_text'
+
+
+class Restaurant
+  include Text
+  
+  attr_reader :menu
```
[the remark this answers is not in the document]
hannahdesmond: Should I only use accessor methods if I need to expose the menu outside of the class? Would I have been better just storing my menu in its own method within this class? (Although having looked at some other people's code I might try playing with storing it as a CSV next time!)

PR #2125, opened 2022-01-03
Kiran's takeaway challenge

KiranKukar:


spec/menu_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'menu'
+
+describe Menu do
+  let(:menu) { Menu.new }
+
+  # instance variables:
+  it { is_expected.to respond_to(:menu_list, :price_list, :dish_list) }
+  # methods:
+  it { is_expected.to respond_to(:print_menu, :load_menu, :prices) }
```
hannahdesmond, line 9: When you refactor, you could leave these expectations out if you test their behaviour e.g. line 16 makes it unnecessary to test whether it responds to #menu_list.

hannahdesmond:
**General comments**

* I really like that you've saved the menu in a CSV file. 
* Your methods are short and don't do too much. 
* Your classes are adhering to the Single Responsibility Principle. 


* It would be good to have a README that shows how to run the program in irb. 
* When I loaded `order.rb` into irb, it could not load the send_sms.rb file as it's missing from the lib folder.

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+  def initialize
+    @menu_list = []
+    @price_list = {}
+    @dish_list = []
+    load_menu
+  end
+
+  def print_menu
+    title = "Kiran's cafe menu:"
+    title + @menu_list.to_s
```
hannahdesmond, line 15: I would further format this output to include some line breaks and make it read a bit nicer.

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+  def load_menu
+    CSV.foreach("menu.csv") do |line|
+      dish, price = line
+      @menu_list << "#{dish}, £#{price}"
+      @price_list[dish.to_sym] = price.to_f
+      @dish_list << dish
+    end
+  end
+
+  def prices
```
hannahdesmond, line 27: This method is doing the same thing as your `price_list` method/attr_reader.

lib/menu.rb
```diff
@@ -0,0 +1,35 @@
...
+    @price_list
+  end
+
+  def dishes
```
hannahdesmond, line 31: Same for this one – you already have a method that does the same thing – `dish_list`.

spec/menu_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  describe '#print_menu' do
+    title = "Kiran's cafe menu:"
+    it "is expected to return the menu list" do
+      expect(menu.print_menu).to eq title + menu.menu_list.to_s
+    end
+  end
+  describe '#prices' do
```
hannahdesmond, line 29: This and the following spec are a bit redundant because the two methods do the same thing. I would suggest trying to test for the output – something like `expect(menu.dishes).to include("nasi goreng")`, for example.

lib/order.rb
```diff
@@ -0,0 +1,67 @@
...
+    @menu = Menu.new
+    @order_list = []
+    @total = 0
+  end
+  
+  def print_menu
+    puts @menu.print_menu
+  end
+
+  def select_dish
```
hannahdesmond, line 17: I would put these as private methods, as when I was running your code, I started trying to use these methods and didn't understand why they didn't add a dish to the `order_list` but I realise they get called when you call `order_item`. So if your method is not called outside of your class and is only used by another method, and you don't want Joe Public messing about with them, you can make them private methods.

lib/order.rb
```diff
@@ -0,0 +1,67 @@
...
+    puts "Please type in the dish you want:"
```
hannahdesmond, line 18: It might be nice to be able to ignore case when responding here, you could add some more logic in to help with that perhaps?

hannahdesmond (commented):
**General comments**

I did not have enough time to go through all of your code at all unfortunately Kiran but I hope you can get some use out of these comments! Your repo looks really great, you've done a lot and your business logic looks really good to me! Sorry I didn't get to spend more time on your tests. Below are some general comments and I have left comments in other parts of your code too. 

* I really like that you've saved the menu in a CSV file. You've inspired me to have another play with that format, thank you! 
* Your methods are generally short and don't do too many things at once, which is great! 
* Your classes are adhering to the Single Responsibility Principle. 
* All specs passing for `menu.rb` 💚


* It would be good to have a README that shows your planning process as well as how to run the program in irb. 
* When I loaded `order.rb` into irb, it could not load the send_sms.rb file as it's missing from the lib folder. I couldn't run the tests for the same reason and had to comment out some of the code to run the `order.rb` in irb.

PR #2126, opened 2022-01-03
Josh Lim - Takeaway Challenge 

sumo-mailman:


Gemfile
```diff
@@ -11,3 +11,5 @@ end
 group :development, :test do
   gem 'rubocop', '1.20'
 end
+
+gem "twilio-ruby", "~> 5.61"
```
kat-mcbrooks, line 15: great (and helpful) to see the twilio gem in the gemfile!

README.md
```diff
@@ -14,70 +14,23 @@ Takeaway Challenge
...
 
-Instructions
+Steps
 -------
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+* irb
```
kat-mcbrooks, line 20: helpful, clear example of how your program runs in irb

lib/order.rb
```diff
@@ -0,0 +1,24 @@
...
+  def add(item)
+    @order_list << { item => @list[item] }
+  end 
+
+  def total
+    @order_list.inject(0) { |sum, hash| sum += hash.values.join.to_i }
+  end
+
+  def confirm
+    @message = Checkout.new.message
```
kat-mcbrooks, line 21: Great that you've followed the SRP and separated out the messaging functionality via Twilio into another class.

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  
+  # let(:message) {:order.confirm}
+  
+  describe '#add' do
+    it { is_expected.to respond_to(:add) }
+
+    it 'increases the order_list by 1' do
+      expect { subject.add("Kim Chi") }.to change { subject.order_list.count }.by(1) 
+    end 
+    # build a test which increases the
```
kat-mcbrooks, line 13: thanks for including comments on your what you next plan to add to the code and tests

spec/order_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  describe '#add' do
+    it { is_expected.to respond_to(:add) }
+
+    it 'increases the order_list by 1' do
+      expect { subject.add("Kim Chi") }.to change { subject.order_list.count }.by(1) 
+    end 
+    # build a test which increases the 
+    # order_list by 1 after add
+  end 
+
```
kat-mcbrooks, line 16: Do you also need a test to check that Order's add(item) method is adding the hash key value pair to the @order_list array ? e.g. something like expect (subject.order_list).to include {"kim chi" => 4}

lib/checkout.rb
```diff
@@ -0,0 +1,24 @@
+# Download the helper library from https://www.twilio.com/docs/ruby/install
+require 'twilio-ruby'
+
+# Find your Account SID and Auth Token at twilio.com/console
+# and set the environment variables. See http://twil.io/secure
+
+class Checkout
+  def initialize(_client = Twilio::REST::Client, _sid = ENV['account_sid'], _auth = ENV['auth_token'])
```
kat-mcbrooks, line 8: I didn't think to inject these as arguments to my Message class. I'm curious as to why you decided to do this, as opposed to just using them directly as variables within the message method? It certainly makes your code clearer, but I'm curious if you had another reason, other than readability, for doing it this way :) (asking for the purposes of my own learning as I'm trying to get to grips with dependency injection so any insight into others' approaches is interesting!)

kat-mcbrooks (commented):
Nice and clear code and great that you implemented the Twilio API

PR #2127, opened 2022-01-03
My Takeaway Challenge

ravensears:
All tests passing with a 98% coverage and linted.

Struggled with these specs:
- Make sure that your Takeaway is thoroughly tested and that you use mocks and/or stubs, as necessary to not to send texts when your tests are run
- However, if your Takeaway is loaded into IRB and the order is placed, the text should actually be sent

Would love some insight on how to have done this.

### DIFF
--- a/lib/messenger.rb
+++ b/lib/messenger.rb
@@ -1,0 +1,17 @@
+require 'rubygems'
+require 'twilio-ruby'
+
+account_sid = 'AC8a40b1c2bad00ceecb1d7a5bf1008585'
+auth_token = 'c5d152e998e2487fa5769c9f5d03dd82'
+@client = Twilio::REST::Client.new(account_sid, auth_token)
+ 
+a = Time.now + 1 * 60 * 60
+hour_later = a.strftime("%I:%M %p, %d %B, %Y")
+
+message = @client.messages.create( 
+    body: "Thank you! Your order was placed and will be arrive on or before #{hour_later}!",  
+    messaging_service_sid: 'MG6c8b3326fcd3fe9d373e5d8f798fb4df',      
+    to: '+447538213479' 
+  ) 
+ 
+puts message.sid
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,60 @@
+require 'messenger'
+
+class Takeaway
+  attr_accessor :menu, :items, :balance
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+    @items = Hash.new(0)
+    @balance = 0
+  end
+
+  def read_menu
+    @menu.dishes
+  end
+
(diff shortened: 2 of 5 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,60 @@
...
+      puts "#{item} #{price}x = £#{@menu.dishes[item] * price}"
+    end
+  end
+
+  def total
+    "Your total is £#{@balance.round(2)}"
+  end
+
+  def complete_order
+    total
```
Haydnstokes92, line 39: would potentially switch these around as the text would go out first confirming the order with total amount

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      ).to_stdout
+    end
+  end
+
+  context 'complete order' do
+    it 'sends a payment confirmation text message' do
+      expect(takeaway).to receive(:send_text)
+      takeaway.complete_order
+    end
+  end
```
Haydnstokes92, line 47: nice clean text written across this last file especially just be careful when refactoring that below 4 lines is known as the ideal amount for functions, perhaps using attr-readers and also private functions may help  to improve this.

lib/messenger.rb
```diff
@@ -0,0 +1,17 @@
...
+ 
+a = Time.now + 1 * 60 * 60
+hour_later = a.strftime("%I:%M %p, %d %B, %Y")
+
+message = @client.messages.create( 
+    body: "Thank you! Your order was placed and will be arrive on or before #{hour_later}!",  
+    messaging_service_sid: 'MG6c8b3326fcd3fe9d373e5d8f798fb4df',      
+    to: '+447538213479' 
+  ) 
+
```
Haydnstokes92, line 16: Need more research on this myself but the code works and a text is sent so thumbs up!

lib/takeaway.rb
```diff
@@ -0,0 +1,60 @@
...
+    @balance = 0
+  end
+
+  def read_menu
+    @menu.dishes
+  end
+
+  def add_order(item, quantity = 1)
+    @items[item] += quantity
+    @balance += @menu.dishes[item] * quantity
```
eoinmakers, line 18: Is it worth keeping an updated balance before the user tries to check out? As it will need to be continuously updated if items are added / removed, vs. calculating the total at checkout once.

lib/takeaway.rb
```diff
@@ -0,0 +1,60 @@
...
+    "Your total is £#{@balance.round(2)}"
+  end
+
+  def complete_order
+    total
+    send_text
+  end
+
+  def send_text
+    require_relative 'messenger'
```
eoinmakers, line 44: Maybe neater to require this at the top of your file, and then make a call to a send text method in the text sending class.

PR #2133, opened 2022-01-04
Kat Takeaway Challenge

kat-mcbrooks:


README.md
```diff
@@ -1,6 +1,12 @@
-Takeaway Challenge
-==================
-```
+## Takeaway Challenge
```
sumo-mailman, line 1: Great! This breakdown was phenomenal

README.md
```diff
@@ -11,73 +17,100 @@ Takeaway Challenge
...
+rice: £2.50
+chapati: £1.50
+ => {"dal"=>6.5, "korma"=>3.0, "rice"=>2.5, "chapati"=>1.5} 
+ ``` 
+
+### Problems/questions
+
+1) My main concerns were around testing. Not sure if I'm using test doubles correctly and therefore if I've got sufficient test coverage. (nb: when I run rspec, it shows 100% coverage, but when I run rspec on individual spec folders, it shows <100% coverage.)
+
+2) Is there a way to test if one class' method creates a new instance of another class? Not sure how to test this is happening correctly (e.g. testing that Takeaway's complete_order method creates Message.new)
```
sumo-mailman, line 100: Wish I could help you with some of these mock/stub questions. Definitely ask a coach

lib/message.rb
```diff
@@ -0,0 +1,19 @@
...
+class Message
+
+  def send(total) 
+    arrival_time = (Time.now + 60*60).strftime("%H:%M")
+    #order_cost = total
+    total = '%.2f' % total
+    Twilio::REST::Client.new(ENV['TWILIO_ID'], ENV['TWILIO_TOKEN']).messages.create(
+      from: ENV['TWILIO_NUM'],
+      to: ENV['MY_NUM'],
+      body: "Thank you for your order, which totals to £#{total}. Your food will arrive by #{arrival_time}"
```
sumo-mailman, line 15: Loved how you added the additional feature of totals in there

lib/dish.rb
```diff
@@ -0,0 +1,9 @@
+class Dish
```
JohnForster, line 1: This currently doesn't seem to be being used by the Menu class, and currently is pretty much equivalent to just being a hash, eg `{ name: "dal",  price:"4.00" }`. 

A Dish class could be useful down the line, if you wanted to allow customisation (Dal without the lentils for me, thanks), but for now I don't think it's necessary.

lib/takeaway.rb
```diff
@@ -0,0 +1,36 @@
...
+  def view_order
+    @order.display
+  end
+
+  def check_total
+    @order.check_total? ? "Your order total has been calculated correctly" : "Your order total does not match the cost of dishes ordered. Speak to the staff."    
+  end
+
+  def complete_order
+    message = Message.new # is this fine here, or do I need to inject it in as an argument? Not sure if dependency injection is really necessary with Twilio class?!
```
JohnForster, line 31: With the code like this, you are creating a new Message (and therefore TwilioClient) every time you complete an order. It would be better to pass the `Message` instance in during initialisation, similar to how you are handling the `order` and `menu`.

spec/order_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'order' 
+describe Order do
+  default_menu = Menu::MENU
+  let(:menu) { instance_double(Menu, dishes: default_menu) }
```
JohnForster, line 4: I'd recommend creating a new menu for your tests, instead of using the default_menu here. Currently if a change happens to `Menu::MENU` it could break your `order_spec` tests, which means that they are not isolated properly.

spec/order_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+  let(:quantity) {2}
+  let(:price) {6.50}
+
+  describe "#add" do 
+  before do 
+    subject.add(dish, quantity)
+  end
+
+    it 'adds dish name, price and quantity to basket' do
+      expect(subject.basket).to include({name: dish, item_price: price, quantity: quantity, total_cost: quantity * price})
```
JohnForster, line 16: I mentioned testing behaviour over state before, but this is an example of testing state. You already have a nice `#view_order` method which displays the order in a way you can control, so you don't want to expose the `#basket` to the outside world. Better to test the _outputs_ of the `#add` method than what it does to the internal state.

If there are no outputs, that's fine, you can use another method to check the result and test the two together.

PR #2139, opened 2022-01-04
Takeaway Challenge

robinucar:


lib/dish.rb
```diff
@@ -0,0 +1,7 @@
+class Dish
+    attr_reader :name, :price
+    def initialize(name, price)
+        @price = price
+        @name = name
+    end
+end
```
JohnForster, line 7: This class could be useful later down the line, but for now it could probably be replaced with a simple hash: `{ name: "Pizza", price: "7.00" }`

lib/order.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'send_text'
+class Order
+    attr_reader :order_list
+    def initialize
+        @order_list = []
```
JohnForster, line 5: Could this name be a bit more descriptive? This name makes it sound like a list of orders.

lib/send_text.rb
```diff
@@ -0,0 +1,25 @@
+require 'twilio-ruby'
+require 'dotenv/load'
+
+class SendText
```
JohnForster, line 4: Ideally classes should be named nouns, and methods verbs. Could this be `TextSender` or `MessageHandler` or equivalent instead?

lib/send_text.rb
```diff
@@ -0,0 +1,25 @@
...
+    def send
+        account_sid = ENV['TWILIO_ACCOUNT_SID']
+        auth_token = ENV['TWILIO_AUTH_TOKEN']
+        @time = Time.now + (1 * 60 * 60)
+
+        @client = Twilio::REST::Client.new(account_sid, auth_token)
```
JohnForster, line 10: You are creating a new Client every time you send a message. Could you instead create this client once and use the same one to send multiple messages?

lib/takeaway.rb
```diff
@@ -0,0 +1,13 @@
+class Takeaway
```
JohnForster, line 1: What does this class represent? Does the name of the class accurately represent that? Can you think of any names that might be more appropriate?

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'order'
+describe Order do
+    let(:hamburger) {double(name: 'Hamburger', price: 5)}
+    let(:chips) {double(name: 'Chips', price: 2)}
+
+    it {is_expected.to respond_to(:add, :total_price, :print_order, :place_order)}
+
+    it 'starts with empty order list' do
+        expect(subject.order_list).to be_empty
```
JohnForster, line 9: You are testing behaviour vs state here. Have a think about how you want this class to be used, should you be able to reach in and look at the order from outside? If an attribute reader isn't used in your feature tests (eg. irb) then it shouldn't be part of the class, and therefore shouldn't be used in tests.

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        subject.add(hamburger)
+        subject.add(chips, 2)
+        expect(subject.total_price).to eq 9
+    end
+
+    it 'print order summary' do
+        subject.add(hamburger)
+        subject.add(chips, 2)
+        expect(subject.print_order).to eq "Hamburger: 5 * 1 = 5, Chips: 2 * 2 = 4"
+    end
```
JohnForster, line 28: This is a great example of testing behaviour instead of state 👍 . This tests both the functionality of the `#add` method as well as the `#print_order` method. You don't need the "add dishes to the order list" test if you have this one instead.

spec/order_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    it {is_expected.to respond_to(:add, :total_price, :print_order, :place_order)}
+
+    it 'starts with empty order list' do
+        expect(subject.order_list).to be_empty
+    end
+
+    it 'add dishes to the order list' do
+        subject.add(hamburger)
+        subject.add(chips, 2)
+        expect(subject.order_list).to include({hamburger => 1}, {chips => 2})
```
JohnForster, line 15: Same again, you should avoid testing `subject.order_list` because this is _state_ not _behaviour_.

JohnForster (commented):
A generally excellent solution. Just a couple of things to watch out for:
- Try to be precise when naming your variables, classes etc.
- Try to avoid testing the internal state of a class, instead test its visible behaviours.
Well done!

PR #2140, opened 2022-01-07
Sam Pickering

samhpickering:


rosiewaite92:
I think this is a really well written piece of code.
The classes, methods and files were all clearly named and easy to understand where things belong. 
Anything I did not understand was due to my own knowledge gaps rather than anything being poorly written. 
I think its is good that you have really utilised resources available in ruby such as twilio and the inbuilt Classes in Ruby. 

My one "clutching at straws" improvement I could suggest would be to use a "personalised' error rather than a RunTime error on line 54 of restaurant_spec.rb just to make things more user friendly.

PR #2142, opened 2022-01-07
My submission for the takeaway challenge

johnny-m-young:
User Stories:
- As a customer, so that I can check if I want to order something, I would like to see a list of dishes with prices
- As a customer, so that I can order the meal I want, I would like to be able to select some number of several available dishes
- As a customer, so that I can verify that my order is correct, I would like to check that the total I have been given matches the sum of the various dishes in my order

Incomplete:
User cannot receive a text confirming order

Improve:
RSpec tests relating to print order fail, not sure how to phrase them.


alicegray33:
I like that the code and tests are nicely written; robust and easy to read. I liked how you considered not just adding a Dish but removing one. And I also liked how you considered the user interacting with the code by creating a user interface.

You had an interesting approach to the delivery time by giving each dish its own prep time, I hadn't considered that as an option!

Some of the tests could have been refactored to avoid repetition with Let or Before; but I understand you run out of time.

lib/order.rb
```diff
@@ -0,0 +1,27 @@
...
+    @prices = []
+    @total_cost = 0
+    @total_prep_time = 0
+  end
+
+  def add_dish(dish)
+    @dishes << dish.name
+    @prices << dish.price
+    @total_cost += dish.price
+    @total_prep_time += dish.prep_time
```
EdwardAndress, line 16: I like what you're doing with prep time here

lib/customer_interface.rb
```diff
@@ -0,0 +1,45 @@
+class CustomerInterface
+
+  attr_reader :menu, :order
+
```
EdwardAndress, line 4: You could inject the menu on initialization to allow some flexibility there

feature_tests.rb
```diff
@@ -0,0 +1,10 @@
+Dir["./lib/*.rb"].each { |file| require file }
+
+interface = CustomerInterface.new
+dish = Dish.new("Katsu Chicken Curry", 7.99, 20)
+another_dish = Dish.new("Sushi Platter", 21.99, 30)
+
```
EdwardAndress, line 6: At the moment, the interface contains the menu but then, to order things, Dishes are created outside of that.  Something doesn't quite feel right there.  I.e. Should there be a more explicit connection between Dishes and the menu?

EdwardAndress: I wonder whether the menu should be a collection of Dishes

EdwardAndress:
Hi Johnny, you've done some great work here and have obviously thought quite hard about how to apply some good OOP.  As mentioned above (?) I think there's probably some refactoring you could do around the relationship between Dish and UserInterface.

PR #2144, opened 2022-01-07
takeaway challenge

Tom-V1:


lib/menu.rb
```diff
@@ -0,0 +1,23 @@
+class Menu
```
simba-tawonezvi, line 1: The code is spaced out well and easy to read.

lib/order.rb
```diff
@@ -0,0 +1,30 @@
+require_relative "sms"
```
simba-tawonezvi, line 1: Code is easy to read and written in a consistent format

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "takeaway"
```
simba-tawonezvi, line 1: I liked how you used "describe" to group the test for the methods

lib/takeaway.rb
```diff
@@ -0,0 +1,53 @@
+require_relative 'menu'
+require_relative 'order'
+
+class TakeAway
```
tomfuller, line 4: I probably would have named this restaurant but thats more my preference than anything.

lib/takeaway.rb
```diff
@@ -0,0 +1,53 @@
...
+    @order = Order.new
+    @finish = false
+  end 
+
+  def show_menu
+    @menu.print_menu
+  end
+
+  def add(input, qua = 1)
+    return puts "I'm afraid we don't have any #{input}" unless valid?(input)
```
tomfuller, line 19: Instead of `return puts "I'm afraid we don't have any #{input}" unless valid?(input)` you can just drop the puts and `return "I'm afraid we don't have any #{input}" unless valid?(input)` or even better surface an error here with `raise “I’m afraid we don't have any #{input}" unless valid?(input)`.

lib/takeaway.rb
```diff
@@ -0,0 +1,53 @@
...
+    return puts "Order already complete, unable to add item" if finish?
+    update_total(input)
+    add_order(input, qua)
+  end
+  
+  def valid?(input)
+    @menu.items.has_key?(input.to_sym)
+  end
+
+  def update_total(input)
```
tomfuller, line 29: Do you think its up to the restaurant to update the order's total or should that behaviour belong to the order?

tomfuller:
Overall your solution is nice and neat and it solves the problem. The only major issue I have is where you have used puts statements instead of return values but I have left some comments to deal with that. In general though you want your functions to return something useful for them to be used by another function. The only time you would use puts as an output is if your are building a command line application otherwise they are only really used for debugging purposes.

PR #2146, opened 2022-01-10
Monday

alicegray33:


jyoung-gds:
- Really good, clean code which is well formatted. All tests pass and code does what the user stories request.
- Use different names for classes to be more representative of their function/role.
- Could create a separate class for send_sms method in Restaurant class, rather than having it as a private method.
- Could write more tests/functionality (e.g. to cover end cases) but that is beyond what was requested in the challenge.

README.md
```diff
@@ -47,37 +24,114 @@ So that I am reassured that my order will be delivered on time
...
-  * Make sure that your Takeaway is thoroughly tested and that you use mocks and/or stubs, as necessary to not to send texts when your tests are run
-  * However, if your Takeaway is loaded into IRB and the order is placed, the text should actually be sent
-  * Note that you can only send texts in the same country as you have your account. I.e. if you have a UK account you can only send to UK numbers.
+## Approach
+
+I read through the user stories and drew up a basic diagram of how I imagined the Classes working and talking to each other:
 
-* Advanced! (have a go if you're feeling adventurous):
-  * Implement the ability to place orders via text message.
+![Diagram 2](https://github.com/alicegray33/takeaway-challenge/blob/main/docs/diagram2.png?raw=true)
```
EdwardAndress, line 31: Nice that you're diagramming!

README.md
```diff
@@ -47,37 +24,114 @@ So that I am reassured that my order will be delivered on time
...
+3.0.2 :021 > restaurant = Restaurant.new
+ => 
+#<Restaurant:0x00007fb7a2928478
+... 
+3.0.2 :022 > restaurant.place(order, "<my_number>")
+ => "Thank you! Your order was placed and will be delivered before 09:21" 
+3.0.2 :023 > 
+```
+
+## Further Thoughts
```
EdwardAndress, line 124: Nice reflection.

lib/list.rb
```diff
@@ -0,0 +1,22 @@
...
+    @list = []
+  end
+
+  attr_reader :list
+
+  def add_dish(dish)
+    @list << dish
+  end
+
+  def view
```
EdwardAndress, line 13: You could probably use map here

```ruby
@list.map { |dish| { dish: dish.name, price: dish.price } }
```

EdwardAndress: In python, you could use list comprehension...

```python
[{ 'dish': dish.name, 'price': dish.price } for dish in self.list ]
```

I've not executed the above to test it though!

lib/order.rb
```diff
@@ -0,0 +1,21 @@
+class Order
+
+  def initialize
+    @meal = []
```
EdwardAndress, line 4: Hmmm... not sure about `meal` as the name here. Could sound like an individual person's dinner or the food that everyone present will share.

lib/restaurant.rb
```diff
@@ -0,0 +1,40 @@
...
+  def place(order, number = nil)
+    @order = order
+    deadline = wait.strftime('%H:%M')
+    message = "Thank you! Your order was placed and will be delivered before #{deadline}"
+    return message if number == nil?
+    send_sms(message, number)
+  end
+
+  private
+
```
EdwardAndress, line 23: You mentioned in your reflection that you could have broken some of your solution down more. I suspect you're talking about this bit, where an SMS is sent...?  That would make sense.  You could use a `Notification` class..?

lib/restaurant.rb
```diff
@@ -0,0 +1,40 @@
...
+  def send_sms(message, number)    
+    # set up a client to talk to the Twilio REST API
+    @client = Twilio::REST::Client.new @account_sid, @auth_token
```
EdwardAndress, line 26: To test this without sending an SMS every time, you'll need to inject the client (so that you can then inject a double in your tests).

spec/restaurant_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    wait_time = time_now + Restaurant::WAIT_TIME
+    ready = Time.at(wait_time)
+  end
+  let(:fake_twilio) { double(:twilio).as_null_object }
+  let(:mobile_number) { '07000 123 123' }
+  let(:message) { "Thank you! Your order was placed and will be delivered before #{wait.strftime('%H:%M')}" }
+
+  it 'receives order and responds with a message' do
+    allow(Time).to receive(:now).and_return(time_now)
+    allow(Twilio::REST::Client).to receive(:new).and_return(fake_twilio)
```
EdwardAndress, line 16: Ah ha! So you found a way to do this without injecting the client :)

EdwardAndress:
Really nice work Alice! I have left a few comments in line against the code but I don't think I have said anything that you hadn't already thought about yourself! :)

PR #2148, opened 2022-01-10
Justine Williams

JALWilliams:
3/4 user stories completed, but didn't have time to attempt API hasn't been attempted

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,47 @@
+class Takeaway
+  attr_reader :basket
+
+  def initialize 
+    @menu = {"pasta": 7, "dhal": 9 , "pizza": 10 , "chocolate": 9} 
+    @basket = { }
+    @total_cost = 0
+    @basket_with_cost = {}
+  end
+
+  def view_menu
+    @menu
+  end
+
+  def order (dish, quantity)
+    price_lookup(dish, quantity)
+    already_ordered_flag= @basket.keys.include? "#{dish}"
+    already_ordered_flag ? updating_order_quanity(dish, quantity) : @basket["#{dish}"] = quantity
+    return "Orderd #{dish} x #{quantity}"
+  end
+
+  def total 
+    @total_cost
+  end
+
+  def basket_summary
+    # "spring roll x4 = £3.96, pork dumpling x3 = £8.97"
+    @basket_with_cost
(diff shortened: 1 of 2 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def initialize 
+    @menu = {"pasta": 7, "dhal": 9 , "pizza": 10 , "chocolate": 9} 
+    @basket = { }
+    @total_cost = 0
+    @basket_with_cost = {}
+  end
+
+  def view_menu
+    @menu
```
OliviaMaxwell, line 12: Potentially could do a .each
So that it doesn't just show the actual hash. To show the items with their price in a list?

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
+class Takeaway
+  attr_reader :basket
+
```
EdwardAndress, line 3: How about this refactor?

```ruby
def initialize(menu)
  @menu = menu
  @basket = {}
  @total_cost = 0
  @basket_with_cost = {}
end
```

It would allow you to change the menu easily when creating new Takeaways

lib/takeaway.rb
EdwardAndress: Nice idea, @OliviaMaxwell !

lib/takeaway.rb
```diff
@@ -0,0 +1,47 @@
...
+    @menu = {"pasta": 7, "dhal": 9 , "pizza": 10 , "chocolate": 9} 
+    @basket = { }
+    @total_cost = 0
+    @basket_with_cost = {}
+  end
+
+  def view_menu
+    @menu
+  end
+
```
EdwardAndress, line 14: There's a space after `order` below, which Ruby doesn't care too much about but sometimes it will trip you up.

EdwardAndress:
Hey Justine, I think this is a good solution – well done! I was particularly impressed by your solution for adding multiple items to the basket. I think the next step, given that it's an OOP exercise, would have been to think about breaking up the application into multiple classes.

Some thoughts on that...

- An Order class could be used to contain all the ordered items and calculate the total
- Menu class might also be useful
- One of the later user stories involves sending a text message and, at that point, an SMS or Notification class is useful.

PR #2149, opened 2022-01-25
Susana Mncwabe

siellsiell:
Hi Susana,

I've created a pull request for you so I can review your challenge. For the next times, please make a full request after you've completed the challenge so that we can find your code more easily :)

Your code show some evidence that you're trying that you aimed to break up your coat into separate responsibilities using different classes. That was a good start! It looks like you've got a bit muddled in the execution of it though, which is common. This stuff is often tricky.

A good start would be to be revisit some of the materials on domain modelling, [this worked example of converting user stories to classes is a good start](https://docs.google.com/document/d/1QR2C3hw1cjZKGNF84Zfy0J5VjAbjCfM4qprWXTUkmfc/edit?usp=sharing), as well as materials on classes and methods.

I encourage you to get support with this from a coach to talk through your reasoning from this challenge and help you figure out where you the stumbling blocks were. 

The coach currently with you this week is the best point of call, but you should feel free to reach out to whoever you feel most comfortable chatting to (they just might be slower at getting back to you).

Do let me know if you have any questions!

### DIFF
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,0 +1,13 @@
+# require "menu"
+
+# describe Menu do
+#   subject(:menu) { describe_class.new(meal) }
+#   let(:meal) do
+#     { paella: 9.99, calamares: 4.99, pinchitos: 3.99, jamon: 9.99 }
+#   end 
+
+#   it "show the menu and prices" do
+#     expect(menu.read).to eq(read_menu)
+#   end 
+
+# end
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,9 +1,8 @@
 require "Menu"
 describe Menu do
   it "allow you to read dishes and prices" do
-    menu = double("menu", dish: "Chicken", price: 4.99)
-    expect(menu.dish).to eq("Chicken")
-    expect(menu.price).to eq(4.99)
+    my_menu = double("Tapas", dish: "Pinchitos", price: 6.99)
+    expect(my_menu.dish).to eq("Pinchitos")
+    expect(my_menu.price).to eq(6.99)
   end 
 end
-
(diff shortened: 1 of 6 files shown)


spec/menu_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "Menu"
+describe Menu do
+  it "allow you to read dishes and prices" do
+    my_menu = double("Tapas", dish: "Pinchitos", price: 6.99)
+    expect(my_menu.dish).to eq("Pinchitos")
+    expect(my_menu.price).to eq(6.99)
```
siellsiell, line 6: By replacing the menu object was a double you're affectively no longer testing the code in your menu class but instead of testing the double.

One way to realise you're doing something like this: delete the code in the class you're testing (in this case the `dish` and `price` methods in your `menu.rb`). If your tests still passes, it's an indication that your test isn't really testing what you want it to test.

You can testing my new class directly without a double:
```
 my_menu = Menu.new("Pinchitos")
 expect(my_menu.dish).to eq("Pinchitos")
``` 

We only want to use doubles to replace classes that interact with the class we are testing. It might be worth revisiting some of the materials on doubles and testing and chatting with a coach about this!

spec/order_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "order"
+  describe Order do
+  let(:food) { double("My Food", add_dish?: true, add_price: true) }
```
siellsiell, line 3: The same comments as above apply to this test.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "takeaway"
+describe Takeaway do
+  let(:my_menu) { double("Tapas", ready?: true, dish: "Pinchitos", price: 6.99) }
```
siellsiell, line 3: The same comments as above apply to this test.

PR #2151, opened 2022-01-30
Takeaway - Luke Collins

Luke-Collins:


lib/takeaway.rb
```diff
@@ -0,0 +1,24 @@
...
+    p input
+    @menu.map { |item, price| if input == item
+    @basket.store(item, price) end }
+  end
+
+  def basket
+    @basket.map { |item, price| return "#{item} - £#{price}" }
+  end
+
+end
```
JonahMassam, line 24: Nice and concise. Can't see any standout issues.

PR #2152, opened 2022-01-30
Oliver Cutting - Takeaway-challenge

OliverCutting:


FEATURETEST.md
```diff
@@ -0,0 +1,21 @@
+#### Takeaway-challenge Feature-test
+
```
abdulrm94, line 2: This document is a great way to explain how to run and operate your code, I really like how you've broke down each function!

lib/dish.rb
```diff
@@ -0,0 +1,9 @@
+class Dish
```
abdulrm94, line 1: Nice, simple, easy to read class. I'm not an expert, but maybe the methods could have been included on a different class? Although I'm sure you had a good reason why you chose this approach!

lib/order.rb
```diff
@@ -0,0 +1,14 @@
+class Order
```
abdulrm94, line 1: Perfect class for the customer to access and push their order to. Does what it says on the tin. Good use of string interpolation and shortcut methods

lib/takeaway.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'dish'
+require_relative 'order'
+require 'twilio-ruby'
+require 'dotenv/load'
+
+class Takeaway
```
abdulrm94, line 6: Again, another impressive approach of clean, effective code that is easy to digest and understand. Your choices for your naming of methods and classes allows someone to walk in and know exactly what you're trying to achieve

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'order'
+
+describe Order do
+  
+  let(:dish) { double :dish, :name => "Pizza", :price => 5 }
+  it 'has a default total of 0' do
+    expect(subject.total).to eq 0
```
abdulrm94, line 7: i'm not sure i've understood this test correctly, we shall discuss, but isn't the total supposed to equal 5? or am i missing something?

spec/order_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  it 'is empty by default' do
+    expect(subject.items).to be_empty
+  end
+
+  it 'can add items to the order' do
+    subject.add(dish)
+    expect(subject.items).to include("Pizza = £5")
```
abdulrm94, line 16: great use of rspec docs to use methods that are appropriate for your test

PR #2153, opened 2022-01-30
Fred Halliwell, Takeaway Challenge

FredHalliwell:
Did somebody say just.each


JaabirM (commented):
Overall, your code is well-written and works. I do think there are some improvements to be made - more notably:

1. You may want to check the indentation and line breaks are used following Ruby convention as this makes code easier to read.
2. You could reduce the number of the files to two - one handles the takeway and other handling the SMS part. Currently your code is quite complicated
3.  Try to reduce the number of unnecessary code in your files such as including GEM files when it's not required for example.

In general, I do think your code is well-written and certainly does the job. Your code is very extensible - i.e. new functions can be added without needing to rework entire codebase which I think is very importaant

PR #2157, opened 2022-01-31
Takeaway Challenge - Mathew Grimward

MathewGrimward:


lib/menu.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def print_dishes
+    @dishes.each do |food, price|
+      puts "#{food}: £#{price}"
+    end
+  end
+
+  def order(*choices)
+    choices.select do |choices|
```
Barrington-Brook, line 17: What do you think about replacing these two lines with the one liner `choices.select {|x|@dishes.keys.include?(x)}` ?

spec/menu_spec.rb
```diff
@@ -0,0 +1,19 @@
+require './lib/menu'
+
+describe Menu do
+  let(:menu) { Menu.new }
```
Barrington-Brook, line 4: Very clearly written and readable tests.

Barrington-Brook, line 4: Could possibly use `subject` here to refactor out `let` statement?

PR #2158, opened 2022-01-31
takeawaychallenge

JamesBuckle93:
Friday Takeaway Challenge attempt.

### DIFF
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,9 @@
+class Takeaway
+
+  attr_reader :menu
+
+  def initialize 
+    @menu = { "pizza" => "£4.50", "chips" => "£2", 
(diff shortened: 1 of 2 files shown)


Edenlsw (commented):
Congratulations on completing the 1st user story. 
It's easy to read and follow the code.

lib/takeaway.rb
```diff
@@ -0,0 +1,9 @@
+class Takeaway
+
+  attr_reader :menu
+
+  def initialize 
+    @menu = { "pizza" => "£4.50", "chips" => "£2",
```
Edenlsw, line 6: Maybe have each menu item on its own line. 
pizza, 
chips,
kebab
It may become difficult to read when the menu becomes longer

PR #2159, opened 2022-01-31
Takeaway challenge - Jaabir Yusuf

JaabirM:
All user stories complete - although the final user story has not been tested with rspec.


FredHalliwell:
Great code. It's really easy to read and it's clear what everything does without comments.  Everything is written in a nice simple way so it would scale well if you were to expand on this project.

One thing to improve on may be to double check all files are being committed when you push. 

Good job 🥇

PR #2160, opened 2022-02-04
Takeaway Challenge - Bromley

nyahehnagi:
Initial Commit

The recieve SMS functionality is still in a state of development. I have it working in isolation but not linked to my feature tests.

I also have further research to do regarding adding new mock objects when there is a loop involved in the implementation. This is the case in my OrderManager class where I loop through and add ordered items to the order.


lib/credentials.rb
```diff
@@ -0,0 +1,12 @@
...
+class Credentials
+
+  attr_reader :account_id, :auth_token
+
+  def initialize
+    @auth_token = ENV['TWILIO_AUTH_TOKEN']
+    @account_id = ENV['TWILIO_ACCOUNT_SID']  
+  end
+
+end
```
CTJMID, line 12: Good use of environment variables in order to protect sensitive info

lib/menu.rb
```diff
@@ -0,0 +1,34 @@
...
+    @menu.find { |dish| dish.name == search_name }
+  end
+
+  def load_menu(filename = './data/menu.csv')
+    @menu = []
+    CSV.parse(File.read(filename)) do |row|
+      name, price = row
+      @menu << @dish_class.new(name, price)
+    end
+  end
```
CTJMID, line 24: Interesting use of csv file to update and display menu

lib/order.rb
```diff
@@ -0,0 +1,72 @@
...
+    @ordered_dishes.each do |ordered_dish, quantity_ordered| 
+      order += format_ordered_item(ordered_dish, quantity_ordered)
+    end
+    order += format_price
+    order.chomp
+  end
+
+  def format_ordered_item(ordered_dish, quantity_ordered)
+    "Dish: #{ordered_dish.name}, Qty: #{quantity_ordered}\n"
+  end
```
CTJMID, line 63: Tidy way of using all information, in useful way.

CTJMID (commented):
Your code looks great, and has great functionality. You have obviously built in alot of features and considered how the app would be used in real-world terms. As discussed,  due to the complexity of the code, I have found it difficult to review meaningfully - however your explanations have helped.

PR #2161, opened 2022-02-05
EDDIE ARNOLD JAN 22 COHORT 

eddiearnoldz:
Very enjoyable even if i did make a complete hash of it. Would love some feedback if and when. Thank you.


lib/food_delivery_ui.rb
```diff
@@ -0,0 +1,57 @@
...
+      end
+    
+    when "5"
+      exit
+    else
+      puts "\nChoose an option, 1, 2, 3 or 4".center(50).colorize(:light_yellow)
+    end
+  end
+end
+OrderItalian.new
```
winters145, line 57: The implementation of an interactive user menu is a really nice idea. Very in-depth and goes above and beyond the simple user story requirements, very impressive!

lib/order.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def initialize
+    @my_order = []
+  end
+
+  def view_menu
+    ITALIAN.each do |key, value|
+      puts "#{key},  £#{value}".colorize(:yellow).center(100)
+    end
+  end
```
winters145, line 24: Perhaps it would have been good to extract this method into it's own class, the Order class seems to be doing a lot here

lib/order.rb
```diff
@@ -0,0 +1,56 @@
...
+  end 
+
+  def view_my_order
+    print_out_my_order
+    puts "You're total is £#{total}".colorize(:light_yellow)
+  end
+
+  def send_sms_confirmation
+    require_relative "./send_delivery_sms"
+  end
```
winters145, line 39: Although this does seem to work, maybe the better practice is to have a SendDeliverySMS class that is called here?

lib/order.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def total
+    total = @my_order.map { |hash| hash.values }.flatten.sum
```
winters145, line 42: Clever use of the .flatten method! Not one I've come across before, definitely a handy one to know!

lib/order.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  attr_reader :my_order
+
+  def initialize
+    @my_order = []
+  end
+
+  def view_menu
+    ITALIAN.each do |key, value|
+      puts "#{key},  £#{value}".colorize(:yellow).center(100)
```
winters145, line 22: A nice idea of using string interpolation to make the menu easy to read & user friendly!

lib/send_delivery_sms.rb
```diff
@@ -0,0 +1,18 @@
+# Download the twilio-ruby library from twilio.com/docs/libraries/ruby
+require 'twilio-ruby'
+require 'dotenv/load'
+TIME = (Time.new + (60 * 30)).strftime("at %I:%M %p")
```
winters145, line 4: Nice clever bit of code for giving a specific time

spec/order_spec.rb
```diff
@@ -0,0 +1,57 @@
+require_relative '../lib/order'
+
```
winters145, line 2: good thorough tests, however could be better practice regarding the separating of classes and having their own spec files

lib/order.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  def add_dish(dish, quantity)
+    quantity.times do
```
winters145, line 27: Clever way of adding multiple quantities of items! Maybe setting a default quantity of 1 would help quash some user errors?

winters145 (commented):
Overall really impressed with how you've gone about this challenge, especially the incredibly in-depth and detailed user menu! Getting a bit of class extraction would really polish everything off. 
Really well done!

PR #2164, opened 2022-02-06
Takeaway challenge - Francesco's attempt

now-what:


lib/menu.rb
```diff
@@ -0,0 +1,32 @@
...
+      "Pollo alla cacciatora" => 7.50,
+      "Torta paesana" => 4.10,
+      "Caffè" => 0.90
+    }
+    @order = []
+  end
+
+  def view
+    @dishes.each do |dish, price|
+      puts "#{dish}: £#{price.to_f}"
```
AliCo-Hash, line 18: Really like the way £ sign was implemented into the puts.

spec/menu_spec.rb
```diff
@@ -0,0 +1,38 @@
+require_relative "../lib/menu"
+
+describe Menu do
+  let(:menu) { Menu.new }
+  # let(:dishes) { {"carbonara" => 3.20, "polpette" => 4.50} }
+  # let(:order) { Menu.new.choose }
+
+  it { is_expected.to respond_to(:view) }
```
AliCo-Hash, line 8: all of these methods are being tested on later. Is there a purpose to responding to them even thought they get tested?

PR #2165, opened 2022-02-06
Takeaway Challenge  Shirley Macfarlane

ShirleyMacfarlane:


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
```
Jaroslaw-Roman-Cudzich, line 1: For code readability, it may be good to split the classes into separate files for both the spec files and actual codes being tested.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,49 @@
+require './lib/takeaway'
```
Jaroslaw-Roman-Cudzich, line 1: Good practice to create a "describe" block specifying the name of the method being tested for each method.

PR #2168, opened 2022-02-06
Takeaway Challenge - Peter O

harmlessgoose:


Gemfile
```diff
@@ -10,4 +10,5 @@ end
```
meghanblyth, line 10: Well done, everything looks great here.

lib/takeaway.rb
```diff
@@ -0,0 +1,40 @@
+# require_relative 'send_sms'
```
meghanblyth, line 1: Maybe the takeaway file and spec could have been divided into separate classes, such as order or menu as well. Apart from that, this is very readable and easy to follow, well done!

lib/takeaway.rb
```diff
@@ -0,0 +1,40 @@
...
+
```
meghanblyth, line 2: I think adding some comments and a README with irb instructions would be great for anyone who doesn't understand your code :)

PR #2171, opened 2022-02-06
Takeaway challenge - Camilla

Camilla-Greene:


lib/phone.rb
```diff
@@ -0,0 +1,13 @@
+
+class Phone
+  attr_reader :text
```
leoht, line 3: You probably don't need to expose the value of the instance variable `@text`, so this `attr_reader` could be omitted — also, it seems the value of `@text` is never used in this class?

lib/phone.rb
```diff
@@ -0,0 +1,13 @@
...
+  def initialize(restaurant = Restaurant.new)
+    @restaurant = restaurant
+    @text = text
+  end  
+  
+  def receive
```
leoht, line 9: I have a few suggestions here:
 * The method `receive` expresses that the class would "receive" a message or a command from somewhere else, where instead what we'd like to do is to have this class *sending* the text message - so renaming the method to an active form (e.g `def send_message`) would be better at communicating this
 * The value of the text message (which is so far returned by `@restaurant.text` should probably be moved out of `Restaurant` into `Phone`, because the class `Restaurant` shouldn't "know" about the text message content, or how to send it — this is a concern for the `Phone` class)
 * This method should probably contain the code that is so far living in `lib/send_sms.rb` (the code that uses the Twilio library to send the text message)

lib/send_sms.rb
```diff
@@ -0,0 +1,18 @@
+# Download the helper library from https://www.twilio.com/docs/ruby/install
+require 'rubygems'
+require 'twilio-ruby'
+
+# Find your Account SID and Auth Token at twilio.com/console
+# and set the environment variables. See http://twil.io/secure
+account_sid = ENV['private']
```
leoht, line 7: This code should be wrapped in the class that is responsible to send the text message, so probably inside the `Phone` class, in the method `receive` (see my comments above on that method)?

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
+class Restaurant
+  
+  attr_reader :dishes_with_prices, :sum, :order
+  
+  def initialize
+    @dishes_with_prices = {
```
leoht, line 6: Nice to use a Hash to store the menu information :) 
You could probably remove the instance variable `@sum` — because you can recalculate it from the `@order` array, since it contains the different dishes in the order, you could change the method `total` so it iterates on the `@order` array to recalculate the total, saving you from having to manage this `@sum` variable. As a good rule of thumb, any value that can be calculated or derived from another one shouldn't be stored

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+  def select(dishes)
+    @order << @dishes_with_prices.select { |k, _v| k == dishes }
+    @sum += @dishes_with_prices.fetch(dishes)
+  end
+
+  def total
+    @sum
+  end
+
+  def ordered
```
leoht, line 35: Since you already used an attribute accessor `:order`, it means you can call `restaurant.order` to get the value of the order — this method `ordered` is doing exactly the same thing, so it can probably be removed since it's redundant?

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+    @order 
+  end
+  
+  def text
```
leoht, line 39: As I explained above, this should probably be moved to the `Phone` class

spec/phone_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'phone'
+require 'takeaway'
+
+describe Phone do
+  describe '#receive' do
+    it 'should receive a text message' do
+      text_double = double :text, text: "Thank you! Your order was placed and will be delivered before 18:00"
```
leoht, line 7: Since you'll move the Twilio code into `Phone`, you'll also have to modify this test so the class from the Twilio gem is mocked — and the job of this test will be to **check that the `Phone` class correctly called the Twilio library** — you can use this guidance about how to stub external libraries: https://github.com/makersacademy/course/blob/main/pills/levels_of_stubbing.md#2-stub-the-gem-associated-with-3rd-party-api

lib/takeaway.rb
```diff
@@ -0,0 +1,42 @@
...
+    @sum = 0
+    @order = []
+    
+  end
+
+  def list
+    @dishes_with_prices 
+  end
+
+  def select(dishes)
```
leoht, line 26: The argument name `dishes` should probably be renamed to be singular, since one dish is selected at a time — I think?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+describe Restaurant do
+  describe '#list' do
+    it 'lists dishes with prices' do
+      restaurant = Restaurant.new
+      expect(subject).to respond_to(:list)
+    end
+  end     
+  describe '#select' do  
+    it 'selects a dish from the menu' do
+      restaurant = Restaurant.new
```
leoht, line 12: This test is called 'selects a dish from the menu' - so what we want to test is that **we can select an item from the menu**  and that its saved correctly in the order

If we think of this test in the "arrange, act, assert" scaffolding, then what we should do in this test is...
1. Arrange: setup a new instance of `Restaurant` (you've done this already)
2. Act: **select** an item on the menu (using the method `.select)
3. Assert: **verify** that this item is present in the order (using the result returned by `restaurant.order` and asserting the selected item is in it)

Try modifying this test so it follows the structure above, and follow the same process of thinking to verify if your other tests below are correctly following this structure too — it's good to always think about *what* does the current test actually needs to verify, and *how* we can verify this

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    let(:key) { "Green Tea " }
+    let(:value) { 2 }
+    it 'tracks the sum of the order' do
+      expect { subject.select(dishes { key }) }.to change { subject.sum }.by value
+    end 
+  end
+  
+  describe '#total' do
+    it 'gives the total amount for the order' do
+      expect(subject).to respond_to(:total)
```
leoht, line 26: For this test as well, we're so far checking that the class implements a method named `:total`, but this is not enough — we want to verify that it gives the **correct total for the order made of the dishes we selected**

Again, let's try to think in "Arrange, Act, Assert":
1. Arrange - setup the `Restaurant` instance
2. Act - select one or two dishes on this instance, to build an order
3. Assert - get the `.total` number value, and verify it's the correct, expected one (using `expect(...).to eq(...)`)

leoht (commented):
Really nice work! :) I've left comments mainly about the code design and interactions between the two classes, and how to move the code using the Twilio library inside the `Phone` class. I also left a few comments on how to improve the tests to make sure they verify correctly what they need to verify, and how to think in "Arrange, Act, Assert" — 
let me know if you have any questions and I'm happy to discuss it further!

Re: the README file — really good to describe how you've approached the problem, it would be even better to add some instructions on how to install & setup, how to run the program (e.g in IRB), how to run the tests, etc. Here's a [nice guide](https://medium.com/@meakaakka/a-beginners-guide-to-writing-a-kickass-readme-7ac01da88ab3) on what to include in a README file for a project (though you don't have to include *all* of these things for the different challenges and exercises, but it's good to aim for at least setup & run instructions + description of the approach and how the program works at a high level)

PR #2174, opened 2022-02-06
Takeaway Challenge

CBuchan5:


lib/menu.rb
```diff
@@ -1,12 +1,14 @@
 class Menu
   attr_reader :list
-  
+
   def initialize(list)
   @list = list
```
nelsonclaire, line 6: I find rubocop -x is great for checking code layout (would it have said this was an extra line?)

spec/menu_spec.rb
```diff
@@ -11,6 +11,11 @@
   it "should have a list of dishes and prices" do 
     expect(subject.list).to eq (list)
   end
+
+  it "prints a list of dishes" do
+    print_menu = "Chicken £4.95, Tofu £3.95, Rice £2.00"
+    expect(subject.print_list).to eq(print_menu)
+  end
```
nelsonclaire, line 18: How build on this consider knowing how many of each is available to order?

lib/menu.rb
```diff
@@ -1,12 +1,14 @@
...
```
nelsonclaire, line 5: Could this be named with something else so its clearer what it is a list of?

PR #2176, opened 2022-02-07
takeaway challenge attemp

AliCo-Hash:


lib/menu.rb
```diff
@@ -0,0 +1,11 @@
+class Menu
+  attr_reader :dishes
+
+  def initialize
+    @dishes = [{:Burger=>"£10"}, {:Fries=>"£3"}]
+  end
+
+  def dishes
+    @dishes
+  end
```
now-what, line 10: Because you have `attr_reader : dishes`, I _think_ this method is redundant?

PR #2178, opened 2022-02-07
Takeaway challenge attempt

lykathleen:
My attempt at the takeaway challenge. User story 4 (Twilio) has not been completed.

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,34 @@
+class Order
+  attr_reader :order_list
+
+  def initialize
+    @total = 0
+    @menu = Menu.new.menu_list
+    @order_list = []
+  end
+
+  def select_dish(dish)
+    @order_list << dish
+  end
+
+  def check_order 
+    raise "An item(s) in your order is not available" if not_on_menu?
+  end 
+
+  def not_on_menu?
+    @menu.each do |item, _price|
+      @order_list.each do |dish|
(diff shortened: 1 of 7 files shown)


lib/takeaway.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  MENU = 
+  { "Pad Thai" => 10,
+    "Butter Chicken" => 10,
+    "Naan Bread" => 5
+  }.freeze
+  
+  def see_menu
+    MENU
```
Rakshak81, line 17: Might want to use iteration

lib/order.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def select_dish(dish)
+    @order_list << dish
+  end
+
+  def check_order 
+    raise "An item(s) in your order is not available" if not_on_menu?
+  end 
+
```
Lorn888, line 17: Clever to create a method that checks if the item is available.

spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
+require_relative '../lib/menu'
+
```
Lorn888, line 2: I appreciate seeing these comments, hinting the codes intention.

Lorn888 (commented):
Clear and logical approach to the user stories.

PR #2179, opened 2022-02-07
Jane's contribution to the takeaway challenge 

jparsons-git:
Jane's contribution to the takeaway challenge


docs/myREADME.md
```diff
@@ -0,0 +1,15 @@
...
+Ruby 3.0.2
+
+## Author
+Jane Parsons
+
+### Progress
+* I had to look at the code review rubric for a steer on this challenge
+
+### TODO
+* This solution does not include the Twilio api and connection
```
SarahM55, line 15: Nice, informative readme!

lib/menu.rb
```diff
@@ -0,0 +1,15 @@
+class Menu
+
+  attr_accessor :dishes
+
+  def initialize
+    @dishes = {}
+  end  
+
+  def load
```
SarahM55, line 9: Good use of a has here to store your menu! I think you could actually have popped the hash itself directly under the 'initialize' method then in your 'load' method, you could have written a simpler line which iterates over each 'pair' in the hash and 'puts' them :)

spec/menu_spec.rb
```diff
@@ -0,0 +1,14 @@
+require_relative '../lib/menu'
+
+describe Menu do
+  
+  subject(:menu) { described_class.new } # allows the use of 'dish' instead of 'subject' for clarity
```
SarahM55, line 5: Nice and clear!

spec/order_spec.rb
```diff
@@ -0,0 +1,53 @@
+require_relative '../lib/order'
+
+describe Order do
+  
+  subject(:order) { described_class.new } # allows the use of 'dish' instead of 'subject' for clarity  
+  
+  describe "#show_menu" do
```
SarahM55, line 7: Really like that you've broken up each 'function' into separate 'describe' blocks, makes it much easier to read!

spec/order_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      order.add_dish("Sweetcorn Patties", 2)
+      order.add_dish("Singapore Noodles (Vegetable)")
+      # order.check_order
+      expect { order.check_order }.to output("You have ordered:\nSweetcorn Patties x2 at £3.99 each\nSingapore Noodles (Vegetable) x1 at £4.99 each\nThe total cost of your order is £12.97\n").to_stdout
+   
+      expect(order.total).to eq 12.97
+    end
+  end
+
+end
```
SarahM55, line 53: These blocks are functional, but they're a wee bit long. For instance, I'm not sure that we need to 'show menu' each time - I think we can assume that if the user wants to check their order that they won't need to see the menu again :) 
I think that you could have removed your 'open_order' method too. You've already initialized your '@order_items' hash within your 'order.rb' file, so instead of calling the 'open_order' method each time, you could have just had your 'add_dish' method push the dish that was being selected into your '@order_items' hash!

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+
+class Order
+
+  attr_accessor :mymenu, :order_items, :total
+
+  def initialize
+    @order_items = {}
+    @total = 0
+    @mymenu = Menu.new
+  end
```
SarahM55, line 11: This looks good!

lib/order.rb
```diff
@@ -0,0 +1,49 @@
...
+    end  
+    puts @order_items
+    return @order_items
+  end  
+
+  def add_dish(dish, qty = 1)
+    q = @order_items[dish.to_sym]
+    q += qty
+    @order_items[dish.to_sym] = q
+  end
```
SarahM55, line 34: We've got a little bit of repitition in these methods - I've talked about how you could have avoided this in the comment I've left on your 'order_spec.rb' file :)

SarahM55 (commented):
Really good effort, looks like you've spent a lot of time working on this - you've set out the code really clearly too and used helpful 'describe' blocks so it's easy to see your thought process! There's a tiny bit of repetition in some of your methods, hopefully my comments help a bit :)

PR #2188, opened 2022-03-05
Maria's Takeaway Challenge

ruiined:
# Marija Gromovaja

# User stories 

- [ x ] User story 1: "I would like to see a list of dishes with prices"
- [ x ] User story 2: "I would like to be able to select some number of several available dishes"
- [ x ] User story 3: "I would like to check that the total I have been given matches the sum of the various dishes in my order"
- [ x ] User story 4: "I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered"

# README checklist

- [ x ] how to install,
- [ x ] how to run,
- [ x ] and how to test your code?


README.md
```diff
@@ -1,83 +1,23 @@
...
+
+_[Source](https://github.com/makersacademy/takeaway-challenge)_
+
+### How to Install
+1. Fork the repository
+2. Git clone to your machine
+3. Run `bundle`
+4. Enjoy
+
+### How to Run
```
arhussain1, line 12: Nice, your README is very clear and straight to the point.

lib/menu.rb
```diff
@@ -0,0 +1,12 @@
+class Menu
+  attr_reader :list
+  
+  def initialize
+    @list = [
```
arhussain1, line 5: Your Menu Class looks great!
- In the future if you wanted to add an item your menu, this class allows you to add that functionality whilst also following the SRP (Single Responsibility Principle)

Potential improvements:
- Might be better to keep individual 'meals' in a 'meal' class.
   - the 'meal' would be used to represent a single meal
   - it could have the 'name' and 'price' as attributes
   - if you needed to change the price or the name of the meal in future this is perfect place to do that as the Meal Class is the one describing any meal you create.

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def total_amount
+    total_amount = 0
+    @order.each { |item| total_amount += item[:price] }
+    total_amount.round(2)
+  end
+
+  def confirmation
+    total_amount
+    total_quantity == 1 ? item = "item" : item = "items"
```
arhussain1, line 29: Nice use of a ternery here

lib/takeaway.rb
```diff
@@ -0,0 +1,44 @@
...
+    # account_sid = 'x'
+    # auth_token = 'x'
+    # @client = Twilio::REST::Client.new(account_sid, auth_token)
+    # message = @client.messages.create(
+    # body: "Thank you! Your order has been confirmed. 
+    #  It will be with you before #{Time.now + 1*60*60}.",
+    # messaging_service_sid: 'x',
+    # to: '+')
+    # puts message.sid
+  end
```
arhussain1, line 43: Overall, I really learned a lot from your methods, they are not over complicated, short and do exactly what they say (they are named well).

Potential Improvement:
- the confirmation text should be extracted from this class into its own class. 
- Currently it looks like your Takeaway class has the following responsibilities:
  - It's responsible for selecting the the users meals
  - It's responsible for creating an order for the user
  - consider separating these responsibilities into separate classes.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'takeaway'
+
+describe TakeAway do
+  let(:takeaway) { TakeAway.new }
```
arhussain1, line 4: I may be wrong here please don't quote me on this. 

I think the doubles should only be used when you are depending on another class and it's methods. 

For example your Menu class would need a double here, just incase there is a mistake in the menu.rb file then it would cause an error here and it would be hard to figure out which file is causing that error.

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  it "shows the total amount of items" do
+    expect(takeaway.total_quantity).to eq 3
+  end
+
+  it "shows the total price of the order" do
+    expect(takeaway.total_amount).to eq 15.67
+  end
+
+  it "gives an order confirmation" do
+    expect(takeaway.confirmation).to be_an_instance_of(String)
```
arhussain1, line 28: I just learned this morning that you can use the RSpec Output Matcher to test to see if a certain string is returned by your method. For example:

```
it "gives an order confirmation" do
    expect do 
      takeaway.confirmation
    end.to_output("Whatever string you want here").to_stdout
end
```

spec/takeaway_spec.rb
ruiined: It's not a double, it's a stub, to be able to call takeaway.new easily

PR #2189, opened 2022-03-06
Jonny Abrams takeaway challenge

jonnyabrams:
Jonny Abrams

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,34 +19,17 @@ I would like to see a list of dishes with prices
 first step - made spec file but no class defined
 
 ```
-An error occurred while loading ./spec/menu_spec.rb.
-Failure/Error:
-  describe Menu do
-  
-  end
-
-NameError:
-  uninitialized constant Menu
-```
-
-it lets you see menu - expect to respond to see - menu.see returns nil in IRB so need to create a menu, array of hashes seems best bet, make it a constant and initialize with it
-
-can now see menu:
-
-```
-jonnyabrams@JONNYs-MacBook-Pro takeaway-challenge % irb -r './lib/menu.rb'
+jonnyabrams@JONNYs-MacBook-Pro takeaway-challenge % irb -r './lib/menu'
 3.0.2 :001 > menu = Menu.new
  => 
-#<Menu:0x000000014f9aa1f0
+#<Menu:0x0000000142857058
 ... 
-3.0.2 :002 > menu.see
-kimchi pancake: 5.99
-tteokbokki: 7.99
-kimchi-jjigae: 12.99
- => 
-[{:dish=>"kimchi pancake", :price=>5.99},
- {:dish=>"tteokbokki", :price=>7.99},
- {:dish=>"kimchi-jjigae", :price=>12.99}] 
+3.0.2 :002 > menu.browse
+bolognese: 7.5
+carbonara: 6.5
+puttanesca: 5.5
+arrabiata: 4.5
+ => {"bolognese"=>7.5, "carbonara"=>6.5, "puttanesca"=>5.5, "arrabiata"=>4.5} 
 ```
 
 2nd user story:
@@ -58,6 +41,28 @@ I would like to be able to select some number of several available dishes
 ```
 
 test drove creation of select method (with 2 args - 'number of dishes' in US) and basket
+
+```
+jonnyabrams@JONNYs-MacBook-Pro takeaway-challenge % irb -r './lib/menu.rb'
+3.0.2 :001 > menu = Menu.new
+ => 
+#<Menu:0x0000000139041fe8
+... 
+3.0.2 :002 > menu.select('tteokbokki', 2)
+ => 2 
+3.0.2 :003 > menu.select('kimchi pancake', 1)
+ => 1 
+3.0.2 :004 > menu.basket
+ => ["tteokbokki", "tteokbokki", "kimchi pancake"] 
+```
+
+3rd user story:
+
+```
+As a customer
+So that I can verify that my order is correct
+I would like to check that the total I have been given matches the sum of the various dishes in my order
+```
 
 
 Takeaway Challenge
--- a/README.md
+++ b/README.md
@@ -49,6 +49,16 @@ kimchi-jjigae: 12.99
  {:dish=>"kimchi-jjigae", :price=>12.99}] 
 ```
 
+2nd user story:
+
+```
+As a customer
+So that I can order the meal I want
+I would like to be able to select some number of several available dishes
+```
+
+test drove creation of select method (with 2 args - 'number of dishes' in US) and basket
+
 
 Takeaway Challenge
 ==================
--- a/README.md
+++ b/README.md
@@ -1,22 +1,73 @@
-set environment variable: https://www.twilio.com/blog/2017/01/how-to-set-environment-variables.html
+# Takeaway Challenge
+==================
+```
+                            _________
+              r==           |       |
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -14,70 +14,173 @@ Takeaway Challenge
...
+#<Menu:0x0000000142857058
+... 
+3.0.2 :002 > menu.browse
+bolognese: 7.5
+carbonara: 6.5
+puttanesca: 5.5
+arrabiata: 4.5
+ => {"bolognese"=>7.5, "carbonara"=>6.5, "puttanesca"=>5.5, "arrabiata"=>4.5} 
+```
+
```
PKilgarriff, line 85: Seeing the output like this is really helpful - both in terms of following your process and having the expected working output visible

README.md
```diff
@@ -14,70 +14,173 @@ Takeaway Challenge
...
+3.0.2 :001 > order = Order.new
+ => 
+#<Order:0x00000001563709b8
+... 
+3.0.2 :002 > order.select_dish('bolognese', 2)
+ => 2 
+3.0.2 :003 > order.select_dish('puttanesca')
+ => 1 
+3.0.2 :004 > order.basket
+ => [["bolognese", 7.5], ["bolognese", 7.5], ["puttanesca", 5.5]]
```
PKilgarriff, line 101: As you mentioned in your approach, this is an ideal data set to transform into a hash

README.md
```diff
@@ -14,70 +14,173 @@ Takeaway Challenge
...
+3.0.2 :006 > order.select_dish('bolognese')
+ => 1 
+3.0.2 :007 > order.basket
+ => {"carbonara"=>0.0, "arrabiata"=>4.5, "bolognese"=>7.5} 
+3.0.2 :008 > order.calculate_total
+ => 12.0 
+```
 
-In code review we'll be hoping to see:
+I also added a 'fail' guard clause in case of unavailable dishes:
```
PKilgarriff, line 133: Does this guard clause stop a user from continuing with their order after selecting an unavailable dish?

README.md
```diff
@@ -14,70 +14,173 @@ Takeaway Challenge
...
+ => 1 
+3.0.2 :004 > order.select_dish('puttanesca', 2)
+ => 2 
+3.0.2 :005 > order.checkout
+carbonara: 19.5
+bolognese: 7.5
+puttanesca: 11.0
+*************
+Your total is 38.0
+ => <Twilio.Api.V2010.MessageInstance body: Sent from your Twilio trial account - Thank you! Your order was placed and will arrive by 11:50 num_segments: 1 direction: outbound-api from: +16626678674 to: +447867513304 date_updated: 2022-03-06 10:51:00 +0000 price:  error_message:  uri: /2010-04-01/Accounts/ACadafd609098a3ba714c6f6e1917a6ce4/Messages/SM3bd3c82ea0f54256898ed9e9b6e47948.json account_sid: ACadafd609098a3ba714c6f6e1917a6ce4 num_media: 0 status: queued messaging_service_sid:  sid: SM3bd3c82ea0f54256898ed9e9b6e47948 date_sent:  date_created: 2022-03-06 10:51:00 +0000 error_code:  price_unit: USD api_version: 2010-04-01 subresource_uris: {"media"=>"/2010-04-01/Accounts/ACadafd609098a3ba714c6f6e1917a6ce4/Messages/SM3bd3c82ea0f54256898ed9e9b6e47948/Media.json"}>
```
PKilgarriff: This is very interesting, I don't think I was able to directly check that a message had been sent correctly - did you manage to use this return value for your tests?

README.md
```diff
@@ -14,70 +14,173 @@ Takeaway Challenge
...
+puttanesca: 11.0
+*************
+Your total is 38.0
+ => <Twilio.Api.V2010.MessageInstance body: Sent from your Twilio trial account - Thank you! Your order was placed and will arrive by 11:50 num_segments: 1 direction: outbound-api from: +16626678674 to: +447867513304 date_updated: 2022-03-06 10:51:00 +0000 price:  error_message:  uri: /2010-04-01/Accounts/ACadafd609098a3ba714c6f6e1917a6ce4/Messages/SM3bd3c82ea0f54256898ed9e9b6e47948.json account_sid: ACadafd609098a3ba714c6f6e1917a6ce4 num_media: 0 status: queued messaging_service_sid:  sid: SM3bd3c82ea0f54256898ed9e9b6e47948 date_sent:  date_created: 2022-03-06 10:51:00 +0000 error_code:  price_unit: USD api_version: 2010-04-01 subresource_uris: {"media"=>"/2010-04-01/Accounts/ACadafd609098a3ba714c6f6e1917a6ce4/Messages/SM3bd3c82ea0f54256898ed9e9b6e47948/Media.json"}> 
+```
+
+## Improvements Needed
+-------
+
+With a basket in the form of a hash, the bill cannot separate multiple orders of the same dish (although it does charge the correct amount).
```
PKilgarriff, line 175: Interesting edge case to work out - do you have any idea how you might approach it?

README.md
```diff
@@ -14,70 +14,173 @@ Takeaway Challenge
...
+-------
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance will make the challenge somewhat easier.  You should be the judge of how much challenge you want this at this moment.
+* Ruby
+* Rubocop for checking code style
+* Simplecov for checking test coverage
 
-Notes on Test Coverage
-------------------
+By [Jonny Abrams](https://github.com/jonnyabrams)
```
PKilgarriff, line 186: Always a good idea to include - makes it easier to browse your other repos!

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
...
+  attr_reader :dishes
+
+  def initialize
+    @dishes = {
+      'bolognese' => 7.50,
+      'carbonara' => 6.50,
+      'puttanesca' => 5.50,
+      'arrabiata' => 4.50
+    }
+  end
```
PKilgarriff, line 11: This works well - do you plan to implement an updateable menu so that the code can be re-used for other restaurants?

lib/menu.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  def browse
+    @dishes.each { |dish, price| puts "#{dish}: #{price}" }
+  end
+end
```
PKilgarriff, line 16: Very neat

lib/order.rb
```diff
@@ -0,0 +1,45 @@
+require_relative 'menu'
+require_relative 'send_sms'
+require 'twilio-ruby'
+
+class Order
+  NOT_ON_MENU_ERROR = 'This is not on the menu'.freeze
```
PKilgarriff, line 6: I always forget to freeze mutable objects in constants! Need to follow your lead

lib/send_sms.rb
```diff
@@ -0,0 +1,21 @@
+require 'rubygems'
+require 'twilio-ruby'
+require 'time'
+
+class SMS
+  def client
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
```
PKilgarriff, line 8: Great use of environment variables - do you have a resource that you can recommend for using them?

PKilgarriff (commented):
Very clean code - methods are clearly named and it's very readable. I'd consider not hard-coding the menu, but in this testing stage, it demonstrates that it works and the classes can interact nicely.

PR #2191, opened 2022-03-06
Nicola Cortese

NicolaCortese:
# Nicola Cortese

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


KennySB1:
Iterators that pull both the key and value can be used instead of storing the dish's hash value as an array

KennySB1:
text confirmation system sensibly implemented with ENV variables

PR #2193, opened 2022-03-06
Gaya's - Takeaway Challenge

gdrmvrm:
Gayathri Dharmavaram

I would really appreciate if you could please comment on how I am to approach mocking & doubles in this particular code.

# User stories 

# README checklist

- [x] how to install,
- [x ] how to run,
- [x ] and how to test your code?

### DIFF
--- a/lib/delivery.rb
+++ b/lib/delivery.rb
@@ -1,0 +1,27 @@
+require_relative 'order'
+require 'time'
+require 'rubygems'
+require 'twilio-ruby'
+
+class Delivery
+
+  def initialize
+    account_sid = ENV['TWILIO_ACCOUNT_SID']
+    auth_token = ENV['TWILIO_AUTH_TOKEN']
+    @client = Twilio::REST::Client.new(account_sid, auth_token)
+  end
+
+  def confirmation(order)
+   raise "Order pending!" if order.order_status == "pending"
+    time = Time.now
+    time += (60 * 60)
+
+    message = @client.messages
+    .create(
+       body: "Thank you! Your order was placed and will be delivered before #{time.strftime("%k:%M")}",
+       from: '+17652565176',
+       to: ENV['GAYA_MOBILE']
+     )
+    puts message.sid
+  end
+end
--- a/lib/dish.rb
+++ b/lib/dish.rb
@@ -1,0 +1,20 @@
+class Dish
+
+  attr_reader :name, :price
+  attr_writer :available
+
+  def initialize(name, price)
+    @name = name
+    @price = price
+    @available = true
+  end
+
+  def available?
+    @available 
+  end
+  
+  def availability
+   return available? ? "available" : "unavailable"
+  end
+
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,38 @@
+require_relative 'dish'
+require_relative 'order_item'
+
+class Order
+
+  attr_reader :basket, :order_status
+
+  def initialize
+    @basket = []
+    @order_status = "pending"
+  end
+
+  def add(order_item)
+    raise "Sorry, dish unavailable" if order_item.dish.available? == false
+    @basket << order_item
+  end
+
+  def print_order
+    @basket.each { |order_item|
+     puts "#{order_item.quantity} x #{order_item.dish.name}"
+    }
+  end
+
+  def calculate_total
+    total = 0
+    @basket.each { |order_item| total += (order_item.quantity * order_item.dish.price) }
+    return total
+  end
+
+  def verify_total(bill)
+    raise "Bill does not match the total sum of dishes ordered" if bill != calculate_total
+    return true
+  end
+
+  def place_order
+    @order_status = "accepted" 
+  end
+end
--- a/spec/delivery_spec.rb
+++ b/spec/delivery_spec.rb
@@ -1,0 +1,25 @@
+require 'delivery'
+require 'time'
+require 'order'
+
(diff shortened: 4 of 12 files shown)


lib/delivery.rb
```diff
@@ -0,0 +1,27 @@
+require_relative 'order'
+require 'time'
+require 'rubygems'
+require 'twilio-ruby'
+
+class Delivery
```
leoht, line 6: Since this class is responsible of sending an order confirmation text, it could be renamed to something like `DeliveryTextSender` to make it more explicit in the name?

lib/dish.rb
```diff
@@ -0,0 +1,20 @@
...
+    @price = price
+    @available = true
+  end
+
+  def available?
+    @available 
+  end
+  
+  def availability
+   return available? ? "available" : "unavailable"
```
leoht, line 17: Minor one - I'd suggest moving this logic into the `print_menu` method of the `Menu` class, because the goal of it is more to provide a human-readable label for the "availability" status, and `print_menu` is probably a better place for this (it would call `.available?` on the item and decide which string to print depending on the result)

lib/order.rb
```diff
@@ -0,0 +1,38 @@
+require_relative 'dish'
+require_relative 'order_item'
+
+class Order
+
+  attr_reader :basket, :order_status
+
+  def initialize
+    @basket = []
+    @order_status = "pending"
```
leoht, line 10: I'd suggest storing these strings `"pending"` and `"accepted"` etc into constants on this class, this way you can refer to these in the code and it avoids typos or other errors:
```ruby
class Order
  STATUS_PENDING = "pending"
  STATUS_ACCEPTED = "accepted"
  # ...
```

lib/order.rb
```diff
@@ -0,0 +1,38 @@
...
+    @basket.each { |order_item| total += (order_item.quantity * order_item.dish.price) }
+    return total
+  end
+
+  def verify_total(bill)
+    raise "Bill does not match the total sum of dishes ordered" if bill != calculate_total
+    return true
+  end
+
+  def place_order
```
leoht, line 35: Minor one - since the behaviour of this method is change the status to `accepted`, a clearer name might be `accept`?

spec/delivery_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      allow(Time).to receive(:now).and_return(@time)
+    end
+
+    it "raises error if order is pending" do
+      order = Order.new
+      expect{subject.confirmation(order)}.to raise_error("Order pending!")
+   end
+
+  #   it "provides confirmation of delivery when order is accepted" do
+  #     order = Order.new
```
leoht, line 20: Instead we should use something like a double to mock the `Order` class instead of directly instantiating the class:
```ruby
# this double will behave like a pending order
order = double(:order, order_status: 'pending')

expect { subject.confirmation(order) }.to raise_error("Order pending!")
```

spec/order_item_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'order_item'
+require 'dish'
+
+describe OrderItem do
+  dish1 = Dish.new("Salmon", 2)
```
leoht, line 5: Here again we should probably use a double that behaves like a `Dish`, instead of directly instantiating the real class

spec/order_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+
+describe Order do
+
+  it { is_expected.to respond_to(:add)}
+
+  it { is_expected.to respond_to(:print_order)}
+
+  order = Order.new
+  dish1 = Dish.new("Soup", 2.5)
```
leoht, line 12: Same remark as above re: mocking here, because `Order` depends on at least `Dish` and `OrderItem`, we should use doubles to mock these two dependencies in our tests

spec/order_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  it "knows that order status is pending by default" do
+    expect(order.order_status).to eq "pending"
+  end
+
+  it "adds a dish to an order basket" do
+    expect(order.add(order_item)).to eq [order_item]
+  end
+
+  it "raises an error if the dish is unavailable" do
+    dish1.available=false
```
leoht, line 24: Rather than reusing the `dish1` and modifying it, it might be better to define a new, different `dish` double that behaves like an unavailable dish instance:
```ruby
unavailable_dish = double(:unavailable_dish, :available? => false)
unavailable_order_item = double(:unavailable_order_item, dish:  unavailable_dish)

expect{order.add(unavailable_order_item)}.to raise_error("Sorry, dish unavailable")
```

spec/order_spec.rb
leoht: Also, it might be better to wrap the initialisation of these different variables in a `before` block:
```ruby
before do 
   @order = Order.new
   @dish1 = Dish.new("Soup", 2.5)
   @order_item = OrderItem.new(dish1, 2)
end
```

By doing this (and using instance variables so these variables are still available in each test example block), we make sure that the values are reinitialised before each test. Since each test should be independent of the others, we want to have freshly created instances of order, dish, etc, for each test that runs

leoht (commented):
Great work :) It would be nice having an extra file, e.g `main.rb` or `example_usage.rb` that sets out how the classes interact together, a bit like a feature test in IRB. Otherwise I had a few comments on how to improve a bit the code and how to use doubles to mock dependencies - let me know if you have any further question!

PR #2194, opened 2022-03-06
Valentina Maggio Takeaway challenge - Week 2

valentina-maggio:
# Your name

Valentina Maggio

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


lib/menu.rb
```diff
@@ -0,0 +1,26 @@
...
+    }
+  end
+
+  def view_dishes
+    @dishes.each { |item, price| puts "#{item} £#{price}" }
+  end
+
+  def check_availability(item)
+    msg = "Dish not available. Please make a new selection."
+    raise msg if @dishes[item].nil?
```
HCVF, line 24: Nice, simple way of checking if the item exists

lib/menu.rb
```diff
@@ -0,0 +1,26 @@
+class Menu
```
HCVF, line 1: SRP followed well by making Menu a different class to Order

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+    @menu.check_availability(item)
+    @basket[item] += @menu.dishes[item]
+  end
+
+  def remove_dish(item)
+    check_basket(item)
+    @basket[item] -= @menu.dishes[item]
+  end
+
+  def view_basket
```
HCVF, line 23: Nice DRY method here

README.md
```diff
@@ -14,20 +14,88 @@ Takeaway Challenge
...
+        from /Users/Valentina/Projects/makers_projects/takeaway-challenge/lib/order.rb:31:in `place_order'
+        from (irb):10:in `<main>'
+        from /Users/Valentina/.rvm/rubies/ruby-3.0.2/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
+        from /Users/Valentina/.rvm/rubies/ruby-3.0.2/bin/irb:23:in `load'
+        from /Users/Valentina/.rvm/rubies/ruby-3.0.2/bin/irb:23:in `<main>'
+3.0.2 :009 > order.place_order(10)
+ => <Twilio.Api.V2010.MessageInstance body: Sent from your Twilio trial account - Thank you! Your order was placed and will be delivered before 17:34 num_segments: 1 direction: outbound-api from: +************* to: +************* date_updated: 2022-03-06 16:34:03 +0000>
+```
+
+Commands explained
```
HCVF, line 79: Very clear explanation of how the program is used

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+class Order
+
+  attr_reader :basket, :menu
+
+  def initialize(menu = Menu.new)
+    @basket = Hash.new(0)
+    @menu = menu
+  end
+
+  def add_dish(item)
```
HCVF, line 13: Maybe this method could take an optional 2nd argument so you can add more than one of the item at a time
Something along the lines of `def add_dish(item, quantity = 1)`

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+    @menu.check_availability(item)
+    @basket[item] += @menu.dishes[item]
+  end
+
+  def remove_dish(item)
```
HCVF, line 18: Same as above so you can quickly remove the 9 items we accidentally added to the basket with the previous method :)

lib/order.rb
valentina-maggio: Yes I totally agree! I thought about it for a bit but eventually decided to keep it simple. Next time I'll do as you suggest :) Thanks!

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+
+  def review_payment(payment)
+    msg = "Please enter the correct payment amount."
+    raise msg if payment != checkout_total
+  end
+
+  def thank_you(payment, confirmation)
+    confirmation.send_message if payment == checkout_total
+  end
+end
```
HCVF, line 77: Edge cases well considered throughout

lib/order.rb
valentina-maggio: True! I didn't think about that scenario!!! :D

spec/text_message_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  it 'can create a TextMessage' do
+    expect(text).to be_a_kind_of(TextMessage)
+  end
+
+  it 'responds to send_message method' do
+    expect(text).to respond_to(:send_message)
+  end
+
+  describe '#send_message' do
```
HCVF, line 15: Twilio API stubbed to prevent sending of messages during testing

spec/order_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    it 'stores a dish in the order' do
+      my_order.add_dish('Pizza')
+      expect(my_order.basket).to include('Pizza')
+    end
+  end
+
+  describe '#remove_dish' do
+    it 'removes a dish from the order' do
+      my_order.add_dish('Burger')
```
HCVF, line 27: Since many methods require a dish to be added to the order for testing you could create a 'context' where all instances start with an order added. However I can see why you might not have decided to do that as your tests are very neatly organised and implementing that would require tests for different methods to be grouped into the same context rather than just all grouped by method.

spec/text_message_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'text_message'
+
+describe TextMessage do
+  subject(:text) { described_class.new(client) }
```
HCVF, line 4: This really helps improve test readability esp. when frequently switching between files

spec/text_message_spec.rb
valentina-maggio: Not sure I did this right. I'll ask the coaches for a feedback...

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+    @basket[item] += @menu.dishes[item]
+  end
+
+  def remove_dish(item)
+    check_basket(item)
+    @basket[item] -= @menu.dishes[item]
+  end
+
+  def view_basket
+    check_empty_basket
```
leoht, line 24: Minor one here - why raise an error, rather than doing a fallback to printing an empty basket? Usually, raising exceptions is done to prevent an error/illegal situation from happening, but we could think of having an empty basket an "ok" case, it would just print an empty basket with a total of 0. But that's probably a subjective discussion here :D

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+    @basket.each do |item, price| 
+      if price.positive?
+        qty = (@basket[item] / @menu.dishes[item]).round
+        puts "#{qty}x #{item} £#{price}"
+      end
+    end
+  end
+
+  def checkout_total
+    @basket.select { |_, price| price.positive? }.values.reduce(:+)
```
leoht, line 52: Nice functional way of calculating the total here!

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+  def print_total
+    puts "----------------"
+    puts "Order Total £#{checkout_total}"
+  end
+  
+  def empty_basket?
```
leoht, line 60: These methods could probably be set as private, as they're not meant to be called from the outside of the class

lib/order.rb
```diff
@@ -0,0 +1,77 @@
...
+    @basket[item] -= @menu.dishes[item]
+  end
+
+  def view_basket
+    check_empty_basket
+    print_subtotals
+    print_total
+  end
+
+  def place_order(payment, confirmation = TextMessage.new)
```
leoht, line 29: Minor one - I'd rename the variable/argument `confirmation` to `text_message`, to keep the naming consistent with the class name

spec/order_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+      expect { my_order.remove_dish('Paella') }.to raise_error msg
+    end
+  end
+
+  describe '#view_basket' do
+    it 'prints the items in the basket with quantities and prices and the total' do
+      my_order.add_dish('Olives')
+      my_order.add_dish('Steak')
+      expect { my_order.view_basket }.to output(<<-output
+1x Olives £2.5
```
leoht, line 43: Nice use of HEREDOC syntax here for the formatted expected output

spec/text_message_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    expect(text).to be_a_kind_of(TextMessage)
+  end
+
+  it 'responds to send_message method' do
+    expect(text).to respond_to(:send_message)
+  end
+
+  describe '#send_message' do
+    it 'sends a confirmation text message' do
+      allow(client).to receive_message_chain(:messages, :create).and_return("Thank you for your order!")
```
leoht, line 17: Oh nice - I didn't know `receive_message_chain`!

spec/text_message_spec.rb
leoht: It looks like you've done it right - stubbing the client, a way to improve it further would be to return a double that behaves like the response object returned by the library. From [their example here](https://www.twilio.com/docs/libraries/ruby), it seems we should be able to call `message.id` on the result, so maybe returning a double that implements the `sid` method (instead of a human-readable string) would be better. It probably doesn't really change things in your case though, as you're not using this `sid` after the library returns the result

```ruby
allow(client).to receive_message_chain(:messages, :create).and_return(
  double(:fake_result, sid: 232424)
 )
```

spec/order_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    it 'raises an error if basket is empty' do
+      msg = "Your basket is currently empty. Please add item first."
+      expect { my_order.view_basket }.to raise_error msg
+    end
+  end
+
+  describe '#place_order' do
+    
+    let(:text) { double :text, send_message: "Thank you for your order!" }
```
leoht, line 59: In your code, it seems `send_message` returns directly the return value of `@client.messages.create`, which is not a human-readable string, but probably some kind of result object from the Twilio gem. If we wanted our double to return something that behaves like this, we could make `send_message` return a double, instead of a string:

```ruby
let(:fake_twilio_result) { double(:fake_twilio_result, sid: 4344343) }
let(:text) { double :text, send_message: fake_twilio_result }
```

spec/order_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'order'
+
+describe Order do
+  subject(:my_order) { described_class.new }
```
leoht, line 4: Since `Order` depends on `Menu` (and this dependency can be injected via `new`), we can use a double here to mock the behaviour of `Menu` - e.g the method `dishes` and `check_availability`, otherwise our tests are dependent on `Menu` being working correctly (so not fully isolated)

leoht (commented):
Great work and really good README! I left a few comments to improve the code and tests on various things - let me know in case you have any questions :)

lib/order.rb
valentina-maggio: You're totally right actually! I think I just like this raise error method too much, so I'm putting it everywhere :D The view_basket method should just print an empty basket as you said. I will amend it. Thanks!

lib/order.rb
valentina-maggio: Those methods are actually private :) The only public methods in the Order class are: add_dish, remove_dish, view_basket, place_order :)

lib/order.rb
valentina-maggio: Good point!

spec/order_spec.rb
valentina-maggio: Is it called like this? I saw it in the debugging card game from the workshop skills and I copied the syntax :D I will investigate further this HEREDOC syntax :)

spec/text_message_spec.rb
valentina-maggio: The message chaining was explained in a resource you shared in our channel last week. I found the command very useful and tried to use it in the this test :)

spec/order_spec.rb
valentina-maggio: I see! I was testing the content of the message rather than the message sid which was the actual return value of the method. Got it!

spec/order_spec.rb
valentina-maggio: I think I tried to do it but I wasn't really able to implement it. I will have a look at it again. Thanks!

lib/order.rb
leoht: Oh yes sorry I completely missed the `private` keyword above 😅

PR #2195, opened 2022-03-06
Chris' take-away challenge

clovellbsc:
Chris Lovell

# User stories 

- [ ] User story 4: "I would like to receive a text such as "Thank you! Your order was placed and will be delivered before 18:52" after I have ordered" # attempted but loaderror with twilio-ruby gem

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,50 @@
+require_relative 'menu'
+# require_relative 'sms'
+
+class Order
+  attr_reader :basket
+
+  def initialize 
+    @basket = []
+    #eventually would be equal to a premade menu object (i.e Italian_restaurant) or would be able to select a premade menu
+    @menu = Menu.new   
+  end
+
+  def add_dish_to_basket(dish)
+    count = 0
+    @menu.menu.each { |food| @basket << food && count +=1 if food.name == dish }
+    raise 'There is no such dish on the menu' if count == 0
+  end
+
+  def remove_dish_from_basket(dish) 
+    matching_dishes = @basket.select { |food| food.name == dish}
+    if matching_dishes.count >= 1 
+      @basket.delete_at(@basket.index(matching_dishes[0])) 
+    else
+      raise "This dish is not currently in your basket"
+    end
+  end
+
+  def calculate_order_price
+    @cost = 0
+    @basket.each { |dish| @cost += dish.price }
+    if basket.empty?
+      puts "Your basket is empty"
+    else 
+      puts "Your order total is currently £#{@cost}"
+    end
+    @cost
+  end
+
+  def complete_order
+    calculate_order_price
+    @basket = []
+    # send_sms #would be added once twilio functionality added
+  end
+
+  # def send_sms  
+  #   text = SMS.new
+  #   message = "Thank you! Your order was placed and will be delivered before #{(Time.now + 3600).strftime("%k:%M")}"
+  # end
+
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,5 +1,5 @@
 require_relative 'menu'
-# require_relative 'sms'
+require_relative 'sms'
 
 class Order
   attr_reader :basket
@@ -39,12 +39,15 @@ class Order
   def complete_order
     calculate_order_price
     @basket = []
-    # send_sms #would be added once twilio functionality added
+    send_sms #would be added once twilio functionality added
   end
 
-  # def send_sms  
-  #   text = SMS.new
-  #   message = "Thank you! Your order was placed and will be delivered before #{(Time.now + 3600).strftime("%k:%M")}"
-  # end
+  private
+
+  def send_sms  
+    text = SMS.new
+    message = "Thank you! Your order was placed and will be delivered before #{(Time.now + 3600).strftime("%k:%M")}"
+    text.send_sms(message)
+  end
 
 end
--- a/spec/menu_spec.rb
+++ b/spec/menu_spec.rb
@@ -1,0 +1,35 @@
+require 'menu'
+
+describe Menu do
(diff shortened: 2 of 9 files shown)


lib/order.rb
```diff
@@ -0,0 +1,53 @@
+require_relative 'menu'
+require_relative 'sms'
+
+class Order
+  attr_reader :basket
+
+  def initialize 
+    @basket = []
+    #eventually would be equal to a premade menu object (i.e Italian_restaurant) or would be able to select a premade menu
+    @menu = Menu.new
```
leoht, line 10: Ideally this would be dependency injected via an argument to `initialize` rather than hardcoded

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  def initialize 
+    @basket = []
+    #eventually would be equal to a premade menu object (i.e Italian_restaurant) or would be able to select a premade menu
+    @menu = Menu.new   
+  end
+
+  def add_dish_to_basket(dish)
+    count = 0
+    @menu.menu.each { |food| @basket << food && count +=1 if food.name == dish }
```
leoht, line 15: This could be simplified by - for example - adding a method `find_dish` on the `Menu` class which would find one dish by name (or return `nil`), and so we could now simply do:
```ruby
dish = @menu.find_dish(dish_name)
raise 'There is no such dish on the menu' if dish.nil? 
@basket << dish
```

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+    raise 'There is no such dish on the menu' if count == 0
+  end
+
+  def remove_dish_from_basket(dish) 
+    matching_dishes = @basket.select { |food| food.name == dish}
+    if matching_dishes.count >= 1 
+      @basket.delete_at(@basket.index(matching_dishes[0])) 
+    else
+      raise "This dish is not currently in your basket"
```
leoht, line 24: Minor one - for readability, prefer "short-circuit" guards like this, so you don't have to write an indented if/else statement:
```ruby
raise "This dish is not currently in your basket" if matching_dishes.empty?

@basket.delete_at(@basket.index(matching_dishes[0]))
```

lib/order.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def calculate_order_price
+    @cost = 0
+    @basket.each { |dish| @cost += dish.price }
+    if basket.empty?
+      puts "Your basket is empty"
```
leoht, line 32: Ideally this method shouldn't print out messages with `puts` - because we'd expect this method to only calculate and return the total price (from its name `calculate_order_price`). If we wanted to print out messages, this would probably belong to a different class, which would be responsible for presenting the data of the program to the user, something like `OrderPrinter`

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'order'
+
+describe Order do
+  let(:pasta) { double :pasta }
+  let(:paneer_tikka_masala) { double :paneer_tikka_masala}
```
leoht, line 5: Good use of doubles here to mock the dependency on `Dish` and `Menu` :)

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  it 'outputs "Your basket is empty if you attempt to calculate the cost of an empty basket' do
+    expect { subject.calculate_order_price }.to output("Your basket is empty\n").to_stdout
+  end
+
+  context 'with mocking menu class' do
+
+    before do
+      allow(Menu).to receive(:new).and_return(menu)
```
leoht, line 22: So you can avoid stubbing `Menu.new` if you are able dependency-inject the `menu` double into `Order.new` (like mentioned by my comment above), rather than having `Order#initialize` controlling how it creates a new Menu

spec/menu_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  it 'can add a named dish to the menu' do
+    allow(Dish).to receive(:new).and_return(dish)
+    subject.add_dish('paneer_tikka_masala', 10)
+    expect(subject.menu).to eq([dish])
+  end
+
+  it 'can remove a specific dish from the menu' do
+    allow(Dish).to receive(:new).and_return(dish)
```
leoht, line 22: Same remark as for the other class, if you were to inject the different dish instances into `Menu.new` (right now, they're hardcoded in `Menu#initialize`), you wouldn't have to stub `Dish.new` and could inject doubles in this test, in place of the concrete `Menu` instances

spec/order_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'order'
+
+describe Order do
```
leoht, line 3: It would be good to add at least one test verifying that the `complete_order` method calls `SMS#send_sms` - ideally you'd inject this dependency as well, for example:
```ruby
class Order
  def initialize(sms)
     @sms = sms 
  end 

  # ... 
  def send_sms  
    @sms.send_sms(message)
  end
```

rather than...
```ruby
def send_sms  
    text = SMS.new
    message = "Thank you! Your order was placed and will be delivered before #{(Time.now + 3600).strftime("%k:%M")}"
    text.send_sms(message)
  end
```

Then you'd be able to inject a double for this dependency in the test. You could then finally assert that the method `send_sms` was called on the double, after `complete_order` was called, using RSpec's expectations on receiving messages: https://relishapp.com/rspec/rspec-mocks/docs/basics/expecting-messages

leoht (commented):
Nice work :) I left a few minor comments on various improvements, and I guess the next step would be to change how your dependencies are injected so you can inject your doubles and do mocking this way - let me know if you have any question on the feedback

PR #2203, opened 2022-03-07
Dennis' takeaway

dennihous:
# Dennis Houston

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,4 +1,6 @@
 class Menu
+  attr_reader :menu
+
   def initialize
     @menu = {
       "Pizza" => 8,
@@ -12,7 +14,6 @@ class Menu
   end
 
   def see_menu
-    @menu.each {|dish, price| puts "#{dish} - £#{price}"}
+    @menu.each { |dish, price| puts "#{dish} - £#{price}" }
   end
 end
-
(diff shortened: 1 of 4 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,19 @@
...
+      "Special Fried Rice" => 5.5,
+      "Kebab" => 6.5,
+      "Ice Cream" => 3.2,
+      "Chocolate Cake" => 2,
+    }
+  end
+
+  def see_menu
+    @menu.each { |dish, price| puts "#{dish} - £#{price}" }
+  end
```
felixtaljaard, line 18: Good string interpolation, nice simple method for seeing menu

lib/order.rb
```diff
@@ -0,0 +1,18 @@
...
+
+class Order
+  def initialize(menu = Menu.new)
+    @new_menu = menu
+    @basket_total = 0
+    @order = []
+  end
+
+  def add_order(dish)
+    raise "Item not on the menu" unless @new_menu.menu.key?(dish)
```
felixtaljaard, line 11: Are these menu and key? methods undefined?

PR #2204, opened 2022-03-07
Henry's takeaway

HCVF:
Henry Ferrari

# User stories 

# README checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,3 +11,6 @@ end
 group :development, :test do
   gem 'rubocop', '1.20'
 end
+
+gem 'twilio-ruby', '~> 5.65.0'
+gem 'rack'
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,29 @@
+class Menu 
+  def initialize
+    @order = []
+    @menu_hash = {
+      "kung po chicken" => 7,
+      "crispy chilli beef" => 9,
+      "crispy duck pancakes" => 8,
+      "crispy seaweed" => 2,
+      "spare ribs" => 5,
+      "egg fired rice" => 4,
+      "chow mein" => 6
+    }
+  end
+
+  def view_menu
+    @menu_hash.each_with_index do |(key, value), index|
+      puts "number: #{index} | #{key} | £#{value}"
+    end
+  end
+
+  def add_to_order(name, amount)
+    amount.times { @order << { name => @menu_hash[name] } }
+    puts "#{@menu_hash[name]}"
+  end
+  
+  def view_order
+    puts @order
+  end
+end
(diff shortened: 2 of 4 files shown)


Gemfile
```diff
@@ -11,3 +11,6 @@ end
 group :development, :test do
   gem 'rubocop', '1.20'
 end
+
+gem 'twilio-ruby', '~> 5.65.0'
+gem 'rack'
```
valentina-maggio, line 16: Good, well organised Gemfile

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+      puts "number: #{index} | #{key} | £#{value}"
+    end
+  end
+
+  def add_to_order(name, amount)
+    amount.times { @order << { name => @menu_hash[name] } }
+    puts "#{@menu_hash[name]}"
+  end
+  
+  def view_order
```
valentina-maggio, line 26: Is this a task the Menu should be in charge of? Or perhaps a different class?

lib/menu.rb
```diff
@@ -0,0 +1,29 @@
...
+      "crispy seaweed" => 2,
+      "spare ribs" => 5,
+      "egg fired rice" => 4,
+      "chow mein" => 6
+    }
+  end
+
+  def view_menu
+    @menu_hash.each_with_index do |(key, value), index|
+      puts "number: #{index} | #{key} | £#{value}"
```
valentina-maggio, line 17: Good use of the hash method to print the menu!
Just one things: I can see the first 'number: index' that is printed is 'number: 0'. Is it really what you wanted? Wouldn't it be better to start from 1?

spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
+require "menu"
+describe Menu do
+
+  describe "#view_menu" do
```
valentina-maggio, line 4: Did you also test that the view_menu method is actually returning what you expect it should return, ie the list of dishes with prices?
Same for the add_to_order method. Did you test that is returns an array with the items that are selected?

spec/menu_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+      expect(subject).to respond_to(:view_menu)
+    end
+  end
+
+  describe "#add_to_order" do
+    it "responds to add_to_order" do
+      expect(subject).to respond_to(:add_to_order)
+    end
+  end
+end
```
valentina-maggio, line 15: Have you tested the view_order method?

valentina-maggio (commented):
Overall, the code is clean and easy to understand. You clearly implemented the first two user stories, so well done! Just one suggestion, I think you could create a new class and move some tasks there (but that's just my opinion :) ). Also I'm not sure you fully tested the Menu class and its method.

PR #2207, opened 2022-03-07
Jasmine's takeaway-challenge

jazzc001:
# Jasmine Cheung

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,20 @@
+class Menu
+  DISHES = {
+    "bruschetta" => 3,
+    "spaghetti carbonara" => 8,
+  }
+
+  def dish?(dish)
+    DISHES.key?(dish)
+  end
+
+  def price(dish)
+    raise "'#{dish}' is not on the menu" unless dish?(dish)
+
+    DISHES[dish]
+  end
+
+  def print
+    DISHES.each { |dish, price| puts "* #{dish}: £#{price}" }
+  end
+end
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,28 @@
+class Order
+  attr_reader :dishes
+
+  def initialize(menu, message_sender)
+    @dishes = Hash.new(0)
+    @menu = menu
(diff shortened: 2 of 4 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,20 @@
+class Menu
+  DISHES = {
+    "bruschetta" => 3,
+    "spaghetti carbonara" => 8,
+  }
+
```
r94o, line 6: Good to use a constant here as the dishes will not change

lib/menu.rb
```diff
@@ -0,0 +1,20 @@
...
+  def dish?(dish)
+    DISHES.key?(dish)
+  end
+
+  def price(dish)
+    raise "'#{dish}' is not on the menu" unless dish?(dish)
+
```
r94o, line 13: nice guard clause, very readable

lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+
+    @dishes[dish] += quantity
+  end
+
+  def total
+    return 0 if @dishes.empty?
+
+    subtotals = @dishes.map { |dish, quantity| @menu.price(dish) * quantity }
+    subtotals.sum
+  end
```
r94o, line 21: Think could can replace the .map block with .sum block and therefore be able to remove line 20

lib/order.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def place
+    raise "cannot place an empty order" if @dishes.empty?
+
```
r94o, line 25: Good edge case checking

r94o (commented):
Nice one! Tests are isolated and cover edge cases. Code is fairly readable. I assume message_sender is another (to be implemented) class which will hold the Twilio code?

The only thing I would suggest is creating a dish class so DISHES could store an array of the Dish class so you could then just reference the price by Dish.price rather than DISHES["spaghetti carbonara"] which may make it slightly more readable

PR #2210, opened 2022-04-03
Conor Hope's Takeaway challenge

Conor-Developer:
# Conor Hope

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


spec/dishes_spec.rb
```diff
@@ -1,9 +1,12 @@
 require_relative '../lib/dishes'
 
 describe Dishes do
-  context '#view_files' do
-    xit 'Checks whether a string has been return' do
```
monenavarro: "returned"

lib/verify_order.rb
```diff
@@ -2,12 +2,18 @@
...
       puts "#{element["Name"]} : £#{element["Price"]}"
     end
-    puts "\nNumber of items: #{basket.count}"
-    puts view_total(basket)
+    number_of_items(basket)
+    view_total(basket)
+    basket
   end
 
   def text_order_confirmation
```
monenavarro, line 19: methods are clear and easy to understand

monenavarro:
very readable code and DRY
great use of the CVS file to display menu items and prices- structure of data is good as it stores it one area

PR #2212, opened 2022-04-03
Takeaway challenge submitted - Joe McNally

jmcnally17:
# Your name

Joe McNally

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -8,6 +8,10 @@ class Menu
   end
 
   def list_items
-    @items.map { |item| "#{item[:name]}: £#{item[:price]}0" }.join("\n")
+    @items.map do |item| 
+      dish = item[:name]
+      cost = item[:price]
+      "#{dish}: £#{cost}0"
+    end.join("\n")
   end
 end
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,13 @@
+class Menu
+  def initialize
+    @items = [
+      { name: "Korma", price: 8.50 },
+      { name: "Jalfrezi", price: 9.00 },
+      { name: "Tikka Masala", price: 8.00 }
+    ]
+  end
+
+  def list_items
+    @items.map { |item| "#{item[:name]}: £#{item[:price]}0" }.join("\n")
+  end
+end
--- a/lib/takeaway.rb
+++ b/lib/takeaway.rb
@@ -1,0 +1,3 @@
+class Takeaway
+
+end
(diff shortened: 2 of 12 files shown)


spec/takeaway_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+                                                  "Jalfrezi: £9.00\n" +
+                                                  "Tikka Masala: £8.00")) 
+  }
+  let (:empty_order_double) { double(:empty_order, total: 0.0, bill: "\nTotal: £0.00") }
+  let (:empty_order_class_double) { double(:empty_order_class, new: empty_order_double) }
+  let (:subject) { Takeaway.new(menu_double, empty_order_class_double) }
+
+  let (:order_double) { double(:order, add: true, bill: "Korma: £8.50\nJalfrezi: £9.00\nTotal: £17.50") }
+  let (:order_class_double) { double(:order_class, new: order_double) }
+  let (:takeaway) { Takeaway.new(menu_double, order_class_double) }
```
mcsuGH, line 14: Good use of doubles

spec/order_spec.rb
```diff
@@ -0,0 +1,31 @@
+require_relative '../lib/order'
+
+describe Order do
+  let(:menu_double) { double(:menu, list_items: "Korma: £8.50\n" +
+                                                "Jalfrezi: £9.00\n" +
+                                                "Tikka Masala: £8.00")
+  }
+
```
mcsuGH, line 9: Forgot one line to instantiate subject using the double created

lib/takeaway.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def check_bill
+    no_order?
+    @current_order.bill
+  end
+
+  def add_to_order(dish)
+    no_order?
+    @current_order.add(dish)
+  end
```
mcsuGH, line 27: maybe can find a way to add quantity in there, so that it is easier to order multiple of a single item

lib/menu.rb
```diff
@@ -0,0 +1,17 @@
...
+    ]
+  end
+
+  def list_items
+    @items.map do |item| 
+      dish = item[:name]
+      cost = item[:price]
+      "#{dish}: £#{cost}0"
+    end.join("\n")
+  end
```
mcsuGH, line 16: maybe try remove ".join("\n")" and instead try to just puts on line 14 instead

mcsuGH (commented):
Good, README has clear instructions, code works, all tests are passed, code readable. Good use of doubles in spec files and also not using attribute readers anywhere - very good.

Can maybe write in README if some things are missing / yet to be done (such as spec for text client) but other than that + small edits here and there, very good.

EDIT: turns out there already was a section in the README for the things yet to be done.

PR #2217, opened 2022-04-04
Monica Navarro - SMS feature not working

monenavarro:
# Your name

Monica Navarro

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


almorcrette:
Doesn't look like twilio-ruby is in the gemfile, so the functionality isn't available to someone else using the program. If you add twilio-ruby to the gemfile, then another user who uses your code can run bundler and get twilio-ruby gem installed to use.

almorcrette:
You could update the README from the instructions to your own explanation to someone looking at your program

almorcrette:
If you update your README it could include instructions on how to run the program from a user perspective.

almorcrette:
You isolated all your class unit tests which is great. None of your class unit tests depend on other real objects

almorcrette:
Your rspec fails to run because of sms_spec.rb (because twilio isn't in the bundler so problems loading that) but when sms_spec.rb and sms.rb are removed from the testing, all your tests pass and you have 100% coverage. Great!

PR #2225, opened 2022-05-02
Timbo's Takeaway

timbo-graham:
# Your name

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


lib/order.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  attr_reader :order, :order_total, :simp_order
+
+  def initialize
+
+    @menu = Menu.new
+    @order = []
+    @order_total = 0
+    @simp_order = ""
+
```
jimmy-lyons, line 29: I'd put all the text stuff into its own class. I did have difficulties getting it to work from the order object, but my feeling was since it's a new noun, it should probably be its own class.

lib/order.rb
```diff
@@ -0,0 +1,103 @@
...
+    "Thanks! Your order has been received and will be with you by #{arrival_time}"
+
+  end
+
+  def order_summary
+
+    "Your order so far is: #{simplify_order}"
+
+  end
+
```
jimmy-lyons, line 74: I think this should be re-written so as not to be hard coded, but I like that you added a function that states the quantity of each item - that's not something I did.

lib/menu.rb
```diff
@@ -0,0 +1,22 @@
+class Menu
+
+  attr_reader :menu
+
+  def initialize
+
```
jimmy-lyons, line 6: This could be re-written to import different menus, but I agree that a new menu instance should be initialised with a specific menu.

PR #2226, opened 2022-05-02
Luiza's Takeaway

LGretzk:
# Luiza

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


lib/menu.rb
```diff
@@ -0,0 +1,44 @@
+
+class Menu
```
eoinmakers, line 2: This class adheres well to SRP, it's succinctly written.

There are some over complexities - like each dish being available or not, but this doesn't affect the rest of the implementation.

lib/order.rb
```diff
@@ -0,0 +1,93 @@
+require_relative 'menu'
+require_relative 'confirmation'
+
+class Order
```
eoinmakers, line 4: This class is doing a lot of things! Each method is concisely written, but there's something of an array of responsibilities that could potentially be grouped further.

If you were to group these methods together following this technique, which further classes would you consider drawing out? https://github.com/makersacademy/skills-workshops/blob/main/practicals/object_oriented_design/encapsulation.md

lib/order.rb
```diff
@@ -0,0 +1,93 @@
...
+  def add_by_updating_quantity(id, quantity)
+    @current_order.each { |dish| dish[:quantity] = (dish[:quantity] + quantity) if dish[:id] == id }
+  end
+
+  def fail_if_invalid_id(id)
+    fail 'Invalid id' unless @@menu.id_valid?(id)
+  end
+
+  def fail_if_dish_unavailable(id)
+    fail 'Dish unavailable' unless @@menu.dish_available?(id)
```
eoinmakers, line 78: You don't necessarily need a new method for each failure being raised. 

SRP in methods would include barriers to that responsibility being carried out - so I would include the fail message inside the relevant method, and maybe draw out the conditional itself to its own method, rather.

lib/menu.rb
```diff
@@ -0,0 +1,44 @@
...
+    { id: 5, name: 'Ceaser Salad', price: 8, available?: true, quantity: 0 },
+    { id: 6, name: 'Pizza', price: 9, available?: true, quantity: 0 },
+    { id: 7, name: 'Vegan Pizza', price: 10, available?: true, quantity: 0 },
+    { id: 8, name: 'Yum Yum Salad', price: 11, available?: true, quantity: 0 },
+    { id: 9, name: 'Singapore Laksa', price: 11, available?: true, quantity: 0 },
+    { id: 10, name: 'Spiced Noodles', price: 11, available?: true, quantity: 0 },
+    { id: 11, name: 'Phad Thai', price: 11, available?: true, quantity: 0 },
+    { id: 12, name: 'Caponata', price: 11, available?: false, quantity: 0 }
+  ].freeze
+
```
eoinmakers, line 18: There are some complexities in using a structure like this, when it comes to traversing each data point later in your implementation to print / change dishes. 

I'll share an example I wrote for another dev: 

Hashes are most useful where you have a key value pair - eg. if the only information you needed to record was the item name and price, a simple hash could look like:

```
@menu = { "1. Mattar Paneer" => 12.50, "Black Daal" => 7.50 }

@menu[item_name] will return the cost of that item.
```

For data requiring more than two values, where key/value isn't enough, I would suggest defining a new object, for example:

```
class Dish
    def initialize(name, price)
        @name = name
        @price = price
    end

    # getter methods for retrieving information
end
```

This way, you could work with an array of Dishes inside your Menu class, eg. @menu = [Dish.new(args), Dish.new(args), etc.]

Singular dish data can be accessed with @menu[0].name, etc. Which can be quite a bit easier than traversing a complex hash!

eoinmakers (commented):
Well done on achieving the functionality of the user stories.

There is an opportunity to further separate the behaviour of the Order class - I recommended a practical around encapsulation that helps to draw together common behaviours. 

Your methods are concisely written, and there's a wide range of unit tests included. 

Well done!

PR #2228, opened 2022-05-02
Russell's TakeAway Challenge

Rmorbey:
Russell's TakeAway Challenge

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


lib/menu.rb
```diff
@@ -20,6 +20,6 @@ def view_dishes
   end
 
   def check_if_available(item)
-    return 'That dish is not available. Please select another dish.' if @dishes[item].nil?
+    raise 'That dish is not available. Please select another dish.' if @dishes[item].nil?
```
Alexandros91, line 23: are both of these messages needed? Would just the 'raise' message work without the return one?

PR #2229, opened 2022-05-02
Slava's Takeaway Challenge

amfibiya17:
# Yaroslava Yates
---

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


lib/text.rb
```diff
@@ -0,0 +1,32 @@
...
+    client = Twilio::REST::Client.new(account_sid, auth_token)  
+    
+    begin
+      client.messages.create(
+        body: "Thank you! Your order was placed and will be delivered before #{delivery_time}",
+        to: @mobile,
+        from: ENV['TWILIO_NUMBER']
+      )
+      puts message.sid
+    rescue
```
penguat, line 25: This empty rescue block just swallows any errors, without telling you / the user about them.

I'd encourage you to look into whether you can rescue a more specific error, and at least `puts` to share that an error occurred.

spec/order_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+    it "puts message confirming the order" do
+      expect { subject.confirm_order }.to output(<<~output
+        Order complete.
+      output
+    ).to_stdout
+    end
+
+    it "checks if method send message was called" do
+      allow(subject).to receive(:text).and_return(text)
+      expect(text).to receive(:send_message)
```
penguat, line 86: This is good use of mocking, well done!

spec/text_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it 'creates a new message' do
+      Timecop.freeze(Time.parse("19:00")) do
+        twilio_messages = double(:twilio_messages)
+        expect(Twilio::REST::Client).to receive(:new).with(ENV['ACCOUNT_SID'], ENV['AUTH_TOKEN']).and_return(client)
+        expect(client).to receive(:messages).and_return(twilio_messages)
+        expect(twilio_messages).to receive(:create).with(
+          body: "Thank you! Your order was placed and will be delivered before 20:00", 
+          to: ENV['CUST_NUMBER'], 
+          from: ENV['TWILIO_NUMBER']
+        ).and_return(true)
```
penguat, line 24: You've required twilio_mock for these tests, but you're not really using it to check what's sent - instead you're mocking Twilio yourself.

Twilio mock isn't especially well documented - I ended up looking at its' own tests to understand how to use it to see the message sent: https://github.com/MaicolBen/twilio_mock/blob/master/spec/mocker_spec.rb#L48

With that, this test becomes much simpler:

```
it 'creates a new message' do
      Timecop.freeze(Time.parse("19:00")) do
        text.send_message

        message = TwilioMock::Mocker.new.messages.last
        expect(message.body).to eq "Thank you! Your order was placed and will be delivered before 20:00" 
        expect(message.to).to eq ENV['CUST_NUMBER']
        expect(message.from).to eq ENV['TWILIO_NUMBER']
      end
    end
```

spec/order_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+  describe "#delete_dishes" do
+    it "removes the dish from an array" do
+      subject.select_dishes(1)
+      expect(subject.delete_dishes(1)).to eq (Array.new(0))
+    end
+    
+    it "removes the dish from an array and leaves other items in the basket" do
+      subject.select_dishes(1)
+      subject.select_dishes(5)
+      expect(subject.delete_dishes(1)).to eq ([{name: "Once in a lifetime Pizza", price: 9.99}])
```
penguat, line 43: I'm not a big fan of how you're testing `delete_dishes` here - it's not obvious to me that returning the basket is a core part of `delete_dishes`'s functionality. This is largely my own opinion - but I think it would be better to add 2 dishes here, delete one, and have your `expect` around something explicitly intended to return the basket - like `subject.basket`.

lib/order.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'dishes'
+require_relative 'text'
+
+class Order
+  attr_reader :basket, :menu, :dishes, :complete, :text
```
penguat, line 5: I don't think all of these attribute readers are needed. Extra available state from inside our objects can cause problems sometimes - especially if someone else is going to use our code as an API (like a gem) - if we allow them to use more parts of our code, they are likely to rely upon them, and we have to work to maintain that too.

penguat (commented):
This is an excellent effort, well done! 

I'm impressed that you have managed to mock the twilio API, and have left some suggested code which would simplify that test.

Aside from that, there are some minor suggested improvements, which don't detract from the fact that this is an excellent effort, and one you should be proud of.

PR #2230, opened 2022-05-02
Steven Spiegl

S-Spiegl:
# Your name

# User stories 

# README checklist

- [x] how to install,
- [x] how to run,


S-Spiegl:
.

lib/order.rb
```diff
@@ -0,0 +1,51 @@
...
+  attr_reader :selection, :items, :total
+
+  def initialize
+    @selection = []
+    @items = Menu.new.items
+    @total = 0
+  end
+
+  def view_menu
+    Menu.new
```
eoinmakers: You previously grab the items on line 10, so you could re-use that instance of Menu to print the dishes instead.

lib/order.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def selection_summary
+    @selection
+  end
+
+  def total_summary
+    "Your total is: £#{@total}. Please use the complete_order function, entering your phone number as an argument, to complete your order"
+  end
+
+  def item_added_confirmation
```
eoinmakers, line 44: This functionality is not mentioned in the spec. You should try to adhere to the spec as closely as possible so you're not introducing unnecessary complexity to your project.

eoinmakers (commented):
This is a cleanly written Takeaway challenge, well done.

I left very minor comments inline, but the objectives have been clearly demonstrated here and the user stories have been achieved.

The unit tests are comprehensive and wide ranging. 

Well done.

lib/order.rb
```diff
@@ -5,18 +5,18 @@ class Order
...
 
-  def initialize
+  def initialize(items = Menu.new.items)
     @selection = []
-    @items = Menu.new.items
+    @items = items
     @total = 0
   end
 
   def view_menu
```
S-Spiegl, line 14: Have implemented suggestion to move this to initialize, but have a question:

Which is better? The implemented initialize or the one below?

  ```
def initialize(items = Menu.new)
     @selection = []
     @items = items.items
     @total = 0
  end
```
  
  Both seem to get the same results... The first reads more clearly but makes for a longer argument

File: lib/order.rb [hunk 4]
```diff
@@ -33,6 +33,13 @@ def check_order_prompt
     "Please check your order against your total:"
   end
```
S-Spiegl, line 35: did I understand User Story 3 correctly or did I go overboard?
```
As a customer
I would like to check that the total I have been given matches the sum of the various dishes in my order'

```
did I satisfy this condition simply by giving the customer a 'total' method?

lib/menu.rb
```diff
@@ -13,4 +13,5 @@ def initialize
   def view_menu
     @items
   end
+  #is this method necessary? Taking it out doesn't break tests...
```
S-Spiegl: is this method necessary? Taking it out doesn't break tests...

spec/sms_spec.rb
```diff
@@ -28,3 +25,6 @@
     expect(subject.sms_sent_confirmation).to eq('A confirmation message has been sent to the number you provided')
   end
 end
+
```
S-Spiegl: do these tests actually check an sms has been sent, or just that the class acts as though a message has been sent (e.g. changes sent_confirmation)

spec/order_spec.rb
```diff
@@ -48,15 +48,18 @@
       it 'should call the SMS class' do
         subject.add(1)
         subject.checkout
-        expect(subject).to respond_to(subject.complete_order).with(SMS.new.send_sms(ENV['MY_PHONE']))
+        expect(subject).to respond_to(:complete_order).with(1).argument
       end
     end
```
S-Spiegl, line 54: how to actually check that complete_order((ENV['MY_PHONE'])) calls SMS.new.send_sms? I realised I was misusing respond_to above and below comment so changed it above, but now all the test does is check that the method takes one argument. I want to implement a test to make sure that when someone enters their phone number it calls the SMS class...

PR #2231, opened 2022-05-03
Sam Button Takeaway Challenge

SamButton12:
# Sam Button

# User stories 

# README checklist

- [] how to install,
- [] how to run,
- [] and how to test your code?


spec/menu_spec.rb
```diff
@@ -0,0 +1,13 @@
+require './docs/menu'
+
+describe Menu do
+  it 'responds to Menu class' do
```
lukestorey95, line 4: Don't think this test is needed. Rspec always initialises the subject to be a new instance of the described class

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
+require './docs/order'
+
+describe Order do
+  let(:menu) { Menu.create }
```
lukestorey95, line 4: Interesting why .create and not .new ?

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  let(:order) { Order.create }
+
+  it 'is an instance of Order class' do
+    expect(subject).to be_an_instance_of Order
+  end
+
+  it 'responds to add_to_basket' do
+    expect(subject).to respond_to :add_to_basket
```
lukestorey95, line 12: Think this is being tested below so could probably be refactored out as well

spec/order_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  describe '#add_to_order' do
+    context 'item is on the menu' do
+      it 'can add dishes to current order' do
+        subject.add_to_basket("Guinness", 1)
+        expect(subject.current_order).to eq(["Guinness"])
+      end
+
+      it 'can add multiple dishes at once' do
```
lukestorey95, line 22: This is clever I hadn't thought about implementing quantity

spec/SMS_spec.rb
```diff
@@ -0,0 +1,28 @@
+require './docs/SMS'
+
+describe SMS do
+
+  subject(:sms) { described_class.new }
+
+  before do
+    allow(sms).to receive(:sms)
```
lukestorey95, line 8: Much prefer this was of stubbing the SMS method directly rather than trying to stub the client like I did

docs/order.rb
```diff
@@ -0,0 +1,35 @@
...
+    @new_menu = Menu.new
+    @order_price = 0
+  end
+
+  def add_to_basket(dish, quantity)
+    raise "Item not on the menu" unless @new_menu.menu.has_key?(dish)
+    quantity.times { @current_order << dish }
+  end
+
+  def calculate_price
```
lukestorey95, line 20: Think some of these methods could be encapsulated into private as the the user doesn't need to calculate the price, the program just needs to do it and the user sees the output

docs/SMS.rb
```diff
@@ -0,0 +1,38 @@
...
+    sms
+    @sent = true
+    sms_confirmation
+  end
+
+  def sent?
+    @sent
+  end
+
+  def sms_confirmation
```
lukestorey95, line 35: Nice idea to have the confirmation message, but again could be in private

docs/SMS.rb
```diff
@@ -0,0 +1,38 @@
+require 'twilio-ruby'
+
+class SMS
+
+  ETA = (Time.now + 3600).strftime("%H:%M")
```
lukestorey95, line 5: Magic number could be defined elsewhere

lukestorey95 (commented):
Some really interesting implementation here with quantity. Would love to see some more encapsulation and use of private methods to really clearly define what public interface is that the user can do and what the program does in the background. Think that would help you refine your tests further as well as you don't need to test private methods. And fingers crossed you can figure out your ENV variables as well.

PR #2234, opened 2022-05-03
takeaway-challenge

laura-voss:
# Laura Voss

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,


README.md
```diff
@@ -14,20 +14,54 @@ Takeaway Challenge
...
 
-Instructions
+Status -> Work in Progress
 -------
 
-* Feel free to use google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+* takeaway currently doesn't allow dishes to be deleted from order -> needs to be updated
```
S-Spiegl, line 20: this is a good idea - wish I'd thought of this!

lib/take_away.rb
```diff
@@ -0,0 +1,68 @@
...
+    @current_dish = []
+  end
+
+  def view_menu
+    @menu.view
+  end
+
+  def add_dish(number, qty)
+    fail "please enter a valid dish number" if @menu.existing_dish(number, qty) != true
+    if duplicate_dish(number, qty) == true
```
S-Spiegl, line 20: This seems like a good idea. I just kept adding to the 'basket' array of hashes, so your idea of actually editing a quantity value seems a lot more like a real user experience

lib/menu.rb
```diff
@@ -0,0 +1,26 @@
...
+  attr_reader :menu
+  
+  def initialize
+    @menu = [
+      { :item => 1, :dish => "Mattar Paneer", :price => 12.50 },
+      { :item => 2, :dish => "Black Daal", :price => 7.50 },
+      { :item => 3, :dish => "Raita", :price => 3.50 },
+      { :item => 4, :dish => "Garlic Naan", :price => 3.50 }
+    ]
+  end
```
eoinmakers, line 12: Hashes are most useful where you have a key value pair - eg. if the only information you needed to record was the item name and price, a simple hash could look like:

`@menu = { "1. Mattar Paneer" => 12.50, "Black Daal" => 7.50 } `

@menu[item_name] will return the cost of that item. 

For data requiring more than two values, where key/value isn't enough, I would suggest defining a new object, for example:

```
class Dish
    def initialize(name, price)
        @name = name
        @price = price
    end

    # getter methods for retrieving information
end
```

This way, you could work with an array of Dishes inside your Menu class, eg. @menu = [Dish.new(args), Dish.new(args), etc.]

Singular dish data can be accessed with @menu[0].name, etc. Which can be quite a bit easier than traversing a complex hash!

lib/menu.rb
```diff
@@ -0,0 +1,26 @@
...
+  
+  def view 
+    @menu.map { |dish| "#{dish[:item]}. #{dish[:dish]} £#{dish[:price]}" }
```
eoinmakers, line 15: Were you to use just a single key/value hash as the one I suggested in my first comment, you could refactor this to something like: 

`puts "#{key}:#{value}"`

This will print the full menu.

lib/take_away.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+  
+  def current_dish(number, qty)
+    (@menu.selection(number, qty).push(:qty => qty, :subtotal => sub_total(number, qty))).reduce(&:merge)
+  end
+
+  def dish_added_message(number, qty)
+    (@menu.selection(number, qty)).map do |dish| 
+      "#{qty} x #{dish[:dish]} £#{dish[:price]}/each, has been added to your order -> £#{(sub_total(number, qty))}"
+    end
```
eoinmakers, line 57: Although this is functional, the functionality to receive a confirmation for each item added to the order is not mentioned in the spec - you should stick as closely as possible to the spec as possible. This helps keep your code clean / simple!

lib/take_away.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  def sub_total(number, qty)
```
eoinmakers, line 60: You could omit this and only calculate the total at runtime / when the user confirms the order.

eoinmakers (commented):
This is a good implementation of Takeaway, well done! 

My comments propose simpler solutions, as I think the use of the structure [{}, {}, {}] for the @menu made interacting with it in your code, and I imagine making changes to your code, challenging in places. 

You achieved the functionality of all of the user stories.

There are some additional features, so while you should feel free to have fun with these challenges and try things out, remember when doing tech tests in the future that the interviewer wants to see you can work to a specification and not do more than is asked. 

I thought there were a good range of unit tests included - including the tests that only check if certain methods are being called.

PR #2235, opened 2022-05-03
Gawain Hewitt Takeaway challenge

gawainhewitt:
Gawain Hewitt

# User stories 

# README checklist
 
My readme is very basic to be honest


archiemartini (commented):
Really like this, will look very nice in action

lib/takeaway.rb
```diff
@@ -0,0 +1,13 @@
+require './lib/menu.rb'
+
+class Takeaway
+  attr_reader :menu
+  def initialize
+    @menu = Menu.new
```
archiemartini, line 9: I'm impressed you definitely know what you're doing, especially with the corresponding tests

I'm no expert in dependency injection, but this article near the beginning

https://medium.com/@Bakku1505/introduction-to-dependency-injection-in-ruby-dc238655a278

explains how linking the classes together so "strongly" ie 'def initialize... @menu = Menu.new' can cause issues down the line

spec/menu_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  it 'should respond to contents' do 
+    expect(subject).to respond_to(:contents)
+  end
+
+  it 'should display menu' do
+    expect do
+      subject.show
+    end.to output("1. Chips - £1\n2. Tofu - £2\n3. Broccoli - £1\n4. Ice_cream - £2\n").to_stdout
+  end
+
```
archiemartini, line 29: your general ruby syntax and string-morphing is fantastic!

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+describe Takeaway do
+  it 'should respond to show menu' do 
+    expect(subject).to respond_to(:menu)
+  end
+
+  it 'should call display method from Menu' do
+    expect(subject.menu).to receive(:show)
+    subject.show_menu
+  end
+
```
archiemartini, line 59: again, already thinking about testing injection (looking at commit 1) here is very switched on. nice. took a lot of research/cheating for me to come to tests like this

spec/takeaway_spec.rb
```diff
@@ -1,15 +1,31 @@
 require 'takeaway'
+require 'stringio'
```
archiemartini, line 2: You'll have to clue me in on this

lib/takeaway.rb
```diff
@@ -19,24 +17,45 @@ def show_menu
...
+    quantity = @input.gets.to_i
+  end
+
+  def log_order(dish, quantity)
+    order_item = {}
+    order_item[:food] = dish
+    order_item[:quantity] = quantity
+    @summary << order_item
+  end
+
```
archiemartini, line 66: Really impressive thought about actual user interface, certainly going beyond what is required. You have my sympathy and respect for keeping your head around it though

lib/menu.rb
```diff
@@ -15,15 +15,13 @@ def show
   end
 
   def check(dish)
-    _dish = dish.to_sym
+    the_dish = dish.to_sym
     @dishes.each do |dish|
-      "food #{dish[:food]}"
-      if _dish == dish[:food]
+      if the_dish == dish[:food]
```
archiemartini, line 20: I feel this could be simplified to a ternary operator

ie 'the_dish == dish[:food] ? true : false

lib/takeaway.rb
```diff
@@ -0,0 +1,67 @@
...
+    @summary = []
+    @total = 0
+  end
+
+  def show_menu
+    @menu.show
+  end
+
+  def order
+    loop do
```
leoht, line 20: Nice that you've worked with user input as well! I'd suggest looking at extracting further this logic into a new class, maybe named `Cli` or `Interface`, which would contain methods such as `clarify_order` or `ask_for_order`. It could probably also handle the display of the menu (which is right now in the `Menu` class). This way you can make the main `Takeaway` and `Menu` classes do a bit less, and it also makes it easier to test the input/output, as there will be only one class that directly uses `puts` or `gets`, so you remove the need to mock these for the other classes

spec/menu_spec.rb
```diff
@@ -0,0 +1,35 @@
+require 'menu'
+
+describe Menu do
+
+  context '#show' do
+    it 'should display menu' do
+      # output = <<~MENU
```
leoht, line 7: I reckon this didn't work because the [heredoc syntax](https://www.rubyguides.com/2018/11/ruby-heredoc/) needs the "ending" tag (in your case `MENU`) to be on its own line, at the very beginning of the line (without any blank space before). This would make it:

```ruby
it 'should display menu' do
  output = <<~MENU
       1. Chips - £1
       2. Tofu - £2
       3. Broccoli - £1
       4. Ice_cream - £2 
MENU

   # ...
end
```

lib/takeaway.rb
```diff
@@ -0,0 +1,67 @@
...
+        quantity = ask_for_quantity
+        log_order(dish, quantity)
+      else
+        clarify_order
+      end
+    end
+  end
+
+  def confirm_order
+    @summary.each do |dish|
```
leoht, line 33: This could also be part of the new class mentioned above - usually it's good practice to split the code that deals with "presenting" or display, from the code that is dealing with pure data or logic. This way, it's easier to change things related to how looks the program outputs (think formatting, translating, or more complex things!) without touching at the core logic of the program, which stays the same no matter how the output is presented and how the input is provided

spec/feature_test_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        Please type each dish you require followed by return. 
+        When you have finished your order press return twice.
+        0 order of Broccoli at £1 each
+        Total order is £1
+        OUTPUT
+    end
+  end
+
+  def place_order_with_input(*order)
+    input = StringIO.new(order.join("\n") + "\n" + "\n" + "\n")
```
leoht, line 24: Really nice way to mock IO!

lib/menu.rb
leoht: Maybe this method could also be renamed to something like `contains_dish?` so it's more explicit on what it does

lib/menu.rb
```diff
@@ -0,0 +1,37 @@
+class Menu
+
+  def initialize(input: $stdin, output: $stdout)
+    @input = input
+    @output = output
+    @dishes = [{ food: :Chips, price: 1 }, 
+      { food: :Tofu, price: 2 }, { food: :Broccoli, price: 1 },
```
leoht, line 7: Minor one here - it's probably better to leave the dish names as strings rather than symbols, as symbols are mostly used for keys, special configuration values, enumerated values, etc, rather than purely "textual" values

spec/feature_test_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'takeaway'
+
+describe 'Featuretest' do
+  context '#order' do
+    it 'will ask again if item mispelt' do
+      output = place_order_with_input("Brocoli", "Broccoli")
```
leoht, line 6: Nice one! This test suite should probably contain a bit more tests, at least one testing the whole integration of the different components with a successful order process, and assert the output of the `takeaway.order` method

leoht (commented):
Well done, that's great work - I left some comments about a few different things to improve, but let me know if you'd like to discuss this feedback more in depth :)

PR #2237, opened 2022-05-03
Alexandros' takeaway challenge

Alexandros91:
Alexandros Papagiannis

# User stories 


lib/order.rb
```diff
@@ -0,0 +1,29 @@
+require_relative './menu.rb'
+
+class Order
```
Rmorbey, line 3: You can initialize the menu in order class so you don't have to have it twice.

def initialise(menu = Menu.new)
@menu = menu
end

spec/menu_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'menu'
+require 'order'
+
+describe Menu do
+	describe '#see_menu' do
```
Rmorbey, line 5: I like how you nestle your describe for your methods, it makes reading your unit tests much easier.

lib/order.rb
```diff
@@ -0,0 +1,29 @@
...
+		"fried eggs with bacon" => 6.00,
+		"eggs benedict with avocado" => 7.50,
+		"scrambled eggs with mozzarella" => 7.00,
+		"pancakes with chocolate" => 7.50,
+		"yoghurt with fresh fruit" => 6.00
+  }
+
+	@order = {}
+
+	def select_items(item, num_of_portions)
```
Rmorbey, line 17: I like how you've added number of portions available, that's something I didn't even think of.

PR #2238, opened 2022-05-03
Takeaway Pull Request

AKCDNG:
# Anish Kakaiya

# User stories 

- [X] User story 1: "I would like to see a list of dishes with prices"
- [X] User story 2: "I would like to be able to select some number of several available dishes"
- [X] User story 3: "I would like to check that the total I have been given matches the sum of the various dishes in my order"

# README checklist

- [X] how to install,
- [X] how to run,
- [X] and how to test your code?


lib/takeaway.rb
```diff
@@ -0,0 +1,54 @@
...
+    fail "Cannot complete order. Nothing has been ordered." if order_empty?
+    @text.send_text
+  end
+
+  private
+
+  def search_menu(dish)
+    @dishes.show_menu.each do |dish_name| 
+      @order.push(dish_name) if dish == dish_name[:name]
+    end
```
tamasmagyarhunor-makers, line 39: Can you think of a more efficient way of selecting the dish instead of having to iterate over all the dishes each time they add something to the order?
with the current solution think about what effect would it have if both the menu and the order would be massive?

spec/takeaway_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    takeaway.order_dishes("Jam On Toast")
+    expect(takeaway.order_total).to eq(15.5)
+  end
+
+  it "sends text after placing order" do
+    takeaway = Takeaway.new
+    takeaway.order_dishes("Avocado On Toast", 2)
+    takeaway.stub(:place_order) { "Order placed. Text sent." }
+    expect(takeaway.place_order).to eq "Order placed. Text sent."
+  end
```
tamasmagyarhunor-makers, line 31: Not sure if I get your question right regarding mocking Anish, but when you start working with mocks first its usual to feel like you're 'cheating' the behaviour. The important bit here that you have to mock/stub the behaviour of the object that your object is depended on. 
You set up the scenario for the order to be placed, but in the future when your test is run, you wont be sure how to Twillio API behaves. Maybe it will be down at some point and if your Takeaway objects tests fail, you'd think your code is wrong. This is why when your tests include dependencies, you want to make sure you mock/stub the necessary behaviour that helps your object (in this case Takeaway) to be tested.

tamasmagyarhunor-makers (commented):
Hey Anish,
Regarding your request to review it from the Single Responsibility Principle I think you did well. Your objects and methods have one responsibility. You also used Encapsulation well, only leaving the necessary public methods. 
I left a comment regarding your question about mocking Twillio.
Mocking can be quite tricky to grasp when you're dealing with it the first time, please do reach out for a quick chat if you feel you'd appreciate a bit more explanation on it.

PR #2239, opened 2022-05-03
Simon Tinsley (Partial)

simon-2357:
# Your name

Simon Tinsley

# User stories 

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/menu.rb
+++ b/lib/menu.rb
@@ -1,0 +1,22 @@
+class Menu
+
+  # For whoever is code reviewing this - I shamelessly stole these dishes from one
+  # of my fav restaurants - Rasa in Stoke Newington, I highly recommend it if you're ever in the area.
+  
+  def initialize(dish_class=Dish)
+    @dishes = [
+      { name: "Rasam", price: 5 }, 
+      { name: "Plain Dosa", price: 6 }, 
+      { name: "Masala Dosa", price: 7 }, 
+      { name: "Rasavangi", price: 7.5 }, 
+      { name: "Plain Rice", price: 3 }, 
+      { name: "Tamarind Rice", price: 3.5 }, 
+      { name: "Mango Halwa", price: 4 }
+    ]
+  end
+
+  def view
+    @dishes.each { |x| print "#{x[:name]}, £#{x[:price]}"+ "\n" } 
+  end
+
+end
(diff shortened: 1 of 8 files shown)


lib/menu.rb
```diff
@@ -0,0 +1,22 @@
+class Menu
+
+  # For whoever is code reviewing this - I shamelessly stole these dishes from one
```
LGretzk, line 3: noted :)

lib/order.rb
```diff
@@ -0,0 +1,29 @@
...
+    dishes << dish
+  end
+
+  def remove(dish)
+    dishes.delete(dish)
+  end
+
+  def total
+    total = 0
+    dishes.each { |x| total += x.price }
```
LGretzk, line 18: total += x[:price] ?

lib/restaurant.rb
```diff
@@ -0,0 +1,46 @@
+class Restaurant
+  m = Menu.new
+  o = Order.new
```
LGretzk, line 3: may be wrong, but shouldn't Restaurant initialize with both of these variables? I.e. 
def initialize
  @m = Menu.new
...

lib/restaurant.rb
```diff
@@ -0,0 +1,46 @@
...
+  else
+    puts "That doesn't seem to be a valid action, please choose add remove or place."
+    action = gets.chomp.downcase
+    next_action
+  end
+
+  def add
+    puts "Here is our menu, what would you like to order?"
+      m.view
+      dishes = gets.chomp.downcase
```
LGretzk, line 29: does the person need to type in the whole dish hash?

lib/sms.rb
```diff
@@ -0,0 +1,12 @@
+require 'twilio-ruby'
+
+  account_sid = "AC4652867e62a8ad85f253fcd1c96016ba" # Your Test Account SID from www.twilio.com/console/settings
+  auth_token = "590379d91fe68d1e887e7dcf91354037"   # Your Test Auth Token from www.twilio.com/console/settings
```
LGretzk, line 4: I think that needs to be put in the environment

spec/order_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'order'
```
LGretzk, line 1: nice clear code, good use of doubles, and test coverage

LGretzk (commented):
nice clear code, good use of doubles, and test coverage

PR #2240, opened 2022-05-03
Takeaway Challenge (partial)

GeorgeDainton:
# George Dainton

# User stories 

- [Y] User story 1: "I would like to see a list of dishes with prices"

# README checklist

- [ ] how to install,
- [ ] how to run,

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,11 @@
+class Order
+
+  def initialize(menu = Menu.new)
+    @menu = menu
+  end
+
+  def choose_item(item)
+  end
+   
+
+end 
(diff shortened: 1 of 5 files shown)


lib/order.rb
```diff
@@ -0,0 +1,11 @@
+class Order
+
+  def initialize(menu = Menu.new)
```
AKCDNG, line 3: `(menu = Menu.new)` may not need to be there, on the next line you could have `@menu = Menu.new`

AKCDNG (commented):
All good! Just the one change from my perspective :)

PR #2241, opened 2022-05-03
Farzan's TakeAway Challenge 3/4 user stories

Farzan-I:
# Farzan Imanzadeh

# User stories 

- [X] User story 1: "I would like to see a list of dishes with prices"
- [X] User story 2: "I would like to be able to select some number of several available dishes"
- [X] User story 3: "I would like to check that the total I have been given matches the sum of the various dishes in my order"

# README checklist

- [X] how to install,
- [X] how to run,
- [X] and how to test your code?

### DIFF
--- a/lib/order.rb
+++ b/lib/order.rb
@@ -1,0 +1,26 @@
+class Order 
+  attr_reader :basket
+
+  def initialize
+    @total = 0
+    @menu = Menu.new.dishes
+    @basket = []
+  end
+
+  def choose_dish(dish)
+    @basket << dish
+  end
+
+  def total
+    @menu.each do |dish, price|
+      @basket.each do |selected_dish|
+        @total += price if selected_dish == dish
+      end
+    end
+    "Your basket's total is £#{@total}."
+  end
+
+  def check_basket
+    return @basket
+  end
+end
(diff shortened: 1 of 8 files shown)


lib/order.rb
```diff
@@ -0,0 +1,26 @@
+class Order
```
ErlantzR, line 1: Name of variables are quite adequate and easy to understand what the code is trying to do

lib/order.rb
```diff
@@ -0,0 +1,26 @@
+class Order 
+  attr_reader :basket
+
+  def initialize
+    @total = 0
+    @menu = Menu.new.dishes
+    @basket = []
+  end
+
+  def choose_dish(dish)
```
ErlantzR, line 10: It might need to include a way ask for more than one of each dish. A hash for @basket could help on that

lib/view_menu.rb
```diff
@@ -0,0 +1,13 @@
+class ViewMenu
```
ErlantzR, line 1: This might have been a method inside Menu class?

spec/order_spec.rb
```diff
@@ -0,0 +1,32 @@
+require 'order'
```
ErlantzR, line 1: All features seem to be tested thoroughly

